PR #14, opened 2017-04-12
removing google console refs and cleaning up instructions

choldgraf:
There have been bugs in the interactive google console which made these instructions not work...we couldn't figure them out and think it's probably better to just have people use the user SDK anyway since it's less clunky. (this also fixes some random syntax / display issues I noticed)


choldgraf:
(in case anyone sees this: I'm gonna be pretty liberal about merging PRs for early stages of these docs...we can incorporate more due-diligence once they've stabilized but stuff is changing really quickly right now)

PR #37, opened 2017-04-28
add pip to requirements, use double equal

amueller:
requirements.txt uses ``==`` (and I can never remember that lol). Also, updating pip is probably a good idea on whatever image that is, so I added that here as a slight hint.

### DIFF
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -81,7 +81,6 @@ configure JupyterHub to build off of this image.
 
    .. code-block:: bash
 
-          pip==9.0.1
           numpy==1.12.1
           scipy==0.19.0
           matplotlib==2.0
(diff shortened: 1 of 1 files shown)


doc/source/extending-jupyterhub.rst
```diff
@@ -81,9 +81,10 @@ configure JupyterHub to build off of this image.
 
    .. code-block:: bash
 
-          numpy=1.12.1
-          scipy=0.19.0
-          matplotlib=2.0
+          pip==9.0.1
```
yuvipanda: I think we should take this out, since this should be managed by the s2i builder rather than by users (same reason you don't have to specify notebook or ipython here)

amueller: it'll have an old version though. It uses pip-8.1.2. If you think it's better not to mess with it, alright.

willingc: Perhaps float pip as `pip>=9.0`

yuvipanda: Nope, that makes the image non-reproducible. We should try to use == as much as possible.

yuvipanda:
I've gotten rid of pip and merged this! Thanks for the patch :)

PR #39, opened 2017-04-28
Add Google OAuth2 example

ryanlovett:
First pass at providing full instructions for setting up Google OAuth2.


willingc:
@ryanlovett Thanks! I'm going to merge and then clean up any formatting 👍

PR #71, opened 2017-05-25
Corrected url to jupyterhub helm-chart

mjbright:
Corrected URL to new address (now jupyterhub-v0.3.x was jupyterhub-0.3.x) and also updated to latest v0.3.1


yuvipanda:
Thank you for the patch! :D

PR #77, opened 2017-05-29
Clean up rst docs and add note for Azure

willingc:


willingc:
@yuvipanda Docs now build with no errors or warnings.

yuvipanda:
w00t, awesome!

PR #83, opened 2017-06-07
Restructure introduction and demark divisions of interest

mpacer:
The primary purpose of this PR is to remove the infinite reference loop that you ran into in following the instructions. 

Basically "getting started" took you to "creating your JupyterHub" telling you to go to the first section of it which was the "getting started" block.

I think what was actually meant was the next block if someone wants a straightforward introduction to a working system (the zero to jupyterhub of the title of the thing). They should just jump right to creating a kubernetes cluster on google cloud.

Everything else is just my attempt at a quick summary without reading the entirety of those documents based on the description of the target. It's not my most careful titling work, but I figure that you'll want to have input on them no matter what they are (a reasonable desire).

### DIFF
--- a/doc/source/getting-started.rst
+++ b/doc/source/getting-started.rst
@@ -17,12 +17,30 @@ following things:
 * **Docker** to use containers that standardize computing environments
 * **JupyterHub** to manage users and deploy Jupyter notebooks
 
-To get started, we recommend beginning with the first item on our
-:ref:`main page <creating-your-jupyterhub>`. If you'd like to know how to
-expand and customize your jupyterhub setup, check out
-:ref:`extending-jupyterhub`. Finally, for a more extensive description of the
-tools and services that JupyterHub utilizes, see our 
-:ref:`tools` page.
+You already are well on your way to understanding what it means (procedurally)
+to deploy Jupyterhub. 
+
+For more guidance
+-----------------
+
+If you want more guidance, the next step on your journey is to :ref:`create a
+Kubernetes cluster on Google Cloud
+<google-cloud>`. 
+
+For info on extensions and customization 
+----------------------------------------
+
+If you'd like to know how to expand and customize your jupyterhub setup, check
+out :ref:`extending-jupyterhub`. 
+
+For more a discussion about software dependencies
+-------------------------------------------------
+
+For a more extensive description of the tools and services that JupyterHub
+utilizes, see our :ref:`tools` page.
+
+Questions or Suggestions?
+=========================
 
 If you have questions or suggestions, please reach out at our `issues page`_
 on GitHub.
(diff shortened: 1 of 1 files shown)


doc/source/getting-started.rst
```diff
@@ -17,12 +17,30 @@ following things:
...
-To get started, we recommend beginning with the first item on our
-:ref:`main page <creating-your-jupyterhub>`. If you'd like to know how to
-expand and customize your jupyterhub setup, check out
-:ref:`extending-jupyterhub`. Finally, for a more extensive description of the
-tools and services that JupyterHub utilizes, see our 
-:ref:`tools` page.
+You already are well on your way to understanding what it means (procedurally)
+to deploy Jupyterhub. 
+
+For more guidance
```
choldgraf: I'd call each of these sections something less conversational and more topic-driven / short. So for this, call it something like "Deployment Guide" and then add a sentence about how we've put together a short walkthrough to deploy JupyterHub from scratch.

doc/source/getting-started.rst
```diff
@@ -17,12 +17,30 @@ following things:
...
+-----------------
+
+If you want more guidance, the next step on your journey is to :ref:`create a
+Kubernetes cluster on Google Cloud
+<google-cloud>`. 
+
+For info on extensions and customization
```
choldgraf: "Extending and Customizing JupyterHub"

choldgraf: Maybe add an extra sentence or two describing what we mean by this. E.g.:

"If you'd like to know how to expand and customize your jupyterhub setup, such as
increasing the hardware available to users or changing authentication services, check
 out :ref:`extending-jupyterhub`."

doc/source/getting-started.rst
```diff
@@ -17,12 +17,30 @@ following things:
...
+Kubernetes cluster on Google Cloud
+<google-cloud>`. 
+
+For info on extensions and customization 
+----------------------------------------
+
+If you'd like to know how to expand and customize your jupyterhub setup, check
+out :ref:`extending-jupyterhub`. 
+
+For more a discussion about software dependencies
```
choldgraf: "Dependencies for Building JupyterHub"

Something like that, since otherwise it's a little confusing whether we mean "dependencies within your jupyterhub deployment" vs "dependencies that you need in order to deploy"

choldgraf:
Nice! I agree that this is a better way of displaying / organizing that information. I added a few comments on titles etc but in general I think it looks great

doc/source/getting-started.rst
mpacer: Hopefully the new title is less ambiguous.

doc/source/getting-started.rst
mpacer: Went for that verbatim.

Issue #87, opened 2017-06-13
Change docs to use default small cluster and document how to change size of cluster


aculich:
The default cluster size [currently documented](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blame/master/doc/source/create-k8s-cluster.rst#L56) for GCP is `n1-highmem-2` but that's a little large to start with for people who want to just test it out and may burn through credits faster than desired.

We should probably provide a smaller default along with instructions on how to dynamically resize the cluster (which is different from auto-scaling which is a separate topic).

willingc:
@aculich @yuvipanda Can this be closed or is there still an action remaining on this?

PR #90, opened 2017-06-19
Setup a cheaper instance size as default

yuvipanda:
Also clarify what 'zone' is

Partial fix for #87


choldgraf:
LGTM - we should get the cost estimator numbers working!

PR #92, opened 2017-06-19
Don't say 'hardware' when we mean 'computational resources'

yuvipanda:
Fixes #39

### DIFF
--- a/doc/source/tools.rst
+++ b/doc/source/tools.rst
@@ -15,7 +15,7 @@ This is whatever will run the actual computation. Generally it means a
 company, university server, or some other organization that hosts computational
 resources that can be accessed remotely. JupyterHub will run on these
 computational resources, meaning that users will also be operating on these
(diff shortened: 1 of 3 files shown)


doc/source/tools.rst
```diff
@@ -12,10 +12,10 @@ Cloud Computing Providers
...
 
 This is whatever will run the actual computation. Generally it means a
-company, university server, or some other organization that hosts computers
-that can be accessed remotely. JupyterHub will run on this hardware, meaning
-that users will also be operating on this hardware if they're interacting
-with your JupyterHub.
+company, university server, or some other organization that hosts computational
+resources that can be accessed remotely. JupyterHub will run on these
+computational resources, meaning that users will also be operating on these
+computational resources if they're interacting with your JupyterHub.
```
choldgraf: "...also be operating on the same resources if they're interacting..."

PR #103, opened 2017-06-27
Freeze current version of docs to 3.1

willingc:
To match the version of helm-chart and then we can tag/release this as 3.1 on RTD.


choldgraf:
Good idea to mirror the release of the helm chart, I'm +1 to merge but will wait to see if @yuvipanda or @minrk have any reasons why we shouldn't do that

yuvipanda:
+1.

yuvipanda:
Going to merge this, then tag 0.3.1! Is that enough for it to show up on RTD, @willingc / @choldgraf?

willingc:
@yuvipanda It's best to tag the release after it successfully builds with changes on RTD.

RTD is having a build issue on their end, so we needed to do a workaround in #106 to have the docs build on RTD. We should create a new tag for #106 commit and then set that to stable in RTD. After merging #105 and making sure the docs build on RTD, we can go ahead and tag that commit for 0.4.

Issue #102, opened 2017-06-25
Update this doc for v0.4 of helm-chart 


yuvipanda:
O'Reilly is going to publish something about the helm-chart and this guide soon, so we should make sure we update this guide for 0.4. We should also make sure people who are still on 0.3 have a way to access the older code.

Easiest and simplest way would be to tag a 0.3.x version of this repo and make 0.4 changes to master. We can then keep making changes to master to reflect 0.4.

Long term, we should figure out #100 and #101.

willingc:
Completely agree with this approach for the short-term. 👍

willingc:
## Changes needed for 0.4

**Creating Your Own JupyterHub Guide**

- [ ] https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub.html#install-jupyterhub Update version in command

**Customization Guide**

- [ ]  https://zero-to-jupyterhub.readthedocs.io/en/latest/extending-jupyterhub.html#applying-configuration-changes Update version in command (or make it generic for current version)
- [ ] https://zero-to-jupyterhub.readthedocs.io/en/latest/extending-jupyterhub.html#authenticating-with-oauth2 Add a full example for GitHub OAuth

**Administration Guide**

- [ ] No changes identified for 0.4

PR #104, opened 2017-06-27
Update documentation version to 4.0.0

willingc:
Partial fix to #102.

- Bump versions in `conf.py`
- Make `helm upgrade` command generic and provide 4.0.0 as an example.


doc/source/extending-jupyterhub.rst
```diff
@@ -17,12 +17,22 @@ The general method is:
 
      .. code-block:: bash
 
-        helm upgrade <YOUR_RELEASE_NAME> https://github.com/jupyterhub/helm-chart/releases/download/v0.3/jupyterhub-v0.3.tgz -f config.yaml
+        helm upgrade <YOUR_RELEASE_NAME> https://github.com/jupyterhub/helm-chart/releases/download/<HELM_CHART_VERSION>/jupyterhub-<HELM_CHART_VERSION>.tgz -f config.yaml
 
    Where ``<YOUR_RELEASE_NAME>`` is the parameter you passed to ``--name`` when
    `installing jupyterhub <setup-jupyterhub.html#install-jupyterhub>`_ with
-   ``helm install``. If you don't remember it, you can probably find it by doing
+   ``helm install`` and ``<HELM_CHART_VERSION>`` is the desired version.
```
choldgraf: If we use `<HELM_CHART_VERSION>` here, then should we do the same below (e.g., line 33)

yuvipanda: This will actually change a bit more for 0.4, I'll make the changes. Plus IMO we should explicitly mention the version here...

yuvipanda:
Going to close in favor of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/105

PR #108, opened 2017-06-28
Add note about verifying that helm has installed

### LINKED ISSUE #48: Add a note about getting at least helm version 2.4.1
On kubernetes 1.6.x clusters, it is needed to avoid https://github.com/jupyterhub/helm-chart/issues/4

yuvipanda:
Fixes  #48

### DIFF
--- a/doc/source/setup-helm.rst
+++ b/doc/source/setup-helm.rst
@@ -32,6 +32,23 @@ cluster. At the terminal, enter:
 
 This command only needs to run once per Kubernetes cluster.
 
+Verify
+------
+
+You can verify you have the correct version + it installed properly by running:
+
+.. code:: bash
+
+   helm version
+
+It should provide output like
+
+.. code::
+   Client: &version.Version{SemVer:"v2.4.1", GitCommit:"46d9ea82e2c925186e1fc620a8320ce1314cbb02", GitTreeState:"clean"}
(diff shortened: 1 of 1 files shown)


doc/source/setup-helm.rst
```diff
@@ -32,6 +32,23 @@ cluster. At the terminal, enter:
 
 This command only needs to run once per Kubernetes cluster.
 
+Verify
+------
+
+You can verify you have the correct version + it installed properly by running:
+
+.. code:: bash
+
```
choldgraf: why a space after the `..` block here but not later on?

doc/source/setup-helm.rst
```diff
@@ -32,6 +32,23 @@ cluster. At the terminal, enter:
...
+   helm version
+
+It should provide output like
```
choldgraf: `should provide something similar to the following output:`

doc/source/setup-helm.rst
```diff
@@ -32,6 +32,23 @@ cluster. At the terminal, enter:
 
 This command only needs to run once per Kubernetes cluster.
 
+Verify
+------
+
+You can verify you have the correct version + it installed properly by running:
```
choldgraf: `that you have the`

Issue #109, opened 2017-06-28
Complete section on customizing environment of users & resources available to them.


yuvipanda:
This is currently the `extending-jupyterhub.rst`, which we should split up and re-format.

Following ToC (adapted from #67):
- [x] User environment
  - [x] Using an existing docker image
  - [x] Creating a new image (repo2docker)
  - [x] Setting environment variables
- [x] User resources
  - [x] Memory limits & guarantees
  - [x] CPU limits & guarantees
  - [x] Storage
    - [x] Enabling & disabling
    - [x] Changing size available
    - [x] SSD vs HDD

choldgraf:
we've now got an extensive section on user storage, so closing this! http://zero-to-jupyterhub.readthedocs.io/en/latest/user-storage.html

PR #112, opened 2017-06-30
Create a separate file for just user-experience changes

yuvipanda:
- Some content is from extending-jupyterhub, which is too scattered.
  Ideally it wouldn't exist and be split into a bunch of different
  files. This is the first split.
- Recommend using docker-stacks instead of the dsep image
- Document how to turn off persistent storage
- Document how to change size of persistent storage
- Document how to set up environment variables

Partial fix for #109


yuvipanda:
@willingc @choldgraf I probably missed a bunch of stuff here, would appreciate a review! Feel free to modify and push if you could :)

One thing to figure out is how often we link people to the 'just changing the file is not enough, you have to apply the change!' section. In every section? unsure.

choldgraf:
That's a good idea - I was also thinking it be good to have an (internal?) document where we can write post-mortems (well, hopefully not mortem, but you get the idea) of bootcamps/workshops/etc that have been run on jhub or binder tech. We could have the general takeaways, challenges, etc in one place so we keep track of what to improve in the future. WDYT?

yuvipanda:
@choldgraf yeah, something like https://github.com/data-8/infrastructure/tree/master/incidents. I see you already have this going :)

choldgraf:
@yuvipanda is this ready for someone to take a look at, or still a WIP?

yuvipanda:
Ready for someone to look at!

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+.. code-block:: yaml
+
+   singleuser:
+     extraEnv:
+       EDITOR: "vim"
+
+This will set the environment variable ``EDITOR`` to the value ``vim``. You
+can set any number of static environment variables here as you want.
+
+Users can read the environment variables in their code in various ways.
```
choldgraf: I feel like this little end section isn't necessary...it's not our job to tell users how to access environment variables within their respective languages

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+--------------
+
+User resources are the CPU / RAM / Storage resources you provide your users.
+
+
+Setting memory and CPU guarantees / limits for your users
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each user on your JupyterHub gets a slice of memory and CPU to use. There are
+two ways to specify how much users get to use: resource *guarantees* and
```
choldgraf: maybe make this bold instead of italics?

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+A resource *guarantee* means that all users will have *at least* this resource
+available at all times, but they may be given more resources if they're
+available. For example, if users are *guaranteed* 1G of RAM, users can
+technically use more than 1G of RAM if these resources aren't being used by
+other users.
+
+A resource *limit* sets a hard limit on the resources available. In the example
+above, if there were a 1G memory limit, it would mean that users could use
+no more than 1G of RAM, no matter what other resources are being used on the
+machines.
```
choldgraf: Maybe a quick note that "if users used more than 1GB of RAM, their kernel would be restarted automatically". Or some note to the effect of saying what the subjective experience would be from the user's perspective

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+    .. code-block:: yaml
+
+       singleuser:
+           memory:
+              limit: 1G
+              guarantee: 1G
+
+This sets a memory limit and guarantee of 1G. Kubernetes will make sure that
+each user will always have access to 1G of RAM, and requests for more RAM will
+fail (your kernel will usually die). You can set the limit to be higher than
```
choldgraf: ah I see it's documented here...

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+the guarantee to allow some users to use larger amounts of RAM for
+a very short-term time (e.g. when running a single, short-lived function that
+consumes a lot of memory).
+
+.. note::
+    Remember `apply the changes <#applying-configuraiton-changes>`_ after changing
```
choldgraf: remember to*

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+    your config.yaml file!
+
+Storage resources
+~~~~~~~~~~~~~~~~~
+
+Each user gets their own, 10Gi disk for storage when they log in. You
```
choldgraf: no comma

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+can customize this in many ways!
+
+Turning off per-user persistent storage
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
choldgraf: OCD alert: section header underline doesn't match the title length :-)

doc/source/user-experience.rst
```diff
@@ -0,0 +1,282 @@
...
+    .. code-block:: bash
+
+          singleuser:
+            lifecycleHooks:
+              postStart:
+                exec:
+                  command: ["/bin/sh", "-c", "test -f $HOME/.copied || cp -Rf /srv/app/src/. $HOME/; touch $HOME/.copied"]
+
+
+Note that this will only copy the contents of the directory to $HOME *once* -
```
choldgraf: .. note::

willingc:
@yuvipanda @choldgraf I've made some pretty heavy edits to this section to formatting and wording. I'm going to merge this PR. I'll cover Chris' review comments in a separate PR with the other changes.

PR #113, opened 2017-07-03
Config yaml example fix

jkuruzovich:
Current version shows example as token/proxy where I believe it should
really be proxy/secretToken.


yuvipanda:
Thanks for catching it!  I'm actually going to remove the examples, since am pretty sure it'll be tempting for people to copy paste it :)

PR #120, opened 2017-07-16
Clarify 'size' when resizing cluster

### LINKED ISSUE #119: Clarify meaning of size and length of time for command in Resizing your Cluster
We should be more explicit that size in this section refers to number of nodes. Also the command `gcloud container clusters list` can take a couple of minutes to reflect the change after a resize.

willingc:
Closes #119 

- Clarify that size is the number of nodes
- Improve formatting of the command for cluster status
- Mention it may take a minute or two for the new cluster size to propagate

### DIFF
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -132,7 +132,8 @@ Expanding and contracting the size of your cluster
 
 You can easily scale up or down your cluster's size to meet usage demand or to
 save cost when the cluster is not being used. Use the ``resize`` command and
-provide a new cluster size as a command line option ``--size``:
+provide a new cluster size (i.e. number of nodes) as a command line option
+``--size``:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


doc/source/extending-jupyterhub.rst
```diff
@@ -141,8 +142,14 @@ provide a new cluster size as a command line option ``--size``:
...
-To display the cluster's name, zone, or current size, use the command
-``gcloud container clusters list``.
+To display the cluster's name, zone, or current size, use the command:
+
+.. code-block:: bash
+
+   gcloud container clusters list
+
+After resizing the cluster, it may take a couple of minutes for the new cluster
+size to be reported back as the service is adding or removing nodes.
```
yuvipanda: You can find the true count of currently 'ready' nodes by doing:

```
kubectl get node
```

This will report all the nodes and their status (Ready or NotReady)

yuvipanda: This might be useful to add? Otherwise +1!

willingc: Adding... Great suggestion.

yuvipanda:
ty!

PR #127, opened 2017-08-11
add timeout to helm install

henchc:
if you're pulling from a large docker image you might get a ``Error: timed out waiting for the condition``, adding the ``--timeout`` parameters fixes this.


yuvipanda:
Thanks for the patch, @henchc!

PR #128, opened 2017-08-11
add whitelist of usernames to github

henchc:
add config for whitelist of github oauth

### DIFF
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -75,7 +75,7 @@ To add a whitelist of usernames add to the config file under `hub`:
 .. code-block:: yaml
 
      hub:
-         extraConfig:
+         extraConfig: |
              c.Authenticator.whitelist = {"user1", "user2"}
 
 Full Example of Google OAuth2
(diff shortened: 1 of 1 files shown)


doc/source/extending-jupyterhub.rst
```diff
@@ -69,6 +69,14 @@ For more information see the full example of Google OAuth2 in the next section.
           clientId: "y0urg1thubc1ient1d"
           clientSecret: "an0ther1ongs3cretstr1ng"
           callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
+          
+To add a whitelist of usernames add to the config file under `hub`:
+
+.. code-block:: yaml
+
+     hub:
+         extraConfig:
```
yuvipanda: You need a | after extraConfig I think?

willingc: I think both with `|` or without is technically fine since there is only one line and not a block following. Both pass http://www.yamllint.com. I'm inclined to put a `|` if you keep the line below or move it all to one line.

henchc: yes i just checked it both ways and can confirm it works for both, I'll add the `|`

PR #130, opened 2017-08-14
Adding a bit of CPU resource management info

cmoscardi:
Just a code snippet including the `cpu` limitation option.

### DIFF
--- a/doc/source/user-experience.rst
+++ b/doc/source/user-experience.rst
@@ -254,6 +254,7 @@ a very short-term time (e.g. when running a single, short-lived function that
 consumes a lot of memory).
 
 Similarly, you can limit CPU as follows:
+
     .. code-block:: yaml
 
        singleuser:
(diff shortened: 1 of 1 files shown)


doc/source/user-experience.rst
```diff
@@ -253,6 +253,16 @@ the guarantee to allow some users to use larger amounts of RAM for
...
 
+Similarly, you can limit CPU as follows:
+    .. code-block:: yaml
+
+       singleuser:
+           cpu:
+              limit: 500m
+              guarantee: 500m
+              
+This would limit your users to a maximum of 500 "milli-cpu" (so 1/2 of a CPU core), as well as guarantee them that same amount - this measure is defined in the kubernetes docs here: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu .
```
choldgraf: I think as a general rule we're hyperlinking text rather than displaying full links. Could you do something like "this measure is defined in the compute resources page of the Kubernetes docs" and hyperlink "compute resources page"?

choldgraf:
looks like a nice addition! thanks! one small comment in there

at some point we should reorganize this section so it's a little bit more structure (e.g., have a section for RAM, a section for CPU, etc) but I think that's in a future PR :-)

cmoscardi:
Change made. Yeah, seems like a reasonable reorganization!

choldgraf:
thanks for the ++ !

PR #132, opened 2017-08-17
Always install helm with its own SA

yuvipanda:
Even if RBAC isn't enabled, this is the right thing to do.

### DIFF
--- a/doc/source/setup-helm.rst
+++ b/doc/source/setup-helm.rst
@@ -28,7 +28,9 @@ cluster. At the terminal, enter:
 
    .. code:: bash
 
-      helm init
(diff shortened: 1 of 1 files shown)


doc/source/setup-helm.rst
```diff
@@ -28,7 +28,9 @@ cluster. At the terminal, enter:
 
    .. code:: bash
 
-      helm init
+      kubectl --namespace kube-system create sa tiller
```
willingc: Perhaps use the more verbose command for user understanding: `kubectl create serviceaccount --namespace kube-system tiller`

PR #139, opened 2017-08-25
Add quick note about institutional support

yuvipanda:


yuvipanda:
I'm going to merge this quickly just now, but @willingc @choldgraf @minrk y'all probably have opinions that might need follow up PRs :)

willingc:
Looks good. Only suggestion @yuvipanda would be to add a link to the website sponsor page too.

choldgraf:
I think we should at least add the moore + sloan foundations (by invoking the MSDSE) too, no? Good idea btw

PR #146, opened 2017-08-28
To be used as the basis for deployment, docker images must have jupyterhub

arokem:
As I recently discovered.


willingc:
Thanks @arokem.

PR #154, opened 2017-09-08
improve CPU/Mem docs

choldgraf:
should close https://github.com/jupyterhub/helm-chart/issues/69 when we figure it out. I don't know that we want to merge this yet so we can double-check that the instructions are correct

### DIFF
--- a/doc/source/user-experience.rst
+++ b/doc/source/user-experience.rst
@@ -264,11 +264,16 @@ Similarly, you can limit CPU as follows:
 
        singleuser:
            cpu:
-              limit: 500m
-              guarantee: 500m
+              limit: .5
+              guarantee: .5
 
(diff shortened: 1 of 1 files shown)


doc/source/user-experience.rst
```diff
@@ -264,11 +264,16 @@ Similarly, you can limit CPU as follows:
...
-              guarantee: 500m
+              limit: .5
+              guarantee: .5
 
-This would limit your users to a maximum of 500 "milli-cpu" (so 1/2 of a CPU core), as well as guarantee them that same amount - this measure is defined on the `compute resources page <https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu>`_ of the kubernetes docs.
+This would limit your users to a maximum of .5 of a CPU (so 1/2 of a CPU core), as well as guarantee them that same amount.
 
+.. note::
+
+   For a more complete description of how you can provision computational resources (and what values are allowed), see
```
yuvipanda: Let's get rid of this note, since we don't actually allow the same values that kubernetes allows.

choldgraf: cool - or we can just punt this for when I make the config.yml -> doc PR

yuvipanda: The current doc is wrong, so we should remove this note and merge it.

choldgraf:
changed

PR #157, opened 2017-09-17
json schema

choldgraf:
this adds documentation for the json schema that @yuvipanda put together

http://predictablynoisy.com/zero-to-jupyterhub-k8s/schema.html

it's a little bit janky, because it's using a sphinx extension that A. expects `json` objects, and B. only works on python 2.7. But I figure that since this is *just* a documentation website, it's not too big of a deal. I may try to push some upstream commits to the sphinx extension to get it working with python 3, it's not too complex.

LMK what you think

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,3 +6,4 @@ dependencies:
 - sphinx>=1.3.6,!=1.5.4
 - pip:
     - recommonmark==0.4.0
+    - git+git://github.com/choldgraf/sphinx-jsonschema.git@jsonpointer
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,4 +6,4 @@ dependencies:
 - sphinx>=1.3.6,!=1.5.4
 - pip:
     - recommonmark==0.4.0
-    - git+git://github.com/choldgraf/sphinx-jsonschema.git@jsonpointer
+    - git+git://github.com/lnoor/sphinx-jsonschema.git@master
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,4 +6,3 @@ dependencies:
 - sphinx>=1.3.6,!=1.5.4
 - pip:
     - recommonmark==0.4.0
-    - git+git://github.com/lnoor/sphinx-jsonschema.git@master
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,15 +42,6 @@ source_parsers = {
     '.md': 'recommonmark.parser.CommonMarkParser',
 }
 
-# Generate JSON version of the helm-chart
-import yaml, json, sys
-import requests
-resp = requests.get('https://raw.githubusercontent.com/jupyterhub/helm-chart/master/jupyterhub/schema.yaml')
-yamlraw = resp.text
-yamlstruct = yaml.load(yamlraw)
-with open('_static/schema.json', 'w') as ff:
-    json.dump(yamlstruct, ff, sort_keys=True, indent=2)
-
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 #
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -32,7 +32,8 @@ import recommonmark
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.mathjax']
+extensions = ['sphinx.ext.mathjax',
+              'sphinx-jsonschema']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
@@ -40,6 +41,15 @@ templates_path = ['_templates']
 source_parsers = {
     '.md': 'recommonmark.parser.CommonMarkParser',
 }
+
+# Generate JSON version of the helm-chart
+import yaml, json, sys
+import requests
+resp = requests.get('https://raw.githubusercontent.com/jupyterhub/helm-chart/master/jupyterhub/schema.yaml')
+yamlraw = resp.text
+yamlstruct = yaml.load(yamlraw)
+with open('_static/schema.json', 'w') as ff:
+    json.dump(yamlstruct, ff, sort_keys=True, indent=2)
 
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -32,8 +32,7 @@ import recommonmark
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.mathjax',
-              'sphinx-jsonschema']
+extensions = ['sphinx.ext.mathjax']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/.gitignore
+++ b/.gitignore
@@ -135,6 +135,3 @@ ENV/
 
 # Emacs
 .#*
-
-# JSON schema
-doc/source/_static/schema.json
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ ENV/
 
 # Emacs
 .#*
+
+# JSON schema
+doc/source/_static/schema.json
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -4,14 +4,15 @@ Extending your JupyterHub setup
 ===============================
 
 The helm chart used to install JupyterHub has a lot of options for you to tweak.
-This page lists some of the most common changes.
+For a semi-complete list of the changes you can apply via your helm-chart,
+see the :ref:`helm-chart-schema`.
 
 .. _apply-config-changes:
 
 Applying configuration changes
 ------------------------------
 
-The general method is:
+The general method to modify your Kubernetes deployment is to:
 
 1. Make a change to the ``config.yaml``
 2. Run a helm upgrade:
@@ -27,6 +28,9 @@ The general method is:
 3. Wait for the upgrade to finish, and make sure that when you do
    ``kubectl --namespace=<YOUR_NAMESPACE> get pod`` the hub and proxy pods are
    in ``Ready`` state. Your configuration change has been applied!
+
+Below are a few common tasks you can accomplish with modifications to your
+helm chart.
 
 Authenticating with OAuth2
 --------------------------
@@ -69,7 +73,7 @@ For more information see the full example of Google OAuth2 in the next section.
           clientId: "y0urg1thubc1ient1d"
           clientSecret: "an0ther1ongs3cretstr1ng"
           callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
-          
+
 To add a whitelist of usernames add to the config file under `hub`:
 
 .. code-block:: yaml
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -37,6 +37,7 @@ JupyterHub deployment.
 
    extending-jupyterhub
    user-experience
+   schema
    tools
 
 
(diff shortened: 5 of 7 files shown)


willingc:
@choldgraf I think it would be simpler and more effective to use Sphinx's [`literalinclude`](http://www.sphinx-doc.org/en/stable/markup/code.html#includes). A few nice benefits:
- less janky output
- works with Python 3
- simpler build
- supports diffing of helm charts when new versions come out
- can use `:start-after:` and `:end-before:` options to isolate sections of the helm chart if desired.

Something like in a rst file (line nos are optional):

```
Schema - simple
===============

.. literalinclude:: schema.yaml
   :language: yaml
   :emphasize-lines: 12,15-18
   :linenos:
```

cc/ @minrk

screenshot 2017-09-17 19 38 11

choldgraf:
hmmm - I hadn't even considered that...maybe I am getting ahead of myself with the HTML output :-)

I think that's a good idea as well! I can see pros to both sides, but agree that `literalinclude` gets us most of the what we want. I guess what the sphinx extension would get us is the ability to customize the way the HTML is rendered, and I think it'd also let us do things like add links within the description fields etc.

Maybe we go with `literalinclude` for now and swap it out with something fancier one day if a good option comes up? What do you think @minrk and @yuvipanda ?

minrk:
I'm a big fan of `literalinclude`.

yuvipanda:
I'm not a big fan of having people read raw json schema in yaml for
reference documentation. However, I don't think we should switch to using
python2 either. My ideal would have html output since that is quite
important for reference docs, but not be python2

choldgraf:
FWIW - I've got a PR into `sphinx-jsonschema` that will remove the python 2 vs 3 problems...was a library they were depending on (and there's another library that does the same thing and is more well-maintained)

doc/environment.yml
```diff
@@ -6,3 +6,4 @@ dependencies:
 - sphinx>=1.3.6,!=1.5.4
```
willingc: Add `PyYAML` so that RTD can build.

doc/source/conf.py
```diff
@@ -41,6 +42,15 @@
     '.md': 'recommonmark.parser.CommonMarkParser',
 }
 
+# Generate JSON version of the helm-chart
+import yaml, json, sys
```
willingc: Small style suggestion: Since `yaml` is coming from PyYAML and not the standard lib, it's probably better to import on its own line.

yuvipanda: +1, let's import things separately all the time :)

doc/source/conf.py
```diff
@@ -41,6 +42,15 @@
...
+import requests
+resp = requests.get('https://raw.githubusercontent.com/jupyterhub/helm-chart/master/jupyterhub/schema.yaml')
```
yuvipanda: This feels quite bad to me, since this won't get rebuilt every time that changes, there's no tagging, etc.

I think we should either:

1. Put this docs in the helm-chart repo
2. Merge the helm-chart repo into this repo

Before we do this. Quickest thing to do is (1), but I'm also happy to just do (2). I'd rather merge helm-chart repo into this, since the branding as 'z2jh' seems to have stuck a lot more than just 'helm chart'

doc/source/conf.py
choldgraf: oops - that was copypasta from a stackoverflow :-)

doc/source/conf.py
choldgraf: yeah I agree this is hacky...let's do number 2 since it seems like this is the direction in which we're headed anyway. What if we renamed the repository to `jupyterhub-k8s` and then `zero to jupyterhub k8s` would be the documentation for it? What do you think @willingc or @minrk ?

willingc: I'm fine with merging in the helm chart. Is there a technical reason to rename the repo? If not, let's leave it as is.

yuvipanda: Let's merge the helm-chart here and leave the name as is!

willingc: @yuvipanda Any downsides that you can see down the road from merging the helm chart here? I'm not thinking of any but 🤷‍♀️

yuvipanda: Licensing is the only one. The helm-chart repo is Apache2 explicitly because that's the upstream charts repo license, and that's where we want to eventually be. This repo is 3BSD. We can dual license this as apache2 and 3BSD to get around that.

willingc: If @minrk is cool with the dual license, then let's do it. Thanks.

choldgraf: the only reason I was thinking of renaming is that in my mind the helm chart is the tech product, and "zero to jupyterhub k8s" is documentation around that product. So in my mind calling the repo `jupyterhub-k8s` would designate this space as 'jupyterhub development around kubernetes'...and that this could be more than the zero to jupyterhub guide. This is just my own idiosyncratic way of thinking about this...I'm happy to keep the naming as-is

yuvipanda: Most of the work of actually integrating with kubernetes (for some definition of 'actually') is in kubespawner repo, so I think we're ok!

choldgraf: sounds good to me :-)  lemme know if/when you merge in helm-chart and then I can rebase / update this PR

willingc: @choldgraf Let's try to keep things simple by not renaming. There's enough churn and changes in other places. Let's not introduce more than necessary.

choldgraf: how about we rename the repo to "zero to zero to jupyterhub k8s for kubernetes"? :-)

choldgraf: (jk, current name sounds good to me)

minrk:
While json schemas can be nice for producing validators, I don't think they are particularly useful for generating user-facing documentation. Other formats tend to be better suited to describing fields for humans.

choldgraf:
@minrk could you suggest an alternative that would work better and not require a lot of duplication of effort?  IMO the rendered schema page doesn't look too bad as a reference we can point to.

yuvipanda:
@minrk @choldgraf I think for reference documentation of a yaml file, JSON Schema seems like the right place. It's similar to swagger / openAPI specs, which internally use JSON Schema. We already use those for CHP for example: https://github.com/jupyterhub/configurable-http-proxy#using-the-rest-api.

There's also general support for adding JSON Schema based validation upstream in helm (https://github.com/kubernetes/helm/issues/2431) and I think we should continue using it for both validation & reference documentation.

choldgraf:
in my comment, I meant purely for documentation. It sounds like json schemas are the right way to go for validation, but it sounded like @minrk had an idea for something different specifically for documentation. I'm just saying that if it is easy to transform the json schema into another format and build documentation off of this at documentation build-time, I can look into making this happen. Though IMO the current approach is a good enough first step.

minrk:
> Though IMO the current approach is a good enough first step.

Yeah, I think it's an AOK place to start. I don't think it's a long term solution.

I don't actually know of a good solution that doesn't duplicate effort. But I've generally found that the goal of avoiding duplicated documentation effort in favor of automatic generation ends up being done at the expense of usefulness of the documentation to users. I think thorough, hand-written (even if generated from a structure that's *not* designed for validation) docs are really the only way to get good documentation of these things. For example, language like `enum` in json schemas should not show up in user documentation.

I think the best documentation for a yaml file is probably a completely filled out (and ideally valid) yaml file with every option defined and explained. If you could generate this from a json schema or some other source, that would be cool, but I don't think the representation we're getting now is very informative to users. It's mainly useful to people who already know and understand what a json schema is.

willingc:
I agree with @minrk that we should look for a better long term solution. I'm hoping to ping Erik Rose and some others to see if they are using something better at Mozilla or OpenStack.

choldgraf:
Good points from both of you 

> I think the best documentation for a yaml file is probably a completely filled out (and ideally valid) yaml file with every option defined and explained

Isn't that the goal of this JSON schema? Yuvi's done a quick first-pass at providing a human-readable description for some of the most common chart fields people need to use, and we were going to grow that documentation as time went on, but wanted to get an MVP in there sooner than later.

I'm happy with a different approach for the docs, though it feels to me like the schema is a natural place to put descriptions etc. In terms of rendering, I agree that there is some information in the output schema that isn't super human-useful, though things like min/max values, data type, etc would be useful to some people (e.g., the person who had a hard time figuring out if they should give floating point or integer numbers to the `cpu` flag)

willingc:
I propose that we do this initially in separate PRs:

- edit the prose comments that Yuvi made in the yaml file.
- then take the edited yaml file and use `literalinclude` to have a simple page of the yaml (temporarily as reference)
- create a user focused page that interleaves prose with yaml or json AND/OR create our own extension to do what we want

minrk:
> Isn't that the goal of this JSON schema?

It isn't. JSON Schemas are not designed for readability by humans. A schema file is not a valid file *of* the schema it describes, especially when the format it describes isn't JSON.

It's entirely possible that we can write our own generator that takes our json schema and emits well commented yaml, or yaml segments interspersed with rst or markdown explanations. I think this is probably possible for us to do as a bespoke, single-purpose thing where I don't think it's possible to do in general for all json schemas. That would still mean that the place to *store* this info is a JSON schema. That would still make it harder to do things like including examples that would be good for docs but not appropriate for a validation-focused format like json-schema.

Right now there's a problem with this as it is because the comments contain some markdown, but the export doesn't know that so the markdown isn't rendered correctly. If we wrote our own transform where our own assumptions are baked into the transform (e.g. description can contain markdown), then

choldgraf:
I guess this is partially going to come down to opinion, as I find this

http://predictablynoisy.com/zero-to-jupyterhub-k8s/schema.html

pretty readable (esp if we include rST in the descriptions) and much more readable than YAML. But I'm not going to block on this if the team prefers a different approach.

re: editing the content of the schema itself, I agree that should be done in a different PR...this is just supposed to be infrastructure to expose it to users on the documentation.

willingc:
@choldgraf The only thing that is blocking this is using the jsonschema sphinx extension. Why don't you remove it and then the rest of this could be merged. 

It's not just a matter of opinion. The PR relies on an extension that the maintainer states:

> To put it crudely,the current implementation solves my problem.

The extension also renders the blocks oddly and the labels extend over boundaries. It looks unprofessional. If you want to correct that then I'm more inclined to merge this.

choldgraf:
ok - I'll remove the extension...I agree it's not a long-term solution, I had thought from 

> Update dependencies in environment.yml. This is fine to merge after that. We should find a better long term solution since the layout isn't particularly user friendly and the elements are overrunning the chart boundaries.

that we were going to keep it only as a stop-gap measure until we come up with something better for displaying the information. Sorry if I'm getting confused about the plan here :-/

yuvipanda:
I wanted something more like http://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.spawner.KubeSpawner rather than the current output. I agree with @minrk that we shouldn't have things like 'enum' and 'object' in user facing documentation. All I really want is 'name' and 'description' - the 'second yaml file with all options filled out' is just  the name and description fields of the JSON schema IMO. We just need to make sure our descriptions are actual documentation (which I've tried to right now) - this way we only have to edit one file, and when we do get validation validation failures will produce useful output.

So it looks like a plan of action could be:

1. Remove the jsonsphinx extension right now from this PR
2. Work on a way to generate a HTML page like http://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.spawner.KubeSpawner as a separate PR

I don't want to expose users directly to the YAML schema file.

How does this sound, @choldgraf, @minrk, @willingc?

(and thank you for all your work, @choldgraf!)

choldgraf:
latest push removes the sphinx extension, this is what the page looks like now:

http://predictablynoisy.com/zero-to-jupyterhub-k8s/schema.html

I don't think it'd be hard to write something that just parses the file and renders nice-looking rST based solely on a few subfields. I'm happy to work on that if folks think it's a good solution.

yuvipanda:
@choldgraf I don't think we should expose that raw blob of JSON to people, so I'd rather us not do that at all until we have nice HTML based output.

willingc:
Thanks @choldgraf. I'm going to merge this and we can iterate the schema page in a separate PR. Thanks all for your patience.

minrk:
@choldgraf I think that's a good idea!

PR #159, opened 2017-09-18
Doc typo

gabelev:


choldgraf:
I think it's actually supposed to be "machinery"...you could add that in? Thanks for the contribution! :-)

choldgraf:
thx for the addition!

PR #166, opened 2017-09-19
Edit readme to add helm charts

willingc:
This is an initial pass at updating the README. I'm going to self merge. My expectation is that there will be future PRs on the README, and this PR is simply to notify users of the current content.


choldgraf:
just saw this - thx @willingc !

PR #169, opened 2017-09-19
move module

choldgraf:
moves the notebook inside of `doc` so that the root folder is less-cluttered...


choldgraf:
(everything should work the same, it just moves the notebook to a different place)

willingc:
@choldgraf What do you want to do with the `environment.yml` in the root directory? Is it still needed? If so can it be renamed to `environment_binder.yml` and perhaps moved along with the other files?

doc/source/cost.rst
```diff
@@ -38,7 +38,7 @@ you.
 
 .. raw:: html
 
-   <a target="_blank" href="http://beta.mybinder.org/v2/gh/jupyterhub/zero-to-jupyterhub-k8s/master?filepath=draw_function.ipynb">
+   <a target="_blank" href="http://beta.mybinder.org/v2/gh/jupyterhub/zero-to-jupyterhub-k8s/move_ntbk?filepath=doc/ntbk/draw_function.ipynb">
```
yuvipanda: Is this a branch it's pointing to? If so you should replace that with either a commit hash or a tag :)

choldgraf: that was actually just to double-check that this PR branch works...I'm gonna switch it to a tag w/ the next push!

choldgraf:
fixed the binder link...I think we should be good to go if @yuvipanda is cool w/ current state

PR #186, opened 2017-09-20
Rename a bunch of stuff that was pointing to helm-chart still

yuvipanda:
Remaining references are still valid


willingc:
Thanks @yuvipanda

PR #165, opened 2017-09-19
[WIP binder] use JupyterHub API to launch containers

minrk:
re-issue of https://github.com/jupyterhub/helm-chart/pull/80

pending jupyterhub/binderhub#110

PR #188, opened 2017-09-20
[0.5.x] support loading service tokens

minrk:
the main part of #165

hub.services is stored in configmap. Token is loaded as SERVICE_TOKEN_[NAME] via hub-secret

For example:

```yaml
hub.services:
  binder:
    admin: true
    apiToken: 'generated-secret'
```

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -25,6 +25,9 @@ spec:
       - name: config
         configMap:
           name: hub-config
+      - name: secret
+        secret:
+          secretName: hub-secret
       {{ if eq .Values.hub.db.type "sqlite-pvc" }}
       - name: hub-db-dir
         persistentVolumeClaim:
@@ -39,6 +42,8 @@ spec:
         volumeMounts:
           - mountPath: /etc/jupyterhub/config/
             name: config
+          - mountPath: /etc/jupyterhub/secret/
+            name: secret
         {{ if eq .Values.hub.db.type "sqlite-pvc" }}
           - mountPath: /srv/jupyterhub
             name: hub-db-dir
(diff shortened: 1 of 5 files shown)


jupyterhub/templates/hub/secret.yaml
```diff
@@ -8,3 +8,8 @@ data:
   {{ if .Values.hub.cookieSecret }}
   hub.cookie-secret: {{ .Values.hub.cookieSecret | b64enc | quote }}
   {{- end }}
+  {{ range $key, $value := .Values.hub.services -}}
+  {{ if $value.apiToken }}
+  services.token-{{$key}}: {{ $value.apiToken | b64enc | quote }}
```
yuvipanda: I'd prefer `services.token.{{$key}}`

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -70,6 +70,15 @@ spec:
             secretKeyRef:
               name: hub-secret
               key: proxy.token
+        {{ range $key, $value := .Values.hub.services -}}
```
yuvipanda: A more robust way to do this would be to mount the secret as a volume (similar to how we do for configmap) and just read it with something similar to get_config we currently use in jupyterhub_config.py. This will break with strange errors if there are non alphanumeric characters in service name, for example.

minrk: Do you mean put the whole `hub.services` in a secret volume instead of splitting like this?

yuvipanda: No, we could mount hub-secret into a volume, similar to us mounting the configmap. So all secrets would be in /etc/jupyterhub/secrets are somesuch, and we can have a `get_secret` method in jupyterhub_config.py similar to `get_config`. 

I'm also happy to make that change as a separate PR after this.

minrk: Roger. I think that's enough for me to give it a go.

minrk: OK, I think I got it. I added a `src` for `get_config` to fetch from either `config` or `secret` and put the whole services dict in the secret since it contains secrets and is much simpler than splitting it between configmap and secrets. Let me know if this implementation makes sense. I tested it and it does work, at least.

yuvipanda:
This is mostly great. However, I'd still like all non-secret config to be in the configMap. That complicates the templates a little bit, but is worth it I think...

minrk:
ok, now only tokens are back in the secret and it's mounted as a volume, loaded via `get_secret` similar to `get_config`.

yuvipanda:
awesome!

PR #189, opened 2017-09-20
remove binderhub chart

minrk:
that will be moved over to binderhub itself in another PR


minrk:
Closing in favor of deleting only in v0.5.x

PR #191, opened 2017-09-20
[0.5.x] base singleuser-sample on docker-stacks

minrk:
Use smallest base-notebook with placeholder for installation of packages.

JUPYTERHUB_VERSION is a build arg and specified in the Makefile

includes #190 to avoid adding unused build arg to binderhub image

### DIFF
--- a/images/singleuser-sample/Dockerfile
+++ b/images/singleuser-sample/Dockerfile
@@ -4,5 +4,5 @@ FROM jupyter/base-notebook:ae885c0a6226
 
 # pin jupyterhub to match the Hub version
 # set via --build-arg in Makefile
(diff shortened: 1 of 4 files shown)


images/singleuser-sample/Dockerfile
```diff
@@ -1,21 +1,8 @@
...
-USER jovyan
-
-RUN python3 -m venv /srv/venv
-RUN /srv/venv/bin/pip3 install --no-cache-dir notebook ipykernel jupyterhub==0.8.0b5
-
-
-WORKDIR /home/jovyan
-
-EXPOSE 8888
+# pin jupyterhub to match the Hub verion
```
yuvipanda: version

minrk:
Updated

PR #195, opened 2017-09-20
[0.5.x] add cull.users flag

minrk:
for culling idle users in addition to their servers (e.g. binder)

cull_idle_users script updated from jupyterhub@0.8.0rc1 to get the flag

depends on #188


images/hub/jupyterhub_config.py
```diff
@@ -194,25 +203,36 @@ def generate_user_name(spawner):
...
-                '--cull_every=%s' % cull_every
-            ]
-        }
+    cull_cmd = [
+        '/usr/bin/python3',
+        '/usr/local/bin/cull_idle_servers.py',
+        '--timeout=%s' % cull_timeout,
+        '--cull-every=%s' % cull_every,
     ]
+    print(get_config('cull.users'), file=sys.stderr)
```
yuvipanda: we should get rid of this debug line

PR #197, opened 2017-09-20
Don't include raw YAML on RTD

yuvipanda:
Neither raw YAML nor raw JSON Schema is human readable, so we should
refrain from showing it until we have better human readable output.


yuvipanda:
The diff is large because I also converted windows style line endings to unix style ones to be consistent with rest of repo.

yuvipanda:
superseeded

PR #198, opened 2017-09-20
nicer schema page

choldgraf:
This adds some python to the end of `conf.py` that generates rST headers based off of the JSON schema. It gives us more control over the information that we expose on the `schema.rst` page, and lets us style it however we like.

e.g. of how it looks now:

http://predictablynoisy.com/zero-to-jupyterhub-k8s/schema.html

thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,7 +103,8 @@ instance/
 .scrapy
 
 # Sphinx documentation
-docs/_build/
+doc/build/
+doc/source/_static/schema.txt
 
 # PyBuilder
 target/
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,3 +6,4 @@ dependencies:
 - sphinx>=1.3.6,!=1.5.4
 - pip:
     - recommonmark==0.4.0
+    - pyyaml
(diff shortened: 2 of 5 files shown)


minrk:
That looks really nice! Do you think we should be inlining rst instead of markdown in the schema comments or handling the markdown somehow?

choldgraf:
should supersede #197 if we're OK w/ it

choldgraf:
yeah I think so - otherwise sphinx can get confused (e.g., adding ``` sometimes makes sphinx think you're trying to create a section header)

choldgraf:
oh, and @willingc we should double check that the header order is correct...I know you sent me a list of this earlier but I couldn't figure out which PR it was attached to...LMK if I should change the ordering that's there

yuvipanda:
This is awesome! \o/

I'd like to keep the markdown if possible, mostly because I would like to upstream bits of this into helm, and that's going to have to be markdown. But it might be easier to switch to rst for us anyway, since we're using our own custom converter.

minrk:
We've already loaded sphinx markdown plugins, so there might be a way to use that to do simple md->rst internally (or md->sphinx nodes, I'm not sure).

willingc:
Since the markdown will only be in descriptions, we can likely filter out the descriptions and use recommonmark to convert the md to rst.

PR #199, opened 2017-09-20
support markdown in schema.yaml

choldgraf:
this is a janky solution to allow us to embed markdown in the schema so that it's rendered nicely. The problem is that we can't render markdown from rST, so we're doing the reverse. We're storing the "text" part of `schema.md` in a text file that *won't* be rendered with sphinx (`schema.txt`). Then this is combined with the markdown version of the schema and `schema.md` is generated at build.

What do folks think?


yuvipanda:
+1 on it would be nice to use markdown in schema.yaml - if we upstream it, it'll be md rather than rst, since rst isn't widely used outside the python ecosystem.

However, I'm also happy to use rst for schema.yaml if others feel strongly about it!

doc/source/conf.py
```diff
@@ -22,6 +22,8 @@
 # sys.path.insert(0, os.path.abspath('.'))
 
 import recommonmark
+from recommonmark.transform import AutoStructify
+import numpy as np
```
minrk: unused numpy import?

minrk:
I think this is a good solution. @willingc?

willingc:
I think what Chris has works. Down the road, we may want to refactor into a Sphinx extension so that it can be reused easily and doesn't clutter the conf.py. For now, it works and looks good so :+1:

PR #201, opened 2017-09-21
define proxy.secretToken and NodePort in minikube-config

minrk:
so it can be used unmodified for testing with a predictable URL


willingc:
Thanks @minrk

PR #205, opened 2017-09-25
Remove makefile

willingc:
@yuvipanda I see you bumped to rc2. Did you want to update the Makefile too?


yuvipanda:
I'm actually tempted to remove the makefile. We don't actually use it anywhere, and build.py does all the work in our CI process.

willingc:
Let's remove it.

willingc:
I could push to this PR.

yuvipanda:
w00t! ty, @willingc

PR #206, opened 2017-09-26
add handling for servername derived from named-servers

Analect:
This is intended to work in conjunction with changes to kubespawner implemented by https://github.com/jupyterhub/kubespawner/pull/83.  I have tested it against hub 0.8.0rc2 and the [`handle-named-servers`](https://github.com/Analect/kubespawner/tree/handle-named-servers) feature branch as per the PR above and it appears to work well.

Within kubespawner, in the case of a named-server getting spawned via the API like:
`curl -v -X POST -H "Authorization: token 0a4991e94510497cb68c6b14392ce665" "http://myjupyterhub.com/hub/api/users/testuser/servers/test1"`

...then a pvc (and requisite pv) gets created as `claim-testuser-test1`.

If a user simply launches a default server via the API like:
`curl -v -X POST -H "Authorization: token 0a4991e94510497cb68c6b14392ce665" "http://myjupyterhub.com/hub/api/users/testuser/server"`

...then a pvc gets created as `claim-testuser` (the existing behaviour).  The `servername` variable only gets a value where a named-server is spawned.  Otherwise, it is set to a blank and therefore doesn't alter the pvc name in the case where a user spawns a default server.


yuvipanda:
This looks fine to me, but we'll have to wait until theappropriate kubespawner PR is merged, and then also bump the kubespawner version here before we merge this. Otherwise it'll cause spawn failures.

Analect:
@yuvipanda  ... maybe this can be merged given that https://github.com/jupyterhub/kubespawner/pull/83 has already been merged. Tks.

yuvipanda:
@Analect can you bump the kubespawner version in the Docker image up too? Otherwise this will fail since it'll still be using an older version of kubespawner.

I can do this later if you don't get to it first :)

Analect:
OK ... I bumped the kubespawner commit hash.  Both dockerfiles already pointed to version 0.8 of hub.  I changed the base-notebook tag to latest.  As per this [commit](https://github.com/jupyter/docker-stacks/commit/2404602e574f862ced5150db26a38bab869a9ee5), that base image includes version 0.8 of hub, so extra step of reinstalling the hub in that image may be overkill.

yuvipanda:
ty for your patience, @Analect! This is good to go...

PR #209, opened 2017-09-29
Create a service account specifically for hub

yuvipanda:
This is more secure than granting permissions to the default sa


yuvipanda:
This is still off by default, since pre-puller doesn't work with RBAC enabled.

PR #212, opened 2017-10-04
Fix ingress routing bug

cmoscardi:
Doh.

### DIFF
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -16,7 +16,7 @@ spec:
   rules:
     - http:
         paths:
-          - path: /
+          - path: /*
             backend:
               serviceName: proxy-public
               servicePort: 80


yuvipanda:
I don't think this should make any difference... The paths are considered prefixes no?

cmoscardi:
@yuvipanda : 
https://t5.thedataincubator.co
https://t6.thedataincubator.co

These are two clusters in identical configuration states save for this `*` character (and the hostname, of course). It's definitely the issue. 
Also FWIW, `kube-lego` does use the `*` character for its routing, which was what gave me the idea... 
```spec:
  rules:
  - host: t6.thedataincubator.co
    http:
      paths:
      - backend:
          serviceName: kube-lego-gce
          servicePort: 8080
        path: /.well-known/acme-challenge/*
      - backend:
          serviceName: proxy-public
          servicePort: 80
        path: /
```

cmoscardi:
I don't have much sense of whether this is platform-specific... however, `http://` does work? Could it be a GKE thing? 

http://t6.thedataincubator.co/
http://t5.thedataincubator.co/

willingc:
@yuvipanda Can you review this for @cmoscardi? Thanks!

yuvipanda:
@cmoscardi it's possible this is a GKE thing, in which case it might / could break on nginx ingresses.

Did you manage to try out nginx ingress? The GKE ingress provider just doesn't seem to work very well for our use case...

cmoscardi:
@yuvipanda I cannot get nginx ingress working for the life of me. Are others having success with it?

willingc:
Hi @cmoscardi, When you get a chance perhaps you and @yuvipanda (he'll be checking messages less frequently for the remainder of the week), take a look at this [configuration from nginx](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/multiple-ingress-controllers).  It looks as if an annotation may be needed to use nginx ingress instead of the GKE default.

cmoscardi:
@willingc yes, that's what I'm using. Here's the relevant section from my config:
```
ingress:
  enabled: true
  annotations:
    kubernetes.io/tls-acme: "true"
    kubernetes.io/ingress.class: "nginx"
  host: t7.thedataincubator.co
  https:
    enabled: true
    type: "kube-lego"
```
I'm also doing a helm install for kube-lego (as well as nginx ingress, though I'm not sure if that's necessary)

willingc:
@cmoscardi To clarify, are you using the [nginx/kubernetes-ingress](https://github.com/nginxinc/kubernetes-ingress) or the [kubernetes/ingress-nginx](https://github.com/kubernetes/ingress-nginx)?

willingc:
Additional configuration example for nginx/kubernetes-ingress https://www.nginx.com/products/nginx/kubernetes-ingress-controller/

cmoscardi:
I'm using this: 
https://github.com/kubernetes/charts/tree/master/stable/nginx-ingress

(via the `helm install` command)

Which appears to be kubernetes/ingress-nginx .

Thus far, nothing beyond that. Since I managed to hack the GCE ingress to work, I'm using that for our deploys at the moment, as it does seem more clear / stable than the nginx setup.

willingc:
Sharing a helpful video with all re: nginx ingress and a fanout ingress for microservices: https://youtu.be/HwogE64wjmw

yuvipanda:
Thanks for your patience with this, @cmoscardi. I've spent all day playing with various ways of setting up HTTPS easily in z2jh. #229 is the result. With this, all you have to do is add to your config.yaml:

```
proxy:
   hosts:
      - myhostname.com
```

And make sure myhostname.com is pointing to the external IP exposed by the proxy-public service.

I've to finish documenting it and test it, and make sure it works in cases where there's no externalIP provider.

yuvipanda:
Note that that PR is a breaking change - there's no more top level 'ingress' key in config.yaml. I'll make sure to document this too tomorrow.

PR #214, opened 2017-10-05
Remove explicit mention of kubernetes version from gke

yuvipanda:
This keeps changing all the time, plus we don't care now that
the default version is 1.7+


choldgraf:
LGTM

PR #219, opened 2017-10-09
Call reference documentation reference & not as schema

yuvipanda:
The schema is an implementation detail, what this is doing is
producing reference docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,7 +104,7 @@ instance/
 
 # Sphinx documentation
 doc/build/
-doc/source/schema.md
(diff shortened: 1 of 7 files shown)


yuvipanda:
@willingc I renamed and did a more thorough job of this, but I haven't had a chance to test it yet. Shall do tomorrow!

yuvipanda:
@willingc @choldgraf I tested this and it looks good to me!

willingc:
Thanks @yuvipanda and @choldgraf 👍

PR #224, opened 2017-10-11
Support for CILogon auth

zonca:
Added support for CILogon auth in the configuration yaml.

There is an issue which is unrelated to k8s with CILogon and Jupyterhub 0.8, see https://github.com/jupyterhub/oauthenticator/issues/132.

I implemented a quick workaround in Jupyterhub (https://github.com/jupyterhub/jupyterhub/pull/1474) and tested that CILogon is working properly in my setup.
A proper fix should be ready soon see https://github.com/jupyterhub/oauthenticator/issues/133, I can then update this PR.

### DIFF
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -74,6 +74,18 @@ For more information see the full example of Google OAuth2 in the next section.
           clientSecret: "an0ther1ongs3cretstr1ng"
           callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
 
+**CILogon**
+
+.. code-block:: yaml
+
+      auth:
+        type: cilogon
+        github:
+          clientId: "y0urc1logonc1ient1d"
+          clientSecret: "an0ther1ongs3cretstr1ng"
+          callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
+
+
 To add a whitelist of usernames add to the config file under `hub`:
 
 .. code-block:: yaml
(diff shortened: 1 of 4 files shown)


yuvipanda:
@zonca ty! I'll wait till the upstream CILogon issue is fixed before merging this :) This otherwise lgtm!

yuvipanda:
@zonca can you also add some docs for this, perhaps in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/extending-jupyterhub.rst#authenticating-with-oauth2?

yuvipanda:
@willingc @choldgraf I am thinking we should have a reference page specifically for the various authenticators, with detailed notes on how to use them. Thoughts?

choldgraf:
I think that's a great idea

zonca:
I think we should mostly reference the `oauthenticator` documentation, just
describe what is specific to zero-to-jupyterhub

willingc:
Thanks @zonca for the PR and all the troubleshooting for CILogon. 

As for documenting authenticators, let's discuss next week in person. I'm inclined to agree with @zonca re: oauthenticator. I'm not sure that we should have detailed installation info on each authenticator as I think it will be difficult to maintain and also may stress out a new user which was the original intent of the guide. I do think that it would be great to have a table that refers users to other authenticators as available but perhaps leave the details to the individual authenticator docs.

zonca:
@minrk can you please release another tagged version of `oauthenticator` so we can use that for `zero-to-jupyterhub` (we need the CILogon fix)?

minrk:
@zonca oauthenticator 0.7.2 is out with the cilogon fixes

zonca:
thanks @minrk 
I've added `oauthenticator` 0.7.2 and an example in the docs. It is basically the same as Github.

yuvipanda:
Thank you!

PR #228, opened 2017-10-16
Allow mutliple extraEnv entries

rschroll:
When I tried to specify multiple entries in extraEnv, the hub pod would get stuck in a crash loop.  The logs indicated a problem parsing `/etc/jupyterhub/config/hub.extra-config.py`.  As best I could guess from the error message, it was getting everything on a single line.

This change in the templating seems to solve the problem.

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -111,9 +111,9 @@ data:
   {{- end }}
   {{ if .Values.singleuser.extraEnv -}}
   singleuser.extra-env: |
-    {{ range $key, $value := .Values.singleuser.extraEnv -}}
+    {{ range $key, $value := .Values.singleuser.extraEnv }}
     {{ $key | quote }}: {{ $value | quote }}
-    {{- end }}
+    {{ end }}
   {{- end }}
   token.cookie_secret: {{ .Values.hub.cookieSecret | quote }}
   hub.base_url: {{ .Values.hub.baseUrl | quote }}


yuvipanda:
Thank you for your patch! This looks good to me! I'll merge it after Travis passes :)

rschroll:
As an editorial, I find it a bit odd to convert from yaml to plain text and back to yaml.  Is there a reason for this?  Had it remained yaml throughout, we would have tripped over this problem earlier, when creating the configmap.yaml file.  This convention is used in a few places, so it may be worth reconsidering it in general.

(I'm sure there was a reason to do it this way, but I don't understand the various layers of JupyterHub, Kubernetes, and Helm well enough to even figure out where to start looking for that reason..)

yuvipanda:
The only reason is that the `toYaml` function in Helm templates is... very trippy. It sometimes seems to indent just the first line of the resulting yaml, thus producing synctatically invalid YAML.

One solution you'll see in places is just using `toJson` instead. I helped add that to helm, and know it produces a single line only - so we can indent it properly. Since all valid JSON is valid YAML, this also is valid YAML. it's hacky...

So our possible solutions are:

1. Debug the interaction between toYaml and indent, and figure out how to fix it.
2. Switch everything to toJson

I agree the current methodology sucks.

yuvipanda:
If you want to try either (1) or (2) for this patch, that's awesome too! It's an easy way to get more patches in :D

yuvipanda:
Awesome. I'll just merge this then.

Thank you for the patch :)

manics:
I thought I'd mention a use case for keeping the backwards compatibility in future too. If `extraEnv` is a dict you can easily merge a public config in a git repo with private variables set on the command line during deployment, e.g.
```yaml
hub:
  extraEnv:
    OAUTH2_AUTHORIZE_URL: "https://login.elixir-czech.org/oidc/authorize"
    OAUTH2_TOKEN_URL: "https://login.elixir-czech.org/oidc/token"
    OAUTH_CALLBACK_URL: https://example.org/hub/oauth_callback
    OAUTH_CLIENT_ID: xxx
```
and set the private variables separately during deployment: `--set hub.extraEnv.OAUTH_CLIENT_SECRET=XXX`, but to my knowledge you can't do this when `extraEnv` is a list.

PR #229, opened 2017-10-17
Add first class support for HTTPS

yuvipanda:
I investigated various ways of providing HTTPS to the hub. I used the following principles
in guiding this implementation:

0. As people building software, it is our responsibility to make sure it is easy for 
    our users to install it in the most secure way possible
1. Prefer ease of use for folks using HTTPS vs folks using just HTTP
2. HTTPS should be very easily added. Only requirement should be getting DNS
    sorted, and nothing more.
3. Write as little code as possible!

With this design, all you need to do is specify a list of hosts that the hub should be
accessible from as `proxy.hosts`, and nginx + kube-lego will do the rest!

This also works as is when HTTPS is not used. There's negligible extra overhead
from the nginx + kube-lego when we aren't using HTTPS, but that's acceptable for
the drastic simplification it brings us.

This makes some breaking changes in where things go in values.yaml. Specifically,
the old `ingress` top level key no longer exists. A lot of the keys under 'proxy' have also
been restructured.

Things to do:

- [x] Make sure this still works for people who just want an ingress that is to be 
      consumed by other ingress controllers (useful for bare metal)
- [x] Test that this works fine without HTTPS everywhere
- [ ] <s>Write detailed notes in CHANGELOG about the breaking changes & why we
       made them.</s>
- [ ] <s>Fill out schema.yaml with new keys</s>
- [x] Set up appropriate RBAC rules for proxy too


yuvipanda:
@choldgraf @willingc I added a very skeleton bit of documentation, I suspect it needs to be improved a lot before being merged!

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,25 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
      size for the workshop. This workflow also helps you avoid scrambling on
      the workshop day to set up the cluster and JupyterHub.
    - **After the workshop:** The cluster can be deleted.
+
+
+Setting up HTTPS
+---------------
```
willingc: Underline is one - too short

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,25 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+
+Zero to JupyterHub makes setting up HTTPS quite easy.
+
+1. Buy a domain name from a registrar. Pick whichever one you want!
```
willingc: Let's remove `!`

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,25 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+2. Create an `A record` from the domain you want to use, pointing to the
+   external IP provided to the `proxy-public` service.
+3. Wait for the change to propogate. This can take several minutes. Wait until
```
willingc: propagate

Propagation can take from several minutes to several hours.

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,25 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+   you can type in the name of the domain you bought, and it shows you the hub.
+4. Add the following to your config.yaml:
```
willingc: Put config.yaml in double backticks ``

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,25 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+
+   ```yaml
+   proxy:
+     hosts:
+       - <your-domain-name>
+   ```
+
+   And run helm-upgrade.
```
willingc: Let's make "run helm-upgrade" its own step.

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
      size for the workshop. This workflow also helps you avoid scrambling on
      the workshop day to set up the cluster and JupyterHub.
    - **After the workshop:** The cluster can be deleted.
+
+
+Setting up HTTPS
+----------------
+
+Zero to JupyterHub makes setting up HTTPS quite easy.
```
choldgraf: Maybe a sentence with a quick reason you might want to do this. 

Also shouldn't we call this the JupyterHub helm chart, instead of Zero to JupyterHub? Or do we think of the helm chart *as* z2jh?

E.g. "HTTPS allows you to XXX. The JupyterHub helm chart makes setting up HTTPS quite easy."

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+
+1. Buy a domain name from a registrar. Pick whichever one you want
```
choldgraf: end period.

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+2. Create an `A record` from the domain you want to use, pointing to the
```
choldgraf: double backticks

choldgraf: "point" instead of "pointing", and make it a new sentence.

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+   external IP provided to the `proxy-public` service.
```
choldgraf: link to the command to `get svc` this IP address?

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+3. Wait for the change to propagate. Propagation can take several minutes to
+   several hours. Wait till you can type in the name of the domain you bought,
+   and it shows you the hub.
```
choldgraf: ~it shows you the hub~ you see the JupyterHub landing page

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+4. Add the following to your ``config.yaml``:
```
choldgraf: ~your~

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+
+   ..code-block:: yaml
+
+     proxy:
+       hosts:
+         - <your-domain-name>
+
+5. Apply the config changes by running ``helm upgrade``.
```
choldgraf: link to helm upgrade command?

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+6. Wait for about a minute, now your hub is HTTPS enabled!
```
choldgraf: Related to the comment at the top, but we should say what this means to a deployer

doc/source/extending-jupyterhub.rst
```diff
@@ -176,3 +176,26 @@ report the current ``Ready/NotReady`` status of all nodes in the cluster.
...
+   several hours. Wait till you can type in the name of the domain you bought,
+   and it shows you the hub.
+4. Add the following to your ``config.yaml``:
+
+   ..code-block:: yaml
+
+     proxy:
+       hosts:
+         - <your-domain-name>
+
```
cmoscardi: Also need 
```
https:
  auto:
    letsEncryptContactEmail: <your-email>
```

Which is validated upon attempting to deploy, but isn't included in this README

yuvipanda: ooo, great catch! ty!

Did this work for you otherwise?

cmoscardi: Looks like the hub is in a `CrashLoopBackOff` state with `v0.4` of the `k8s-hub` image.

(I just took the stock `values.yaml` from this branch to try to make it work).

My attention is being pulled by a fire, but I assume this is the same thing as before where I just need to put in a more recent hub image.

doc/source/extending-jupyterhub.rst
rpwagner: HTTPS allows you to secure your authentication flows.

That's the specific need we have for Globus Auth, since it requires an HTTPS callback URL.

cmoscardi:
Interestingly, redeploying seems to kill the ingress - but then another redeploy recreates the ingress. Sort of. I still can't get either the lego or the jupyterhub ingress to connect to the controller that's supposed to run it.

cmoscardi:
So I've been following along with the deployment guide here...
https://github.com/kubernetes/ingress-nginx/blob/master/deploy/README.md ...
And it looks like you're doing things somehwat differently than what they're doing for a GKE deploy, which includes a Service (that takes care of turning on some kind of GKE load balancer).

However, in the default case of just running this chart, it looks like the ingress IP gets set to the IP for one of the nodes (the one where the proxy nginx pod lives). However, SSHing into that node and just looking at port usage, nothing is listening on port 80 (let alone 443). I did set a firewall rule to open up 80 and 443, but there's no point when the node itself doesn't have anything listening on that port.

I'm continuing to futz with it, including adding another service as per the ingress-nginx deployment guide (which seems to spin up a GCP load balancer) but this really Just Works for you on GKE @yuvipanda ?

doc/source/extending-jupyterhub.rst
yuvipanda: I think of the chart as z2jh, mostly by dint of the fact that that's what I've been calling it in talks and what not...

doc/source/extending-jupyterhub.rst
yuvipanda: You can not create an A record without it pointing anywhere, so I'm going to leave this one as is for now unless you have strong opinions :)

doc/source/extending-jupyterhub.rst
yuvipanda: how do I do so? I've never managed to figure out how to use rst properly :'(

yuvipanda:
@cmoscardi yeah, you don't need to install nginx-ingress or anything. I've done some renaming so things are a bit more obvious now. You should pretty much forget any other HTTPS instructions I might've provided an just use this one specifically instead. We're bundling nginx ingress + kube-lego here, and making sure that they only act on our internal ingresses (that is now in ingress-internal.yaml).

doc/source/extending-jupyterhub.rst
yuvipanda: I've added some text here that's vague and buzzworddy, since I don't think we can summarize why you care about HTTPS in sufficient detail without being vague and buzzwordy. I think it's ok for us though.

yuvipanda:
This is ready to be reviewed again!

1. Make sure helm is installed in your cluster
2. Prepare a config.yaml per https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub.html#prepare-configuration-file
3. Install it almost like in https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub.html#install-jupyterhub, except use `v0.5.0-aee2160` instead of v0.4 as version string. 
4. This should give you a normal working JupyterHub without HTTPS
5. Follow the instructions now in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/229/files#diff-f9b292ea7db2ddbd8379bff6ee295ce3R179

This should give you HTTPS! No need to futz with installing other software (ingress providers or kube-lego). If these steps do not give you working HTTPS, that is a bug that we should fix.

Pinging @minrk @cmoscardi @choldgraf @willingc @rschroll @henchc @rpwagner for helping test :)

yuvipanda:
@willingc before we merge this though, I'd like to tag a v0.4 version of the docs and mark them as 'stable' so people who are using v0.4 can still access old docs. This doc change requires v0.5...

rpwagner:
@NickolausDS, see if you can follow Yuvi's steps above. I'll do the same.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -1,40 +1,127 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: LEGO_POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: LEGO_EMAIL
+              # {{ required "proxy.https.auto.letsencryptContactEmail is a required field" .Values.proxy.https.auto.letsencryptContactEmail }}
+              value: {{ .Values.proxy.https.auto.letsencryptContactEmail | quote }}
```
NickolausDS: The field `letsencryptContactEmail` doesn't quite match the docs: `letsEncryptContactEmail`

minrk:
Followed instructions and worked like a charm, other than the `e` case mismatch on `letsEncryptContactEmail` between docs and chart. I guess the docs (capital E) makes the most sense?

Feedback:

- after enabling https, I can still access my cluster via http. I would have expected http to redirect to https at this point
- The structure of the helm config is a little confusing. It seems a little bit odd for `hosts` to be top-level on proxy. Since the presence of proxy.hosts triggers a bunch of https config and doesn't seem relevant when running http-only (is that true?), should it perhaps be `proxy.https.hosts`? Or if it's lego-specific, even `https.letsencrypt.hosts`?
- `https.auto.letsEncryptContactEmail` seems like it might not be the right level, either. What about:
   `https.letsencrypt.email`?
- Does `auto` mean `letsencrypt`, or does it mean automatic based on other config (e.g. ssl credentials present = plain files, otherwise letsencrypt)? If it's automatic at the helm level, auto seems a good name. If it's just an alias for letsencrypt, maybe calling it `letsencrypt` would be clearer
- I don't see how to enable SSL with a provided key/cert. It would be useful to have that as an example as well.

I don't think we need to address all of these in this PR, but I think nailing down the yaml structure would be good. This is the yaml I would expect, if I understand the logic as it is now:

```yaml
proxy:
  https:
    hosts:
      - <your-domain-name (without https://)>
    type: letsencrypt # default
    letsencrypt:
      email: <your-email-address>
```

NickolausDS:
I went through all the instructions and was able to get a cert working.

yuvipanda:
Thanks for testing, @NickolausDS and @minrk!

@minrk Hmm, for me hitting the http endpoint autoredirects me to https after HTTPS has been set up:

```
(zero-to-jupyterhub-k8s) yuvipanda@courage:~/code/zero-to-jupyterhub-k8s$ curl -I wat.boo.omgwtf.in
HTTP/1.1 301 Moved Permanently
Server: nginx/1.13.5
Date: Wed, 01 Nov 2017 18:28:08 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Location: https://wat.boo.omgwtf.in/
Strict-Transport-Security: max-age=15724800; includeSubDomains;
```

And HSTS seems enabled by default too.

Auto did mean letsencrypt, so I've just switched it to say letsencrypt instead. So it is proxy.https.letsencrypt.contactEmail now.

There's no support for manual TLS certs in this PR. We can add that as a separate PR, but probably not for v0.5? Don't think that's worth blocking...

Re: hosts location as proxy.hosts vs proxy.https.hosts, I agree it's a little awkward. Am a little torn about it, since hosts can be used for other things in the future. However, at this moment they're only used for https, so I'm going to change this so it's proxy.https.hosts.

yuvipanda:
@minrk the YAML structure now is almost the same that you proposed (just contactEmail rather than just email).

minrk:
tested again today and it works great!

PR #234, opened 2017-10-24
Add missing submenu in turn off verification on Google Cloud

willingc:


yuvipanda:
Great catch, @willingc!

PR #236, opened 2017-10-25
pass storage class parameter through

zonca:
it was used by `KubeSpawner` but not available in `helm`. I tested it with a `rook-block` storage class and worked fine.


zonca:
sure!

yuvipanda:
Hmm, I think this will still fail, since I am not sure what's the output of passing 'null' to quote.

Can you put an 'if' guard around this to make sure that the entry is never made in the configmap if the value is null?

zonca:
done

yuvipanda:
I'm going to wait on this until next week, since I realized we already support specifying storageclass for the hub's PVC, and I wanna make sure we keep the structure similar. Sorry about the delay, @zonca

yuvipanda:
@zonca I've done some moving-around of how things are named, but merging this now!

PR #238, opened 2017-10-26
Add note about making yourself superuser on GKE

yuvipanda:
This is needed if you don't want the helm init sequence to fail.
Note that with Kubernetes 1.8 GKE mandates RBAC, so we can't go
back to the simpler times of everything being root.


choldgraf:
is this in addition to https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-helm.html#initialization

or instead of it?

yuvipanda:
In addition to.

PR #240, opened 2017-10-31
Automatically upgrade the db if we're on sqlite!

yuvipanda:
Probably needs a patch release of JupyterHub including
https://github.com/jupyterhub/jupyterhub/pull/1507

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -31,7 +31,7 @@ RUN adduser --disabled-password \
 ARG JUPYTERHUB_VERSION=0.8
 
 RUN pip3 install --no-cache-dir \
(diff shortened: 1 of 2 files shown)


yuvipanda:
How to test this:

1. Install v0.4 of the helm chart:

helm install jupyterhub/jupyterhub --name=old11 --namespace=old11 --version=v0.4 -f config.yaml 

2. Upgrade to v0.5.0-89371a0-6b9ec9f of helm chart

helm upgrade old11 jupyterhub/jupyterhub --version=v0.5.0-89371a0-6b9ec9f -f config.yaml

The upgrade should work, and the new pods should come up.

PR #241, opened 2017-10-31
Simplify secretToken + cookieSecret specification

yuvipanda:
- No need to ask users to generate 2 random strings
- Not specifying proxy.secretToken will actually produce an informative error message

### DIFF
--- a/doc/source/setup-jupyterhub.rst
+++ b/doc/source/setup-jupyterhub.rst
@@ -23,31 +23,24 @@ For the following steps, use your favorite code editor. We'll use the
    entering ``nano config.yaml`` at the terminal will start the editor and
    open the config file.
 
-2. Create two random hex strings to use as security tokens. Run these two
-   commands (they’re the same command but run them twice) in a terminal:
+2. Create a random hex string to use as a security token. Run this command
+   in a terminal
 
    .. code-block:: bash
 
        openssl rand -hex 32
-       openssl rand -hex 32
 
-   Copy the output each time, we’ll use these hex strings in the next step.
+   Copy the output for use in the next step
 
 3. Insert these lines into the ``config.yaml`` file. When editing YAML files,
    use straight quotes and spaces and avoid using curly quotes or tabs.
-   Substitute each occurrence of ``RANDOM_STRING_N`` below with the output of
-   ``openssl rand -hex 32``. The random hex strings are tokens that will be used
-   to secure your JupyterHub instance (make sure that you keep the quotation
-   marks):
+   Substitute ``RANDOM_STRING`` below with the output of ``openssl rand -hex 32``
+   from step 2.
 
    .. code-block:: yaml
 
-      hub:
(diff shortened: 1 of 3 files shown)


yuvipanda:
We shouldn't merge this until we have a way of making sure it doesn't propagate to RTD yet...

yuvipanda:
@minrk @choldgraf @willingc @betatim this is ready to review!

doc/source/setup-jupyterhub.rst
```diff
@@ -23,31 +23,24 @@ For the following steps, use your favorite code editor. We'll use the
...
 
    .. code-block:: yaml
 
-      hub:
-        # output of first execution of 'openssl rand -hex 32'
-        cookieSecret: "RANDOM_STRING_1"
       proxy:
-        # output of second execution of 'openssl rand -hex 32'
-        secretToken: "RANDOM_STRING_2"
+        secretToken: "RANDOM_STRING"
```
choldgraf: could save a few lines if we removed 37/38 and juts put `secretToken: "<output of openssl rand -hex 32>"`

jupyterhub/templates/hub/secret.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: hub-secret
 type: Opaque
 data:
+  # {{ required "proxy.secretToken must be a 32byte random string generated with openssl rand -hex 32" .Values.proxy.secretToken }}
```
choldgraf: ` ticks around "openssl rand -hex 32". Also should there be a space between 32 and byte? Sorry if I'm nitpicking here

choldgraf:
2 minor comments but I won't block on it if somebody else wants to merge. Otherwise after those are addressed I'm happy to merge it

minrk:
Makes sense. Just so I understand the implications of removing the initial cookie secret from the docs:

1. if hub.db.type = sqlite-pvc, the cookie is generated on first launch and stored in the pvc alongside jupyterhub.sqlite
2. otherwise, each relaunch of the Hub will logout all users
3. users can still specify cookieSecret to avoid this

Is that right?

doc/source/setup-jupyterhub.rst
yuvipanda: I think the current wording is better, since I've seen people literally put in < and > in these before.

jupyterhub/templates/hub/secret.yaml
yuvipanda: Done

yuvipanda:
@minrk yep, that is correct!

minrk:
It would probably be good to document the cases where cookieSecret wants to be set (i.e. only db via url), but that's a pretty advanced use case, so I'm not quite sure where it would go.

choldgraf:
@minrk perhaps we can inject this back into the docs if/when we get enough usage scenarios for it. Thanks @yuvipanda !!!

PR #246, opened 2017-11-02
Support HTTPS via user-provided key/cert

### LINKED ISSUE #244: Support SSL key/cert files
#229 adds handy support for SSL via letsencrypt. We should have similarly simple support when users have existing SSL key/cert files. config should probably look something like:

```yaml
proxy:
  https:
    hosts:
      - needed?
    kind: files
    files:
      key: |
        -----BEGIN RSA PRIVATE KEY-----
        ...
        -----END RSA PRIVATE KEY-----
      cert: |
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
```

minrk:
usage:

```yaml
proxy:
  https:
    type: files
    files:
      cert: |
        <paste cert>
      key: |
        <paste key>
```

closes #244

### DIFF
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -222,7 +222,8 @@ You can also purchase your own SSL certificates from a certificate provider.
           letsencrypt:
             contactEmail: <your-email-address>
 
-  b. If you have your own SSL certificate, add to ``config.yaml``:
+  b. If you have your own SSL certificate, you can configure SSL manually.
+     Add to ``config.yaml``:
 
     .. code-block:: yaml
 
@@ -230,7 +231,7 @@ You can also purchase your own SSL certificates from a certificate provider.
         https:
           hosts:
             - <your-domain-name>
-          type: files
+          type: manual
 
     and paste the contents of your ssl key and certificate to ``secrets.yaml``:
 
@@ -238,15 +239,15 @@ You can also purchase your own SSL certificates from a certificate provider.
 
       proxy:
         https:
-          files:
-           key: |
-             -----BEGIN RSA PRIVATE KEY-----
-             ...
-             -----END RSA PRIVATE KEY-----
-           cert: |
-             -----BEGIN CERTIFICATE-----
-             ...
-             -----END CERTIFICATE-----
+          manual:
+            key: |
+              -----BEGIN RSA PRIVATE KEY-----
+              ...
+              -----END RSA PRIVATE KEY-----
+            cert: |
+              -----BEGIN CERTIFICATE-----
+              ...
+              -----END CERTIFICATE-----
 
 5. Apply the config changes by running ``helm upgrade ...``.
 6. Wait for about a minute, now your hub is HTTPS enabled! Congratulations, your
--- a/doc/source/extending-jupyterhub.rst
+++ b/doc/source/extending-jupyterhub.rst
@@ -202,7 +202,7 @@ You can also purchase your own SSL certificates from a certificate provider.
 
 1. Buy a domain name from a registrar. Pick whichever one you want.
 2. Create an ``A record`` from the domain you want to use, pointing to the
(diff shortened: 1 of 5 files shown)


jupyterhub/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Multi-user Jupyter installation
 name: jupyterhub
-version: v0.5.0
+version: v0.5.0-3be4d58
```
yuvipanda: Accidental?

yuvipanda:
This is awesome!

What do you think about calling this 'manual' or something like that rather than 'files'? Am unsure what 'files' signifies in this context...

minrk:
I was thinking 'files' because they came from files, but of course they aren't files anymore when they go in your .yml. manual seems okay, but it's not super obvious what it means.

Does it make sense for the default behavior to use `https.key/cert` if defined, or is that too many choices to select between automatically?

doc/source/extending-jupyterhub.rst
```diff
@@ -198,26 +198,58 @@ your users & the world at large. Zero to JupyterHub makes doing so quite
 easy since version 0.5, integrating with `Let's Encrypt <https://letsencrypt.org/>`_
 for free HTTPS certificates.
 
+You can also purchase your own SSL certificates from a certificate provider.
+
 1. Buy a domain name from a registrar. Pick whichever one you want.
 2. Create an ``A record`` from the domain you want to use, pointing to the
    external IP provided to the `proxy-public` service.
```
choldgraf: could explicitly say this is the `<EXTERNAL-IP>` field. For whatever reason people almost always get this confused and copypaste the internal IP

choldgraf: (I know this isn't related to this PR so we can punt it to later if you like)

jupyterhub/templates/proxy/ingress-internal.yaml
```diff
@@ -24,8 +23,15 @@ spec:
...
     {{- end }}
   tls:
+    {{- if eq .Values.proxy.https.type "letsencrypt" }}
     - secretName: kubelego-tls-proxy-{{ .Release.Name }}
+    {{- else if eq .Values.proxy.https.type "files" }}
+    - secretName: files-tls-proxy-{{ .Release.Name }}
+    {{- else }}
+    # unhandled type
+    # it would be nice if helm had an `error` function
+    {{ required (printf "https.type must be 'files' or 'letsencrypt', not '%s'" .Values.proxy.https.type)  ._undefined }}
```
choldgraf: just a note to change this if `files` changes to `manual` or whatever else

choldgraf:
as a non-letsencrypt person, files and manual are equally non-informative to me, just FYI. I'm 33/33/33 on files/manual/something else.

yuvipanda:
@minrk yeah, I think that'd be too much magic. Plus, I think there's a good pattern of:

```
type: someString
someString:  
   <someString specific options>
```

This occurs throughout the k8s ecosystem, and is a good one to follow for us.

I agree manual seems not entirely clear, but can't think of anything else.

yuvipanda:
FWIW I also consider provisioning your own certificates an 'advanced' use case.

jupyterhub/values.yaml
```diff
@@ -82,6 +82,8 @@ proxy:
     type: letsencrypt
     letsencrypt:
       contactEmail: ''
+    key:
```
yuvipanda: this should be manual.key and manual.cert IMO. I think we should be doing:

```
type: X
X:
   <all-X-related-options>
```

choldgraf:
agree it's advanced - that's why I'm just giving a datapoint but not weighing in heavily one way or another :-)

jupyterhub/values.yaml
minrk: done

PR #248, opened 2017-11-03
Move whitelist & admin options under 'auth'

yuvipanda:
This is where they logically belong, rather than under hub or as a top level entry.


yuvipanda:
Maybe I should make this be backwards compatible...

minrk:
> Maybe I should make this be backwards compatible...

I really wish there were a good way to have things like deprecation warnings and errors in the helm charts.

As it is now, I think making sure we cover well "migrating from 0.4 to 0.5" in a doc is enough for where we are.

PR #255, opened 2017-11-06
Provide defaults for master that at least sort of work

### LINKED ISSUE #252: jupyterhub/k8s-hub:v0.4 does not have a user with UID 1000
/jupyterhub/values.yaml wants to run commands in the container as UID 1000. In /etc/passwd in the image however there is no user with that ID.

yuvipanda:
Currently running master directly will fail, since the image
being specified is too old. This fixes it a little by updating
to more recent image. This isn't a long term solution, since
we need to keep bumping it manually every time we change the image.
But it helps people out right now who wanna run master...

Fixes #252

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -26,7 +26,7 @@ hub:
   extraEnv: {}
   image:
     name: jupyterhub/k8s-hub
-    tag: v0.4
+    tag: 0f87fa7
   resources:
     requests:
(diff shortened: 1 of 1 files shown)


minrk:
Recapping from gitter: This is a fine solution in the short-term, but improving discoverability of the latest build of the helm chart is probably the right solution in the long term, plus documenting build.py for those who want to develop the chart.

PR #269, opened 2017-11-10
Allow configuring the start timeout of singleuser pod

yuvipanda:
Might be increased for really large images that are working
with prePuller off


yuvipanda:
Forgot to tag @cmoscardi too!

PR #272, opened 2017-11-15
Allow explicitly setting 'hub.db.upgrade' to force upgrades

yuvipanda:
This is for people *not* on SQLite. This allows them to do an upgrade
by doing a deploy with this setting set to true, rather than having
to hunt down which version of JupyterHub we're running in the helm
chart and use that to upgrade.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -44,9 +44,22 @@ spec:
           - jupyterhub
           - --config
           - /srv/jupyterhub_config.py
-          {{ if or (eq .Values.hub.db.type "sqlite-pvc") .Values.hub.db.upgrade }}
-          # Automatic db upgrades for sqlite!
+          # We want to do automatic upgrades for sqlite-pvc by default, but allow users
+          # to opt out of that if they want. Users using their own db need to 'opt in'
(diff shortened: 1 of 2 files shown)


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -44,7 +44,7 @@ spec:
           - jupyterhub
           - --config
           - /srv/jupyterhub_config.py
-          {{ if eq .Values.hub.db.type "sqlite-pvc" }}
```
minrk: As this is, users *cannot* disable upgrade for sqlite-pvc. Do we want them to? If so, a null/'auto' default might be more appropriate so we can preserve opt-in for others and opt-out for sqlite-pvc

yuvipanda: Hmm, go templates is a shitty mixture of 'static typing' and 'false, nil, 0, "" are all false'. I'm trying to figure out how to differentiate between nil and false. 'auto' as a string runs into the static typing problem, where the template var is string some of the time but a bool some of the time...

yuvipanda: Done! lmk what you think of this :)

minrk: oof. That sure is not as nice as `is None`, but I think the logic is correct, and it gives the option, which is good.

PR #275, opened 2017-11-15
Setup RBAC for pre-puller

yuvipanda:
This lets us fully use z2jh on a cluster with RBAC enforced


minrk:
Conflict, but 👍

yuvipanda:
@minrk resolved conflict!

minrk:
Downstream note: rbac config in testing files will need to be updated when this propagates to the binderhub chart

PR #276, opened 2017-11-15
Use hub.baseUrl for ingress path

manics:
If `hub.baseUrl` is set this is probably because a user wants to host multiple web services under a domain. The Ingress controller should therefore use this as the path instead of `/`.

### DIFF
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -14,11 +14,12 @@ metadata:
     {{- end }}
 spec:
   rules:
+    {{- $root := . -}}
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host }}
       http:
         paths:
-          - path: /
+          - path: {{ $root.Values.hub.baseUrl }}
             backend:
               serviceName: proxy-public
               servicePort: 80


jupyterhub/templates/ingress.yaml
```diff
@@ -14,11 +14,12 @@ metadata:
     {{- end }}
 spec:
   rules:
+    {{- $root := . -}}
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host }}
       http:
         paths:
-          - path: /
+          - path: {{ $root.Values.hub.baseUrl }}
```
yuvipanda: Can this just be .Values.hub.baseUrl? Does `range` modify `.`?

manics: `.` is modified in loops. It's poorly documented, it took me a while to figure out why `Values.hub.baseUrl` doesn't work: https://github.com/kubernetes/helm/issues/1311

yuvipanda: Wow that's crazy I haven't been bitten by this before haha! Thanks for finding it!

yuvipanda:
Thank you very much for the patch!

Issue #278, opened 2017-11-16
Documentation for practitioners


choldgraf:
One thing that came out of conversations at the MSDSE summit was that while the Z2JH guide is great for setting up your own JupyterHub, there isn't a good resource for the squishier questions like *why* you'd want to use it, *how* it benefits teaching in the classroom, and *how* to use it most effectively.

I think resources like this would be really valuable to the community. I wonder if it'd be worth creating a space where this kind of information can live. I imagine it being a place where people could store their experiences, share tips (e.g. the existence of nbgitpuller), etc.  Ideally it'd be more community-driven than just maintained by the jupyterhub team.

Anyone have thoughts/ideas on something like this? A wiki? A "Zero to JupyterHub for Education" site like this one? (and similar kinds of sites if other obvious use-cases pop up)

consideRatio:
This post successfully made me know the existence of nbgitpuller. ;D

I'd love a way to get in tough with teachers and similar who wants to educate students using JupyterHub. I'm a teacher in math / physics, my goal is to use JupyterHub and nbgrader in order to allow my students to do hand in exercises presented in Jupyter Notebooks about math. I want it to be able to scale, so that multiple schools could use the same system and get help with integrating programming with math without knowing a lot about programming themselves.

Did you know about this group by the way? https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!forum/jupyter-education

choldgraf:
Ah yes - I imagine this as some kind of place to codify the wisdom learned in the jupyter for education community!

choldgraf:
do any of the pages underneath https://zero-to-jupyterhub.readthedocs.io/en/latest/index.html#resources-from-the-community work for this? What are ways that we could modify what's there so that it's more useful?

consideRatio:
**Actionplan:**
Create a stub where we gather wisdoms regarding usage or non-usage of a z2jh deployment for educational contexts under resources-from-the-community.

consideRatio:
Closed and encompassed as a wish for improvement in the issue linked above.

PR #279, opened 2017-11-17
Add a list of 'users of this chart'

yuvipanda:
Useful for showing wide usage

### DIFF
--- a/doc/source/users-list.md
+++ b/doc/source/users-list.md
@@ -1,0 +1,9 @@
+# Zero to JupyterHub users
+
+This is a community maintained list of organizations / people using the Zero to
+JupyterHub guide / helm-chart to maintain their JupyterHub. Send us a Pull
+Request to add yourself to this alphabetically sorted list!
+
+* Data Science Education Program's DataHub at University of California, Berkeley
+* [MyBinder.org](https://mybinder.org)
+* [PAWS](https://www.mediawiki.org/wiki/PAWS) at [Wikimedia Cloud Services](https://www.mediawiki.org/wiki/Wikimedia_Cloud_Services_team)
--- a/doc/source/users-list.md
+++ b/doc/source/users-list.md
@@ -1,4 +1,4 @@
-# Zero to JupyterHub users
+# Zero to JupyterHub Gallery
 
 This is a community maintained list of organizations / people using the Zero to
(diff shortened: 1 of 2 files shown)


yuvipanda:
We might maybe want to add a one-line description. Not sure.

/cc @choldgraf @willingc if you know other people who should be on this list (plus thoughts on how to structure this).

choldgraf:
I think it's a good idea...we could turn this into some kind of gallery if this list ever gets long enough w/ public-facing deployments

yuvipanda:
+1. Do we also want to count the workshops and what not that people use this for?

choldgraf:
Why not? We can have a workshops section? Or maybe this would go into some kind of place like #278 specifies?

Also we should provide some way for others to tell us about public jupyterhub deployments. Maybe ask people to make a PR if they want to mention their deployment?

yuvipanda:
Cool, I'm all for having a workshop section :) wanna add a few to this list? Between you, @arokem and @aculich that's a few. Let's only include workshops about various subjects that was conducted using z2jh, rather than z2jh workshops themselves...

choldgraf:
yeah I agree, off the top of my head:

* Neurohackweek
* Data on the mind
* The South Data Hub talk


but there are definitely others.

willingc:
I recommend something similar to: http://jupyterhub.readthedocs.io/en/latest/gallery-jhub-deployments.html

doc/source/users-list.md
```diff
@@ -0,0 +1,9 @@
+# Zero to JupyterHub users
```
willingc: I would change the title to Gallery to give more flexibility over time with content

choldgraf:
+1 to merge after we rename to @willingc 's "Gallery" suggestion. That way we don't have to create a new thing whenever we get around to creating images etc for the deployments.

choldgraf:
I went ahead and made @willingc 's suggestion, merging now!

choldgraf:
Is travis being funky right now? I restarted it once but have run into two time-out errors...

yuvipanda:
@choldgraf travis was being funky, I filed https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/434 to investigate and a more permanent fix in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/435

choldgraf:
Cool! Merged this dangling PR :-)

PR #281, opened 2017-11-17
configmap admin.users should be prefixed by auth.

manics:
`admin.users` in the hub config map doesn't match `auth.admin.users` in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/3ebd38bf6b3b564ecead63b30606499f4cb41fc1/images/hub/jupyterhub_config.py#L220


manics:
As an aside, I think it would be nice if `auth.admin.users` (and `auth.whitelist.users`) could be passed as a list such as `[adminuser]` instead of a json string `'["adminuser"]'`

PR #284, opened 2017-11-20
bump kubespawner

minrk:
get the latest patch that should fix orphaned containers on spawn timeout

https://github.com/jupyterhub/kubespawner/pull/98


yuvipanda:
Thanks for the follow up merge, @minrk!

PR #286, opened 2017-11-21
Change heading levels in 'turn-it-off'

yuvipanda:
The current setup, where we have:

```
<h3> "For all clouds"

<instructions>

<h4> <specific-cloud-instructions>
```

Seem strange to me, since it's putting "GKE" or "AWS" 'under'
"For all clouds". It also makes the ToC a bit strange.

Not sure if this is the right thing to do!


minrk:
I think this is right.

PR #288, opened 2017-11-23
Make the published times in index.yaml correct

yuvipanda:
Currently they aren't, they're all set to the last chart's updated
date. This is because they're set from mtime by helm, but our
git clone does not preserve mtimes. So we 'fix it up' by using
information from git to find last modified times.


yuvipanda:
Note that we'll need to make this fix in the binderhub repo too. This makes me think it's time to factor out build.py into its own package maybe.

build.py
```diff
@@ -43,31 +54,81 @@ def build_images(prefix, images, commit_range=None, push=False):
...
 
     with open('jupyterhub/values.yaml', 'w') as f:
-        yaml.dump(values, f)
+        rt_yaml.dump(values, f)
 
 
 def build_chart():
-    version = last_git_modified('.')
+    rt_yaml = YAML()
+    rt_yaml.indent(offset=2)
```
minrk: Was there a reason for duplicating the yaml initialization here and above?

minrk: nevermind, I see it below.

minrk:
Factoring out build.py sounds like a great idea.

PR #297, opened 2017-11-28
disable rbac on minikube

minrk:
since it's usually not available and now on by default


yuvipanda:
Do you get an error if you try to use it with it on on minikube?

PR #302, opened 2017-12-01
Do not shut down user pods when hub is restarted

yuvipanda:
No idea why we need to do this explicitly just now when it has
been working just fine till now?!

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -40,6 +40,9 @@ c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
 c.ConfigurableHTTPProxy.api_url = 'http://{}:{}'.format(os.environ['PROXY_API_SERVICE_HOST'], int(os.environ['PROXY_API_SERVICE_PORT']))
 c.ConfigurableHTTPProxy.should_start = False
 
+# Do not shut down user pods when hub is restarted
+c.JupyterHub.cleanup_servers = False
+
 # Check that the proxy has routes appropriately setup
 # This isn't the best named setting :D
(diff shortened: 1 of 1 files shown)


betatim:
I recognise that as a legit jhub option, but I don't know why you need to set it now. Have you looked at the the config file that is generated from the secrets/config map? Maybe with a new(?) version of k8s/helm it generates a different value to before?

minrk:
This *should* be required to get the behavior we want. If it ever wasn't, it's probably up to how the pods are getting shutdown. I seem to recall something funny about how long KubeSpawner pods take to stop when politely shutdown. That indicated something amiss about how either we or Kubernetes were sending signals to the pods, i.e. that SIGTERM was not being received and pods where only stopping when it resorted to SIGKILL.

If this was happening in fact for all pods, it could have been a bug and one that's been fixed, either in Kubernetes itself or how we are using it. That could be the difference:

- Before: something prevented SIGTERM from being delivered, so Hub was always halted with SIGKILL
- After: whatever was preventing SIGTERM 

This is mostly a wild guess based on a vague memory of the symptoms of slow singleuser pod shutdown a while ago, but it would explain the behavior.

From [kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods), it could also be a grace period: deleting a pod with `--grace-period=0` should send SIGKILL immediately (I don't know if it will send SIGTERM and SIGKILL in rapid succession, or skip SIGTERM). So if something was previously passing `--grace-period=0` and isn't anymore, that could also have the same effect of bypassing the cleanup, or at least killing it before it gets to far.

PR #313, opened 2017-12-03
Update issue template post merge

willingc:
- Removes the comment so the text displays a bit cleaner
- Added an additional set of "```" backticks since folks often delete them when copy/pasting data.

cc/ @choldgraf Please take a look. Thanks.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,3 @@
-<!--
 Welcome to Zero to JupyterHub!
 
 Before filing an issue, please search through the issues to see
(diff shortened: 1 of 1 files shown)


willingc:
Ah, I see why you had the comment in there so it won't display after filed. That's cool. Feel free to close if you want.

choldgraf:
ah I see what you mean about users deleting the backticks. I wonder if there's a way to get them to post inside the backticks without adding an extra set of backticks? Only reason I mention is because it means that if they don't modify that text at all, or if they don't delete one of the sets of backticks we'll get a kind of weird output. WDYT?

image

willingc:
Let's just go with your original PR and I'll close this one. We can always redo if needed 👍

PR #314, opened 2017-12-04
Expose the configuration options of auth state

tothandras:


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -105,6 +105,13 @@ spec:
         - name: {{ $key | quote }}
           value: {{ $value | quote }}
         {{ end -}}
+        {{ if .Values.auth.state.enable }}
```
yuvipanda: This should probably be put in the hub secret and then put here via a ref I think.

jupyterhub/values.yaml
```diff
@@ -102,6 +102,9 @@ auth:
     users:
   dummy:
     password:
+  state:
+    enable: false
```
yuvipanda: we use 'enabled' elsewhere. Use that here too to be consistent?

yuvipanda:
This is awesome, thanks for the patch! I'd like to hold off merging this until we get 0.5 out the door (today or tomorrow), hope that is ok!

jupyterhub/templates/hub/deployment.yaml
tothandras: I think I already do that. Can you clarify?

jupyterhub/values.yaml
tothandras: Okay!

jupyterhub/templates/hub/deployment.yaml
yuvipanda: Nope, I just hadn't fully woken up yet. Aaa, sorryy!!

jupyterhub/values.yaml
tothandras: Updated! 🙂

tothandras:
@yuvipanda Any updates on when it might be?

yuvipanda:
Apologies for the delay! I've merged this in now!

PR #318, opened 2017-12-05
Get a version of kubernetes new enough for RBAC by default

yuvipanda:
1.7 doesn't support the RBAC object versions we are using.


doc/source/create-k8s-cluster.rst
```diff
@@ -63,7 +63,8 @@ connect your credit card or other payment method to your google cloud account.
       gcloud container clusters create <YOUR_CLUSTER> \
           --num-nodes=3 \
           --machine-type=n1-standard-2 \
-          --zone=us-central1-b
+          --zone=us-central1-b \
+          --cluster-version=1.8.4-gke.0
```
minrk: Is it worth a comment about why?

yuvipanda: @minrk done!

PR #321, opened 2017-12-05
Add note about error when ingress config needs to be migrated

yuvipanda:
And link to docs on how to do so


choldgraf:
LGTM. Will merge once I'm at the computer unless folks say we should hold off!

Issue #289, opened 2017-11-24
test deployment


minrk:
We should be running a test deployment on Travis, along the lines of what we have in https://github.com/jupyterhub/binderhub/pull/282.

At the very least, smoke testing that `helm install` produces running pods would catch most templating errors. I think that's enough for a first PR.

After that, we can look at one or two tests that verify that an actual user launch succeeds, to verify that the whole stack works.

manics:
I had the same thought after also running into https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/287. Work-in-progress: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/master...manics:travis-minikube-allbranches

manics:
This is more complicated than I expected due to the issues @minrk ran into in https://github.com/jupyterhub/binderhub/pull/282 (e.g. lack of `nsenter`). I'm still happy to help out if you give me some guidance, e.g. should I just copy https://github.com/minrk/binderhub/blob/b4f3aa45325cd45606e65d13f067c3871ef2c5a5/ci/install.sh ? Or does someone else want to do this?

yuvipanda:
@manics awesome! Thank you for working on it :)

I haven't really touched it, so @minrk would have no ideas. I think just copying whatever was done for binderhub into here seems like a good way to go?

minrk:
Copying `install.sh` probably makes sense. I think we'll need all the same stuff here. I don't love copying/pasting stuff, but it's probably better than adding a repo and having to work in two repos at once.

PR #322, opened 2017-12-05
Fix hub service pod selector

manics:
I just ran into this whilst working on travis/minikube tests (#289)

Related to the change in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/319


yuvipanda:
Whoops! That's a great catch!

PR #323, opened 2017-12-05
Install helm chart in travis

### LINKED ISSUE #289: test deployment

manics:
Install the helm chart in travis.

- Copies ci/install.sh from jupyterhub/binderhub with some minor changes: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/983dbb5f5f6eb9e72938e15e1004309f5ce34ae3
- Adds a Vagrantfile and script to roughly approximate the travis trusty docker environment (makes configuring minikube easier)
- Modifies `.travis.yml` to always run instead of being limited to `master` and `v0.4.x`, but it should only push to those branches (untested)

Closes #289

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,18 +12,14 @@ install:
 - helm init --client-only
 script:
 - |
-  if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+  if [[ "$TRAVIS_PULL_REQUEST" == "false" && ( "$TRAVIS_BRANCH" = master || "$TRAVIS_BRANCH" == v0.4.x ) ]]; then
     docker login -u ${DOCKER_USERNAME} -p "${DOCKER_PASSWORD}"
     export PUSH="--push --publish-chart"
   fi
 - "./build.py build --commit-range ${TRAVIS_COMMIT_RANGE} $PUSH"
-branches:
-  only:
-  - master
-  - v0.4.x
 before_install:
 - |
-  if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+  if [[ "$TRAVIS_PULL_REQUEST" == "false" && ( "$TRAVIS_BRANCH" = master || "$TRAVIS_BRANCH" == v0.4.x ) ]]; then
     openssl aes-256-cbc -K $encrypted_c6b45058ffe8_key -iv $encrypted_c6b45058ffe8_iv -in travis.enc -out travis -d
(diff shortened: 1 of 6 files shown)


manics:
For comparison here's an example of a travis build which failed without https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/322: https://travis-ci.org/manics/zero-to-jupyterhub-k8s/builds/311983195

.travis.yml
```diff
@@ -5,25 +5,19 @@ services:
...
 - "./build.py build --commit-range ${TRAVIS_COMMIT_RANGE} $PUSH"
-branches:
-  only:
-  - master
-  - v0.4.x
+- ./ci/test.sh
 before_install:
 - |
-  if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+  if [[ "$TRAVIS_PULL_REQUEST" == "false" && ( "$TRAVIS_BRANCH" = master || "$TRAVIS_BRANCH" == v0.4.x ) ]]; then
```
minrk: This TRAVIS_BRANCH logic is redundant with the `branches.only` logic above that was removed. Can we switch back to that? Or did it not work for some reason?

minrk:
This is great! I asked one question about removing the travis branch check in favor of what should be an identical check. Note that the `branches.only` config should *not* affect pull requests, which will still be tested.

.travis.yml
manics: I changed that to ensure travis would always run on your own branches when developing. Otherwise you have to either develop on your own `master` branch, have a separate commit to modify travis and remove it before opening a PR, or only have it tested after opening the PR.

If I've understood correctly the original aim was to only do `docker push` for the two branches which makes sense, but I think unconditionally running the tests is useful- I usually have travis setup on my own repos/forks to run on every push on any branch. I'm happy to revert it though.

minrk: Good point. What we should probably do is move these bits to a specific `deploy` step, like we do [on repo2docker](https://github.com/jupyter/repo2docker/blob/master/.travis.yml#L14).

I think this is AOK as it is right now.

PR #325, opened 2017-12-06
Generalize build.py to be not z2jh specific

yuvipanda:
I'm calling it 'chartpress', like a printing press but for charts.

- Factor out repo specific info into a separate chartpress.yaml file,
  so we can move build.py to a separate repo eventually
- Fixup travis to use new commandline params

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ script:
     docker login -u ${DOCKER_USERNAME} -p "${DOCKER_PASSWORD}"
     export PUSH="--push --publish-chart"
   fi
-- "./build.py build --commit-range ${TRAVIS_COMMIT_RANGE} $PUSH"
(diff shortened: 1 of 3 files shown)


yuvipanda:
Next step here is to make sure this works for binderhub's chart CI too, and then split it up into its own repo.

Also perhaps clean up the code :)

minrk:
Anything else you want to do here? e.g. test that it works with binderhub first?

yuvipanda:
I tested the publish_chart bit (which I couldn't on the flight) and fixed it. It's good to go, @minrk. Needs verification post merge tho.

minrk:
`helm package` [failed](https://travis-ci.org/jupyterhub/zero-to-jupyterhub-k8s/builds/312464887)

Issue #251, opened 2017-11-04
Document max length on docker image names


choldgraf:
From https://gitter.im/jupyterhub/jupyterhub?at=59fd7c3ee44c43700ab70c45

it sounds like having docker image names above a certain length will cause timeouts

choldgraf:
this was fixed by #327!

PR #327, opened 2017-12-06
Solves #251: add a note to the doc on 63 chars limit

tracek:
Trivial change in doc to address https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/251

### DIFF
--- a/doc/source/user-experience.rst
+++ b/doc/source/user-experience.rst
@@ -58,7 +58,7 @@ existing image, such as the ``scipy-notebook`` image, complete these steps:
 
    .. note::
    
-      Container name cannot be longer than 63 characters.
+      Container image name cannot be longer than 63 characters.
 
       Always use an explicit ``tag``, such as a specific commit.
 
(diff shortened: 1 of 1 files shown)


doc/source/user-experience.rst
```diff
@@ -57,6 +57,8 @@ existing image, such as the ``scipy-notebook`` image, complete these steps:
            tag: 8e15d329f1e9
 
    .. note::
+   
+      Container name cannot be longer than 63 characters.
```
yuvipanda: This should probably say that the container *image* name can not be longer than 63 characters?

tracek: Indeed, thanks. Fixed.

yuvipanda:
Thanks for the patch, @tracek! Welcome to the community, and I hope to see more patches from you 'round here :)

yuvipanda:
Test failures are unrelated, since this is a doc change.

PR #328, opened 2017-12-06
move deploy to deploy stage

minrk:
avoids restricting testing to deployable branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 - . ci/minikube.env
 - ./ci/install.sh
 script:
-- ./build.py build --commit-range ${TRAVIS_COMMIT_RANGE}
+- ./build.py --commit-range ${TRAVIS_COMMIT_RANGE}
 - ./ci/test.sh
 
 deploy:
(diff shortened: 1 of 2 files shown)


minrk:
Needs to wait for #325 and rebase, since it changes how build.py is called slightly.

Will need to double check that the deploy succeeded after merging this one!

minrk:
Rebased on #330, so if that passes, this should be okay.

.travis.yml
```diff
@@ -8,19 +8,17 @@ install:
 - . ci/minikube.env
 - ./ci/install.sh
 script:
-- |
-  if [[ "$TRAVIS_PULL_REQUEST" == "false" && ( "$TRAVIS_BRANCH" = master || "$TRAVIS_BRANCH" == v0.4.x ) ]]; then
-    docker login -u ${DOCKER_USERNAME} -p "${DOCKER_PASSWORD}"
-    export PUSH="--push --publish-chart"
-  fi
-- "GIT_SSH_COMMAND='ssh -i travis' ./build.py --commit-range ${TRAVIS_COMMIT_RANGE} $PUSH"
+- ./build.py build --commit-range ${TRAVIS_COMMIT_RANGE}
```
yuvipanda: need to get rid of the second 'build', that's why this is failing

yuvipanda:
Thanks, @minrk

PR #330, opened 2017-12-06
use absolute path to travis ssh key

minrk:
git is run from a few different working directories, so abspath should be more portable

This should fix the build after #325


yuvipanda:
Ty!

PR #336, opened 2017-12-07
[REVIEW BUT NOT MERGE] Don't restart hub pod to pick up config change

yuvipanda:
Instead, whenever configmap changes, let's just restart the
hub *process*. We launch it with the watchdog manager, which
can restart processes when files change.

We try to have a 5s timeout to make sure that we don't restart
too many times for the same file. We also ignore files starting
with a '.' and directory changes for the same reason.

Note that this means changes can take upto 20s to propagate, 
due to [caching](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically). However,
unlike restarting the pod, it has the following advantages:

1. The hub is still serving traffic during this time
2. No scheduling latency
3. With sqlite on a persistent disk, it could take upto a minute
    for them to restart

So I think it's worth it to make it happen!

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -42,7 +42,8 @@ RUN pip3 install --no-cache-dir \
          mwoauth==0.3.2 \
          globus_sdk[jwt]==1.2.1 \
          oauthenticator==0.7.2 \
-         cryptography==2.0.3
+         cryptography==2.0.3 \
+         watchdog==0.8.3
 
 RUN pip3 --no-cache-dir install git+https://github.com/jupyterhub/kubespawner@v0.7.1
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
Note that if we want to do away with the 20s change, we can write a simple script that uses the kubernetes API to watch for changes to the configmap and secret, and restart when those change. That might be much better long term anyway, since this 20s thing seems to introduce an amount of uncertainity here...

minrk:
I guess the only disadvantage here is that it's harder to know when the new version is started, right? When we've done a deploy that changed the configmap, how do we know when we are talking to the new instance?

yuvipanda:
@minrk you'd have to look at the logs, it'll have a message when it's restarting.

I'm convinced this isn't good enough, since we don't get enough benefits for giving up the easy observabilitiy. 

I think instead what we need is something that can detect the configmap change instantly and then restart the process, instead of waiting for pod-sync-interval (often 15s more more). Something like https://github.com/kelseyhightower/konfd but doing:

1. Configmap projection to a dir (like we do now)
2. Process supervision & restart when it does change

I haven't found any such thing yet though

Issue #333, opened 2017-12-06
Add section on upgrading.


choldgraf:
We should add a section on upgrading in the z2jh guide.

PR #334, opened 2017-12-06
Documentation improvements for v0.5

choldgraf:
As I was going through the v0.5 deployment, I thought there were a bunch of places where we could reorganize to improve readability etc. This PR takes a general pass at the documentation, with a focus on:

* removing empty sections (and creating GitHub issues instead)
* reducing redundancy
* refactoring so that sections are more discoverable and intuitive
* improving language / explanations of various components of the guide
* making sure the z2jh getting started guide works with v0.5 of the helm chart.

I'm happy to iterate on this a bit, as it's a large-ish change to the docs. I haven't removed much content, but have reordered / reorganized / reclassified a number of things. In particular I'd love to hear what you think @willingc and @yuvipanda !

Built docs here:

http://predictablynoisy.com/zero-to-jupyterhub-k8s/

PR #338, opened 2017-12-07
Move upgrade docs to rendered docs

minrk:
I was able to do an install following the v0.4 instructions and the upgrade instructions in the CHANGELOG, with minor edits I made here. The first thing I did was put the upgrade instructions into the docs, since I don't think the CHANGELOG is very visible to users because it is not included in the docs. We could also include the CHANGELOG itself in the docs, like we do for jupyterhub itself. I'm not sure if the full changelog has the right audience for doing that or not.

The main missing bit of info was the fact that user servers must be stopped during the upgrade. It *sort of* works if you don't do this, but it behaves a bit weird, so I think we should recommend stopping user servers for major upgrades (at least ones like this that change how Hub and server communicate).

Relevant bits:

1. Following v0.4 instructions today result in v1.7.8 cluster,
   so upgrade fails unless `rbac.enabled = false` is added
   (already covered in the upgrade docs, yay!)
2. Running pods were left semi-orphaned during the upgrade.
   They would be cleaned up and replaced on next login of a user,
   but the old pod wouldn't get cleaned up before then.
   We don't need to handle this, we just need to be clear that users
   must stop during the upgrade.


choldgraf:
I think it's a good idea. One note: #334 removes the `upgrading` section from the index, so I should revert that if/when this gets merged.

choldgraf:
sounds good!

willingc:
Closes #333

PR #339, opened 2017-12-07
Add commands to secure tiller.

ryanlovett:
Tiller is open to connections inside the cluster and someone could try to exploit that. This problem is documented at https://engineering.bitnami.com/articles/helm-security.html.

Until there is an easier fix, I think we need to tell people to essentially `type this` to secure their clusters.

### DIFF
--- a/doc/source/setup-helm.rst
+++ b/doc/source/setup-helm.rst
@@ -60,16 +60,7 @@ Ensure that `tiller is secure <https://engineering.bitnami.com/articles/helm-sec
 
    .. code:: bash
 
-      kubectl -n kube-system patch deployment tiller-deploy --patch '
-      spec:
-        template:
-          spec:
-            containers:
-              - name: tiller
-                ports: []
-                command: ["/tiller"]
-                args: ["--listen=localhost:44134"]
-      '
+      kubectl --namespace=kube-system patch deployment tiller-deploy --type=json --patch='[{"op": "add", "path": "/spec/template/spec/containers/0/command", "value": ["/tiller", "--listen=localhost:44134"]}]'
 
 Next Step
 ---------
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thank you for starting this, Ryan! 

I want us to reduce the patch line to one line. Can you try *only* modifying the command param of tiller? And not remove the service? I think if it listens on localhost then it should be secure enough...

yuvipanda:
I think this should go in before we publicize 0.5.

yuvipanda:
we will eventually remove it after we work with upstream to get this into helm itself...

ryanlovett:
Yes, I just tried on another cluster and patching the deployment without first deleting the service did work. I'll update the PR.

ryanlovett:
I've updated it, but I think I need to follow through on this a bit more:
> Can you try only modifying the command param of tiller?
I think you're saying that the patch should only append the listen param and not set it as the one param.

doc/source/setup-helm.rst
```diff
@@ -53,6 +53,24 @@ It should provide output like
...
 Make sure you have at least version 2.4.1!
 
+Secure Helm
+~~~~~~~~~~~
+
+Ensure that `tiller is secure <https://engineering.bitnami.com/articles/helm-security.html>`_ from access inside the cluster:
+
+   .. code:: bash
+
+      kubectl -n kube-system patch deployment tiller-deploy --patch '
```
yuvipanda: You can replace this with the single line:

```
kubectl --namespace=kube-system patch deployment tiller-deploy --type='json' --patch='[{"op": "replace", "path": "/spec/template/spec/containers/0/command", "value": ["/tiller", "--listen=localhost:44134"]}]'
```

ryanlovett: Okay, in my testing, I had to use the "add" operation rather than "replace".

yuvipanda:
Thank you very much, @ryanlovett!

PR #346, opened 2017-12-11
Allow setting a custom max body size in nginx proxy

jacobtomlinson:
When importing some particularly big libraries into our notebook sessions or when generating some large plots we've found that the notebook file can get too large for the proxy. This causes saving to fail with a `413 (Request Entity Too Large)` error.

This PR allows you to customise the `proxy-body-size` in the nginx proxy and set a custom max file size for the notebook. I've set the default value to `8m` which is probably larger than the current max (we were unable to save notebooks >2MB).

Credit to @AlexHilson for finding exactly where to put this setting.


yuvipanda:
Whoops, that's a great catch! Thank you for finding it...

Instead of adding an extra configmap, can we just set the annotation on the internal-ingress instead? that's probably simpler.

Let's also bump the default up to 64m, just in case :)

yuvipanda:
We were already setting annotation on internal ingress, but it doesn't seem to work :(

I've just bumped the default to 64m and did a minor rename. This seems to work for me, I'll merge shortly.

Thank you very much for your work, @jacobtomlinson!

yuvipanda:
@jacobtomlinson thanks for the patch! <3 welcome to the community, hope to see more patches from you :)

This is going to be in the v0.5 release, I'll make sure to add your name to the contributors list!

jacobtomlinson:
Awesome thanks! Please also add @AlexHilson as we worked on this PR together.

PR #354, opened 2017-12-15
Show rbac code block

ryanlovett:
The block is hidden when raw.


yuvipanda:
The test failure seems spurious, going to merge this now. Thanks, @ryanlovett!

PR #359, opened 2017-12-18
Update Azure documentation to use AKS.

ryanlovett:
The instructions give the option of obtaining some values with jq or looking them up manually. Not sure whether to change this to just look them up manually or just install jq and type these commands.

The service principal creation was necessary for me because of how and when my institution setup AD integration with Azure. I suspect that many others won't need to do this. I can't confirm since I don't have Azure credits on non-work accounts.

### DIFF
--- a/doc/source/create-k8s-cluster.rst
+++ b/doc/source/create-k8s-cluster.rst
@@ -100,7 +100,7 @@ connect your credit card or other payment method to your google cloud account.
 
 .. _microsoft-azure:
 
-Setting up Kubernetes on Microsoft Azure Container Service (ACS)
+Setting up Kubernetes on Microsoft Azure Container Service (AKS)
 ----------------------------------------------------------------
 
 .. note::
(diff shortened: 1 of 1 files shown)


choldgraf:
This is on my todo list! Just haven't had time to get to it yet bc of holiday stuff but my flight back to the states is tomorrow

yuvipanda:
I'm going to merge this now since I think it's better than what we have now and it's been here a while :) We can iterate after!

willingc:
Thanks @ryanlovett and sorry I didn't get back to this yet.

ryanlovett:
No worries @willingc. :)

choldgraf:
Likewise - I start w/ the instructions but got bogged down installing the CLI which took a *really* long time for some reason :-/

ryanlovett:
That is good feedback by itself @choldgraf. There are apt mirrors which host packages for azure-cli -- I could provide some info on that as well as a note about how long it takes to install the cli.

choldgraf:
yep - I will give it another go when I have a minute, and will open an issue where we can concatenate pain points etc that we can use for a subsequent PR...sounds good?

ryanlovett:
Yes, sounds good.

PR #362, opened 2017-12-21
Add a fuller fledged security document

yuvipanda:


doc/source/security.md
```diff
@@ -0,0 +1,80 @@
...
+## Secure the access to Helm
+
+In its default configuration, helm pretty much allows root access to all other
+pods running in your cluster. See this [Bitnami Helm security article](https://engineering.bitnami.com/articles/helm-security.html)
+for more information. As a consequence, the default allows all users in your cluster to pretty much have root access to your whole cluster! 
+
+You can mitigate this by limiting public access to the Tiller API. To do so, use the following command:
+
+```bash
+kubectl --namespace=kube-system patch deployment tiller-deploy --type=json --patch='[{"op": "add", "path": "/spec/template/spec/containers/0/command", "value": ["/tiller", "--listen=localhost:44134"]}]'
```
choldgraf: wonder if it's worth highlighting that this namespace is not whatever namespace that they chose when setting up their jupyterhub

choldgraf:
this is a great addition, thanks!

PR #363, opened 2017-12-21
Re-add info about HTTPS with some clarifications

yuvipanda:
Cleaned it up to make it very clear that most people should use the
let's encrypt option

### DIFF
--- a/doc/source/security.md
+++ b/doc/source/security.md
@@ -4,7 +4,64 @@ The information in this document focuses primarily on cloud based deployments. F
 
 Brad Geesamen gave a wonderful talk titled [Hacking and Hardening Kubernetes by Example](https://kccncna17.sched.com/event/CU6z/hacking-and-hardening-kubernetes-clusters-by-example-i-brad-geesaman-symantec) at Kubecon NA 2017. You can [watch the talk](https://www.youtube.com/watch?v=vTgQLzeBfRU) or [read the slides](https://schd.ws/hosted_files/kccncna17/47/Hacking%20and%20Hardening%20Kubernetes%20By%20Example%20v1.pdf). Highly recommended that you do so to understand the security issues you are up against when using Kubernetes to run JupyterHub.
 
-## Secure the access to Helm
+## HTTPS
+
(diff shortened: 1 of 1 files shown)


doc/source/security.md
```diff
@@ -4,7 +4,64 @@ The information in this document focuses primarily on cloud based deployments. F
 
 Brad Geesamen gave a wonderful talk titled [Hacking and Hardening Kubernetes by Example](https://kccncna17.sched.com/event/CU6z/hacking-and-hardening-kubernetes-clusters-by-example-i-brad-geesaman-symantec) at Kubecon NA 2017. You can [watch the talk](https://www.youtube.com/watch?v=vTgQLzeBfRU) or [read the slides](https://schd.ws/hosted_files/kccncna17/47/Hacking%20and%20Hardening%20Kubernetes%20By%20Example%20v1.pdf). Highly recommended that you do so to understand the security issues you are up against when using Kubernetes to run JupyterHub.
 
-## Secure the access to Helm
+## HTTPS
+
+Enabling HTTPS for your hub is pretty easy, since we automatically integrate with [Let's Encrypt](https://letsencrypt.org/).
```
choldgraf: can we avoid using words like "easy" in the documentation? This is a highly user-specific word :-)

choldgraf: instead maybe just "This section describes how to enable HTTPS on your JupyterHub. The easiest way to do so is by using [Let's Encrypt](https://letsencrypt.org/), though it is also possible to set up your own HTTPS credentials.".

choldgraf: I really like the times when you've added links to a blog post etc in motivating the various topics in z2jh (e.g., like in the recent security PR). Can you think of something we could link to from here for users that may not have an intuition for why they'd use HTTPS? (not blocking on this PR but if one comes to mind I think it'd be helpful!)

choldgraf:
nice addition! a couple comments and then I'm +1  (only the first I think is blocking on the PR)

choldgraf:
made the small edits myself, thanks @yuvipanda ! :-)

yuvipanda:
Thanks, @choldgraf! I like this workflow more than us going back and forth, I appreciate the time you spent cleaning this! <3

PR #365, opened 2017-12-21
Bump kubespawner version

yuvipanda:
https://github.com/jupyterhub/kubespawner/pull/64 got merged!

Note that this is a noop for z2jh users for now. Sometime in the 0.6 or 0.7 timescale
we will enable using the Kubernetes Ingress controllers as an option (not default).


yuvipanda:
This test failure is real, I'm investigating.

yuvipanda:
Test failure is same failure as other failures, but am investigating this time :D

PR #367, opened 2017-12-22
Make tests a little more stable

yuvipanda:
- Print out more debugging info when we are waiting for the hub
  to be up, to make it easier for us to debug failing tests
- Make the curling code a little better

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -4,6 +4,19 @@ metadata:
   name: hub
 spec:
   replicas: 1
+  strategy:
+    rollingUpdate:
+      {{ if eq .Values.hub.db.type "sqlite-pvc" }}
+      # If we're using a PVC, set maxUnavailable to 1
+      # This is required, since the pod attached to PVC must die
+      # before new pod can run.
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks @willingc! I'll merge this now...

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -4,6 +4,19 @@ metadata:
...
+      {{ if eq .Values.hub.db.type "sqlite-pvc" }}
+      # If we're using a PVC, set maxUnavailable to 1
+      # This is required, since the pod attached to PVC must die
+      # before new pod can run.
+      maxUnavailable: 1
+      {{ else }}
+      # If we aren't using a PVC, it's ok to spawn up a new pod
+      # before killing the currently running one. This also makes the '--wait'
+      # flag of helm work properly
+      maxUnavailable: 0
```
minrk: This may not be okay for postgres/mysql databases. Starting a new pod while the other is shutting down could potentially result in multiple Hubs talking to the same db and produce inconsistent state. It's likely only safe for the case where db.type is null for in-memory-only.

PR #368, opened 2017-12-27
Fix styling of yaml code block in HTTPS section

mrocklin:
This fixes a rendering error on http://zero-to-jupyterhub.readthedocs.io/en/latest/security.html#set-up-manual-https


choldgraf:
nice catch!

PR #373, opened 2018-01-02
Fix #372 - build.py corrupts yaml

### LINKED ISSUE #372: build.py corrupts yaml
## Description
I ran into an issue with entries in values.yaml file after using the build.py script. My `extraVolume` entries were corrupted.

## Reproduction

```yaml
# create this file and name it values.yaml
root-1:
  # offset=2 will indent the dash 2 spaces (default: 2)
  # sequence=4 will indent indent-status-2 4 spaces and ignore what offset is used (default: 2)
  - indent-status-1: ok
    indent-status-2: bad after ./test.py runs
root-2:
  # mapping=2 will indent the following entry 2 spaces (default: 2)
  name: old-name
```

```python
# create this runnable.py file, and run it
import os
from ruamel.yaml import YAML

def main():
    yaml = YAML()
    yaml.indent(mapping=2, offset=2, sequence=4)

    with open('values.yaml') as f:
        values = yaml.load(f)

    values['root-2'].update({'name': 'new-name'})

    with open('values.yaml', 'w') as f:
        yaml.dump(values, f)


main()
```

#### Look at `indent-status-2`, it has a indentation causing the yaml to become corrupt.
```yaml
# create this file and name it values.yaml
root-1:
  # offset=2 will indent the dash 2 spaces (default: 2)
  # sequence=4 will indent indent-status-2 4 spaces and ignore what offset is used (default: 2)
  - indent-status-1: ok
  indent-status-2: bad after ./test.py runs
root-2:
  # mapping=2 will indent the following entry 2 spaces (default: 2)
  name: new-name
```

consideRatio:
So the issue is that `sequence=4` is required to give valid yaml when for you have for example the following within your `values.yaml`. It becomes corrupt [when the `dump` method is invoked](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/373/commits/9701a909ad2ff3a01962116ac95083bc1d9879da#diff-2955d5257f635de1df53f55a171ca5c7R103).

```yaml
  # ...
  extraVolumeMounts:
    - name: volume-nbgrader
      mountPath: /home/jovyan/exchange
```

if not, it becomes this corrupt yaml using `sequence=2` as default indentation for some reason [discussed here](https://stackoverflow.com/questions/44388701/how-get-the-sequences-properly-indented-with-ruamel-yaml).

```yaml
  # ...
  extraVolumeMounts:
    - name: volume-nbgrader
    mountPath: /home/jovyan/exchange
```

NOTE: `typ=rt`, `mapping=2`, and `offset=2` are default values and could be removed, but I left them there...

closes #372

### DIFF
--- a/build.py
+++ b/build.py
@@ -9,7 +9,7 @@ from ruamel.yaml import YAML
 
 # use safe roundtrip yaml loader
 yaml = YAML(typ='rt')
-yaml.indent(offset=2)
+yaml.indent(mapping=2, offset=2, sequence=4)
 
 def last_modified_commit(*paths, **kwargs):
     return subprocess.check_output([


minrk:
Wow, that seems like a major bug in ruamel.yaml, if roundtrip's defaults are invalid.

consideRatio:
@minrk oops - I added a followup comment in #372 now.

> I was mistaken about the default value of offset: it was 0, and not 2. But it is still problematic. Increasing the default `offset=0` to `2 or more` without increasing `sequence` as well will corrupt valid yaml if the yaml contains lists of objects.
https://bitbucket.org/ruamel/yaml/issues/164/invalid-output-from-round-trip-dumper-with#comment-42330748

minrk:
Ah, gotcha. That makes more sense. Thanks!

PR #386, opened 2018-01-10
Pack strategy reduces node count.

ryanlovett:


yuvipanda:
Interesting. I guess it depends on what resources you are talking about! It increases utilization of specific nodes (since there are more pods on them than spread out), but when used with an autoscaler it decreases overall 'resource' (money? node count?). Not sure how to properly phrase it... /cc @choldgraf

ryanlovett:
Ah, right. I thought it was a typo but now see what you were getting at. Maybe something like:

"This increases the number of nodes that may be freed up by an autoscaler."

But I defer to you two.

choldgraf:
I like Ryan's description!

yuvipanda:
I was trying to figure out why CI was failing. It looks like it's because:

1. Ryan is using a branch in this repo rather than his fork
2. We are running branch and PR tests both. We only want PR tests normally, branch tests should be restricted to master.

I'll make a PR for (2)

yuvipanda:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/388 should fix it.

ryanlovett:
Sorry! Will fork next time. I used the Edit > "Create branch and make PR" process.

yuvipanda:
@choldgraf your comment was unclear, do you want to merge this as is?

PR #394, opened 2018-01-12
Setup a pdb for hub and proxy

### LINKED ISSUE #390: Create Pod Disruption Policies for Hub and Proxy by default
https://kubernetes.io/docs/tasks/run-application/configure-pdb/ allow us to say that hub and proxy pods should be left alone as much as possible. We should do so by default!

yuvipanda:
This provides a hint to autoscalers that nodes with these
services running should be considered last for scaling down. This
helps keep them from being disrupted unless needed.

Fixes #390


stefansedich:
@yuvipanda the PDB basically makes it impossible to drain nodes when performing node upgrades in an automated way, does it make sense that it is the default?

PR #396, opened 2018-01-12
move image requirements to requirements.txt

minrk:
makes it easier for bots and such to track out-of-date dependencies. This will let us inspect the requirements of our image with tools like requires.io or libraries.io. Still investigating what the best options are. It looks like libraries.io monitors dependencies, but can only do email notification. Requires.io appears to be able to attempt automatic pull requests like greenkeeper.


yuvipanda:
+1 to merging after! We might also want to use a frozen file? Or maybe use pipenv which supports locks?

PR #397, opened 2018-01-12
Move get_config and get_secret into a z2jh module

### LINKED ISSUE #374: extraConfigMap - get_config takes 0 positional arguments
[it sais in the docs](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/advanced.md#hubextraconfigmap) I should be able to use the `get_config` function (defined in jupyterhub_config.py) within hub.extraConfig, but when I do, an error is thrown saying: `TypeError: get_config() takes 0 positional arguments but 1 was given`

I currently workaround this by copying the `get_config` definition from `jupyterhub_config.py` into `hub.extraConfig`.

---

__An excerpt from my `values.yaml`__

```yaml
    extraConfig: |
      # Useful functions duplicated from z2jh/images/hub/jupyterhub_config.py
      # See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/374
      # -----------------------------------------------------------------------
      import os
      import yaml

      def get_config(key, default=None):
          """
          Find a config item of a given name & return it

          Parses everything as YAML, so lists and dicts are available too
          """
          path = os.path.join('/etc/jupyterhub/config', key)
          try:
              with open(path) as f:
                  data = yaml.safe_load(f)
                  return data
          except FileNotFoundError:
              return default

      def get_secret(key, default=None):
          """Get a secret from /etc/jupyterhub/secret"""
          path = os.path.join('/etc/jupyterhub/secret', key)
          try:
              with open(path) as f:
                  return f.read().strip()
          except FileNotFoundError:
              return default
      # -----------------------------------------------------------------------
```

yuvipanda:
This serves multiple purposes:

- Allows subconfigs loaded via hub.extraConfig to use get_config
  and get_secret
- Stops overloading the `get_config` method that is injected in
  config files by traitlets

Fixes #374


images/hub/z2jh/__init__.py
```diff
@@ -0,0 +1,31 @@
+"""
```
minrk: This can be `z2jh.py`. No need for a directory when a module is one file. In general, nontrivial code in `__init__` is good to avoid.

yuvipanda: Done

images/hub/Dockerfile
```diff
@@ -36,6 +36,9 @@ RUN pip3 install --no-cache-dir \
          -r /tmp/requirements.txt
 
 ADD jupyterhub_config.py /srv/jupyterhub_config.py
+
+# FIXME: Just adding this to current dir doesn't seem to work
```
minrk: Current directory is only on PATH when doing `python -i` or other such interactive things. CWD is not normally on PYTHONPATH when launching programs. Putting it here would work if we had:

```python
import os, sys
sys.path.insert(0, os.path.dirname(__file__))
```

in config.py.

yuvipanda: Hmm, but it worked when I did `jupyterhub -f jupyterhub_config.py` locally too...

yuvipanda: I'll re-test in a bit.

yuvipanda: never managed to re-test, but current code is good enough I guess :)

minrk:
Approved, pending Travis coming back to life.

PR #398, opened 2018-01-12
Support extraConfig being a dict

### LINKED ISSUE #320: Allow multiple extraConfig
Binder relies on `hub.extraConfig` to configure the Spawner. If we want to customize a given Binder *deployment* via `hub.extraConfig`, we have to copy/paste the hub.extraConfig from the binderhub chart and add to it.

It would be handy if extraConfig were something that could be added to without having to copy the previous value. Are dicts the only structure that can do this?

yuvipanda:
Using an alphabetically ordered dict also allows meta-charts to
*insert* config between specific upstream configs.

Fixes #320

### DIFF
--- a/doc/source/advanced.md
+++ b/doc/source/advanced.md
@@ -92,8 +92,10 @@ Code. Some examples of things you can do:
    supported in the helm chart
 
 Unfortunately, you have to write your python *in* your YAML file. There's no way
-to include a file in `config.yaml`. Remember to use a `|` for multi-line strings
-in YAML:
+to include a file in `config.yaml`.
+
+You can specify `hub.extraConfig` as a raw string (remember to use the `|` for multi-line
+YAML strings):
 
 ```yaml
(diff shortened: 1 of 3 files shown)


doc/source/advanced.md
```diff
@@ -104,6 +106,19 @@ hub:
     }
 ```
 
+You can also specify `hub.extraConfig` as a dictionary, if you want to logically
+split your customizations. The code will be evaluated in alphabetical sorted
+order of the key.
+
+```yaml
+hub:
+  extraConfig:
```
choldgraf: wouldn't there need to be a `|` after `extraConfig:`?

yuvipanda: Nope, because it's a dictionary. The | is a YAML feature that allows for multi-line strings, so we use it for the keys.

minrk:
Nice!

PR #399, opened 2018-01-12
Rewrite pre-puller & introduce image-puller

yuvipanda:
Currently, the pre-puller code is in https://github.com/yuvipanda/kube-image-puller,
which started out as a hopeful go project before becoming a bash project.

This commit is based off that repo, but completely rewrites most of it,
simplifying things drastically.

A daemonset is used to pull images onto nodes. The singleuser container image
is specified as an initContainer, causing it to be pulled. We run something
trivial in it (an `echo`), since what we really want to do is pull the image. This is
a much superior way than the old way of using a script to talk directly to the
docker API for multiple reasons:

1. We do not need access to the `/var/run/docker.sock` socket. This closes a 
    big security access point, since any attacker who gains access to `/var/run/docker.sock`
    can escalate to root on the entire cluster.
2. Authentication to various cloud providers' registries is handled for us by kubernetes,
    so we don't have to handle authentication. Previously we only support authentication
    for gcr.io, with the need to explicitly add support for other providers. This sidesteps
    all of that completely!
3. Less shell scripting in the universe, especially in secure contexts, is always a good thing!

In addition, we have a [helm hook](https://github.com/kubernetes/helm/blob/master/docs/charts_hooks.md)
that runs a Job on each `helm install` or `helm upgrade`. This is a small shell script
that checks to see if the image required is present on all nodes, runs a copy of the
daemonset if needed, waits for the images to be present on all nodes & then exits.
This is written as a shell script with jq and curl to keep the size of it as small as possible,
so we don't pull in a big image to pre-pull smaller images.

The `continuous` daemonset can / should be used with autoscaling clusters, since it
pre-pulls the image whenever a new node comes up. The `hook` job is enabled by
default for all installs, since we want to wait till the user image is in all nodes by default.
This makes the user servers start up much faster after install / upgrade, which is
behavior we want.

TODO:

- [x] Stop using labels, just look at images list in node status
- [x] Add documentation
- [x] Add RBAC support
- [x] Heavily comment the shell script
- [x] Add better logging and cleanup shell script
- [x] Fix imagePuller persistent daemon to work


jupyterhub/values.yaml
```diff
@@ -157,3 +164,5 @@ cull:
   timeout: 3600
   every: 600
 
+pullBlocker:
+  image: {name: yuvipanda/k8s-pull-blocker, tag: 2b02899}
```
minrk: why use inline dict here, rather than yaml-style like everywhere else?

yuvipanda: ah, pullBlocker is no longer used, this was a quick hack from earlier commits. I'll remove.

jacobtomlinson:
This is really cool.

Quick question, why does the DaemonSet get deleted at the end? Surely this will prevent the image being pulled to any new nodes added to the cluster.

yuvipanda:
@jacobtomlinson there are *two* daemonsets:

1. One created by the prePuller that is deleted at the end. This is used to block helm install / upgrade until images are prePulled
2. Another that's the imagePuller, which is a different daemonset instance that is left on (if enabled explicitly - this is disabled by default). This is for pulling on dynamic new nodes

There's some fromYaml / toJson trickery so we don't have to keep this daemonset definition in two different places, but they are used in two different purposes!

yuvipanda:
Am currently cleaning up the bash code a little bit more, will write up docs right after!

yuvipanda:
I can also confirm that this pre-puller works on Azure AKS, which was not the case for the previous one!

yuvipanda:
In the course of debugging this, I've found and filed a helm chart error: https://github.com/kubernetes/helm/issues/3348

yuvipanda:
I spent some more time thinking about it, and requiring raw access to the docker socket for pre-pulling seems like something that wouldn't fly in any secure cluster. The ability to have $CURL_EXTRA_OPTIONS basically allows shell injection in ways that would allow users to get root, so I think we need an alternate approach! I've an idea for such an alternate approach that wouldn't require docker.sock, and am working on it now.

yuvipanda:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/399/commits/8ab304718b0638e8d68d0d6ed47f289b29a57608 implements the shellscript-less approach, which feels sortof obvious in hindsight!

yuvipanda:
ok, the code is in good shape now. I'm going to write docs + rewrite the PR description to match final reality!

yuvipanda:
Added docs, redid PR description! This is ready for review now!

minrk:
Code approved by me, pending Travis recovery. The minikube change means the new prepuller will run on the Travis tests, right?

Requested review of new docs from @willingc.

yuvipanda:
@minrk thanks! yes, the minikube change should run and test the prepuller too once travis wakes up!

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
 singleuser:
     schedulerStrategy: pack
 ```
+
+## Pre-pulling Images for Faster Startup
+
+When user images are of non-trivial sizes (as most of them are), **pre-pulling**
```
willingc: Pulling and building a user's images forces a user to wait before the user's server is started. Sometimes, the wait can be 5 to 10 minutes. **Pre-pulling** the images on all the nodes can cut this wait time to a few seconds. Let's look at how pre-pulling works.

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+the images on all nodes significantly decreases the amount of time it takes to
+start a user's server. Without pre-pulling, sometimes it can take upto 5-10
+minutes for a user's server to launch. With pre-pulling, it is cut down to a few
+seconds in most cases.
+
+By default, the user container image is pulled on to all nodes whenever a `helm
```
willingc: With **pre-pulling**, which is enabled by default, the user's container image is pulled on all nodes...

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+install` or `helm upgrade` is performed. This causes `helm install` and `helm
```
willingc: While this causes..., this time makes the user startup experience faster and more pleasant.

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+upgrade` to take several minutes, but mostly that is worth it since it makes
+user startup performance much faster. By default `helm install` or `helm
```
willingc: New paragraph:

With the default **pre-pulling** setting, a `helm install` or `helm upgrade` will cause the system to wait for 5 minutes to begin pulling the images before timing out. This wait time is configurable by passing the `--wait <seconds>` flag to the `helm` commands.

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
 ```
+
+## Pre-pulling Images for Faster Startup
+
+When user images are of non-trivial sizes (as most of them are), **pre-pulling**
+the images on all nodes significantly decreases the amount of time it takes to
+start a user's server. Without pre-pulling, sometimes it can take upto 5-10
+minutes for a user's server to launch. With pre-pulling, it is cut down to a few
+seconds in most cases.
+
```
willingc: `### Pre-pulling usage and configuration`

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+By default, the user container image is pulled on to all nodes whenever a `helm
+install` or `helm upgrade` is performed. This causes `helm install` and `helm
+upgrade` to take several minutes, but mostly that is worth it since it makes
+user startup performance much faster. By default `helm install` or `helm
+upgrade` will wait for 5minutes before giving up. You can configure this by
+passing the `--wait <seconds>` flag to either of them.
+
+To disable the pre-pulling during `helm install` and `helm upgrade`, you can use
```
willingc: We recommend using pre-pulling. For the rare cases where you have a good reason to disable it, pre-pulling can be disabled. To disable the pre-pulling during `helm install` and `helm upgrade`, you can use...


~~This is not recommended...it!~~

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+
+```yaml
+prePuller:
+   hook:
+     enabled: false
+```
+
+This is not recommended unless you have specific good reasons to disable it!
+
+### Pre-Pulling for dynamically sized clusters
```
willingc: Pre-Pulling in a dynamically sized cluster

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+
+If you add new nodes to a cluster (manually or with autoscaling), the new nodes
```
willingc: Cluster size can change through manual addition of nodes or autoscaling. When a new node is added to the cluster, the new node does not yet have the user image. A user using this new node would be forced to wait while the image is pulled from scratch. Ideally, it would be helpful to pre-pull images when the new node is added to the cluster.

By enabling the **continuous pre-puller** (default state is disabled), the user image will be pre-pulled when adding a new node. When enabled, the **continuous pre-puller** runs as a
 daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) to force kubernetes to pull the user image on all nodes as soon as a node is present. The continuous pre-puller uses minimal resources on all nodes and greatly speeds up the user pod start time.

doc/source/advanced.md
```diff
@@ -199,3 +199,50 @@ To explicitly specify this strategy, use the following in your `config.yaml`:
...
+will not have the user image. When a user pod starts on the new node, it will
+need to pull the image from scratch - making it very slow. This is undesirable.
+To fix this, you can enable the **continuous pre-puller**, which will run as a
+[daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) that
+takes minimal resources on all nodes. This will force kubernetes
+to pull the user image on all nodes as soon as a node is present, thus making
+the user pod start time much faster.
+
+The continuous pre-puller is not enabled by default. To enable it, use the
```
willingc: ~~not enabled~~ disabled

willingc:
Happy to have you merge and I can update the docs after.

yuvipanda:
@willingc thank you for the review! I would love the workflow where this gets merged and the docs get edited by you after :) I often feel that's a much quicker and less frustrating editing experience for everyone!

I'm investigating the travis failures now to see if that's a result of the Travis outage yesterday or a problem here.

yuvipanda:
It was a legit failure, I think I fixed it in latest push. Let's wait for Travis now...

yuvipanda:
@willingc I'll address the comments about comments in the scripts now!

images/pre-puller/block.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/sh
+# Script that does the following actions:
+#  1. Create an image-puller daemonset to fetch the user image on all nodes
+#  2. Check if all nodes have the user image present in them in a loop
```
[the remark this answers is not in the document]
yuvipanda: I've rephrased it!

images/pre-puller/block.sh
```diff
@@ -0,0 +1,94 @@
...
+#     A single line JSON of the image-puller daemonset to create
+#  2. KUBERNETES_SERVICE_HOST, KUBERNETES_SERVICE_PORT
+#     The hostname and port to use for talking to the k8s API.
+#     When running in cluster, this is automatically set by Kubernetes
+#  3. IMAGE
+#     Full name of user image
+#  4. CURL_EXTRA_OPTIONS
+#     Extra commandline options to pass to curl
+#
+# jq & curl are required to run this script. Complex jq scripts are kept as separate files.
```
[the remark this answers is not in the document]
yuvipanda: Done

images/pre-puller/block.sh
```diff
@@ -0,0 +1,94 @@
...
+#
+# This script is designed to be run from inside a kubernetes cluster only.
```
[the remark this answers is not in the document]
yuvipanda: Done

yuvipanda:
w00t, tests caught a real failure that I couldn't detect locally and helped us fix it. Yay tests :) Travis is green now!

willingc:
Yay tests! Merging

yuvipanda:
Thanks for the review, @willingc and @minrk! And thanks for the original inspiration, @jacobtomlinson!

PR #411, opened 2018-01-15
Cleanup Azure AKS install instructions

yuvipanda:
- Most people don't need to create ad accounts manually,
  so remove it.
- Do not use environment variables. People might not know what
  they are, and for the people who do, they can use it even if
  we do not mention it
- Add several missed steps that someone on a new account might
  need, such as enabling the APIs and creating a special SSH key  

Things to do:

- [ ] Note AKS limitations - no RBAC and no prePuller, and explain how to
      work around for both helm init and jupyterhub itself

### DIFF
--- a/doc/source/create-k8s-cluster.rst
+++ b/doc/source/create-k8s-cluster.rst
@@ -216,6 +216,15 @@ Setting up Kubernetes on Microsoft Azure Container Service (AKS)
 
    The response should list three running nodes and their kubernetes versions!
 
+.. note::
+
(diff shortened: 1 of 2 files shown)


yuvipanda:
@ryanlovett note that I used my berkeley.edu account and did not have to create any special purpose ad accounts.

ryanlovett:
> I did not need to run the az provider register commands.

Interesting, I did need to do that today for a new cluster.

PR #412, opened 2018-01-15
Fixup AKS instructions + include helm instructions for AKS

yuvipanda:
Followup to #411


choldgraf:
this looks good to me - I'll merge this now since I'll take a pass through these instructions this week as well and can spot-check confusion points then

PR #416, opened 2018-01-17
Block access to cloud metadata endpoints by default

### LINKED ISSUE #407: Block access to cloud metadata endpoint by default
Currently, we expect users to do extra steps to secure their clusters from users accessing the cloud metadata endpoints (https://zero-to-jupyterhub.readthedocs.io/en/v0.5-doc/security.html#audit-cloud-metadata-server-security). IMO, we should instead do that by default and allow users to opt out of it. Most users won't actually be doing the blocking right now, and run insecure clusters...

yuvipanda:
Uses an iptables block to block access to the cloud metadata
ip used by the most common cloud providers (Google, AWS, Azure).
This is enabled by default so we can be secure by default.

Fixes #407 

TODO:

- [x] Pre-pull this image too
- [x] Fixup security docs to refer to this

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -1,6 +1,7 @@
 import os
 import glob
 from tornado.httpclient import AsyncHTTPClient
+from kubernetes import client
 
 from z2jh import get_config, get_secret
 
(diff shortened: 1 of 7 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -104,7 +105,7 @@
 
 init_containers = get_config('singleuser.init-containers')
 if init_containers:
-    c.KubeSpawner.singleuser_init_containers = init_containers
+    c.KubeSpawner.singleuser_init_containers += init_containers
```
Carreau: Does that work on traitlets ?

minrk: `list_trait.extend(other_list)` explicitly works. If `+=` calls extend, this is fine. I'm not sure if that's explicit/official, though so might be reliant on an unguaranteed implementation detail. Switching this to .extend would make me more comfortable.

yuvipanda:
w00t, tyvm @Carreau!

PR #417, opened 2018-01-17
Fixup z2jh module for py3.6 switch

yuvipanda:


yuvipanda:
Incorporated into #419 since both commits are needed together to fix travis failures.

PR #418, opened 2018-01-17
Add pre-puller content to docs

willingc:
Follow up to #399

### DIFF
--- a/doc/source/advanced.md
+++ b/doc/source/advanced.md
@@ -1,7 +1,14 @@
 # Advanced Topics
 
 This page contains a grab bag of various useful topics that don't have an easy
-home elsewhere. Most people setting up JupyterHubs on popular public clouds
+home elsewhere:
+
+- Ingress
+- Arbitrary code in `jupyterhub_config.py`
+- Picking a Scheduler Strategy
+- ## Pre-pulling Images for Faster Startup
+
+Most people setting up JupyterHubs on popular public clouds
 should not have to use any of this information, but it is essential for more
 complex installations.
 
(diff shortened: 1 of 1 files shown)


doc/source/advanced.md
```diff
@@ -1,7 +1,14 @@
 # Advanced Topics
 
 This page contains a grab bag of various useful topics that don't have an easy
-home elsewhere. Most people setting up JupyterHubs on popular public clouds
+home elsewhere:
+
+- Ingress
```
choldgraf: Here could we just use:

```
.. contents::
   :local:
   :depth: 1
```
?

choldgraf: (that way it'd auto-update and also add links to the sections)

yuvipanda: This is recommonmark, not sure it supports it?

yuvipanda:
Thanks, @willingc! Going to  merge this despite travis failures (#419 to address travis failures)

doc/source/advanced.md
willingc: @choldgraf I couldn't get this to work using recommonmark. If you can, please do :-)

choldgraf: I just gave it a shot and couldn't get it to work either. I have a feeling it's a bug in recommonmark :-/

willingc: I did some investigation the other day for the JupyterLab folks re: recommonmark and it hasn't had a release since last May. There are some issues with compatibility and other things. We may need to fix recommonmark :( at some point or look at a docutils/sphinx solution. For now, it seems fine if we don't get fancy. @choldraf :(

PR #423, opened 2018-01-18
improving user culling docs

choldgraf:
Documentation on user culling behavior.

### DIFF
--- a/doc/source/user-management.md
+++ b/doc/source/user-management.md
@@ -9,7 +9,8 @@ for a period of time. This helps free up computational resources and keeps
 costs down if you are using an autoscaling cluster.
 When these users navigate back to your JupyterHub, they will
 have to start their server again, and the state of their previous session
-will be lost. This is known as "culling".
+(variables they've created, any in-memory data, etc)
+will be lost. This is known as *culling*.
 
 ```eval_rst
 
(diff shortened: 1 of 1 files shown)


doc/source/user-management.md
```diff
@@ -3,6 +3,51 @@
 This section describes management of users and their
 permissions on JupyterHub.
 
+## Culling user pods
+JupyterHub will automatically delete any user pods that have no activity
+for a period of time. This helps free up computational resources and keeps
+costs down if you are using an autoscaling cluster.
+When these users navigate back to your JupyterHub, they will
+have to start their server again, and the state of their previous session
```
yuvipanda: Should we explain what is 'state' here?

doc/source/user-management.md
```diff
@@ -3,6 +3,51 @@
...
+will be lost. This is known as "culling".
```
yuvipanda: culling should be bold or italics?

doc/source/user-management.md
```diff
@@ -3,6 +3,51 @@
...
+
+```eval_rst
+
+.. note::
```
yuvipanda: Is there a reason this is a ..note rather than just a paragraph?

choldgraf: to me the paragraph about inactivity is *related* to culling but kind of breaks the flow of the culling configuration instructions. It feels like an aside, which is why I made it a note.

choldgraf:
comments addressed!

PR #424, opened 2018-01-18
Update create-k8s-cluster.rst

bdshadow:
1.8.4-gke.1 is not valid anymore. 1.8.6-gke.0 must be used


choldgraf:
@yuvipanda do you know if google will have some kind of LTS version available at some point? Or is this software just iterating too quickly at the moment?

@bdshadow can you confirm that the installation instructions all still work with the new version?

choldgraf:
(and thanks for noticing / making the PR! much appreciated!)

choldgraf:
Looks from here: https://cloud.google.com/kubernetes-engine/supported-versions like this is indeed true. Merging!

PR #426, opened 2018-01-19
updating azure instructions, standardizing capitalization, adding azure teardown

choldgraf:
This does the following:

* Standardizes capitalization of variables in the instructions
* Improves the Azure installation instructions (and confirmed that they work)
* Adds a section on the Azure interactive CLI
* Adds an Azure tear-down section at the end.


doc/source/setup-helm.rst
```diff
@@ -26,7 +26,13 @@ Initialization
 --------------
 
 After installing helm on your machine, initialize helm on your Kubernetes
-cluster. At the terminal, enter:
+cluster. There are two pathways to do this. One if you are running on Microsoft
+Azure, and another if you're running on another provider.
+
+If you are not on Azure
```
yuvipanda: hmm, I'm loathe to make this happen for several reasons:

1. It talks about particular vendors rather than particular features deep inside the docs
2. It is only true for AKS, not all of Azure (there are other ways to set up Kubernetes this on Azure).

I think I'm happy mentioning Azure AKS as an example here, but not as a subheading.

doc/source/create-k8s-cluster.rst
```diff
@@ -233,7 +266,8 @@ Setting up Kubernetes on Microsoft Azure Container Service (AKS)
 
    1. You have to `not use RBAC <security.html#use-role-based-access-control-rbac>`_, since AKS does not support it
       yet.
-   2. You should skip step 2 (granting RBAC rights) :ref:`when setting up helm <helm-rbac>`.
+   2. You cannot grant RBAC rights on Azure, so double-check that you are running
```
yuvipanda: "You cannot grant RBAC rights on Azure" also seems wrong, since RBAC is entirely no working there, so it seems redundant. Also, it's only true for Azure AKS.

doc/source/setup-helm.rst
choldgraf: Sounds good - mostly I did this because the current instructions are confusing as to what we mean by "step 2". Is it the second section of the `setup-helm` page? Or is it *just* number "2" within the "initialization" section https://zero-to-jupyterhub-with-kubernetes.readthedocs.io/en/latest/setup-helm.html#initialization

yuvipanda: it's only number 2 within the initialization section.

I agree it's confusing, but given AKS is in preview now and they're constantly fixing it I think it's better to keep our vendor specific stuff confined...

yuvipanda: You *could* skip the whole thing and just run `helm init` actually, 'just skip step 2' was the 'minimal diff from what we do for more mature vendors'.

choldgraf: I'm happy to keep the instructions the same, just think they could be more explicit. I'll push that change in a sec

choldgraf:
updated per your comments

doc/source/setup-helm.rst
```diff
@@ -38,7 +38,11 @@ cluster. At the terminal, enter:
 
 2. Give the ``ServiceAccount`` `RBAC
    <https://kubernetes.io/docs/admin/authorization/rbac/>`_ full permissions to
-   manage the cluser. While most clusters have RBAC enabled and you need this
+   manage the cluser.
+
+   **If you're on Azure AKS, you should skip directly to step 3.**
```
yuvipanda: Shouldn't this be at the end of step 1 and not step 2? I'm happy either way.

choldgraf: good point - updated

choldgraf:
comment addressed - you don't think that the RBAC errors I was getting are due to incorrect instructions? I was gonna hesitate on merging this until we can be sure that there's not something misleading in there.

yuvipanda:
@choldgraf yeah, am pretty sure it's from performing step 2, since it explicitly states that the error is `"cluster-admin" not found` and that's the 'cluster role' we gave to tiller in step 2.  I left more comments in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/410

choldgraf:
ok, latest push adds updated instructions, including a section on local CLI + online interactive shell. I think this is ready to go from my end.

doc/source/turn-off.rst
```diff
@@ -70,10 +73,36 @@ Google Cloud Platform
...
+
+1. Perform the steps in :ref:`delete-namespace`. These cloud provider agnostic steps will
+   delete the helm chart and delete the hub's namespace. This must be done before proceeding.
+
+2. Delete your resource group. You can list your active resource groups with
+   the following command
+
+   .. code-block:: bash
+
+      az aks list --output table
```
yuvipanda: Should this be `az group list`?

doc/source/turn-off.rst
```diff
@@ -70,10 +73,36 @@ Google Cloud Platform
...
+
+   You can then delete the one you want with the following command
+
+   .. code-block:: bash
+
+      az group delete -n <YOUR-GROUP-NAME>
```
yuvipanda: We specify '--name' everywhere, so specify this here too?

doc/source/turn-off.rst
```diff
@@ -70,10 +73,36 @@ Google Cloud Platform
...
+
+3. Double check to make sure all the resources are now deleted, since anything you
+   have not deleted will cost you money! You can check the `web portal <https://portal.azure.com>`_
+   (check the "Resource Groups" page) to verify that everything has been deleted.
+
+   These might take several minutes to clear up, but they shouldn't have anything
+   related to your JupyterHub cluster after you have deleted the cluster.
```
yuvipanda: deleted the resource group?

doc/source/turn-off.rst
yuvipanda: We should also mention that this is the resource group you created, and a note of danger if you delete a different resource group

doc/source/setup-jupyterhub.rst
```diff
@@ -40,11 +40,22 @@ For the following steps, use your favorite code editor. We'll use the
...
+
+   .. code-block:: yaml
+
+      rbac:
+         enabled: false
+
+   See the `RBAC documentation <security.html#use-role-based-access-control-rbac>`_
+   for more details.
+
+5. Save the ``config.yaml`` file. If using the nano editor, hit ``Ctrl-X`` and
```
yuvipanda: Do we have nano specific commands in other places? If not IMO we should remove that from here too.

doc/source/setup-jupyterhub.rst
```diff
@@ -40,11 +40,22 @@ For the following steps, use your favorite code editor. We'll use the
    .. code-block:: yaml
 
       proxy:
-        secretToken: "RANDOM_STRING"
+        secretToken: "<OUTPUT-OF-`openssl rand -hex 32`>"
 
 .. Don't put an example here! People will just copy paste that & that's a security issue.
 
-4. Save the ``config.yaml`` file. If using the nano editor, hit ``Ctrl-X`` and
+4. **Azure AKS only** If you're on Microsoft Azure AKS, you must disable
```
yuvipanda: This feels very hacky, but I guess there isn't really a clear way around it. Hopefully AKS will get a little bit more mature soon and we can get rid of it.

choldgraf: Yeah I'm assuming we'll re-write this as the process changes

choldgraf:
comments addressed!

yuvipanda:
This LGTM! I'll wait a bit for @willingc to take a look if she wants, otherwise can merge!

Thanks for churning this through, @choldgraf!

doc/source/turn-off.rst
```diff
@@ -14,10 +14,14 @@ Tearing down your JupyterHub entails:
...
 
-
 For all cloud providers
 -----------------------
 
+.. _delete-namespace:
+
+Delete the helm namespace
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
willingc: Should we go ahead and be consistent with the caps on step 1 and 2.

choldgraf: Hey - could you clarify what you mean? I'm not sure which caps you're talking about

willingc: In the next few lines (which are not edited in the initial PR):

helm delete <your-helm-release-name> --purge

helm delete <YOUR-HELM-RELEASE-NAME> -- purge

kubectl delete namespace <YOUR-NAMESPACE>

choldgraf:
@willingc ahhh I see - good point, I missed that one! ok changes made

PR #438, opened 2018-01-23
Rewrite prepuller in go

yuvipanda:
- This lets us easily wait for multiple images to show up,
  rather than just one image
- Add ability to specify multiple images to wait for in prePuller
  via config.yaml. This is very useful for people who are using
  multiple images via extraConfig.
- We don't use the kubernetes client library or any dependencies,
  because Go's dependency management is very immature. Check this
  out for example, on how to use the kubernetes go client as a
  dependency: https://github.com/kubernetes/client-go/blob/master/INSTALL.md#dependency-management-for-the-serious-or-reluctant-user
- This is smaller than our previous alpine + bash image, at about 4MB (vs 6MB)


doc/source/advanced.md
```diff
@@ -279,3 +279,21 @@ prePuller:
...
+
+By default, the pre-puller only pulls the singleuser image & the networktools image (if
+access to cloud metadata is disabled). If you have customizations that need additional
+images present on all nodes, you can ask the pre-puller to also pull an arbitrary number
+of additional images.
+
+```yaml
+prePuller:
+   extraImages:
+     ubuntu:
```
choldgraf: can each entry just after `extraImages` be named anything? It's a bit hard to tell since you're calling it `ubuntu` but it's also just pulling `ubuntu`

choldgraf:
I am probably +1 on using Go, but just a general note that however small the # of scientists is that are skilled in bash, the number of scientists that understand Go is a lot smaller...

yuvipanda:
@choldgraf I agree re adoption, however I think far more people are going to be able to understand making:

1. HTTPS requests with client and server certificates
2. Doing JSON munging

with Go than with Bash, since it's not just bash you have to know but bash + somewhat deep curl + jq. With Go we have a chance of getting this right, with bash IMO the chance is virtually 0. Nobody really knows bash, you only think you do. You need to understand kubernetes API to touch this code, and IMO that's a much bigger barrier than language.

I'm going to add more comments justifying *why* this is on go, and why our options were basically Go / Rust / Bash (mostly because this image needed to be as small as possible).

If someone else more skilled in bash wants to rewrite this back into bash I'm happy for them to do so :)

doc/source/advanced.md
yuvipanda: Yes, it can be named anything (no underscores I think). Do you think you'll have time to edit this and push changes too? :)

yuvipanda:
@choldgraf I've added a README now!

consideRatio:
@choldgraf I'm not very skilled in Bash or Go, I've had to learn a bit about both, and i felt really bad about learning Go.

Assuming only say 10% knows anything, I'd love to see 90% get to learn Go instead of Bash... When i realized googling for bash help was a challenge in itself (what does -e do in [ -e ... ] for example) i got really sad about learning bash.

Make as few people as possible learn bash: BIG WIN.

images/pre-puller/daemonset.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	_, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
```
cam72cam: Couldn't this be simplified as "return err" ?  Are you doing this to stick with the above error handing convention?

images/pre-puller/main.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		if imagesPresent(nodes, images) {
+			log.Printf("All images present on all nodes!")
+			break
+		}
+
+		time.Sleep(2 * time.Second)
```
cam72cam: Should there be a flag for the timeout between checks or is 2s a sane enough default not to worry about it?

cam72cam:
@yuvipanda I think we should either add some comments describing why we are not using the k8s client library or file a ticket to adopt it across the board once we choose a dep management tool.  I understand that they are in flux with choosing the next tool and converting too it, but I also think that redoing the work should be avoided if possible.

doc/source/advanced.md
choldgraf: yeah I can take an editing pass before we merge, on this point I just wasn't sure what the actual behavior was. We don't yet have docs on actually using these extra images, right?

images/pre-puller/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+1. Bash
+2. Go
+3. Rust
+
+All these can produce images < 10MB. Originally this was in Bash, but it very
+quickly became hard to maintain (since we need to set up TLS and talk to a JSON
+HTTP API & munge the resulting JSON). There was very little error handling code,
+and extending it was very bad. This made it clear it needed to be in a real
+programming language. Go is a little bit more 'mature' now than Rust is, and so
```
choldgraf: Just a thought here - there are a few moments where you (or the docs) give opinionated statements about the relative merits of Bash / Go / etc. Do you know of at least a blog post we could link people to that discusses any of this? I believe you that Go is a better language to implement this in, but I could see, e.g., a person who had recently taken software carpentry (which teaches bash and tells students that they should *really* learn bash in doing their analyses) getting frustrated by the opinionated language w/o any material to back it up

images/pre-puller/Dockerfile
```diff
@@ -1,7 +1,9 @@
...
+FROM golang:1.9.2
 
-RUN apk add --no-cache curl jq
+COPY *.go /go/
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags '-w -s' -installsuffix cgo -a -o /go/block /go/*.go
 
-ADD . .
+FROM scratch
+
+COPY --from=0 /go/block /block
```
Carreau: Nice ! Multi stage build !

images/pre-puller/README.md
Carreau: I dont' think that in any way this diminishes the value of Bash or Rust, I think it does just show that different times in the lifetime of a project might require different languages. Bash was nice at the beginning (and is ubiquitous), now we just need a more robust language. 

There is of course the fact that the language you know is likely better suited as you know it.

athornton: I don't think there's any shame in saying "Hey, this particular component outgrew bash.  Once you find yourself really wanting a map data type, it's probably time to consider whether sticking to shell is the right thing."

IMHO Rust is about 4-5 years behind Go in terms of maturity of its toolset and ecosystem.  I also find its object lifecycle management harder to wrap my head around than Go, but I cut my teeth on C and so I have a bias towards Go, since in many ways it really is C with 30 years of lessons learned.

choldgraf: I totally agree with you - I'm just saying we need to be careful about saying "X is better than Y" without providing evidence for it. It's important to be careful not to accidentally introduce the kind of "my tools are better than yours" mentality that can sometimes have unintended consequences on the reader, especially those who are relatively inexperienced in the devops world (e.g. all scientists). E.g., linking out to a page like [this](http://goroutines.com/shell) or [this](https://eng.heetch.com/go-as-a-scripting-language-eeaa0da842ea) could help.

yuvipanda: I wrote out a really long reply here but then realized I was arguing about relative benefits of programming languages, which I've a personal ban on :)

I've instead pushed a commit that removes all references to bash or rust from the README. I wrote it that way to have it pre-empt any language wars, but now it's clear that it'll have exactly the opposite effect. LMK if this works for you, @choldgraf?

Also, I appreciate you thinking about how this might feel incendiary to a population that I do not have much insights to. I agree with what everyone else has said in this thread, but also agree that we should take extra steps to avoid putting anything that might be even vaguely interpreted as 'X sucks' in an official, non-personal capacity. Thanks!

yuvipanda: The only sentiment I'm trying to convey here is:

1. Why is this in Go while the rest of the project is in Python?
2. Because we want to have a very small image size.

The fact that this code was in bash at some point is a historical accident that seems best to ignore in our docs.

yuvipanda: Ok, I simplified it even more, making it into a FAQ that addresses both @choldgraf's concerns here and @cam72cam's questions about why we aren't using the kubernetes go client library.

images/pre-puller/daemonset.go
yuvipanda: @cam72cam ah, I think for now I just want to stick to the convention if that's ok.

choldgraf:
+1 @yuvipanda , lemme know if/when you're ready for a merge and then I can take a quick pass through docs for clarity etc. Appreciate yall bearing with me

doc/source/advanced.md
yuvipanda: Yeah, we don't have any docs. That can come later I think - my hope is to have @ryanlovett figure out all the best practices here and then we can document them.

images/pre-puller/README.md
yuvipanda: I've also edited this PR title and description to remove things that could possibly be used as ammo in any 'X sucks' conversation.

yuvipanda:
@choldgraf ready to go from my perspective :) Thank you for helping us avoid pitfalls that might not necessarily be visible to us :)

images/pre-puller/Dockerfile
consideRatio: Another thing learned by subscribing to this repo's development! :heart:

choldgraf:
made a commit for the readme, I'm +1

willingc:
@yuvipanda Will take a look at the Go code today.

yuvipanda:
Thanks for the clarifying commit, @choldgraf. +1!

And thanks for the review, @minrk! Thanks for taking the time to review this, @willingc! I appreicate it :)

images/pre-puller/main.go
willingc: The default currently seems reasonable. Let's keep it simple for now. But if we see it's not adequate, we can take up @cam72cam's suggestion in the future.

yuvipanda:
w00t, thanks @willingc! I'll merge now!

PR #439, opened 2018-01-23
bump singleuser-sample base image

minrk:
should fix path/permissions issues when container is configured to run with elevated permissions, which were briefly broken in docker-stacks and we just so happened to bump the base image while that was true.

cc @cam72cam who reported this on gitter


cam72cam:
I can do a quick test run in my setup to confirm this resolves the issue.
Thanks for setting up the PR @minrk

minrk:
Thanks for testing!

PR #441, opened 2018-01-23
Add a link to Dockerfile for default image

betatim:
Took me a while to find the configuration for the default image of the chart so added a link to the guide.

### DIFF
--- a/doc/source/user-environment.rst
+++ b/doc/source/user-environment.rst
@@ -36,7 +36,7 @@ Use an existing Docker image
 .. note::
 
    You can find the configuration for the default Docker image used in this
-   guide `here <https://github.com/jupyterhub/zero-to-jupyterhub-k8s>`_.
+   guide `here <https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/images/singleuser-sample>`_.
 
 Using an existing Docker image, that someone else has written and maintained,
 is the simplest approach. For example, Project Jupyter maintains the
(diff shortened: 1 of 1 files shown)


doc/source/user-environment.rst
```diff
@@ -33,6 +33,11 @@ Use an existing Docker image
    match the version installed by the helm chart that you're using. For example,
    ``v0.5`` of the helm chart uses  ``jupyterhub==0.8``.
 
+.. note::
+
+   You can find the configuration for the default Docker image used in this
+   guide `here <https://github.com/jupyterhub/zero-to-jupyterhub-k8s>`_.
```
choldgraf: I'm a bit confused - there's no Dockerfile in the root of the z2jh repository...is that the right link?

cam72cam: Might be looking for https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/singleuser-sample/Dockerfile which is broken until https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/439 is merged

cam72cam:
This would have saved me at least a half hour :+1:

doc/source/user-environment.rst
betatim: Indeed, pasting links while not paying attention 🚓

betatim: I linked to the directory so people also get the README etc.

yuvipanda: We need to rewrite the README completely, since it's 100% out of date...

yuvipanda:
Thanks for the PR, @betatim!

PR #444, opened 2018-01-23
Bump kubespawner version

yuvipanda:
This will be released as KubeSpawner 0.8.


minrk:
Wait for the release itself, rather than pulling from git?

yuvipanda:
@minrk yep, released and updated commit here!

PR #445, opened 2018-01-24
Minor fixes on 'Use JupyterLab by default' section

martjanz:


choldgraf:
Much appreciated! These fixes are particularly useful as we are about to release v0.6 so there will be another doc freeze at that point. 🥇

PR #458, opened 2018-01-29
Handle the 'NoSchedule' taint on nodes properly

### LINKED ISSUE #454: prePuller doesn't understand noschedule taint
In the pre-puller, we only check for the `unschedulable: true` spec to see if a daemonset will get scheduled there. This accounts for cordoned nodes, but not for tainted nodes.

We should check for:

```
  taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    timeAdded: null
    value: "true"
```

as well in the pre-puller.

yuvipanda:
This is usually used to prevent pods from scheduling on a node
without explicit tolerations. The defaults for many cluster management
setups (such as https://github.com/Azure/acs-engine and I think
many amazon setups) use this to prevent master from being scheduled,
so we should respect this.

Fixes #454

### DIFF
--- a/images/pre-puller/node.go
+++ b/images/pre-puller/node.go
@@ -13,19 +13,24 @@ import (
 // Only contains fields we will actively be using, to make JSON parsing nicer
 type NodeList struct {
 	Kind  string `json:"kind"`
-	Items []struct {
-		Metadata struct {
-			Name string `json:"name"`
-		} `json:"metadata"`
-		Spec struct {
-			Unschedulable bool `json:"unschedulable"`
-		} `json:"spec"`
-		Status struct {
-			Images []struct {
-				Names []string `json:"names"`
-			} `json:"images"`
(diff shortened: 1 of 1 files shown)


yuvipanda:
I'm pretty sure we need this for 0.6

images/pre-puller/node.go
```diff
@@ -64,12 +69,23 @@ func getNodes(transport *http.Transport, server string, headers map[string]strin
...
+func isNodeSchedulable(node *Node) bool {
+	if len(node.Spec.Taints) != 0 {
+		for _, taint := range node.Spec.Taints {
+			if taint.Effect == "NoSchedule" {
+				return false
+			}
+		}
+	}
+
+	return !node.Spec.Unschedulable
```
willingc: Gotta love the double negatives.

PR #463, opened 2018-01-30
Edit azure docs for cluster set up

willingc:
Addresses #410


willingc:
@choldgraf Ready for review since we won't migrate from 0.5->0.6 for AKS.

choldgraf:
looks great! found one small type-o. thanks for taking a pass :-)

willingc:
Good catch @choldgraf

PR #467, opened 2018-01-31
Dev instructions for new releases

choldgraf:
This adds more fleshed out instructions for cutting a new release, and was inspired by the most recent (v0.6) release of JupyterHub's helm chart. Somebody other than @choldgraf should test this out the next time we need to make a release!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,19 +54,144 @@ development.
    under `images`, but only step 7 if you changed things only under `jupyterhub`
 
 
-## Releasing a new version of the Helm Chart
+## Releasing a new version of the helm chart
 
-In order to release a new version of the Helm Chart, make sure to perform each
-of the following steps:
+The following steps can be followed to release a new version of the Helm Chart.
+This should happen approximately once every 5-7 weeks.
 
-1. Generate a release document. This should follow the structure of previous
-   entries in the [CHANGELOG](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md).
-2. Double-check that there aren't any un-documented breaking changes.
-3. If applicable, include a section in the [CHANGELOG](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md)
-   that contains upgrade instructions (usually only applicable for major changes to the code).
-4. Generate a list of contributors since the latest release. Use the script
-  in `tools/contributors.py` to list all contributions (anyone who made a
-  commit or a comment)    since the latest release. Update the dates in that
-  script, run it, and paste the output into the changelog. For an example,
-  see [the v0.5 list of contributors](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/v0.5/CHANGELOG.md#contributors).
-5. ...
+### Make a CHANGELOG
+
+This needs to be manually created, following the format of
+current [CHANGELOG](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md). The general structure should be:
+
+* A short description of the general theme / points of interest for
+ this release.
+* Breaking changes + a link to the [upgrade instructions](https://zero-to-jupyterhub.readthedocs.io/en/v0.5-doc/upgrading.html) in the docs
+* A list of features with brief descriptions under each.
+* The contributor list mentioned in the section below.
+
+### Add list of contributors
+
+We try to recognize *all* sorts of contributors, rather
+than just code committers.
+
+Use the script in `tools/contributors.py` to list all
+contributions (anyone who made a commit or a comment)
+since the latest release. For each
+release, you'll need to find the versions of all repos
+involved:
+
+* [z2jh](https://github.com/jupyterhub/zero-to-jupyterhub-k8s)
+* [KubeSpawner](https://github.com/jupyterhub/kubespawner)
+* [JupyterHub](https://github.com/jupyterhub/jupyterhub)
+* [OAuthenticator](https://github.com/jupyterhub/oauthenticator))
+
+edit `contributors.py` to have the appropriate dates
+for each of these versions. Then, run the script and paste
+the output into the changelog. For an
+example, see [the v0.5 list of contributors](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/v0.5/CHANGELOG.md#contributors).
+
+
+### Push built images to DockerHub + bump version
+
+The JupyterHub helm chart uses a Docker image that's registered
+on DockerHub. When releasing a new version of the helm chart,
+you also need to push a new version of this image. To do so,
+you must have:
+
+1. Docker running locally
+2. An account on DockerHub that you are logged into from
+  your local docker installation.
+3. Push rights for images under `jupyterhub/` on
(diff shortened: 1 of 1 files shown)


choldgraf:
I added a checklist that people can copy/paste into issues to manage this process!

CONTRIBUTING.md
```diff
@@ -54,19 +54,189 @@ development.
...
-4. Generate a list of contributors since the latest release. Use the script
-  in `tools/contributors.py` to list all contributions (anyone who made a
-  commit or a comment)    since the latest release. Update the dates in that
-  script, run it, and paste the output into the changelog. For an example,
-  see [the v0.5 list of contributors](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/v0.5/CHANGELOG.md#contributors).
-5. ...
+## Releasing a new version of the helm chart
+
+The following steps can be followed to release a new version of the Helm Chart.
+This should happen approximately once every 5-7 weeks.
```
willingc: Presently, we expect a release approximately every 5-7 weeks.

CONTRIBUTING.md
```diff
@@ -54,19 +54,189 @@ development.
...
+
+This issue will be used to coordinate the next release of the helm
+chart, {{release-name}}. Instructions for creating the release can be found in
+[CONTRIBUTING.md](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CONTRIBUTING.md#releasing-a-new-version-of-the-helm-chart).
+Below is the checklist for this release.
+
+- [ ] Make a CHANGELOG
+- [ ] Generate and add the list of contributors
+- [ ] Build and push a new Docker image to DockerHub
+- [ ] Commit version bump in `Chartyaml` and `Values.yaml`
```
willingc: Chart.yaml

CONTRIBUTING.md
```diff
@@ -54,19 +54,189 @@ development.
...
+- [ ] Update references in documentation to the new version
+- [ ] Confirm that a new deployment using the updated instructions works
+- [ ] Create and push a new tag for this release
+- [ ] Create and publish a new GitHub release
+- [ ] Write / publish a blog post based largely off of the CHANGELOG
+- [ ] Set ReadTheDocs to begin using `latest` by default
```
willingc: Is there a reason for this?

willingc: Ah nevermind. I see why a few lines down.

CONTRIBUTING.md
```diff
@@ -54,19 +54,189 @@ development.
...
+since the latest release. For each
+release, you'll need to find the versions of all repos
+involved:
+
+* [z2jh](https://github.com/jupyterhub/zero-to-jupyterhub-k8s)
+* [KubeSpawner](https://github.com/jupyterhub/kubespawner)
+* [JupyterHub](https://github.com/jupyterhub/jupyterhub)
+* [OAuthenticator](https://github.com/jupyterhub/oauthenticator))
+
+edit `contributors.py` to have the appropriate dates
```
willingc: Edit

CONTRIBUTING.md
choldgraf: think we could get away with just having docs that point to the last version, then updating straight from `last-version` to `new-version` instead of going through `latest`?

willingc: I'm inclined to just leave enough time before release (i.e. RC) to have the docs ready at the same point as the release of the charts. If the docs are radically off, then there should be a dot release perhaps. I'll defer to you and @yuvipanda but it would be great to release all at once as any minor docs changes will be available in latest anyway.

choldgraf:
ok, latest push addresses comments + tries to clarify a bit how the documentation should be handled with new releases. LMKWYT!

CONTRIBUTING.md
```diff
@@ -54,19 +54,198 @@ development.
...
+
+```
+Title: Release {{release-name}}
+Content:
+
+This issue will be used to coordinate the next release of the helm
+chart, {{release-name}}. Instructions for creating the release can be found in
+[CONTRIBUTING.md](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CONTRIBUTING.md#releasing-a-new-version-of-the-helm-chart).
+Below is the checklist for this release.
+
```
willingc: I would like to an explicit step 0: Code, tests, and documentation to support a release are stable.

CONTRIBUTING.md
```diff
@@ -54,19 +54,198 @@ development.
...
+- [ ] Make a CHANGELOG
+- [ ] Generate and add the list of contributors
+- [ ] Build and push a new Docker image to DockerHub
+- [ ] Commit version bump in `Chart.yaml` and `Values.yaml`
+- [ ] Update references in documentation to the new version
```
willingc: Update references in documentation to the new version (note: documentation should be stable and there should be no anticipated major changes to content).

CONTRIBUTING.md
```diff
@@ -54,19 +54,198 @@ development.
...
+release, then make minor modifications to make it more blog-friendly.
+
+Don't forget to tell the JupyterHub community about the new release, and
+to encourage people to talk about it on social media!
+
+That's it! Congratulations on making a new release of JupyterHub!
+
+### Extra step - release a documentation release
+
+It is common that documentation changes are made shortly after a new release.
```
willingc: I'm inclined to say let's just do this as part of the release process (making the time window: 2 weeks or when the next release docs begin to be written).

Otherwise, we are generating more work for us and more confusion for the end user (a large portion will looks for the current version docs).

choldgraf: IIRC the reasoning was that there are still a lot of large-ish changes to the docs happening, so it'd be best if the docs always show the latest updates until there's something written that doesn't apply to the currently-released helm chart. 

What if we kept this behavior for one or two release cycles, and then once the docs stabilize we can start to just jump straight from `last-version` to `next-version`?

cc @yuvipanda in case you've got thoughts as well

CONTRIBUTING.md
willingc: I'm firm on wanting this explicitly stated. This should be the goal even if it takes a release or two to get there.

CONTRIBUTING.md
willingc: As long as the comment (Code, tests, and documentation to support a release are stable.) goes in as the first checklist step, I'm fine doing this as you propose for a couple of releases.

CONTRIBUTING.md
choldgraf: Definitely agree w/ you on this point 👍 my question only pertained to the other comment

choldgraf:
@willingc what do you think about the latest push? I added a note at the end suggesting that this was temporary behavior, and added a first step of "stable code / docs / etc" to the checklist

choldgraf:
@yuvipanda wanna push the green button on this if you're 👍 on the language?

CONTRIBUTING.md
yuvipanda: Yeah, I think we can do this once enough of our docs get stable enough, but I think for now, we should keep this unfortunately. For example, we run into problems with GKE version needing to be bumped quite often still, so if we don't switch to latest users won't be able to even complete a setup

yuvipanda:
Thanks for making this happen, @choldgraf!

PR #469, opened 2018-01-31
Remove duplicated section about whitelist

clkao:
this seems to be a run-away section that was moved to the end of the doc.

### DIFF
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -84,16 +84,6 @@ tape archive, public cloud, or your own laptop. Start a Globus app
           identityProvider: "youruniversity.edu"
 
 
-To add a whitelist of usernames add to the config file under `auth`:
-
-.. code-block:: yaml
-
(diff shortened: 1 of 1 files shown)


choldgraf:
hey @clkao - I am a little bit confused :-)  I don't see that text duplicated elsewhere in the page...could you point to the other place where this content is duplicated?

choldgraf:
ahhh nvm - I indeed was confused but it was my fault not yours! OK I see where the duplication is. Nice catch @clkao !! 🎉

clkao:
it's at the end of the doc: https://github.com/clkao/zero-to-jupyterhub-k8s/blob/9e83e2a3463d5c4c3d8f758a877fe6658480da27/doc/source/authentication.rst#adding-a-whitelist ;)

choldgraf:
yep, I found it, was just a little under-powered as I hadn't yet had my morning coffee ;-)

PR #470, opened 2018-01-31
Add keycloak odic auth config example.

clkao:


doc/source/authentication.rst
```diff
@@ -83,16 +83,30 @@ tape archive, public cloud, or your own laptop. Start a Globus app
...
+      hub:
+        extraEnv:
+          OAUTH2_AUTHORIZE_URL: https://${host}/auth/realms/${realm}/protocol/openid-connect/auth
+          OAUTH2_TOKEN_URL: https://${host}/auth/realms/${realm}/protocol/openid-connect/token
+      auth:
+        type: custom
+        custom:
+          className: oauthenticator.generic.GenericOAuthenticator
+          config:
+            client_id: "y0urc1logonc1ient1d"
```
choldgraf: we usually use variable-naming like this `<YOUR-VARIABLE-HERE>`...could you do the same for the client_id and secret fields?

doc/source/authentication.rst
```diff
@@ -83,16 +83,30 @@ tape archive, public cloud, or your own laptop. Start a Globus app
           callbackUrl: "https://<your_jupyterhub_host>/hub/oauth_callback"
           identityProvider: "youruniversity.edu"
 
+Keycloak (OpenID Connect)
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To add a whitelist of usernames add to the config file under `auth`:
+OpenID Connect endpoint discovery is not supported by oauthentiator,
+but you can still configure JupyterHub to authenticate with it.
```
choldgraf: Could you provide a link out to documentation on how folks could learn more about this authentication option (and/or how to use it)?

doc/source/authentication.rst
clkao: sure, but do you mean host and realm? because i copied those stub client_id/secret from other sections.

choldgraf: nah I just mean `client_id: "<YOUR-CLIENT-ID>"` and the same for the secret just beneath

choldgraf: gah, actually, nevermind. I just realized all of our other examples use this too...just leave it the way it is and we should revise that in a separate PR for the whole page

choldgraf:
looks great @clkao - I am 👍 but will wait a few hours for others to take a pass, otherwise will merge later on!

clkao:
awesome, thanks! Folks reading https://github.com/jupyterhub/oauthenticator/issues/107 can now find a cleaner setup here.

choldgraf:
woo, thanks @clkao !

ablekh:
Does this solution (thanks, @clkao!) presented above and documented [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/authentication.rst#openid-connect) still have this **logout issue**: https://github.com/jupyterhub/jupyterhub/issues/1805? Please advise.

consideRatio:
@ablekh the github auth will have the logout issue as you are using login as a "is this app authorized by me as a user?" rather than "hello github let me login now".

If you use OIDC auth with Krycloak, you wont have this issue with keycloak itself as it will have a logout endpoint that will destroy all "user is still logged in" state.

Logout aspects are in generally quite complicated as compared to login ones

ablekh:
@consideRatio Thank you very much for your clarification. I've just recently started evaluating Keycloak in terms of its integration with our JupyterHub cluster(s), plus I'm not an expert in the authentication domain, so I still have a lot to learn (and try). :-)

consideRatio:
@ablekh you are very welcome! i have struggled a lot with this stuff lately and this question triggered a lot of thoughts that make me feel a piece of the puzzle fell into place.

The example above btw does not seem to provide a logout endpoint, so that will cause failure.

I want to provide a fresher example

consideRatio:
Oh btw i edited my comment, you were so fast to read the initial one so i failed to edit in time

ablekh:
@consideRatio Re: edit - will re-read, no worries. :-) Thank you for additional clarifications. I'm very glad that my question helped you, in some way, to aggregate and unify your thoughts on the subject. I look forward to seeing your fresher / better example and other insights.

PR #471, opened 2018-01-31
Add note about ingress workaround.

ryanlovett:
This might be more suitable for general troubleshooting section of z2jh as opposed to a release-specific issue. But since the upstream issue coincides with this particular release it is probably fine to put it here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ If you are running v0.4 of the chart, you should upgrade to v0.5 first
 before upgrading to v0.6. You can find out what version you are using
 by running `helm list`.
 
+#### Troubleshooting
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
I think this should be somewhere else, since it can possibly happen to any object, including some that you should not delete (PVC, for example).

choldgraf:
@yuvipanda any suggestions on where? (more generally, I like the idea of being able to incorporate user feedback about upgrade struggles into the docs)

choldgraf:
also @ryanlovett just a note that I made a commit directly to your PR :-)

yuvipanda:
I think a more general troubleshooting section?

But if you think it's ok where it is I'll defer to you, especially since it' tied to a specific helm bug. If this keeps happening for other objects and people report it we can add them here I guess.

choldgraf:
my thinking is "one report, treat it as an exception and make a one-off doc fix, > 1 report, treat it as a more general problem and consider building general docs around it"

yuvipanda:
+1, @choldgraf :) Thanks for the report, @ryanlovett

PR #473, opened 2018-02-03
Move troubleshooting to FAQ

yuvipanda:
- There is already a 'debug' that should have all the debugging info
- The Google Cloud specific stuff in the Troubleshooting section are
  out of date and no longer accurate
- We make this into a FAQ, which can be used for both general and
  cloud specific questions people might have.


choldgraf:
+1 I think it's a good refactor, especially since the old version made it confusing where to look for something

Issue #10, opened 2017-04-11
Create a diagram showing how the services interact


choldgraf:
This would be useful to give people a "big picture" idea of how it all fits together

PR #481, opened 2018-02-06
adding diagram for jupyterhub

choldgraf:
This adds a diagram for a high-level view of JupyterHub. I think there are probably many improvements to be made, but I feel like it's better to have something we can iterate on. Would love to hear what people think!

The google slide used to create the image can be found here: https://docs.google.com/presentation/d/1t5W4Rnez6xBRz4YxCxWYAx8t4KRfUosbCjS4Z1or7rM/edit#slide=id.p

edits or comments are welcome!

closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/10


yuvipanda:
Hey @choldgraf, I added comments + made some minor changes in the google slides you linked to. Thank you very much for making these!

consideRatio:
I considered some changes, it's really tough to make a clean concise diagram though. I reverted to the original version, the changes I considered is available as a revision by Anon user Feb 17 at 3:14 pm.

choldgraf:
I incorporated a bunch of @consideRatio 's changes and cleaned things up a bit more in general...LMK what people think.

choldgraf:
one question: on the jupyterhub side, what serves the actual user interface? is it `hub-` or `proxy-`?

consideRatio:
Guessing it works like the proxy will ask he hub if the user is logged in
based on cookies, then redirect based on that without serving any HTML
content. Don't really know though.

consideRatio:
@choldgraf is all HTML etc always server through the proxy? Will the binder use the jupyterhub's proxy or have one additional layer of proxy on top perhaps?

Im thinking that the user will keep speaking with the same IP all the time, for the JH and the BH, is that correct?

choldgraf:
@consideRatio I think it's a different IP since if you notice, on binder you're redirected to https://hub.mybinder.org which handles subsequent directing to user pods

choldgraf:
FYI this one is ready to go from my end. I think the diagram could use improvements but is a helpful addition. I'm happy to spot-check things moving forward...

consideRatio:
Looks great!

willingc:
@choldgraf Sorry this one slipped off my radar. Merging.

Issue #482, opened 2018-02-06
Moving the side-bar search to the top


choldgraf:
Right now the search bar gets lost if there are lots of sub-sections on the page...e.g.:

![2018-02-05_17-39-34](https://user-images.githubusercontent.com/1839645/35837529-872975e0-0a9b-11e8-94a3-183204a99b06.gif)

I find the search bar quite useful, and it's easy to miss when it's buried like that. Could we move it to the top of the TOC so that it's always visible at the top? What do you think @willingc ?

willingc:
Sure. I might even consider having it at the top and the bottom if the contents are long. @choldgraf

choldgraf:
oo cool idea, LMK make a quick PR to prototype it and we can see how it feels

choldgraf:
huh - I actually can't get the search bar to show up twice...if I copy/paste to the top of the sidebar, then it only shows up on top and not the bottom

willingc:
Move it to the top for now and if folks ask we can investigate both. 👍

PR #487, opened 2018-02-06
sidebar improvements

choldgraf:
This makes a couple of small improvements to the sidebar:

* Moving the search bar to the top of the sidebar so it's more discoverable
* Tweaking the CSS so that section headers are larger than content underneath + bold so that they're more easily separable visually

Old

New

Here's a live version: http://predictablynoisy.com/zero-to-jupyterhub-k8s/

What do folks think about the CSS tweaks? @willingc ?

closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/482

### DIFF
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -64,23 +64,36 @@ img.logo {
 
 /* Sidebar TOC and headers */
 
+div.sphinxsidebarwrapper div {
+    margin-bottom: .8em;
+}
 div.sphinxsidebar h3 {
     font-size: 1.3em;
-    padding-top: .8em;
+    padding-top: 0px;
     font-weight: 800;
+    margin-left: 0px !important;
 }
 
 div.sphinxsidebar p.caption {
     font-size: 1.2em;
     margin-bottom: 0px;
+    margin-left: 0px !important;
+    font-weight: 900;
+    color: #767676;
 }
 
 div.sphinxsidebar ul {
     font-size: .8em;
     margin-top: 0px;
     padding-left: 3%;
+    margin-left: 0px !important;
 }
 
 div.relations ul {
     font-size: 1em;
+    margin-left: 0px !important;
+}
+
+div#searchbox form {
+    margin-left: 0px !important;
 }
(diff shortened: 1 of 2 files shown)


choldgraf:
closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/482

willingc:
I would be inclined to remove the github watch at the top of the sidebar or move it toward the bottom instead of the link to GitHub repo.

I like the restyled headings (Creating your JupyterHub, etc.). Perhaps bump the contrast a bit more to a darker black or dark grey for those. 

This looks good on mobile too.

choldgraf:
See latest push for updates! (I also updated the image)

doc/source/_static/custom.css
```diff
@@ -61,3 +61,39 @@ img.logo {
...
+    font-weight: 800;
+    margin-left: 0px !important;
+}
+
+div.sphinxsidebar p.caption {
+    font-size: 1.2em;
+    margin-bottom: 0px;
+    margin-left: 0px !important;
+    font-weight: 800;
+    color: gray;
```
willingc: We need to increase the contrast to at least #767676 color for accessibility.

FYI. I use AXE extension on Chrome to test. http://accessible-colors.com/ is useful for if I have... what do I need...

willingc: I vaguely remember @carreau had feelings about `!important` but I'm not sure why.

choldgraf: whoah! that's cool! I'd never heard about that, OK will change it in a sec

choldgraf: the reason that we're using it here is because `!important` is used above in the CSS so it's overriding everything else...which is probably both the reason why using `!important` is discourage, and also the reason why we need to use it here :-P

choldgraf:
ok, we're at 767676 now!

doc/source/_static/custom.css
Carreau: Yeah... if really we get too many `!important` we can decide to refactor.

choldgraf:
travis timed out, so restarting

Issue #274, opened 2017-11-15
Add suggestions for node size + disk space interactions


choldgraf:
A recent bit of confusion came up on the gitter, where a person was hitting resource limits on disk size. It seems that on GCP nodes are tied to a default disk size that (maybe?) isn't configurable. We should recommend fewer nodes w/ more resources per node.

PR #494, opened 2018-02-10
node disk size interaction docs

choldgraf:
Adds a warning to the user storage section about node disk size limits.

closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/274

### DIFF
--- a/doc/source/user-storage.md
+++ b/doc/source/user-storage.md
@@ -67,8 +67,8 @@ resources you request.
 
    Some cloud providers have a limited number of disks that can be attached to
    each node. Since JupyterHub allocates one disk per user for
-   persistent storage, this limits the number of users that can be present in
-   each node (and usually this is not configurable). If you need users to have
+   persistent storage, this limits the number of users that can be running in
+   a each node at any point of time. If you need users to have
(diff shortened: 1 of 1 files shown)


doc/source/user-storage.md
```diff
@@ -62,6 +62,17 @@ may be simultaneously attached to a node in your cluster. Check
 your cloud provider for details on the limits of storage
 resources you request.
 
+```eval_rst
+.. note::
+
+   Some cloud providers have a limited amount of allowed storage per node.
```
yuvipanda: Cloud providers limit the *number* of disks that can be attached per node, regardless of how much storage they have. This limits the number of users that can be present in each node, and usually it is not configurable at all. So the limiting factor is number of users per node, not amount of storage.

choldgraf:
see latest push...how's that?

yuvipanda:
@choldgraf I made a minor change and pushed to this PR, can you LMK if it still LGTY? I can merge after!

choldgraf:
found and fixed a small typo but LGTM now!

yuvipanda:
Good catch, thanks @choldgraf!

PR #497, opened 2018-02-13
reorganizing a few sections in docs

choldgraf:
This reorganizes some of the material in the z2jh guide, specifically:

* A section on HTTPS and authentication is moved out of `advanced` and into `security`
* A section on pre-pulling and packing config is moved to a new section called `speed and optimization`
* Renamed a section on upgrading so it's clear what we're talking about


doc/source/authentication.rst
```diff
@@ -191,3 +191,64 @@ You can specify this list of usernames in your `config.yaml`:
        users:
          - user1
          - user2
+
+Advanced: adding your own Ingress object
```
yuvipanda: IMO, Ingress is unrelated to authentication / security and should remain in advanced.

choldgraf: yep, already fixing it, didn't think you'd notice that so quickly haha

doc/source/authentication.rst
```diff
@@ -191,3 +191,64 @@ You can specify this list of usernames in your `config.yaml`:
...
+You can specify multiple hosts that should be routed to the hub by listing them
+under ``ingress.hosts``.
+
+Note that you need to install and configure an
+`Ingress Controller <https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-controllers>`_
+for the ingress object to work. We recommend the
+`nginx ingress <https://github.com/kubernetes/charts/tree/master/stable/nginx-ingress>`_
+controller maintained by the community.
+
+Automatic HTTPS with ``kube-lego`` & Let's Encrypt
```
yuvipanda: Shouldn't this be in security and not in 'authentication'?

choldgraf: yep, had a brain fart

choldgraf:
@yuvipanda it's in `security` now, which is where my brain thought I'd put it but apparently I need to stop staring at a computer screen for bit...

doc/source/security.md
```diff
@@ -155,3 +155,63 @@ This is a sensitive security issue (similar to writing sudo rules in a
 traditional computing environment), so be very careful.
 
 There's ongoing work on making this easier!
+
+## Advanced: Adding your own ingress object
```
yuvipanda: I think this should still be in advanced and not security, since Ingress is not about security at all :)

choldgraf:
hmmm, that's a fair point - in general I am trying to make sure "advanced" only has material that really doesn't have a home elsewhere (partially since it'll be less discoverable in "advanced"). If you feel the ingress stuff should be in there, I'll revert it

yuvipanda:
@choldgraf yeah, Ingress is specifically 'I am on a home-made cluster, how do I get traffic into it?' so I think it fits very squarely in advanced. If we ever have a section specifically for running on non-public cloud providers, I think Ingress can move there. 

And thank you for taking effort to make sure advanced does not become a grab bag!

choldgraf:
ok ingress is back where it started :-)

yuvipanda:
@choldgraf I'm confused now :) You moved Ingress *and* HTTPS back to advanced? I thought you were only going to move Ingress back to advanced but leave HTTPS in security (which seems appropriate)

yuvipanda:
Oh no, I see the problem - the Ingress section had a HTTPS subsection! It all makes sense now :) I think this is good as is.

choldgraf:
I saw that the section was a subsection of "ingress" so I moved them together. Think that it could be factored out?

choldgraf:
ah ya - ok :-)

Issue #477, opened 2018-02-05
Pre-puller can't pull images in v0.6


GerbenWelter:
As discussed on gitter:

Pre-puller fails to pull in required docker images or recognise the images are already available. It only keeps logging the following over and over again:

```
 2018/02/02 09:59:09 0 of 2 images pulled in node worker1.company.com
```

This happens either on an install or upgrade from v0.5 using Helm 2.7.2 and Kubernetes 1.8.4. It also happens on a test-cluster at home using Helm 2.8.0 and Kubernetes 1.9.2.

## Contents of `config.yaml`
```
proxy:
  secretToken: "random_hex_string"
  service:
    type: NodePort
```

The NodePort change is because this is an on-premise test-cluster.

And as requested by @yuvipanda the output (sanitized) of 'kubectl get nodes -o yaml':

```
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  metadata:
    annotations:
      flannel.alpha.coreos.com/backend-data: '{"VtepMAC":"5e:6b:ac:33:61:d5"}'
      flannel.alpha.coreos.com/backend-type: vxlan
      flannel.alpha.coreos.com/kube-subnet-manager: "true"
      flannel.alpha.coreos.com/public-ip: 10.1.1.1
      node.alpha.kubernetes.io/ttl: "0"
      volumes.kubernetes.io/controller-managed-attach-detach: "true"
    creationTimestamp: 2018-01-16T15:31:26Z
    labels:
      beta.kubernetes.io/arch: amd64
      beta.kubernetes.io/os: linux
      kubernetes.io/hostname: master.company.com
      node-role.kubernetes.io/master: ""
    name: master.company.com
    namespace: ""
    resourceVersion: "2355345"
    selfLink: /api/v1/nodes/master.company.com
    uid: 50e2a2d1-fad2-11e7-a5bb-0050568db8e3
  spec:
    externalID: master.company.com
    podCIDR: 10.244.0.0/24
    taints:
    - effect: NoSchedule
      key: node-role.kubernetes.io/master
      timeAdded: null
  status:
    addresses:
    - address: 10.1.1.1
      type: InternalIP
    - address: master.company.com
      type: Hostname
    allocatable:
      cpu: "2"
      memory: 3761852Ki
      pods: "110"
    capacity:
      cpu: "2"
      memory: 3864252Ki
      pods: "110"
    conditions:
[...]

PR #502, opened 2018-02-19
Rework of hook- / continuous-image-puller and image-awaiter

consideRatio:
# What this PR does
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/476#issuecomment-363972524 - This PR solves the issue that could arise by having more than 50 images on a node.
#477 - I now quite confident it will solve @GerbenWelter issue even though I never pinpointed his issue
#476 - This PR considers my initial issue of finding myself with abandoned k8s resources after helm-upgrade failures by failing future upgrades unless it has been cleaned up.

### About the changes - lower complexity, robust functionality
- I reduced a lot of complexity while maintaining function in the prepuller image code.
- It no longer is responsible for creating and destroying a daemonset. It will simply wait for the deamonset's status to stabilize ([`desiredNumberScheduled == numberReady`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#daemonsetstatus-v1-apps)). This requires K8s 1.8+.
- Two less RBAC resources are now required, and the permission no longer includes DaemonSet creation.

### Techniques utilized
__Helm hooks__
```yaml
kind: ServiceAccount
metadata:
  annotations:
    # Create this resource before the actual helm upgrade is made...
    # ... and before the image-awaiter job (weight 10) but after the hook-image-puller (weight -10) ...
    "helm.sh/hook-weight": "0"
    # ... and then delete it when the image-awaiter job has finished one way or another
    # NOTE: This requires helm 1.7+
    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
```

### Verified behaviour
- [x] Helm upgrades are blocked until fresh images are pulled (Demo 1)
- [x] Hook resources are deleted after successful upgrades (Demo 1)
- [x] Fresh nodes get images pulled by the continuous image puller (Demo 2)
- [x] Tainted nodes are supported out of the box (Demo 3)

### TODO
- [x] Verify algorithm's function if nodes are scaling down.
- [x] Verify algorithm's function if nodes are cordoned (gets an `unschedulable` attribute).
- [x] Attempt to solve cleanup if a helm upgrade is aborted
- [x] Test usage in general

### Documentation
- [x] Update changelog
- [x] Update documentation in optimization.md

### Concerns 
Aborted helm upgrades might result in abandoned k8s resources that will make future helm upgrades fail saying the a certain resource already exist. If this happens, manual cleaning is needed like below. It is somewhat robust already due to the annotation `"helm.sh/hook-delete-policy": hook-succeeded,hook-failed` but I've seen it fail still.

```shell
# manual cleanup of abandoned resources
kubectl delete all --selector hub.jupyter.org/deletable=true
```

### Dependencies
- [x] Helm 1.7+
- [x] K8s 1.8+

### Demo 1 - Hook image puller - You will see...
1. A helm upgrade will initiate two helm `pre-upgrade` hooks, the hook-image-puller daemonset first, and instantly thereafter the image-awaiter job.
2. The hook-image-puller daemonset's pod will pull the required images by using dummy Init containers, and the image-awaiter will wait for the daemonset status to seem alright and then exit.
3. The hook resources will autodelete themselves.
4. The actual helm upgrade will begin, a new hub pod is created and the old is terminated.

hook-image-puller2

### Demo 2 - Continuous image puller + Cluster autoscaler - You will see...
__Warning - the GIF is 3-4 minutes__
1. Many user arrives
2. An additional node is made available by the cluster autoscaler
3. All user leave
4. (Not seen in GIF - cluster scales down)
continuous-image-puller-up

### Demo 3 - Tainted and cordoned nodes - You will see...
1. The DaemonSet status before...
2. ... and that it reacts great after a NoSchedule taint is applied

__Related commands__
```shell
# apply taint
kubectl taint node <nodename> key=value:NoSchedule
# remove taint
kubectl taint node <nodename> key:NoSchedule-

# not in demo, but tested (makes the node marked unschedulable)
kubectl cordon <nodename>
kubectl uncordon <nodename>
```

tainted


consideRatio:
### About the k8s API
So I'm currently listing pods from the hook-image-puller daemonset, but one can also use the beta api for daemonset status and get this highly relevant information. This API will be in the core API as of k8s 1.9.

__Setup access to the cluster API__
```shell
kubectl proxy --port 8080
```

__The currently used query__

```shell
# Current API
http://localhost:8080/api/v1/pods?labelSelector=component=hook-image-puller
```
__response__
```json
{
  "kind": "PodList",
  "apiVersion": "v1",
  "metadata": {
    "selfLink": "/api/v1/pods",
    "resourceVersion": "5428524"
  },
  "items": [
    {
      ...
      "status": {
        "phase": "Running",
        ...
```

__The perhaps better query__

```shell
# Beta API (k8s 1.8)
curl http://localhost:8080/apis/apps/v1beta2/namespaces/prod/daemonsets/continuous-image-puller/status

# Core API (k8s 1.9)
curl http://localhost:8080/api/v1/namespaces/prod/daemonsets/continuous-image-puller/status
```
__response__
```json
{
  "kind": "DaemonSet",
  "apiVersion": "apps/v1beta2",
  ...
  "status": {
    "currentNumberScheduled": 2,
    "numberMisscheduled": 0,
    "desiredNumberScheduled": 2,
    "numberReady": 2,
    "observedGeneration": 1,
    "updatedNumberScheduled": 2,
    "numberAvailable": 2
  }
}
```

consideRatio:
## Helm upgrade demo
__What you will see in this gif__
1. Helm hooks initiated, notice the `hook-image-puller` (from a helm-hooked daemonset) and the `image-awaiter` (from a helm-hooked job) pods show up briefly, the puller is pulling the images to the node (already present), and the image-awaiter is a job that will quit when the daemonset's pods are ready.
2. When the image-awaiter job finishes, the helm hooks are completed and everything created in this phase is scheduled to be deleted, so they disappear along with some RBAC resources not seen.
3. Then the new hub terminates and the old hub terminates

hook-image-puller

yuvipanda:
hey! This is great work! 

Checking if all pods are ready seems much more stable than the current setup of checking for nodes :) I'm very happy with it!

I realized a more sure-fire way to delete the daemonset when the hook job is completed is to use [Kubernetes Garbage Collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).

Here is how it would work:

1. The spawning job **pod** would know its own uid, probably via the [downward api](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)
2. Set `ownerReference` (from https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) in the daemonset with this info before spawning daemonset
3. This means whenever the pod gets deleted, it becomes kubernetes' responsibility to delete the daemonset!

This simplifies our code and kills the part that's been the most painful (missing deletions). I haven't verified this theory however, so doing that would also be good!

What do you think of incorporating that into this PR too?

yuvipanda:
I also like @athornton's approach of spawning a bunch of pods rather than a daemonset, but I think we can incorporate that in a different PR once this is working :)

jupyterhub/templates/image-puller/_helper.yaml
```diff
@@ -1,28 +1,40 @@
...
 metadata:
   name: {{ .name }}
   labels:
     app: jupyterhub
-    component: image-puller
+    component: {{ .name }}
     release: {{ .top.Release.Name }}
     heritage: {{ .top.Release.Service }}
+    {{ if .hook }}
+    hook-type: temporary
```
yuvipanda: Labels that we make up should generally be namespaced. So perhaps something like `hub.jupyter.org/deletable: "true"`?

yuvipanda:
Nevermind, my GC idea will not actually work with this approach, since the image-awaiter does not actually create the daemonset. Reading through again, this time with more Tea.

README.md
```diff
@@ -43,6 +43,7 @@ Thank you to the following contributors:
 - Aaron Culich
 - Carol Willing
 - Chris Holdgraf
+- Erik Sundell
```
yuvipanda: <3

yuvipanda:
ok, I'm officially a fan of this clever approach! :) Ignore my messages about GC.

What happens if the helm upgrade fails? If I understand this correctly, currently this forces users to manually run a kubectl command to clean things up before another upgrading. I think this might cause problems for continuous deployment cases...

Instead, can we put something like `{{ .Release.Time.Seconds }}-{{.Release.Name}}-{{.Release.Revision}}` so that helm upgrades can continue happening even if they are leaving behind dead objects? We could further add another script that deletes leftover objects from previous attempts if needed, but otherwise is ok.

What do you think?

consideRatio:
:D I'm glad!

So in certain cases, that I have fail to reproduce but have seen, the `"helm.sh/hook-delete-policy": hook-succeeded,hook-failed` isn't handled and some objects are stranded. I figure it is because the helm upgrade was aborted.

Let us add what you suggest to the name in order to to allow for continued smooth upgrades and only when the user is bothered by a cluttered k8s namespace require the user to clean them up by a relatively simple manner.

```shell
# manual cleanup of abandoned resources
kubectl delete all --selector hub.jupyter.org/deletable=true
```

I'll add `{{.Release.Name}}-{{.Release.Revision}}-{{ .Release.Time.Seconds }}`!

jupyterhub/templates/image-puller/_helper.yaml
consideRatio: Excellent done!

yuvipanda:
Excellent! is this still WIP? 

I can spend some time this afternoon testing and merging this if not.

Do you think we need to update any user docs from this change?

consideRatio:
@yuvipanda I added Demo 3 in the post above and ensured kubectl cordon and kubectl taint NoSchedule did not mess up the logic of comparing [`desiredNumberScheduled == numberReady`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#daemonsetstatus-v1-apps).

Perhaps some additional review could be useful but I'm becoming confident in this PRs readiness now.

consideRatio:
@yuvipanda I'm not sure about the user documentation, I'll look it through later but will be taking a coding break now.

Working together towards a common goal... love it! You have a very encouraging attitude Yuvi, I really appreciate that - it makes it even more motivating!

choldgraf:
and thanks for making a lot of great contributions @consideRatio <3

consideRatio:
# Continuous image pulling, enabled by default?
@yuvipanda @choldgraf 
Continuous image pulling, should we have it enabled by default? I have a hard time imagining someone being upset about minuscule resources lost by a pause container on their fixed amount of nodes. I enabled it by default in the latest commit.

Btw, I rebased my PR branch on top of an updated master branch thinking that would be good as it would make the PR merge be frictionless. Was that alright or did I fail to account for some serious downside?

consideRatio:
Removing the WIP --- All my TODOs are done.

yuvipanda:
@consideRatio let's have it off by default in this PR, can you open an issue to discuss it?

consideRatio:
@yuvipanda okay done!

yuvipanda:
@consideRatio ok, am testing this locally now!

yuvipanda:
I found an unrelated issue about LimitRange objects, but this is good to go.

Congratulations on getting your first big patch merged into the chart, @consideRatio!

PR #505, opened 2018-02-19
Fixes #504 - Singleuser pod packing at hub node first

### LINKED ISSUE #504: 'pack' scheduling of nodes should have affinity for the hub
The 'pack' scheduling strategy for nodes should create a very high affinity for the hub, as this will be the node to scale down last of all.

Learning more, will send a PR when I got it right.

consideRatio:
## Main idea of PR
To make the 'pack' scheduling strategy so that singleuser-pods first schedule on the hub node, and only if it is full, have them schedule on other nodes, and if so they schedule on the nodes with most singleuser images already.

---

With `schedulerStrategy: pack`, I expect my pods to pack into nodes allowing for a kubernetes cluster scaler to scale up and down efficiently. It cannot scale down if there is a hub on the node, so therefore the pods should start packing on the hub node, then onto other nodes.

I got this issue while having 0 users and 2 nodes. When I added my first user it was spread to the non-hub node, then every other user was packed into that node, and later they would populate the hub node. With only one user, I would not be able to scale down even though I had enough nodes to handle the situation, because the user was not packed to the hub node.

This PR solves such issues by adding a big affinity for being in the hub node when the `schedulerStrategy: pack` is set.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -285,7 +285,7 @@ if scheduler_strategy == 'pack':
         'affinity': {
             'podAffinity': {
                 'preferredDuringSchedulingIgnoredDuringExecution': [{
-                    'weight': 1000,
+                    'weight': 50,
                     'podAffinityTerm': {
                         'labelSelector': {
                             'matchExpressions': [{
@@ -297,7 +297,7 @@ if scheduler_strategy == 'pack':
                         'topologyKey': 'kubernetes.io/hostname'
                     }
                 }, {
-                    'weight': 100,
+                    'weight': 5,
                     'podAffinityTerm': {
                         'labelSelector': {
                             'matchExpressions': [{
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -285,6 +285,18 @@ if scheduler_strategy == 'pack':
         'affinity': {
             'podAffinity': {
                 'preferredDuringSchedulingIgnoredDuringExecution': [{
+                    'weight': 1000,
+                    'podAffinityTerm': {
+                        'labelSelector': {
+                            'matchExpressions': [{
+                                'key': 'component',
+                                'operator': 'In',
+                                'values': ['hub']
+                            }]
+                        },
+                        'topologyKey': 'kubernetes.io/hostname'
+                    }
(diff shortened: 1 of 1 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -285,6 +285,18 @@ def generate_user_name(spawner):
         'affinity': {
             'podAffinity': {
                 'preferredDuringSchedulingIgnoredDuringExecution': [{
+                    'weight': 1000,
+                    'podAffinityTerm': {
+                        'labelSelector': {
+                            'matchExpressions': [{
+                                'key': 'component',
+                                'operator': 'In',
+                                'values': ['hub']
```
choldgraf: do these match regular expression-style? We should match "startswith" hub if possible, in case a user creates a pod w/ the word "hub" in it

consideRatio: Good question, I have presumed not, but i'll find out.

consideRatio: `matchLabels` is a map of `{key,value}` pairs. A single `{key,value}` in the `matchLabels` map is equivalent to an element of `matchExpressions`, whose `key` field is “key”, the `operator` is “In”, and the `values` array contains only “value”. 

--- https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ ---

In other words, matchExpressions with key: component, operator: In, values: [hub] is equal to matchLabels: [component:hub], and that probably is an exact match rather than wildcarded like *hub* or similar.

---

The matchExpressions naming is probably referring to that the operator can vary, it can be `In`, `NotIn` or `Exist` expressions. They only seem to support equality or inequality.

consideRatio: When I read the k8s docs, I recall an impression that one was encouraged to use matchExpressions rather than matchLabels. Perhaps just confirmation bias by now, but they could do all the things matchLabels can do and more.

consideRatio: TL;DR; They match with equality/inequality rather than regex matching/not matching.

consideRatio:
Woops corrected a mistake...
`Invalid value: 1000: must be in the range 1-100"` --- Fixed.

choldgraf:
Way to go!!!

Issue #503, opened 2018-02-19
[WIP] Autoscaling - a living development documentation


consideRatio:
# Goal
Allow easy setup of node autoscaling suitable for jupyterhub k8s clusters.

## An intro to autoscaling
If we have many users, we will have many singleuser pods, each guaranteed some resources. The more users, the more pods, the more resources required. We might want to add nodes when we are unable to schedule more pods, or perhaps when were are close to being unable to schedule more pods. There are various tools available related to scaling, here is a summary.

### Instance group autoscaler ([google](https://cloud.google.com/compute/docs/autoscaler/)) - nah...
Does not require or understand kubernetes, so it does not account for kubernetes concepts like pods and their requested resources. It will instead look at CPU load of the nodes and add more nodes when the CPU usage is too high. Not relevant to us, as we want to guarantee users to have a certain set of resources on the node they are scheduled.

### [Horizontal pod autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) - nah...
Will increase the number of pods doing a certain task if needed based on for example CPU utilization. This is relevant if you need more copies of the same pod. Each user should only have one pod, not multiple pods, so scaling up in this way makes little sense for the singleuser pods.

### Kubernetes cluster autoscaler (aka CA) ([google](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler)) - YES
This is the relevant autoscaler! What do we want to scale? The available nodes. When do we want to scale? When we are about to run out of resources to schedule new users, or even when we predict that will happen. The kubernetes cluster autoscaler will add nodes to a cluster if it has pods that cant fit in the previous nodes. Awesome!

consideRatio:
# Observations
- It seems like the google kubernetes cluster autoscaler will not scale down to one pod without additional setup... https://github.com/kubernetes/autoscaler/issues/582#issuecomment-362540339
- It seems like enabling autoscalers will create a pod labeled with `k8s-app=kube-dns-autoscaler` and requesting 20millicores. It would be great if this always was scheduled on the same pod as the hub... Or hehe the other way around, that the hub was scheduled on the node with the cluster autoscaler!

# Inspection of google's k8s cluster autoscaler

```shell
kubectl describe pod --selector k8s-app=kube-dns-autoscaler --namespace kube-system
```

```shell
# output
Name:           kube-dns-autoscaler-7db47cb9b7-bw79l
Namespace:      kube-system
[...]

PR #508, opened 2018-02-21
Cluster autoscaling - PDB for singleuser-server pods

consideRatio:
## PR summary
Solves one relevant part in allowing clusters to autoscale. Denying the cluster autoscaler the ability to disrupt singleuser-server pods.

This part: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/503#issuecomment-367073761

### DIFF
--- a/jupyterhub/templates/singleuser/pdb.yaml
+++ b/jupyterhub/templates/singleuser/pdb.yaml
@@ -1,0 +1,15 @@
+{{ if .Values.singleuser.pdb.enabled }}
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: singleuser-server
+  labels:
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks for the PR! Should the cluster autoscaler not already do that, since these pods are not generated from a replicaset or replication controller? 

From https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

> Pods that are not backed by a controller object (so not created by deployment, replica set, job, stateful set etc). *

consideRatio:
Ah excellent I hope you are right! I'll confirm my pods doesn't crash again and it must have been a timing fluke with the culler that mislead me. I'll be back.

PR #509, opened 2018-02-21
Fixes #506, gid=100 instead of 1000, causes storage permission issues?

### LINKED ISSUE #506: Group ID of jovyan in singleuser image
Should the group ID used by jovyan in the singleuser image be 1000 or 100?

[Here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/76fcf7b208d19593ff5ce536a8319f6a3400e714/jupyterhub/values.yaml#L131) it is set to 1000 but in the base image that the singleuser image is based on jovyan is in group 100.

I think this is a typo.

consideRatio:
@betatim do you know if this will affect peoples storage solution?

Perhaps we could add this for next release and declare in the upgrade instructions to manually set the gid to 1000 again if they will have some old user storage that has a memory of the gid=1000 situation.

closes #506

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -229,7 +229,7 @@ properties:
           Use this only if you are building your own image & know that a group with this gid
           exists inside the hub container! Advanced feature, handle with care!
 
-          Defaults to 1000, which is the uid of the `jovyan` user that is present in the
+          Defaults to 1000, which is the gid of the `jovyan` user that is present in the
           default hub image.
   proxy:
     type: object
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -149,7 +149,7 @@ singleuser:
   initContainers:
   nodeSelector: {}
   uid: 1000
-  fsGid: 1000
+  fsGid: 100
   serviceAccountName:
   schedulerStrategy:
   storage:


jupyterhub/values.yaml
```diff
@@ -5,7 +5,7 @@ hub:
   cookieSecret:
   publicURL:
   uid: 1000
-  fsGid: 1000
+  fsGid: 100
```
minrk: this isn't the right gid for the hub

jupyterhub/schema.yaml
```diff
@@ -229,7 +229,7 @@ properties:
           Use this only if you are building your own image & know that a group with this gid
           exists inside the hub container! Advanced feature, handle with care!
 
-          Defaults to 1000, which is the uid of the `jovyan` user that is present in the
+          Defaults to 100, which is the uid of the `jovyan` user that is present in the
```
minrk: s/uid/gid/

consideRatio: I was unfamiliar with the shorthand notation `s/uid/gid/` - based on context I figure I should replace uid with gid now that I look at it!

jupyterhub/values.yaml
consideRatio: Ah, finally starting to grasp this better! I'll fix this!

jupyterhub/schema.yaml
consideRatio: Oh this was actually within the schema for the hub values, so I'll let this remain to 1000-

consideRatio:
To summarize @minrk @betatim - this is a very basic PR code wise, but what are the consequences?

The idea was to use the same group ID utilized by the jupyter docker-stacks declared in the [base-notebook Dockerfile](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L40) that all images are built from. The main reason I cared was because of an error message when opening a terminal saying that the group with id of 1000 did not exist.

### PR Code Summary
1. Typo fix in the schema file
2. values.yaml - singleuser.fsGid changed to 100 from 1000

minrk:
Great, thanks!

PR #513, opened 2018-02-21
[WIP] Edit typo in template

willingc:


yuvipanda:
I suspect it's not a typo, because Go templates are weird! I think 'type' returns a string that is formatted that way, and Go Templates doesn't let us use type literals (like what is in the patch) anyway.

yuvipanda:
@willingc is it ok if I close this?

PR #515, opened 2018-02-22
small updates to auth docs

choldgraf:
Makes GitHub first (since it's the easiest+most free of the auth steps) and adds a bit of extra information

### DIFF
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -52,7 +52,8 @@ Below is the structure to use in order to authenticate with GitHub.
 
 .. note::
 
-   The org_whitelist is optional and will require users to accept the read:org github oath scope when logging in.
+   The org_whitelist is optional and will require users to accept the read:org github OAuth scope when
+   logging in.
 
 Google
 ^^^^^^
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -21,28 +21,22 @@ declare the values in the helm chart (``config.yaml``).
 
 Here are example configurations for common authentication services. Note
 that in each case, you need to get the authentication credential information
-before you can configure the helmchart for authentication.
-
-Google
-^^^^^^
-
-For more information see the full example of Google OAuth2 in the next section.
-
-.. code-block:: yaml
-
-    auth:
-      type: google
-      google:
-        clientId: "yourlongclientidstring.apps.googleusercontent.com"
-        clientSecret: "adifferentlongstring"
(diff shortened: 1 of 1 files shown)


doc/source/authentication.rst
```diff
@@ -55,6 +49,30 @@ The org_whitelist is optional and will require the use to accept the read:org gi
           org_whitelist:
             - "SomeOrgName"
 
+
+.. note::
+
+   The org_whitelist is optional and will require users to accept the read:org github oath scope when logging in.
```
manics: Typo in `oath`

choldgraf:
thanks for noticing @manics !!

doc/source/authentication.rst
```diff
@@ -52,7 +52,8 @@ Below is the structure to use in order to authenticate with GitHub.
 
 .. note::
 
-   The org_whitelist is optional and will require users to accept the read:org github oath scope when logging in.
+   The org_whitelist is optional and will require users to accept the read:org github OAuth scope when
+   logging in.
```
cam72cam: Perhaps a note here that it is not required if all the users membership is public

choldgraf:
great point @cam72cam !

cam72cam:
That does not exactly address my concern, perhaps something like:
```
The ``org_whitelist`` section is optional, it lets you only allow users from particular GitHub organizations that you list. If you do not add the auth scope read:org, only users with public membership in the org will be allowed to log in. If it is *not* included, then all GitHub users will be allowed to access your JupyterHub.
```

choldgraf:
@cam72cam hmm - I'm actually a bit unclear as to what `read:org` means, and wasn't able to find options for this when creating a github OAuth app. Could you link to docs on what you're talking about here? Or more generally let me know what this means in more lay-person's terms?

manics:
@choldgraf I've got a related PR open: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/523
Assuming this goes in I could either open a separate docs PR discussing `org_whitelist` and `scope` once this PR is merged in which case just leave your changes as they are, or I could give you a commit to cherry-pick here?

choldgraf:
Let's leave this open and you're welcome to suggest language here that I can add once you merge it in! That work for you?

choldgraf:
(just please remember to ping me when it's merged)

cam72cam:
@choldgraf read:org means that the application can lookup what organizations the user is a member of (both public and private memberships)

manics:
@choldgraf Suggested text in https://github.com/choldgraf/zero-to-jupyterhub-k8s/pull/1

PR #516, opened 2018-02-22
Don't make users specify k8s version with GKE

yuvipanda:
1.8 is now the default, so we do not need this anymore!


choldgraf:
hooray! 🎉

PR #519, opened 2018-02-22
Bump base version to v0.7

yuvipanda:
This would make CI produce charts like v0.7-<something>
which is more accurate

We should amend the release process to make this be a part
of it!


betatim:
LGTM.

Where are the release instructions for this chart? Are they part of the jupyterhub releasing?

consideRatio:
@betatim https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CONTRIBUTING.md#releasing-a-new-version-of-the-helm-chart

PR #501, opened 2018-02-16
Add the ability to whitelist github orgs

cam72cam:
Fixes #382

PR #523, opened 2018-02-22
Allow configuration of OAuthenticator.scope

manics:
Follow-up to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/501

This allows [`OAuthenticator.scope`](https://github.com/jupyterhub/oauthenticator/blob/0.7.3/oauthenticator/oauth2.py#L199) to be set. The default is ~`[read:org]` if `auth.github.org_whitelist` is set, otherwise~ empty.

## Examples with Github.

`auth.github.org_whitelist` unset, `auth.scopes` unset:
screen shot 2018-02-22 at 13 28 53

`auth.github.org_whitelist` set, `auth.scopes` set to `[read:org]`:
screen shot 2018-02-22 at 13 28 23

`auth.github.org_whitelist` set, `auth.scopes` set to `[read:org, read:user, user:email]`:
screen shot 2018-02-22 at 13 27 29

In theory this property should also work with other OAuth providers, but I don't have an easy way to test.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -141,7 +141,7 @@ elif auth_type == 'github':
     if len(org_whitelist) != 0:
         c.GitHubOAuthenticator.github_organization_whitelist = org_whitelist
         if not auth_scopes:
-            c.Authenticator.scope = ['read:org'] # required for private membership
+            c.OAuthenticator.scope = ['read:org'] # required for private membership
 elif auth_type == 'cilogon':
     c.JupyterHub.authenticator_class = 'oauthenticator.CILogonOAuthenticator'
     c.CILogonOAuthenticator.oauth_callback_url = get_config('auth.cilogon.callback-url')
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -121,7 +121,6 @@ c.KubeSpawner.cpu_guarantee = get_config('singleuser.cpu.guarantee')
 
 # Allow switching authenticators easily
 auth_type = get_config('auth.type')
-auth_scopes = get_config('auth.scopes')
 email_domain = 'local'
 
 if auth_type == 'google':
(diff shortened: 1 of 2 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -139,7 +140,8 @@
     org_whitelist = get_config('auth.github.org_whitelist', [])
     if len(org_whitelist) != 0:
         c.GitHubOAuthenticator.github_organization_whitelist = org_whitelist
-        c.GitHubOAuthenticator.scope = ['read:org'] # required for private membership
+        if not auth_scopes:
+            c.Authenticator.scope = ['read:org'] # required for private membership
```
manics: This feel like the wrong place to embed the logic for default scopes, but I don't have a better suggestion.

images/hub/jupyterhub_config.py
```diff
@@ -139,7 +140,8 @@
     org_whitelist = get_config('auth.github.org_whitelist', [])
     if len(org_whitelist) != 0:
         c.GitHubOAuthenticator.github_organization_whitelist = org_whitelist
-        c.GitHubOAuthenticator.scope = ['read:org'] # required for private membership
+        if not auth_scopes:
+            c.OAuthenticator.scope = ['read:org'] # required for private membership
```
manics: This feels like the wrong place to embed the logic for default scopes. Maybe this should be left unset since it should work with public members of an org, and an example added to the docs?

cam72cam: That makes sense

manics:
I've removed the default GitHub scope. I can update the docs after https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/515 is merged

images/hub/jupyterhub_config.py
```diff
@@ -186,6 +185,8 @@
 else:
     raise ValueError("Unhandled auth type: %r" % auth_type)
 
+c.OAuthenticator.scope = get_config('auth.scopes', [])
```
minrk: This overrides the default of the chosen OAuthenticator to an empty list if unset in the helm chart.

To avoid this, use a condition:

```python
auth_scopes = get_config('auth.scopes')
if auth_scopes:
    c.OAuthenticator.scope = auth_scopes
```

manics: Fixed

minrk:
Thanks!

PR #524, opened 2018-02-22
Doc: fix a link in the glossary

bzz:
Drive-by updated, previous link to jupyterhub authenticator docs returned 404.


choldgraf:
Thanks! Nice catch

PR #531, opened 2018-02-23
avoid culling users whose servers haven't stopped

### LINKED ISSUE #522: culler is failing and exiting when culling users and servers are slow to stop
Two issues:

1. culler script seems to exit when the cull request fails. It's unclear why this happens, but we should confirm and fix this
2. the 'real' issue is that the culler is hitting 400 errors in the first place. The cause is servers that are slow to stop (DELETE /users/:name gives 400 if the user's server is running and cannot stop promptly). The previous request to stop the server will have returned 202 ACCEPTED instead of 204 DELETED in this case. If we delay deleting users if we get 202 ACCEPTED from the server deletion, we should be safe here.

minrk:
attempting to cull a user whose server is slow to stop results in a `400 DELETE` error, which seems to exit the culler process.

user will be deleted in a later round when the server has finished stopping.

The initial request in `call_sync` was failing, which causes the script to exit. This is replaced with an `add_callback`, in which case errors are logged instead of exiting.

closes #522

### DIFF
--- a/images/hub/cull_idle_servers.py
+++ b/images/hub/cull_idle_servers.py
@@ -113,16 +113,17 @@ if __name__ == '__main__':
         help="""Cull users in addition to servers.
                 This is for use in temporary-user cases such as tmpnb.""",
     )
-    
+
     parse_command_line()
     if not options.cull_every:
         options.cull_every = options.timeout // 2
     api_token = os.environ['JUPYTERHUB_API_TOKEN']
-    
+
     loop = IOLoop.current()
     cull = lambda : cull_idle(options.url, api_token, options.timeout, options.cull_users)
-    # run once before scheduling periodic call
-    loop.run_sync(cull)
+    # schedule first cull immediately
+    # because PeriodicCallback doesn't start until the end of the first interval
+    loop.add_callback(cull)
     # schedule periodic cull
     pc = PeriodicCallback(cull, 1e3 * options.cull_every)
     pc.start()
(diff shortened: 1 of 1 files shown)


images/hub/cull_idle_servers.py
```diff
@@ -83,12 +89,17 @@ def cull_one(user, last_activity):
         if not user['server'] and not cull_users:
             # server not running and not culling users, nothing to do
             continue
+        if not user['last_activity']:
+            continue
         last_activity = parse_date(user['last_activity'])
         if last_activity < cull_limit:
+            if user['pending']:
```
betatim: `pending` means that the state of the spawner is changing (from stopped to starting or the other way around)?

images/hub/cull_idle_servers.py
```diff
@@ -70,7 +70,13 @@ def cull_one(user, last_activity):
                 method='DELETE',
                 headers=auth_header,
             )
-            yield client.fetch(req)
+            resp = yield client.fetch(req)
+            if resp.code == 202:
+                msg = "Server for {} is slow to stop.".format(user['name'])
+                if cull_users:
+                    app_log.warning(msg + " Not culling user yet.")
+                    return
```
betatim: Note: this is the magic line that avoid culling the user if the server is still shutting down. It took me two reads to spot this, so leaving a note :) maybe add a comment for future Tims that will overlook this again.

minrk: comment added for this and above

PR #533, opened 2018-02-23
get chartpress from standalone repo

minrk:
now at https://github.com/jupyterhub/chartpress

removes build.py


consideRatio:
Ah nice! One additional level of complexity to understand removed for the z2jh user!

PR #535, opened 2018-02-23
Update user-environment.rst now that JuptyerLab is in beta

rsignell-usgs:
Change note now that JupyterLab is in beta.


choldgraf:
🎉 Thanks for noticing!

PR #540, opened 2018-02-26
culler: await every Future when culling

minrk:
avoids orphaning Futures when a cull request fails

orphaned futures could end up never finishing their request to the Hub, resulting in a failure to cull servers after the failing request


minrk:
Additionally, I loaded pycurl in the cull-idle-users script since it seems to help with lots of requests.

yuvipanda:
ty, @minrk!

PR #541, opened 2018-02-26
fill image tags values.yaml with placeholders

minrk:
these values (both name and tag) are autogenerated by chartpress, so the values stores in values.yaml will never be used.

We could also replace these with an empty dict, so that it's extra clear that the image data needs to be filled out by running `chartpress`, and the values.yaml tracked by the repo is never valid.

cc @consideRatio who I think may have run into an issue related to this.


consideRatio:
Ah... Thanks @minrk !

PR #543, opened 2018-02-27
[WIP] Add a secondary scheduler with policy we can tweak

### LINKED ISSUE #542: Make 'pack' schedulerStrategy work properly
When running jupytehub on k8s, we ideally want the scheduler to *pack* pods together onto nodes as much as possible. This helps with autoscaling a lot.

We currently *sortof* enable this by setting `schedulerStrategy: pack`. It uses podAffinity to do its thing. But podAffinity does is not weighted - if you have two nodes with 2 and 80 pods each, a new pod can get scheduled on either one. This limits its effectiveness a fair bit!

After digging around and talking to more people in the kubernetes community, I believe a real solution is:

1. Run another copy of the `kube-scheduler` in our cluster, with a `--scheduler-name` set to something custom (so this will only schedule our pods)
2. Put the name of our kube-scheduler in `schedulerName` for all our pods
3. Use a custom policy.json file to configure kube-scheduler to use https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithm/priorities/most_requested.go as a priority function. This should do what we need it to do!

https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/ has more general info on this approach.

Defaults of the current algorithms in https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithmprovider/defaults/defaults_test.go

yuvipanda:
Still running into https://github.com/kubernetes/kubernetes/issues/60469,
which makes this unusable with an autoscaler.

policy.json content comes from https://github.com/kubernetes/kubernetes/pull/59401

Fixes #542 

- [x] Make sure we can pick version of scheduler that matches kubernetes version
      automatically
- [x] Test RBAC to make sure it works properly
- [x] Play with various policy.json config to see what works best for us
- [x] Sort out what's causing https://github.com/kubernetes/kubernetes/issues/60469
- [ ] Use the 1.11 way of configuring the scheduler binary
- [ ] Make policy tweakable from config.yaml


yuvipanda:
Setting `MostRequestedPriority` predicate's weight to something like 10 or 100 seems to give us the behavior we want, but everything is stonewalling because of https://github.com/kubernetes/kubernetes/issues/60469

consideRatio:
@yuvipanda KubeCon 2018 videos are out and this was very relevant for me too look at considering implementing a scheduler, especially for the singleuser-server pods! I got excited! :D

https://www.youtube.com/watch?v=mepNec5OpjY

jupyterhub/templates/scheduler/rbac.yaml
```diff
@@ -0,0 +1,66 @@
...
+metadata:
+  name: {{ .Chart.Name }}-{{ .Release.Name }}-scheduler
+subjects:
+- kind: ServiceAccount
+  name: {{ .Chart.Name }}-{{ .Release.Name }}-scheduler
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: {{ .Chart.Name }}-{{ .Release.Name }}-scheduler
+  apiGroup: rbac.authorization.k8s.io
```
consideRatio: We can do `name: system:kube-scheduler` instead.

Picked that up from [this presentation](https://www.youtube.com/watch?v=4TaHQgG9wEg&t=14m22s&index=210&list=PLj6h78yzYM2N8GdbjmhVU65KYm_68qBmo).

jupyterhub/templates/scheduler/configmap.yaml
```diff
@@ -0,0 +1,115 @@
+{{ if .Values.scheduler.enabled -}}
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ .Chart.Name }}-scheduler-config
+data:
+  policy.json: |
```
consideRatio: Should this be policy.cfg ?

[kube-scheduler documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/):
image

jupyterhub/templates/scheduler/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+        configMap:
+          name: {{ .Chart.Name }}-scheduler-config
+      containers:
+      - command:
+        - /usr/local/bin/kube-scheduler
+        - --address=0.0.0.0
+        - --leader-elect=true
+        - --scheduler-name={{ .Chart.Name }}-{{ .Release.Name }}-scheduler
+        - --lock-object-namespace={{ .Release.Namespace }}
+        - --lock-object-name={{ .Chart.Name }}-{{ .Release.Name }}-scheduler-lock
```
consideRatio: ### Research note
Is the `--leader-elect` is supposed to be enabled "[...] when running replicated components for high availability." and it is defaulting to true.

Reading in [configure multiple schedulers - section 3](https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/#3-run-the-second-scheduler-in-the-cluster) they write that if you want to setup leader election, you must update the following...

```
--leader-elect
--lock-object-namespace
--lock-object-name
```

And that one must also add the name of the scheduler to the ClusterRole under rules.apiGroups.resourceNames.

jupyterhub/templates/scheduler/rbac.yaml
consideRatio: But... if we want to support having replicas >1 of our scheduler we should add stuff to the ClusterRole under resourceNames as described in [configure multiple schedulers - part 3](https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/#3-run-the-second-scheduler-in-the-cluster).

jupyterhub/templates/scheduler/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - -v=4
```
consideRatio: I did not see information about this in the documentation, it some verbosity level i figure.

jupyterhub/templates/scheduler/deployment.yaml
consideRatio: While reading in the [kube-scheduler documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/) I understood it as the `--lock-object-name` should be an existing Endpoint object by default, but can also be a ConfigMap if we also pass `--leader-elect-resource-lock=configmaps`.

Since we don't have an endpoint for our scheduler, since we don't have a Service for it currently and I'm unaware of the need for one, we should specify the configmap we use.

```yaml
        command:
        - /usr/local/bin/kube-scheduler
        - --address=0.0.0.0
        - --scheduler-name=jupyterhub-scheduler
        - --policy-configmap=scheduler-config
        - --policy-configmap-namespace={{ .Release.Namespace }}
        - --leader-elect=true
        - --leader-elect-resource-lock=configmaps
        - --lock-object-name=scheduler-config
        - --lock-object-namespace={{ .Release.Namespace }}
        - --v=4
```

consideRatio:
## Notes
- [jupyterhub/kubespawner#190](https://github.com/jupyterhub/kubespawner/pull/190) - Allow configuration of the scheduler.

consideRatio:
I wanted to check the box "make sure we **can** set the version..."

By using either the `.Chart` objects provided by Helm for templates that gives access to Chart.yaml information, or by using `.Capabilities` objects related to Kubernetes.

I'm not confident on what I think make most sense, using `.Capabilities` would be more in tune with the actual cluster and avoid some mismatches but require a `helm upgrade` to actually stay in tune. While using `.Chart` would allow us to use a single kube-scheduler binary per helm chart release, which may be good.

I'm thinking we try to use `.Chart` over `.Capabilities` to decide on the binary. What do you think @yuvipanda ?

consideRatio:
## About RBAC
I'm not happy about needing to create a ClusterRoleBinding, but I figure we must in order to have a scheduler that works.

## About the policies
The scheduler can have performance issues. How can we minimize them? I figure we might be able to remove some node node filters aka. `predicates` or some node preferences aka. `priorities` and that would reduce the workload.

The heavy work is probably done in the preferences / priorities.

consideRatio:
## About kube-scheduler
https://github.com/kubernetes/community/blob/master/contributors/devel/scheduler.md
https://github.com/kubernetes/community/blob/master/contributors/devel/scheduler_algorithm.md

consideRatio:
Wieeeeeeeeeeeeee this took some time but your checklist was excellent @yuvipanda !

consideRatio:
- [`kube-scheduler` binary's documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/) deprecated stuff such as the `- --policy-configmap=<name of configmap>` without saying what to do instead. But I found [this code](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/componentconfig/v1alpha1/defaults.go) and figure they will support setting the Policy configmap through a `KubeSchedulerConfiguration` of the `apiGroup: componentconfig/v1alpha1`.

consideRatio:
Continued on in #758

PR #544, opened 2018-02-28
Rephrase definition of GitHub

willingc:
Remove "hub" as its an overused term and potentially confusing.


choldgraf:
great point

PR #551, opened 2018-03-01
Add info about minikube install for MacOS

willingc:
Add same wording that we have in BinderHub contributing re: installing from binaries instead of homebrew for MacOS.


choldgraf:
Thanks for the tip !

willingc:
Thanks for the review and merge on these two PRs @choldgraf.

PR #557, opened 2018-03-03
Add informational fields and minimal comments to jupyterhub chart

willingc:
Addresses #556.

- Adds additional information fields to Chart.yaml
- Adds minimal descriptions to values.yaml (See [Chart Best Practices](https://docs.helm.sh/chart_best_practices/#document-values-yaml))

### DIFF
--- a/jupyterhub/Chart.yaml
+++ b/jupyterhub/Chart.yaml
@@ -1,4 +1,16 @@
-apiVersion: v1
-description: Multi-user Jupyter installation
 name: jupyterhub
 version: v0.7-dev
+apiVersion: v1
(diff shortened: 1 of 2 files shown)


choldgraf:
this is great! thanks!

choldgraf:
oops! accidentally merged before realizing that tests weren't passing. I've reverted, so maybe you can re-open this PR so we can figure out what's up?

Here's the error:

```
helm install --name jupyterhub-test --namespace jupyterhub-test ./jupyterhub/ -f minikube-config.yaml
Error: error unmarshaling JSON: json: cannot unmarshal array into Go struct field Metadata.home of type string
```

PR #570, opened 2018-03-08
Make daemonsets do a rolling update

yuvipanda:
Without this, the continuous pre-puller pods do not work,
since they're stuck in the old image version forever.

https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/
has more info

### DIFF
--- a/jupyterhub/templates/image-puller/_helper.yaml
+++ b/jupyterhub/templates/image-puller/_helper.yaml
@@ -33,9 +33,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      # Arbitrarily large number, since we want all the old daemonsets to die immediately
-      # and new ones to be born
-      maxUnavailable: 1000
+      maxUnavailable: 100%
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


jupyterhub/templates/image-puller/_helper.yaml
```diff
@@ -30,6 +30,12 @@ spec:
         component: {{ .name }}
         release: {{ .top.Release.Name }}
         heritage: {{ .top.Release.Service }}
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      # Arbitrarily large number, since we want all the old daemonsets to die immediately
+      # and new ones to be born
+      maxUnavailable: 1000
```
consideRatio: We can set this to be 100% instead I think:
https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/#rollingupdatedaemonset-v1beta2-apps

consideRatio:
Nice catch! I remember thinking it did not matter for the hook-image-puller and left it out at one point... Doh!

LGTM!

jupyterhub/templates/image-puller/_helper.yaml
yuvipanda: Great catch, @consideRatio! fixed.

PR #571, opened 2018-03-09
Customization options for dynamic storage

enolfc:
This allows the pvc and volumes used for single user to have custom names, which is useful is the user names are long (> 60 characters) as this will fail in kubernetes.


enolfc:
Just realised an issue on my test deployment, will resubmit

enolfc:
After re-checking, things look ok. Was not using the latest version of on the deployment.

yuvipanda:
This LGTM, thank you very much for your PR! Congratulations on getting your first PR here merged, hope to see more :)

PR #574, opened 2018-03-12
use urllib quote to encode username in culler FIX #573

### LINKED ISSUE #573: Cull_idle fails for some users with non ascii usernames
At the moment Cull Idle queries the API for the users with inactive running servers and sends a `DELETE` to the endpoint users/USERNAME/server. https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/706a8f7637881ace06b66c90c5267019d75d2cf0/images/hub/cull_idle_servers.py#L69 This works for most users, but fails with users with non ASCII usernames as the api responds with a 404.

By url-encoding the username we can make sure it works for every user.

chicocvenancio:
Closes #573

### DIFF
--- a/images/hub/cull_idle_servers.py
+++ b/images/hub/cull_idle_servers.py
@@ -31,6 +31,11 @@ import datetime
 import json
 import os
 
+try:
+    from urllib.parse import quote
+except ImportError:
+    from urllib import quote
+
 from dateutil.parser import parse as parse_date
 
 from tornado.gen import coroutine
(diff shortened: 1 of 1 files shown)


chicocvenancio:
Fwiw, I tested this and it kept working with usernames with ascii characters and started working with usernames with non-ascii characters https://github.com/chicocvenancio/paws/commit/9a77d1df995d41aa41233aa4e50c95b4144c7269

consideRatio:
I get this happy feeling about the code base when I think of how many is participating in testing it and contributing to it!

minrk:
Thanks! This should be right. Since JupyterHub itself doesn't support Python 2 (3.5 for 0.9, 3.3 for 0.8), I don't think we need to support Python 2 in example scripts, but there's no harm in having an import check like this.

PR #576, opened 2018-03-13
Revert "Set maxUnavailable appropriately for hub & proxy pods"

yuvipanda:
As @minrk pointed out in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/367#pullrequestreview-86373648,
this is actually unsafe. My comment about the proxy being safe
also do not make sense in hindsight.

This reverts commit 7d03d6459ff8d186abe2e854d524c71eca3c51d3.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -9,12 +9,6 @@ metadata:
   name: proxy
 spec:
   replicas: 1
-  strategy:
-      rollingUpdate:
-          # There's no PVCs attached here, so is ok to set maxUnavailable to 1
-          # This means new pod will be spun up and traffic transferred to it
-          # before old pod is shut down.
-          maxUnavailable: 1
   template:
     metadata:
       annotations:
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -4,19 +4,6 @@ metadata:
   name: hub
 spec:
   replicas: 1
-  strategy:
-    rollingUpdate:
-      {{ if eq .Values.hub.db.type "sqlite-pvc" }}
-      # If we're using a PVC, set maxUnavailable to 1
-      # This is required, since the pod attached to PVC must die
-      # before new pod can run.
-      maxUnavailable: 1
-      {{ else }}
-      # If we aren't using a PVC, it's ok to spawn up a new pod
-      # before killing the currently running one. This also makes the '--wait'
-      # flag of helm work properly
-      maxUnavailable: 0
-      {{ end }}
   template:
     metadata:
       labels:


gsemet:
can you elaborate why is it unsafe? or make it optional?

To my understanding, there is no db migration performed while the old pod is running, because there is not livenessProbe.
it should do:
- old hub is running and accepting users
- new hub is starting, docker image is downloading and so
- new hub is started, since there is not livenessProbe on the hub deployment, the old hub is stopped
- new hub perform user scan and db migration (can you confirm it?). Service may be unavailable here
- new hub is completely started, service is restored.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -9,12 +9,6 @@ metadata:
   name: proxy
 spec:
   replicas: 1
-  strategy:
```
gsemet: why removing the rolling update on the proxy as well? There is not the same potential issue and it does have a livenessprobe to switch at the right moment

yuvipanda:
I'll let @minrk answer the questions around hub, since he originally pointed it out. My understanding is that the hub keeps state in memory and in the db, and if a different process (like another hub) modifies the db then silent data corruption might occur.

jupyterhub/templates/proxy/deployment.yaml
yuvipanda: I agree, but I think the proxy might *already* have this behavior, and my patch had the opposite effect of what the comment implies. The hub is constrained by the PVC binding, but the proxy isn't - so the defaults should 'just work'.

minrk:
I think the sqlite-pvc case was fine, as long as my understanding of maxUnavailable=1 is correct that there's no time during which two hub pods are running. The main unsafe case was external db (e.g. postgres), which has the same requirements that no two hubs are talking to the same db, but maxUnavailable=0 allowed the second hub to start before tearing down the first one, which can result in the second hub's state being inconsistent with the db.

The main issue here seems to be that the proxy had maxUnavailable=1, which means that the old proxy had to be torn down before the new one, rather than maxUnavailable=0, which allows the old one to stay up while the new one takes its place, so the value was the opposite of what was described in the comment.

gsemet:
Do you expect database migration to cause old hub to corrupt the DB? What kind of "corruption" do you think to?

minrk:
Not corrupted per se. The database itself will remain valid, but may become *inaccurate*. JupyterHub assumes that it has total ownership and control over the database, so it maintains an in-memory cache of certain state for performance benefits. There is also some purely in-memory state, such as pending events (start, stop, etc.). The result of this is that if the second jupyterhub launches and loads db state and then the first one makes some transactions before shutting down, those may not be reflected in the new hub until certain subsequent actions are taken to reconcile Hub state with db state (e.g. restarting the second hub a second time).

The main likely scenarios are, for instance, a user pod starting but the Hub state suggesting that the user's pod is not running, or the other way around (this is harder, I think).

choldgraf:
@minrk I wonder if we should clarify the relationship between the hub and the DB somewhere for the sake of k8s...e.g. the image on the jupyterhub page:



makes it seem like the database is a *part* of the jupyterhub, whereas it sounds like with k8s the database is a separate thing than the "hub pod" and a single database can be shared by two hubs

gsemet:
I wonder what could be done to have the ORM reflect more accurately the DB state, what I mean it is pretty often we see backends that can be scaled up to several instances without conflict. In short, given the currrent architecture of the jupyterhub, it should not be that hard to acheive

minrk:
> In short, given the currrent architecture of the jupyterhub, it should not be that hard to acheive

If we want jupyterhub to canonically scale beyond a few thousand users without external projects like [this one](https://github.com/berkeley-dsep-infra/data8xhub), we need to do this. It is possible, but it is not trivial. 100% of JupyterHub state would have to leave the Hub and live on the database (including pending state, etc.), and event handlers like user stopped -> remove from proxy would need to be able to deal with multiple Hubs potentially attempting to handle the same event. We would also need to remove all optimizations regarding the local cache of database state, which dramatically reduces performance of individual requests. See [this analysis of startup and request performance](https://nbviewer.jupyter.org/github/minrk/jupyterhub-startup-perf/blob/master/analysis.ipynb) of various optimizations.

One thing that's hard to ensure is handled correctly when there are multiple Hubs on one database: two requests to start the same user's server coming in quick succession.

There needs to be a lock that prevents the second request from attempting any transactions, or resolve duplicate actions kubernetes-style with 409 responses. Neither of these is simple to rigorously guarantee. It's been hard enough to guarantee it with the single-threaded in-memory state we have, where we can make solid guarantees about what code is executed before another request is handled.

@choldgraf good idea on clarifying the relationship. In a way, the relationship is currently signaled in that diagram, in that the database is very much a part of a single Hub instance and should be considered wholly owned by and part of the Hub, even if storage happens to be external. Hubs should not be expected to cooperate with each other or other processes that write to the Hub's database (reading is fine).

Issue #561, opened 2018-03-04
In docs, reference.txt links to non-existing page (user-experience.html)


tylere:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/3c23fd2860aa885c283b3549fe9d673019c25222/doc/source/reference.txt#L13

whosford:
I'd like to work on this issue if nobody else is

choldgraf:
Please feel free to make a pr!

choldgraf:
just realized that @whosford already fixed this - thanks a bunch!

PR #579, opened 2018-03-15
Update references.txt with new user environment link

whosford:
This fix is for issue #561 

Updated references.txt replacing the old `user-experience.html` link with new `user-environment.html` link. Changed the link text to be 'user environment'


choldgraf:
Thanks for the patch!!

PR #582, opened 2018-03-15
default extraConfig is a dict

minrk:
avoids warning:

> warning: destination for extraConfig is a table. Ignoring non-table value

does not appear to affect behavior, though, only the warnings. I believe the ignored value is the default `extraConfig` empty string, which has no effect.

closes https://github.com/jupyterhub/binderhub/issues/483


gsemet:
+1

yuvipanda:
Thanks, @minrk

PR #588, opened 2018-03-16
add max-age pod culler

minrk:
deletes user pods older than `cull.maxAge` seconds. Default is 0 (disabled).

Culler watches pods via logic copied from kubespawner. Pods older than cull.maxAge are culled every `cull.every`.

This should enable temporary-service deployments like Binder to shutdown users that run for too long that the idle culler doesn't catch.


jupyterhub/templates/pod-culler/cronjob.yaml
```diff
@@ -0,0 +1,41 @@
+{{ if and .Values.cull.enabled .Values.cull.maxAge }}
```
yuvipanda: can we instead make this into a deployment that just maintains state itself? I think that's much easier for debugging purposes in cases like this (just look at logs) than a cronjob.

minrk: Sure! That's what I had first, but I thought a cron job might make more sense than a long-running job always squatting resources. Inspecting the logs is definitely easier that way, though, so I'll switch it back.

minrk:
pod-culler is now a long-running container that uses a watch based on kubespawner's pod reflector

yuvipanda:
Is this good to go, @minrk?

PR #592, opened 2018-03-20
Separate automatic HTTPS to its own deployment

yuvipanda:
The extra network hop is worth the following advantages:

1. The code is much easier for us to understand and reason about.
2. We have pod affinity set up to try to place all three components
   (hub, proxy, autohttps) in the same node if possible. This
   should cut down on hops!
3. For users with at least one proxy in front of the hub,
   this is a much cleaner way to disable the extra nginx cost -
   both in terms of resources and extra processing overhead.

Longer term, we should build / use a proxy implementation that
can automatically get certificates from ACME and store them
in Kubernetes Secret objects.

We also terminate HTTPS in CHP when using manual HTTPS, avoiding
an extra hop unnecessarily. IIRC this means we lose http/2 when
using manual HTTPS, but that feels fine for now.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -1,3 +1,5 @@
+{{ $HTTPS := (and .Values.proxy.https.hosts .Values.proxy.https.enabled ) }}
+{{ $autoHTTPS := (and $HTTPS (eq .Values.proxy.https.type "letsencrypt" ) ) }}
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
@@ -21,7 +23,7 @@ spec:
         component: proxy
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
-        {{- if eq .Values.proxy.https.type "letsencrypt" }}
+        {{- if $autoHTTPS }}
         # required for kube-lego to work
         app: kube-lego
         {{- end }}
@@ -47,6 +49,7 @@ spec:
                       operator: In
                       values: [ {{ .Release.Name | quote }} ]
       containers:
+      {{ if $HTTPS }}
         - name: nginx
           image: "{{ .Values.proxy.nginx.image.name }}:{{ .Values.proxy.nginx.image.tag }}"
           imagePullPolicy: {{ .Values.proxy.nginx.image.pullPolicy }}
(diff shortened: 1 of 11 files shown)


yuvipanda:
I've tested this without HTTPS, but need to test it with HTTPS.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -86,7 +89,6 @@ spec:
             - name: https
               containerPort: 443
               protocol: TCP
-        {{ if and .Values.proxy.https.hosts (and .Values.proxy.https.enabled (eq .Values.proxy.https.type "letsencrypt" ) ) -}}
```
minrk: missing `if $autoHTTPS` here?

minrk:
Makes sense. This would affect Binder, right? as we do our own https there, so the internal nginx would be disabled?

yuvipanda:
@minrk yep, this would make life simpler in Binder too.

jupyterhub/templates/proxy/deployment.yaml
yuvipanda: I just realized this is a rat's nest of conditionals. I'm just going to split the autoHTTPS part into its own deployment, which should make things easier for us to maintain for now. Longer term, we can replace CHP with something that does ACME (or add a sidecar that does ACME and stores the certificates in a kubernetes Secret object)

yuvipanda:
@minrk I've pretty much redone this entire PR. Have also updated description. I've tested both manual and automatic let's encrypt modes.

LMK what you think!

yuvipanda:
This is also the last change I need for my EdX launch I think :)

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -15,22 +16,14 @@ spec:
...
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
-        hub.jupyter.org/https-type: {{ .Values.proxy.https.type }}
       labels:
         name: proxy
         component: proxy
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
-        {{- if eq .Values.proxy.https.type "letsencrypt" }}
-        # required for kube-lego to work
-        app: kube-lego
```
consideRatio: We could add `app: jupyterhub` as a label now that this hack is out of the way :)

consideRatio: I get a warm feeling about this PR! I don't grasp whats done fully, but it seems like you are successfully separating concerns and making the code more understandable!

minrk:
Nice improvement! I'll let @consideRatio update the `app: jupyterhub` labels in his upcoming label PR.

Issue #593, opened 2018-03-20
Figure out what Kubernetes installation method we 'support'


yuvipanda:
We currently have instructions for Google, Azure and AWS to create a kubernetes cluster. We've had people try to set up clusters in other places:

1. JetStream (from @zonca https://zonca.github.io/2017/12/scalable-jupyterhub-kubernetes-jetstream.html)
2. Digital Ocean (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/401)
3. Bare metal clusters in various places (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/591, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/277, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/493 and various questions on the Gitter Channel)
4. OpenStack (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/88, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/340, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/237#issuecomment-341523196 and various discussions)
5. OpenShift

I'd love for us to answer the following questions:

1. When do we have instructions in the guide, and when do we just link to an external location?
2. What do we do when users have problems with kubernetes installations and come to us? Do we help try to fix them? Do we redirect them to the k8s community? What is the line here between 'kubernetes problem' and 'JupyterHub' problem?
3. For the providers we have instructions for, should we grade them with 'levels of support'? Which ones do we test with Kubernetes version updates and JupyterHub chart version updates?

I think we need to make sure we communicate clearly that you can use z2jh without depending on Cloud Providers. This is especially important in the current climate of intense critique of privacy of various platforms. Several people have run k8s clusters successfully on all these platforms, but everyone has to re-learn the same lessons without any place to share them.

Here are the various options I see:

1. Introduce various installation methods in the z2jh guide that can be contritbuted to by users. We would answer question (3) from above, and mark these with appropriate levels. Particularly important is how much testing we do with new releases, since we might not have access to test on these platforms.
2. Have a place to put links to various blog posts. This takes all decisions out of our hands, and 'reader beware, check the date before trying!' mode.
3. Have a wiki or a community repo for doing option (1), to make it clearer that it is not part of the main z2jh guide. This can have much wider merge rights - perhaps anyone who lands a PR gets merge rights? 

[...]

PR #594, opened 2018-03-20
Refactor docs for Step Zero, a K8 cluster

willingc:
Related to #593.

This refactor adds:

- a Step Zero which is configuring your Kubernetes cluster.
    - add directories in source for cloud vendors with existing docs and placeholder directories for those that we still would like docs
    - add placeholder directories with `.gitkeep` so they can be checked in
    - move single page k8 install docs into step zero and break into vendor specific docs
- Decrease the toctree level displayed in "Creating your JupyterHub" as the subitems were not particularly helpful vs. the space on the page they took
- Add a community resources section
    - put community resources (additional resources) here
    - move gallery of deployments used here

What was not done:
- break up the vendor specific shutdown cluster doc
- fix the annoying caption displaying for toctrees on the index page (that will need some css work). We can't remove the caption as it is used in the sidebar nav.
- new content is not added

[Rendered docs](https://test-zerotojh.readthedocs.io/en/cluster-jupythub-doc/index.html)


minrk:
I like the dedicated step zero pages a lot! It definitely feels nicer than scrolling past all of the instructions for providers that aren't yours.

I think it's still unclear which additional providers we should cover, but there's no harm in placeholders.

doc/source/additional-resources.rst
```diff
@@ -1,18 +1,19 @@
 .. _additional-resources:
 
-Additional resources
-====================
+Community-authored documentation
```
choldgraf: this is a great idea

doc/source/amazon/step-zero-aws.rst
```diff
@@ -0,0 +1,172 @@
+.. _amazon-aws:
```
choldgraf: I'm assuming that you haven't added *new* content to the provider-specific sections, so not giving them as close of a look for wording etc, let me know if that's wrong and you'd like a closer look!

doc/source/index.rst
```diff
@@ -11,27 +11,46 @@ hub remotely.
...
-questions, feel free to reach out at the `issues page`_.
+tools may change quickly.
+
+If you have tips or deployments that you would like to share, see
+:ref:`community-resources`. If you see anything that is incorrect
+or have any questions, feel free to reach out at the `issues page`_.
+
+.. _getting-to-zero:
+
+Getting to Step Zero: your Kubernetes cluster
```
choldgraf: I like this wording :-)

doc/source/amazon/step-zero-aws.rst
willingc: Yep. No big changes to provider content except links to images from moving the file.

doc/source/index.rst
```diff
@@ -78,9 +97,37 @@ cloud-based deployments and tips for maintaining your deployment.
...
+This section gives the community a space to provide information on setting
+up, managing, and maintaining JupyterHub.
+
+.. important::
+   We recognize that Kubernetes has many deployment options. As a project team
+   with limited resources to provide end user support, we rely on community
+   members to share their collective Kubernetes knowledge and JupyterHub
+   experiences.
+
+We hope that you will use this section to share deployments with on a variety
```
choldgraf: with/on - we should choose one!

doc/source/index.rst
```diff
@@ -78,9 +97,37 @@ cloud-based deployments and tips for maintaining your deployment.
...
+of infrastructure and for different use cases.
```
choldgraf: does "variety of infrastructure" parse? maybe it should be "variety of deployments"?

willingc: Yeah. I see a typo too. Please open a PR to reword.

choldgraf:
I had a few small comments that shouldn't be blocking on merge (which is good, since this is now merged haha). Feel free to spot-check the individual comments, or we can just take care of these little things over time. I'm finishing up my last day of euro-traveling and conferencing so will have more time for this starting soon.

choldgraf:
(and thanks very much @willingc , I think this is a much-needed refactor)

willingc:
Thanks @choldgraf. Safe travels.

willingc:
That would be fantastic @jzf2101. Would you mind taking care of that?

jzf2101:
I'll also followup with additional suggestions if I have anything I'd like to add in the review

PR #595, opened 2018-03-21
Add property to control debug logging

manics:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/589

If `debug.enabled` this turns on debug logging for JupyterHub, Spawner, and the included nginx ingress container. configurable-http-proxy is unchanged since it's [already at debug](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6b5bf6dc13020baa089355fdd36408165e6f3582/jupyterhub/values.yaml#L74) (though should it default to info instead?)


jupyterhub/values.yaml
```diff
@@ -204,3 +204,6 @@ cull:
   users: false
   timeout: 3600
   every: 600
+
+debug:
+  enabled: false
```
minrk: Is there a reason for this to be `debug.enabled` instead of just `debug: false`?

manics: I copied the style used for [`rbac.enabled`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/e2fbdc247953d17563fb017e4e3d00e66d2c1d1d/jupyterhub/values.yaml#L49)

PR #596, opened 2018-03-21
Allow singleuser_image_pull_policy to be modified

manics:
This is useful during development of singleuser images. Most other images in this chart already have a `pullPolicy` or `imagePullPolicy` property.


yuvipanda:
ty for the PR, @manics!

PR #604, opened 2018-03-27
enable debug for hub

clkao:


yuvipanda:
Thanks for the PR, @clkao!

PR #609, opened 2018-03-27
Add nodePort option helm chart for hub

summerswallow:
The hub section of the helm chart permits the assigning of type to the Service for hub. By default this is ClusterIP. However, it may be desirable for the hub to be exposed so that an external load balancer can be used. Overriding the type to "NodePort" through a config.yaml file does work. However, without the ability to configure the nodePort specifically Kubernetes selects a random port between 30000 and 32767. For an external load balancer it is desirable to keep this port fixed.

### DIFF
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -13,3 +13,6 @@ spec:
     - protocol: TCP
       port: 8081
       targetPort: 8081
+      {{ if .Values.hub.service.ports.nodePort -}}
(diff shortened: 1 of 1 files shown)


summerswallow:
If I didn't do this correctly feel free to correct it. I tarred the helm chart seems to do what I intended.

summerswallow:
I've added the requested change. I originally had this in there, but tried it without and thought it worked. Well, let's see if this does the trick

consideRatio:
@summerswallow build success! 🎉

minrk:
Thanks @summerswallow!

PR #611, opened 2018-03-30
Create kops direct guide for AWS

### LINKED ISSUE #243: Switch to recommending kops for AWS installation
kops no longer requires that we use route53 + DNS, and is generally more robust + production worthy than the current quickstart we're using.

https://github.com/kubernetes/kops/blob/master/docs/aws.md is the doc for using it. we should use the 'k8s.local' option for DNS and change our instructions to use kops!

cam72cam:
Fixes #243

I have gone through my internal docs for setting up a k8s cluster using kops and ported them to RST.

I have created a second file instead of overwriting the existing guide.  I can change that before merge if we want to completely replace the heptio guide.


doc/source/amazon/step-zero-aws-kops.rst
```diff
@@ -0,0 +1,130 @@
...
+AWS does not have native support for Kubernetes, however there are
+many organizations that have put together their own solutions and
+guides for setting up Kubernetes on AWS.
+
+This guide uses kops to setup a cluster on AWS.  This should be seen as a rough template you will use to
+setup and shape your cluster.
+
+Procedure:
+
+1. Create a AIM Role
```
minrk: IAM role?

cam72cam:
At this point the only downside to the kops guide (IMO) is that I don't have a solid understanding of the exposed ports / security settings.

choldgraf:
TBH I don't think anybody has a really keen understanding of the ins/outs of Heptio either, so I'd be +1 on replacing the heptio guide as long as we know that it's straightforward to deploy.

@cam72cam is your guide "ready" from your perspective? If so, then I think the path forward is for some other folks to try deploying following your steps, and make sure that it all works out. WDYT?

cam72cam:
Running through that now

cam72cam:
Guide works, I am playing with the VPC and private networking options.  Not sure if I want to include that in the tutorial

cam72cam:
@minrk @choldgraf I've done a few updates and overridden the existing aws guide.

The new guide includes some notes on security and provides the option of using a private topology

doc/source/amazon/step-zero-aws.rst
```diff
@@ -7,166 +7,162 @@ AWS does not have native support for Kubernetes, however there are
...
+         --node-size t2.medium \
+         --node-volume-size 10 \
+         --yes
+
+   For a more secure setup add the following params to the kops command::
+   
+         --topology private \
+         --networking weave \
+
+   This creates a cluster where all of the masters and nodes are in private subnets and don't have external IP addresses.  A mis-configured security group or insecure ssh configuration is less likely to comprimize the cluster.
```
summerswallow-whi: comprimize->compromise

doc/source/amazon/step-zero-aws.rst
```diff
@@ -7,166 +7,162 @@ AWS does not have native support for Kubernetes, however there are
...
+7. Set the region to deploy in
+
+   ``export REGION=`curl -s http://169.254.169.254/latest/dynamic/instance-identity/document|grep region|awk -F\" '{print $4}'```
+
+8. Set the availability zones for the nodes
+
+   For this guide we will be allowing nodes to be deployed in all AZs::
+  
+       export ZONES=$(aws ec2 describe-availability-zones --region $REGION | grep ZoneName | awk '{print $2}' | tr -d '"')
+       export ZONES=$(echo $ZONES | tr -d " " | rev | cut -c 2- | rev)``
```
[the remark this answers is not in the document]
cam72cam: The lines already have a , in them.  It's a json struct and the line looks like:
```
"ZoneName": "somename",
```

doc/source/amazon/step-zero-aws.rst
```diff
@@ -7,166 +7,162 @@ AWS does not have native support for Kubernetes, however there are
...
+    few basic services are up and running.
+   
+    Keep running 'kops validate cluster' until you see "Your cluster $NAME is ready" at the end of the output.
+   
+    ``time until kops validate cluster; do sleep 15 ; done`` can be used to automate the waiting process.
+    
+    If at any point you wish to destroy your cluster after this step, run ``kops delete cluster $NAME --yes``
+    
+
+11. Confirm that ``kubectl`` is connected to your Kubernetes cluster.
```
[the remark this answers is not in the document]
cam72cam: I have never needed to do those steps?  What does it facilitate?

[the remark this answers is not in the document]
summerswallow-whi: I've done the Cloudformation with Heptio a number of times. 
This is just to give the same functionality as described in steps 3-5 of the Heptio guide.

Part of the reason I bring it up is that it's not the easiest thing to figure out.
But if you copy step 3 from old Heptio. Insert this command, then step 4 and 5 can proceed as described.
This will enable the user later using helm to run that from their own machine rather than copy any customizations and configurations to the host machine.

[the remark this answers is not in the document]
summerswallow-whi: It will integrate this section better with the next section regarding the use of helm

doc/source/amazon/step-zero-aws.rst
```diff
@@ -7,166 +7,162 @@ AWS does not have native support for Kubernetes, however there are
...
+        parameters:
+          type: gp2
+
+    Next, run this command:
+
+        .. code-block:: bash
+
+           kubectl apply -f storageclass.yml
+
+    This enables `dynamic provisioning
```
summerswallow-whi: In a discussion with @willingc, she thought that including the instructions for  creating an encrypted storage class would be useful in the documentation.
An easy way to do this is:
1) add to the storageclass.yml file in the parameters section:
 encrypted: "true"  .. should note that the quotes are important
2)  run kubectl delete storageclass gp2   --- This step is important as you can not modify a storage class apparently
3)  kubectl appy -f storageclass.yml

willingc:
Thanks for the review @summerswallow-whi. 

@cam72cam It would be great to iterate on this PR as we have a lot of folks interested in AWS. Great job getting the ball moving :sunny:

doc/source/amazon/step-zero-aws.rst
cam72cam: Can we defer this to a future PR?  I'd like to keep this as close to the original as possible.

doc/source/amazon/step-zero-aws.rst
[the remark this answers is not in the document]
cam72cam: Ah ok, our plans are to lock down access to the CI machine used to run the KOPS commands, but this would be good to re-include.  I'll make the change and test on monday when I am back from vacation.

doc/source/amazon/step-zero-aws.rst
summerswallow-whi: Sure. Thought since I was in here to add it.

cam72cam:
@willingc @summerswallow-whi @minrk @choldgraf I think this PR is close to being ready to merge?  Any thoughts?

doc/source/amazon/step-zero-aws.rst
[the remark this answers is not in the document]
cam72cam: I've added and tested e27d5b9

choldgraf:
@summerswallow-whi (or anybody else) can you confirm that these steps work for you in creating a new cluster?

I think we'll need to update these instructions to be more user-friendly, but don't want that to block this PR.

PS: @cam72cam in the future could you please create a branch and work from that, rather than working from `master`?

doc/source/amazon/step-zero-aws.rst
```diff
@@ -3,183 +3,171 @@
...
+11. Confirm that ``kubectl`` is connected to your Kubernetes cluster.
+
+    Run::
+
+       kubectl get nodes
+
+    You should see a list of two nodes, each beginning with ``ip``.
+    
+    If you want to run kubectl from a box not on AWS, you can use run the following on AWS: ``kubectl export kubecfg``
+
```
summerswallow-whi: I think you want kops export kubecfg not 
kubectl.

summerswallow-whi:
I followed these steps. They seem to work with the one exception I mentioned above about the ZONES, but I was easily able to work around it.

I do agree that the instructions are a bit hard to follow, especially if you are new with AWS.

choldgraf:
ok cool - then I'm +1 on merging this after @cam72cam spot-checks @summerswallow-whi 's one comment. Then we can open an issue about making the `kops` guide more user-friendly, and someone else (maybe me) can use that as an opportunity to run through these instructions with fresh eyes and iterate on another PR.

@willingc what do you think?

willingc:
I'm taking a full day K8 on AWS course on Monday. I'm okay if we merge now or wait until after.

doc/source/amazon/step-zero-aws.rst
cam72cam: Thanks for catching that :+1:

choldgraf:
woot! thanks for making the fix @cam72cam , I'll merge this now since it sounds like the information is technically correct and quite useful! We can iterate on future PRs!

Thanks so much @cam72cam !!! :-)

PR #614, opened 2018-03-30
NetworkPolicy access to proxy 8000 / 8443

manics:
I think https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/592 means direct access to port 8000 on the proxy may be required. Without it I get a 504 gateway error from my external K8s ingress.

I'm not sure if this port needs to be always open, or only if the internal HTTPS ingress is disabled


yuvipanda:
Good catch, @manics. Yes, this is requied only when internal https ingress is disabled. I didn't catch it because I already had a network policy doing the equivalent of this (and some additional stuff - https://github.com/berkeley-dsep-infra/data8xhub/blob/master/hub/templates/networkpolicy.yaml).

Can you make a conditional for it?

minrk:
Nice!

PR #621, opened 2018-04-01
debug chartpress, disable shallow-clone

### LINKED ISSUE #612: Failed to pull image "jupyterhub/k8s-network-tools:d6340ab"
Using JupyterHub version v0.7-be3d3b7 https://jupyterhub.github.io/helm-chart/jupyterhub-v0.7-be3d3b7.tgz

`jupyterhub/k8s-network-tools:d6340ab` isn't on [Docker Hub](https://hub.docker.com/r/jupyterhub/k8s-network-tools/tags/), the most recent tag is `81c2613` from 7 days ago

minrk:
- run `git diff` after chartpress to show changes, in case something suspicious is in values.yaml
- pin specific chartpress commit so that chartpress only changes explicitly (there's also a slight possibility of caching being responsible for the recent weird chartpress behavior in #612)
- rerender values.yaml with recent changes that deviated slightly in style to avoid churn when chartpress rerenders values.yaml

closes #612

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
 - ./ci/install.sh
 script:
 - chartpress --commit-range ${TRAVIS_COMMIT_RANGE}
+- git diff
 - ./ci/test.sh
 
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
 - 3.6
+git:
+  depth: false
 services:
 - docker
 install:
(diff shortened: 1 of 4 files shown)


minrk:
chartpress is indeed getting the wrong commits for network-tools and singleuser-sample on Travis. The rest are correct. Interestingly enough, it's getting the right commits when I run it locally. Investigating...

minrk:
This was caused by shallow-cloning done by travis, which resulted in misidentifying the latest commits of images that haven't changed in a bit. Disabling the shallow-clone behavior has fixed the image tagging.

PR #629, opened 2018-04-03
clarify expectations for cost estimation

### LINKED ISSUE #626: Users not sure about what Cost Estimator means
Faculty from some universities view it as a estimate of how much we are charging for the service which stops them from proceeding further. It should be explained how we are not charging for providing cloud services and that many cloud vendors give free resources to academic partners.

willingc:
Closes #626 

- Moves cost projection section to an appendix of the administrator guide.
- Adds a clarification section at the beginning
- Reorder to have all factors precede the cost estimator

### DIFF
--- a/doc/source/cost.rst
+++ b/doc/source/cost.rst
@@ -1,7 +1,19 @@
 .. _cost:
 
-Estimating costs
-================
+Appendix: Projecting deployment costs
+=====================================
+
+.. important:: Clarification on cost projections
+
+   As a non-profit research project, Project Jupyter does not offer,
+   recommend, or sell cloud deployment services for JupyterHub.
+
+   The information in this section is offered as guidance as requested
+   by our users. We **caution** that costs can vary widely based
+   on providers selected and your use cases.
+
+Factors influencing costs
+-------------------------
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
I like it, @willingc!

doc/source/cost.rst
```diff
@@ -48,35 +83,14 @@ you.
...
-all use the JupyterHub at once, such as during a large class workshop?
-will users use JupyterHub at different times of day?
-
-The usage patterns and peak load on the system have important implications for
-the resources you need to provide. In the future JupyterHub will have
-auto-scaling functionality, but currently it does not. This means that you need
-to provision resources for the *maximum* expected number of users at one time.
-
 
 Examples
```
betatim: Maybe we move this further to the top (or reference it from the beginning) when we have a few more examples. It probably provides the best examples/guidance on how to work out how much it might cost you. As the hard part is converting "N students in my semester long course" into "number of CPUs and RAM to have available". After that you "just " have to look up your cloud providers prices ;)

betatim:
Ah, I overlooked the review request. Chris if you want something changed let me know and we can make a new PR.

PR #631, opened 2018-04-03
Minor typo fix: cluser -> cluster

jtpio:
Just a minor typo that caught my eyes while going through the documentation


betatim:
Thanks for catching this!

PR #632, opened 2018-04-03
Document network policies

manics:
Documentation for https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/546#issuecomment-375025036

### DIFF
--- a/doc/source/security.md
+++ b/doc/source/security.md
@@ -159,11 +159,11 @@ There's ongoing work on making this easier!
 
 ## Kubernetes Network Policies
 
-Kubernetes has optional support for [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) which let you restrict how pods can communication with each other and the outside world. This can provide additional security within JupyterHub, and can also be used to limit network access for users of JupyterHub.
+Kubernetes has optional support for [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) which lets you restrict how pods can communicate with each other and the outside world. This can provide additional security within JupyterHub, and can also be used to limit network access for users of JupyterHub. By default, the JupyterHub helm chart **disables** network policies.
 
-Your Kubernetes cluster may have partial, full, or no support for network policies, and Kubernetes will **silently ignore** policies that aren't supported. If you rely on a policy to restrict what users can access you should verify the policies are enforced on your cluster.
+**Important**: If you decide to enable network policies, you should be aware that a Kubernetes cluster may have partial, full, or no support for network policies. Kubernetes will **silently ignore** policies that aren't supported. Please use **caution** if enabling network policies and verify the policies behave as expected, especially if you rely on them to restrict what users can access.
 
-Network policies are disabled by default, you can enable them in your `config.yaml`:
+You can enable network policies in your `config.yaml`:
 
 ```yaml
 hub:
@@ -177,7 +177,7 @@ singleuser:
     enabled: true
 ```
 
-The default singleuser policy allows all outbound network traffic. To restrict outbound traffic to DNS, HTTP and HTTPS:
+The default singleuser policy allows all outbound network traffic, meaning JupyterHub users are able to connect to all resources inside and outside your network. To restrict outbound traffic to DNS, HTTP and HTTPS:
 
 ```yaml
 singleuser:
@@ -195,4 +195,4 @@ singleuser:
         protocol: TCP
 ```
 
-See the Kubernetes documentation for further information on defining policies.
+See the [Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/) for further information on defining policies.
--- a/doc/source/security.md
+++ b/doc/source/security.md
@@ -155,3 +155,44 @@ This is a sensitive security issue (similar to writing sudo rules in a
 traditional computing environment), so be very careful.
 
 There's ongoing work on making this easier!
+
+
+## Kubernetes Network Policies
+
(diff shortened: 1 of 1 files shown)


doc/source/security.md
```diff
@@ -155,3 +155,44 @@ This is a sensitive security issue (similar to writing sudo rules in a
...
+    enabled: true
+proxy:
+  networkPolicy:
+    enabled: true
+singleuser:
+  networkPolicy:
+    enabled: true
+```
+
+The default singleuser policy allows all outbound network traffic. To restrict outbound traffic to DNS, HTTP and HTTPS:
```
consideRatio: I'm currently commenting on a field where I have little knowledge, but that is a common point of view as well to consider :D 

Perhaps you could mention something about the default network policies for hub/proxy/singleuser as well as kubernetes?

Also have a question myself based on not being read up about network policies, but when you say "outbound traffic", does that mean "cluster outbound" or "pod outbound"? It could be helpful for a person like me to comprehend a difference there if there are.

doc/source/security.md
```diff
@@ -155,3 +155,44 @@ This is a sensitive security issue (similar to writing sudo rules in a
...
+        protocol: UDP
+    - ports:
+      - port: 80
+        protocol: TCP
+    - ports:
+      - port: 433
+        protocol: TCP
+```
+
+See the Kubernetes documentation for further information on defining policies.
```
consideRatio: Perhaps provide a link here as well. This URL is the most relevant documentation on k8s right?
- https://kubernetes.io/docs/concepts/services-networking/network-policies/

consideRatio: @manics I forgot to communicate a big warm appreciation for your contribution !! Thank you =D

doc/source/security.md
```diff
@@ -155,3 +155,44 @@ This is a sensitive security issue (similar to writing sudo rules in a
 traditional computing environment), so be very careful.
 
 There's ongoing work on making this easier!
+
+
+## Kubernetes Network Policies
+
+Kubernetes has optional support for [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) which let you restrict how pods can communication with each other and the outside world. This can provide additional security within JupyterHub, and can also be used to limit network access for users of JupyterHub.
```
willingc: @manics Thanks for the PR. The style for Markdown documents in this guide is to keep to an 80 character limit for paragraphs of text. Please update the PR to split this into multiple lines. Thanks!

A few substitutions:

s/let/lets s/communication/communicate

doc/source/security.md
```diff
@@ -155,3 +155,44 @@ This is a sensitive security issue (similar to writing sudo rules in a
...
+
```
willingc: By default, the JupyterHub helm chart **disables** network policies.

### Enabling network policies

**Important:** If you decide to enable network policies, you should be aware that a Kubernetes cluster may have partial, full, or no support for network policies.  Kubernetes will **silent ignore** policies that aren't supported. If you rely on a policy to restrict what users can access you should verify the policies are enforced on your cluster. Please use **caution** if enabling network policies to make sure that the policies behave as expected.

doc/source/security.md
manics: I've expanded the line about outbound singleuser policies. I'm unsure what to add for hub and proxy though since I at the moment I think it's more relevant for developers of this chart, and I don't want to clutter the doc with too much information.

doc/source/security.md
manics: I'd already included that link in the first paragraph, but I guess there's no harm in repeating it.

doc/source/security.md
manics: @willingc I've reworded your suggested text slightly to avoid repetition. Does this look OK?

doc/source/security.md
```diff
@@ -159,9 +159,19 @@ There's ongoing work on making this easier!
...
 
-Kubernetes has optional support for [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) which lets you restrict how pods can communicate with each other and the outside world. This can provide additional security within JupyterHub, and can also be used to limit network access for users of JupyterHub. By default, the JupyterHub helm chart **disables** network policies.
-
-**Important**: If you decide to enable network policies, you should be aware that a Kubernetes cluster may have partial, full, or no support for network policies. Kubernetes will **silently ignore** policies that aren't supported. Please use **caution** if enabling network policies and verify the policies behave as expected, especially if you rely on them to restrict what users can access.
+Kubernetes has optional support for [network
+policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+which lets you restrict how pods can communicate with each other and the outside
+world. This can provide additional security within JupyterHub, and can also be
+used to limit network access for users of JupyterHub. By default, the JupyterHub
```
willingc: Let's make the sentence "By default..." be its own paragraph so it stands out to the reader.

doc/source/security.md
```diff
@@ -159,9 +159,19 @@ There's ongoing work on making this easier!
...
+helm chart **disables** network policies.
+
+**Important**: If you decide to enable network policies, you should be aware
```
willingc: Let's add a subtitle before the Important block `### Enabling network policies`

manics:
Updated

consideRatio:
@manics thank you so much for your PR's! I think they are a great addition to the project! I did not know about network policies before. It took quite some time for me to finally grasp how they work but I think I've got it now. I'm quite excited about it for the simple reason I've invested quite some time to learn about it now haha :D

I've become eager to stand on your shoulders and add to this work, is it okay if I tag in @manics? I'd like to make it so that if you enable the network policies, they will restrict communication to the essentials (but still allowing augmenting the policies with additional whitelisting as before). I'd also like to add policies for the image-puller pods.

@willingc this looks good to me!

willingc:
Thanks @manics for the PR and @consideRatio for the review. I'm happy to merge this. :cake:

PR #633, opened 2018-04-03
pod.status.start_time can be None

minrk:
if the pod hasn’t started yet

closes https://github.com/jupyterhub/binderhub/issues/510

### DIFF
--- a/images/pod-culler/culler
+++ b/images/pod-culler/culler
@@ -148,6 +148,9 @@ class PodCuller:
         for pod in pods:
             name = pod.metadata.name
             # API results always use UTC timezone
+            if pod.status.start_time is None:
+                # no start time, it hasn't started yet!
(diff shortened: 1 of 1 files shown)


consideRatio:
@minrk this looks good to me, and I'm very confident that this not cause issues if I'd press merge so here goes!

My internal voice:
* Hmmm how many should review a PR before it is merged?
* I figure it depends, in this case, perhaps only one quick look by someone else?

betatim:
> How many should review a PR?

I think if you aren't sure you can comment that it looks good to you and invite someone else to take a look as well. Otherwise mostly one person seems to be enough for small things. For bigger things maybe give people time to look at it and comment.

consideRatio:
@betatim sounds great! Thank you for giving me input on this!

PR #635, opened 2018-04-04
[MRG] values.yaml: extraEnv and extraImages

consideRatio:
- hub.extraEnv's default value now {}, before it was [].
- prePuller.extraImages was referenced in the image-puller daemonsets, and now has an empty placeholder.
- I also made some spacing consistent throughout the file.

[@manics wrote](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/597#issuecomment-378667725):
> I thought I'd mention a use case for keeping the backwards compatibility in future too. If extraEnv is a dict you can easily merge a public config in a git repo with private variables set on the command line during, e.g.
>
>```yaml
> hub:
>   extraEnv:
>     OAUTH2_AUTHORIZE_URL: "https://login.elixir-czech.org/oidc/authorize"
>     OAUTH2_TOKEN_URL: "https://login.elixir-czech.org/oidc/token"
>     OAUTH_CALLBACK_URL: https://example.org/hub/oauth_callback
>     OAUTH_CLIENT_ID: xxx
>```
> and set the private variables separately during deployment: --set hub.extraEnv.OAUTH_CLIENT_SECRET=XXX, but to my knowledge you can't do this when extraEnv is a list.

@manics / @yuvipanda --- will changing extraEnv from [] to {} cause more trouble than future benefit?

I can imagine someone having assumed it to be an array rather a dict, and that could cause issues. At the same time I'd like to encourage people to use it as an dictionary due to benefits described by @manics in the comment above.

__Update 1__
I realized that the default value of extraEnv was changed from a dict to a list in #597 very recently, so changing it back should be super safe.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -26,7 +26,7 @@ hub:
   labels: {}
   extraConfig: {}
   extraConfigMap: {}
-  extraEnv: []
+  extraEnv: {}
   extraContainers: []
   extraVolumes: []
   extraVolumeMounts: []
@@ -48,8 +48,10 @@ hub:
           - ipBlock:
               cidr: 0.0.0.0/0
 
+
 rbac:
   enabled: true
+
 
 proxy:
   secretToken: ''
@@ -104,7 +106,6 @@ proxy:
               cidr: 0.0.0.0/0
 
 
-# Google OAuth secrets
 auth:
   type: dummy
   whitelist:
@@ -173,6 +174,7 @@ singleuser:
   cmd: jupyterhub-singleuser
   defaultUrl:
 
+
 prePuller:
   hook:
     enabled: true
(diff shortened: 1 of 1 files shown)


manics:
@consideRatio My reading of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/597 is that it supports both dicts and lists. I added my comment because of the reference to backwards-compatibility which implies it may be removed in future, and I wanted to make the case that it should always remain supported. Sorry for any confusion!

consideRatio:
Ah thanks @manics now I better grasp the situation! BTW I also noticed that the extraEnv's default value was changed to [] from {} very recently.

In other words, this PR simply revert that default value change, which I think caused me warnings of type conversions, and adds a placeholder for extraImages.

PR #637, opened 2018-04-08
kubeVersion in Charts.yaml works now

consideRatio:
This works for now. I started some discussion on kubernetes/helm and  regarding the surprise I got when the ">=1.8.0" constraint failed to match "v1.9.4-gke.1".

For reference, see [kubernetes/helm issue #3810](https://github.com/kubernetes/helm/issues/3810) and [Masterminds/semver issue #69](https://github.com/Masterminds/semver/issues/69).


yuvipanda:
Thank you for engaging with the upstream community and finding a nice solution! :tada:

PR #625, opened 2018-04-02
.yaml refresher

consideRatio:
#### The PR's size
Making many separate PRs would have been good for the ability to review them. I had that ambition initially but realized that making many PRs each affecting a lot of files and the same files over and over was hard and I gave up on that, I'm hoping it won't cause too much trouble - sorry in advance!

## In this PR
I have aimed make the `.yaml` consistent in various ways.

- [x] Labels (using helper)
- [x] Selectors (using helper)
- [x] Block indentation (`nindent` introduced)
- [x] List element indentation
- [x] Whitespace chomp left (`{{- }}`)
- [x] Template function (`include` replace `template`)
- [x] Yaml and Helm comments (`#` vs `{{/* */}}`)
- [x] Passing scope (with `merge (dict "key1" "val1") .`)
- [x] Utilize [`helm lint`](https://docs.helm.sh/helm/#helm-lint) 
- [x] Utilize [`yamllint`](https://github.com/adrienverge/yamllint) [like kubernetes/charts](https://github.com/kubernetes/charts/blob/master/test/circle/lintconf.yml)
- [x] Setup values for `yamllint` (`lint-chart-values.yaml`)
- [x] Utilize [`kubeval`](https://github.com/garethr/kubeval)
- [x] Linting / Validation tooling (`tools/lint.py`)
- [x] Update guidelines (`CONTRIBUTING.md`)
- [x] Consider dynamic .yaml in [chartpress](https://github.com/jupyterhub/chartpress)
- [x] Consider dynamic .yaml in [pod culler](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/images/pod-culler)
- [x] Consider dynamic .yaml in [kubespawner](https://github.com/jupyterhub/kubespawner) ([kubespawner#159](https://github.com/jupyterhub/kubespawner/issues/159), [kubespawner#161](https://github.com/jupyterhub/kubespawner/pull/161))
- [x] Test run cluster upgrades
- [x] Merge [kubespawner#161](https://github.com/jupyterhub/kubespawner/pull/161) then bump to utilize it.

#### Issues and PR's
[...]

PR #640, opened 2018-04-09
Dont trim whitespace before toYaml to avoid extra indentation

actionshrimp:
The whitespace trim here causes newline to be removed, so the first line of the `toYaml` block receives the indent from the `if` line. This produces broken output when `extraEnv` of `[]` is present.


consideRatio:
Thank you for the PR @actionshrimp, you are right!

PS: I'm hoping to fix most issues like this in #625.

PR #642, opened 2018-04-12
import updated culler

minrk:
from https://github.com/jupyterhub/jupyterhub/pull/1779

- use semaphore to limit Hub API activity, which may improve hub performance with lots of culling (e.g. mybinder.org)
- supports max-age culling (with jupyterhub 0.9), which will allow us to deprecate the podCuller image when we bump jupyterhub


betatim:
Could there be someone who is surprised (stuff breaks suddenly) by these changes?

Otherwise: merge, as this is a copy of what is in jhub.

yuvipanda:
Culler should be its own package now, IMO :)

betatim:
-> [subtle nudge that someone should create a package](https://github.com/jupyterhub/jupyterhub/pull/1779#issuecomment-380743640) :D

PR #649, opened 2018-04-16
Support LDAP login (LDAPAuthenticator)

manics:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/264

Based on https://github.com/jupyterhub/ldapauthenticator/blob/1bb93f3ad258fdcc464c5e5c88f36005ae579ce0/README.md
I've tried to follow the general pattern in `jupyterhub_config.py` when naming LDAP parameters but not everything is clear to me.

LDAPAuthenticator seems to have many more parameters than the other auth providers, so I added a new function `set_config_if_not_none` in z2jh.py to make it easier to keep the built-in defaults from LDAPAuthenticator when a value isn't overridden.

I've only tested this with my own setup:
```yaml
auth:
  type: ldap
  ldap:
    server:
      address: ldap.EXAMPLE.org
      port: 389
      ssl: False
    dn:
      templates:
        - 'cn={username},ou=edir,ou=people,ou=EXAMPLE-UNIT,o=EXAMPLE'
    allowedGroups:
      - 'cn=example-group,ou=groups,ou=EXAMPLE-UNIT,o=EXAMPLE'
```

Warning: I have **not tested** the other parameters, in particular the active directory setup from https://github.com/jupyterhub/ldapauthenticator/blob/1bb93f3ad258fdcc464c5e5c88f36005ae579ce0/README.md#active-directory-integration is untested by me.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -3,7 +3,7 @@ import glob
 from tornado.httpclient import AsyncHTTPClient
 from kubernetes import client
 
-from z2jh import get_config, get_secret
+from z2jh import get_config, get_secret, set_config_if_not_none
 
 # Configure JupyterHub to use the curl backend for making HTTP requests,
 # rather than the pure-python implementations. The default one starts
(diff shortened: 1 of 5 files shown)


manics:
The equivalent of the example given in https://github.com/jupyterhub/ldapauthenticator/blob/1bb93f3ad258fdcc464c5e5c88f36005ae579ce0/README.md#active-directory-integration should be
```yaml
auth:
  ldap:
    dn:
      lookup: true
      search:
        filter: '({login_attr}={login})'
        user: 'ldap_search_user_technical_account'
        password: 'secret'
        dnAttribute = 'cn'
      user:
        searchBase: 'ou=people,dc=wikimedia,dc=org'
        escape: False
        attribute: 'sAMAccountName'
```

PR #655, opened 2018-04-18
import updates to culler

minrk:
bugs fixed in the latest culler version which as been updated to support 0.9

gets fixes from https://github.com/jupyterhub/jupyterhub/pull/1807


minrk:
I think the issues fixed here may be related to the recent binder outage, so I'm going to deploy it all the way to mybinder.org.

choldgraf:
Thanks for catching this!

PR #657, opened 2018-04-18
Revert "bump kubespawner required revision"

minrk:
This reverts commit fad0f64936727ebafbdc8d55c68221627815de53.

Temporary fix because I think the event-monitoring is producing a memory leak on the Binder deployment.

Once confirmed on Binder, will start investigating in kubespawner.


choldgraf:
@minrk have we re-bumped kubespawner? Trying to see if I can write up some docs on how to make the docker image selectable, but it looks like we still need to build this into the helm chart

PR #665, opened 2018-04-25
Add AWS workshop material link to Step Zero

willingc:
It's been a while since we updated the AWS section, and AWS has posted workshop materials on GitHub.

### DIFF
--- a/doc/source/amazon/step-zero-aws.rst
+++ b/doc/source/amazon/step-zero-aws.rst
@@ -3,15 +3,22 @@
 Step Zero: Kubernetes on Amazon Web Services (AWS)
 --------------------------------------------------
 
-AWS does not have native support for Kubernetes, however there are
-many organizations that have put together their own solutions and
-guides for setting up Kubernetes on AWS.
+AWS support for Kubernetes is evolving quickly. Therefore, you have several
+options for setting up a Kubernetes cluster using AWS.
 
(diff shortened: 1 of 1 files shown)


choldgraf:
nice! how does the workshop guide compare to the kops guide in #611 ? Are the complimentary?

yuvipanda:
Thanks for finding these resuorces, @willingc! 

I feel the workshops are somewhat complex, and would not be something that most people would be able to follow if all they wanted to do was run JupyterHub on it. There is probably an uncountable number of ways to set up Kubernetes on AWS, so what we link to / recommend brings back #593.

I'd personally prefer we only get #611 merged, and link to the heptio guide only.

willingc:
It think we need to update the intro paragraph and I'm fine putting the workshop in an FYI box and removing the text. I do think that we should mention it as this workshop is what Amazon is promoting for AWS. From the user's perspective, I would like to know that such a tutorial/workshop exists even if just mentioned in passing.

I will update with the text box.

yuvipanda:
+1 for putting the link in a box! 

Thanks!

choldgraf:
also +1 for adding a link and reference

choldgraf:
+1 on on the language changes, I'll merge in a few unless @yuvipanda would like to see something else! We can also spot check when #611 lands

PR #668, opened 2018-04-28
Add example ldapauthenticator configuration

manics:
Documentation for https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/649

As with the original PR I haven't tested the AD example.


minrk:
Nice!

PR #670, opened 2018-05-02
Allow manual set of service account

gsemet:
This allows to disable the automatic creation of the service account (that requires clusterbinding priviledge) and define the service account for jupyter (hub and proxy) to use

Signed-off-by: Gaetan Semet <gaetan@xeberon.net>

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -51,7 +51,7 @@ hub:
 
 rbac:
   enabled: true
-
+  customServiceAccount: ''
 
 proxy:
   secretToken: ''
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -326,3 +326,18 @@ properties:
               The tag of the image to use.
 
               This is the value after the `:` in your full image name.
+
+  rbac:
+    type: object
+    properties:
+      enabled:
+        type: boolean
+        description: enable RBAC configuration
+      customServiceAccount:
+        type: string
+        description: |
+            Allow to force the service account to use for the puller, culler and https proxy.
+            If let to the default value (empty string), the service account will be automatically
+            created.
+            Please note it does not impact the singleuser pod
+            (see `singleuser.serviceAccountName`).
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -57,7 +57,11 @@ spec:
           claimName: hub-db-dir
       {{- end }}
       {{- if .Values.rbac.enabled }}
+      {{- if eq .Values.rbac.customServiceAccount "" }}
       serviceAccountName: hub
+      {{ else }}
+      serviceAccountName: {{ .Values.rbac.customServiceAccount }}
+      {{- end }}
       {{- end }}
       securityContext:
         runAsUser: {{ .Values.hub.uid }}
--- a/jupyterhub/templates/hub/rbac.yaml
+++ b/jupyterhub/templates/hub/rbac.yaml
@@ -1,4 +1,5 @@
 {{ if .Values.rbac.enabled -}}
+{{ if eq .Values.rbac.customServiceAccount "" }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -28,4 +29,5 @@ roleRef:
   kind: Role
   name: hub
   apiGroup: rbac.authorization.k8s.io
+{{- end }}
 {{- end }}
--- a/jupyterhub/templates/image-puller/job.yaml
+++ b/jupyterhub/templates/image-puller/job.yaml
@@ -1,5 +1,5 @@
 # This job has a part to play in a helm upgrade process. It simply waits for the
-# hook-image-puller daemonset which is started slightly before this job to get 
+# hook-image-puller daemonset which is started slightly before this job to get
 # its' pods running. If all those pods are running they must have pulled all the
 # required images on all nodes as they are used as init containers with a dummy
 # command.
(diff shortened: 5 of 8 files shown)


jupyterhub/values.yaml
```diff
@@ -51,7 +51,7 @@ hub:
 
 rbac:
   enabled: true
-
+  serviceAccount: ''
```
consideRatio: We will actually have a kubernetes `ServiceAccount` object even though the value `rbac.serviceAccount` in `values.yaml` is left blank. I'm thinking that if we name the value `rbac.customServiceAccount` we might indicate the behavior of the setting better. What do you think?

gsemet: i'm fine with it

consideRatio:
@gsemet thank you for the PR! It exposes a piece of missing knowledge to me that I need to figure out, I love it!!

I'm currently trying to grasp `"that requires clusterbinding priviledge"` part and simply understand the situation better overall.

#### Thinking aloud...
A cluster administrator or someone who would be allowed to create kubernetes objects through helm / kubectl. Is that user allowed to do certain things, but not all, like create a pod, but not a `ServiceAccount` object perhaps? I'm thinking this may be my main knowledge gap - where is cluster access defined and what rights do such user have?

A `ClusterRoleBinding` couples a `ServiceAccount` with a `ClusterRole` as seen in [image-puller's `rbac.yaml`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/10538f2c746a9b5a93c4f849e334a69288b0d0ee/jupyterhub/templates/image-puller/rbac.yaml) for example.

How I can progress...
- [ ]  I should look at the z2jh guide on the initial setup of kubectl to a cluster created.
- [ ] Eh... later...

manics:
My understanding is a ClusterRole has privileges across all namespaces, whereas a Role can only operate inside a single namespace. I don't know why the image-puller requires a ClusterRole instead of a Role.

consideRatio:
@gsemet you helped me understand things better! Thank you!

@manics good point, hmmm, so it will speak with the kubernetes api server, asking it about a daemonset, but it does not need to find daemonsets outside its namespace... It previously required privilege to create deaemonsets but no longer...

consideRatio:
@gsemet as I understand it. Your `User`'s cluster privilege is declared somewhere, and that does not allow you as great privilege as the predefined ClusterRole `cluster-admin`. But I think it will allow you to create ServiceAccount objects and Role's, but you may not bind ClusterRole's to it with a ClusterRoleBinding. I think this because i think ServiceAccounts have no permissions unless they have been coupled with a Role or ClusterRole through a RoleBinding or ClusterRoleBinding.

I think @manics is correct in that the image-awaiter job utilizing the ServiceAccount / ClusterRole / ClusterRolebinding in [image-puller's `rbac.yaml`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/image-puller/rbac.yaml).

If you fix #671, I think you solved your problem!

gsemet:
There are several places where the service account are created, at least: puller (that I don't use), culler (that I use) and the https proxy (that I don't use). For simplicity I have set all to use the same hardcoded service account (I actually use the service account used by Tiller). So I agree my solution is not clean, but it works.

I however do not have the time to work on it for the moment, but if one manage to restrict the roles to the namespace I think that would do the job perfectly!

gsemet:
we don't have https against our internal jupyter... but I think it will not have any impact, since we do'nt use nginx ingress. Traefik is really a good ingress. It already handles some https so it may not be difficult to turn https on against jupyter when the day comes :)

gsemet:
Yes service account can be easily created, it is the role that are restricted ! I have disabled the prepuller because it didn’t Worked (didn’t it need the docker socket at one point?). I can try it again but yes, I have disabled it on our prod

consideRatio:
@gsemet I'm still quite confused, but this is my current understanding of the situation.

1. You can create ServiceAccounts, Roles and RoleBindings as found in `/hub/rbac.yaml` ([Roles and ClusterRoles](https://kubernetes.io/docs/admin/authorization/rbac/#role-and-clusterrole))
2. You cannot create ClusterRoleBindings which are found within `/image-puller/rbac.yaml` and within `/proxy/autohttps/rbac.yaml`
3. Everything works if you have `prePuller.hook.enabled: false` and `proxy.autohttps.enabled: false` even though you don't specify the usage of pre-existing ServiceAccounts.

gsemet:
Yes :) I will try #674 patch ASAP (probably not before 2 weeks) ! And I probably been over hacky when I disabled everything that wasn’t working instead of trying to make it work.
The thing is we have a pretty cool but not so common environment:
- bare metal kubernetes
- traefik ingress (very very good). I would like to give a try with envoy as well. At the end we will probably have several ingresses
- namespace creation is restricted, and we only have the default role restricted with permission on our own namespace. I still need to investigate if we have enough role or not in our own namespace, but I really think this is the right pass. Nobody should be able to play with a clusterrole, it allows escalation too easily.
- so we have one tiller per namespace

We are investigating on how we can extra level of security (TLS on helm, ...) so we are learning everyday ! Thanks for your patience and support :)

consideRatio:
@gsemet I'm excited about the setup you have! I figure it may be essential to be able to use z2jh-k8s on baremetal for many administrators in the end and you are leading the way to making things work! I appreciate the possibility to learn from your experiences! :D

consideRatio:
Hey again @gsemet, have you tried non-custom ServiceAccounts after using #674 ?

gsemet:
Hi, not tried. I will try it today

consideRatio:
@gsemet my current understanding I've arrived to by learning through your experiences is that we should aim to avoid creating any `ClusterRoleBindings` as they require more permissions than some have over their cluster.

Currently, I think only the autohttps setup requires a `ClusterRoleBinding`, so if anything should be setup to use a custom predefined ServiceAccount with high allowed permissions it should be the autohttps deployment.

I'll close PR this for now. I'm very thankful that you made it, I got to learn a lot and now I know what we should aim for better.

Perhaps an upgrade from [kube-lego](https://github.com/jetstack/kube-lego) to [cert-manager](https://github.com/jetstack/cert-manager/) or what @minrk suggests in #710 might do the trick of getting rid of the last `ClusterRoleBinding` used for a `ServiceAccount` there as well.

Issue #178, opened 2017-09-19
Allow manually setting up TLS certificates


yuvipanda:
Would be nice, for cases where Let's Encrypt isn't a viable option.

yuvipanda:
@minrk did this in #246!

PR #674, opened 2018-05-03
Tighten up hook-image-puller's image-awaiter jobs requested RBAC permissions

### LINKED ISSUE #671: Too permissive rbac for hook-image-puller's image-awaiter job?
@manics realized that we might be able to narrow the permissions requested by the image-awaiter waiting for the hook-image-puller to finish pulling images!

The previous implementation of image-puller asked questions to the kubernetes api server about available nodes, and that _I think_ required a ClusterRole bound to a ServiceAccount.
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/64348a7d7d3c2372c354f1d7c74c4d9635ea874e/jupyterhub/templates/pre-puller/rbac.yaml

The newer implmentation does not require that, and might work well with a Role restricted to the namespace, as it simply speaks with the api server regarding namespaced resources, _I think_...
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/image-puller/rbac.yaml

@manics would you like to fix it and verify the pullers still works? It is the image-awaiter job created by alongside the hook-image-puller daemonset that needs the access the api-server, if that still functions after switching from ClusterRole to Role etc. I think the update is a success!

consideRatio:
The `image-awaiter` job used in conjunction with the `hook-image-puller` does not need a `ClusterRole` and a `ClusterRoleBinding` as it will no longer inspect the nodes in the cluster, but instead simply inspect a `DaemonSet` within the current namespace. This PR simply replaces `ClusterRole` with `Role`, and `ClusterRoleBinding` with `RoleBinding`.

* This PR is based on the insight of @manics and fixes #671.
* I think it will reduce but not fully fix the challenges for @gsemet (#178) who has a JupyterHub on a cluster but is limited to namespace-wide administrative permissions (Role) for the cluster rather than cluster-wide permissions (ClusterRole).

### DIFF
--- a/jupyterhub/templates/image-puller/rbac.yaml
+++ b/jupyterhub/templates/image-puller/rbac.yaml
@@ -14,7 +14,7 @@ metadata:
     "helm.sh/hook-weight": "0"
 ---
 # needs this role...
-kind: ClusterRole
+kind: Role
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: hook-image-awaiter-{{ .Release.Name }}-{{ .Release.Revision }}-{{ .Release.Time.Seconds }}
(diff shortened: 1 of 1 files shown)


choldgraf:
this seems reasonable to me! though will hold off on merging some someone with a bit more k8s expertise than I

PR #677, opened 2018-05-07
install jupyterhub 0.9 beta

minrk:
We've discussed making an 0.7 release of this chart prior to jupyterhub 0.9 and then an 0.8 release with jupyterhub 0.9 shortly after. We can still do that, but the jupyterhub 0.9 beta is out now, so we might want to fold them together, since we aren't quite ready to make the helm chart release yet.

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -9,7 +9,7 @@ charts:
       hub:
         valuesPath: hub.image
         buildArgs:
-          JUPYTERHUB_VERSION: 0.8.1
+          JUPYTERHUB_VERSION: 0.9.0b1
       network-tools:
         valuesPath: singleuser.networkTools.image
       image-awaiter:
@@ -17,6 +17,6 @@ charts:
       singleuser-sample:
         valuesPath: singleuser.image
         buildArgs:
-          JUPYTERHUB_VERSION: 0.8.1
+          JUPYTERHUB_VERSION: 0.9.0b1
       pod-culler:
         valuesPath: cull.podCuller.image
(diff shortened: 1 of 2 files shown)


choldgraf:
I'm +1 -  will hold off on merge for a bit to see if others have differing opinions!

minrk:
I'm + 1, too. Selfishly, I'd like to use Binder as a test bed for the 0.9 betas. We have a few backed-up changes ready to deploy for the scheduled upgrade today (https://github.com/jupyterhub/binderhub/pull/541 and https://github.com/jupyter/repo2docker/pull/298 are the main ones, I think), so I'd like to hold this one until after today's deploy settles. Maybe deploy it on Thursday or next week.

I will also try to do a test deploy of this myself on my own very-low load deployment to verify that upgrade works with running servers.

yuvipanda:
Go for it!

chartpress.yaml
```diff
@@ -9,14 +9,14 @@ charts:
       hub:
         valuesPath: hub.image
         buildArgs:
-          JUPYTERHUB_VERSION: 0.8.1
+          JUPYTERHUB_VERSION: 0.9.0b1
```
consideRatio: Is this correct? Writing the release as `0.9.0b1` rather than for example `0.9.0-b1`?

Is there a reason for allowing a non-number in the last position (major, minor, patch - note that I'm assuming SemVer here). I've seen something like "MAJOR.MINOR.PATCH-PRERELEASE" where prerelease could be alpha.1 or say beta.1.

willingc: @consideRatio Prereleases in Python typically do not include dashes. https://www.python.org/dev/peps/pep-0440/

PR #680, opened 2018-05-09
Bump kubespawner version

yuvipanda:
Routine bump to bring in latest kubespawner master.

Precursor to doing performance testing with JupyterHub 0.9.

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -10,5 +10,5 @@ mwoauth==0.3.2
 globus_sdk[jwt]==1.2.1
 oauthenticator==0.7.2
 cryptography==2.0.3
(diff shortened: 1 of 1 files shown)


consideRatio:
Let's do it!

For reference, the changes introduced in kubespawner in between latest version can be seen in the following link: https://github.com/jupyterhub/kubespawner/compare/86386e8...6d5993c

choldgraf:
this is the reversion of the reversion from a few weeks ago when we accidentally introduced the memory bug, yes? If @minrk doesn't get to it by tomorrow AM I'll give it a merge then.

choldgraf:
mergey merge merge

PR #686, opened 2018-05-15
authentication github org clarification

### LINKED ISSUE #684: 403 errors with github auth whitelist
I'm just capturing this github auth 403 whitelist problem in case other folks deploying find it useful, if I did something incorrect, or if this behavior is unexpected.  

On our JH instance at pangeo.esipfed.org, we initially were whitelisting users via the `auth: admin` section in `jupyter-config.yaml`, which makes them `admin` users, and that worked fine -- people on this list had no problem logging in.    

```
  auth:
    admin:
      access: true
      users:
        - jreadey
        - rsignell-usgs

    type: github
    github:
      clientId: "SECRET"
      clientSecret: "SECRET"
      callbackUrl: "http://pangeo.esipfed.org/hub/oauth_callback"
      org_whitelist:
        - "HDFGroup"
        - "pangeo-data"
        - "USGS-CMG"
    scopes:
      - "read:org"
```
I then decided to add `org_whitelist:` entries, and after this Helm upgrade, people who logged out to got 403 errors when they tried to log back in.  
Looking at the logs for the hub pod:
```
helm list
kubectl get pods -n esip-dev 
kubectl -n esip-dev logs hub-5647fc9dcd-m86gx
```
revealed that it said "User rsignell-usgs is not in org whitelist" even though I was listed as an admin, and also was a member of a whitelisted org.    

When I went to confirm my membership in the org on github, I saw that my membership in the org was "private", and when I switched it to "public":

I was able to finally login.

Originally reported on pangeo as https://github.com/pangeo-data/pangeo/issues/256#issue-323220095

choldgraf:
closes #684

@rsignell-usgs what do you think?

### DIFF
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -48,8 +48,18 @@ Below is the structure to use in order to authenticate with GitHub.
           callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
 
 
-By default this will allow *any* GitHub user to access your JupyterHub.
-You can restrict access to members of one or more GitHub organisations.
+Giving access to organizations on GitHub
+++++++++++++++++++++++++++++++++++++++++
+
+The configuration above will allow *any* GitHub user to access your JupyterHub.
+You can also restrict access to members of one or more GitHub organizations.
+To do so, see the configuration below.
+
+.. note::
+
+   A user's membership with the GitHub organization `must be public
+   <https://help.github.com/articles/publicizing-or-hiding-organization-membership/>`_
+   for organization-based whitelisting to work.
 
 .. code-block:: yaml
 


betatim:
Merged, if we need to we can make a new PR to adjust things further.

minrk:
> the JuptyerHub oath app client secret needs to be reset or user tokens revoked on github?

Yes, the tokens must be revoked. I believe GitHub has a "revoke all tokens" button for a given oauth application. That's probably what you want.

rsignell-usgs:
I revoked all the user tokens.   But with auth set up as:
```
  auth:
    admin:
      access: true
      users:
        - jreadey
        - rsignell-usgs

    type: github
    github:
      clientId: "SECRET"
      clientSecret: "SECRET"
      callbackUrl: "http://pangeo.esipfed.org/hub/oauth_callback"
      org_whitelist:
        - "HDFGroup"
        - "pangeo-data"
        - "ESIPFed"
        - "USGS-CMG"
    scopes:
      - "read:org"
```
I find that members of the above organizations, unless they are public, get challenged with:
pangeo_err
pangeo_err2

Is that expected, even though `auth.scopes` is `read:org`?

choldgraf:
hmmm, that's a good question. Could you open a new issue and copy/paste the message above so we don't lose track of it? (since this issue is closed)

PR #695, opened 2018-05-24
call the hub container 'hub'

minrk:
hub-container container is redundant

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -66,7 +66,7 @@ spec:
       {{- if .Values.hub.extraContainers }}
 {{ toYaml .Values.hub.extraContainers | indent 6 }}
       {{ end }}
-      - name: hub-container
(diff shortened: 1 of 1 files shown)


consideRatio:
=D I've also considered adding or not adding of say "-container" or "-config" or "-secret" to the names of various k8s resources, in #625.

Would you recommend we systematically avoid adding the resource type to the name? It is the most common practice within z2jh currently, but it is not systematic. The downside for the change, is that it might affect something that I don't figure out it might affect in all circumstances, so I've been cautious to do it systematically. 

In this case, I cant imagine there will be a problem as the container name isn't referenced anywhere within z2jh afaik, but if we change for example "hub-config" to "hub" we might need to update some documentation and people might have prewritten snippets with kubectl referencing "hub-config" etc, I guess the point of my caution for that is that I don't fully know.

betatim:
People who use hub-container containers also have PIN numbers :D

PR #707, opened 2018-05-31
Added encryption instructions for AWS deployment

summerswallow-whi:
My RST editing is not quite right. I appreciate comments on how to fix the formatting. I don't have a good previewer outside of git.

PR #708, opened 2018-06-01
update instructions for spinning down aws resources

oneklc:
Was unable to use the CloudFormation page above to delete resources (nothing was there to delete, yes region was correct) 
Terminating resources from ec2 console causes resources to be re-spawned.
Only wan to remove was to run `kops delete cluster`

### DIFF
--- a/doc/source/turn-off.rst
+++ b/doc/source/turn-off.rst
@@ -108,14 +108,23 @@ Amazon Web Services (AWS)
 1. Perform the steps in :ref:`delete-namespace`. These cloud provider agnostic steps will
    delete the helm chart and delete the hub's namespace. This must be done before proceeding.
 
-2. The easiest way to delete your cloud resources on AWS is to use their
-   website. Go to the ``CloudFormation`` page. This should have a list of all
-   running AWS stacks that you've created.
+2. on CI host:
 
-   If you followed the JupyterHub guide, there should be two items, both containing the name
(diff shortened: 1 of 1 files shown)


choldgraf:
hey @oneklc - thanks for pointing this out!

@cam72cam do you have any thoughts on this? It seems the cloudformation page may not always be a way to delete the k8s cluster.

I can take a pass to clean up the wording a bit, but wanna make sure we have the right information in there first.

cam72cam:
This looks like the right way to go, I did not know that this page existed when I updated the AWS docs.

Perhaps a note on how to do this if you loose your CI node would be helpful as well.
You can do this from any instance with kops installed as long as you do the same exports (NAME, KOPS_STATE_STORE)

oneklc:
Actually i had this issue as i spun down the ci host before realizing i needed it to delete the cluster. Had a tuff time getting back in cause i lost the private key when i terminated ci host.  Luckily i was able to get to the bastion host (cause it respawned) after some monkey work with aws (security and external ip), then set up kops again. Awscli was also need to be set up and configured. 

Thanks for the great documentation. Feel free to reword. Im no english major :)

willingc:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/707#issuecomment-396407004

Thanks!

willingc:
Congrats @oneklc for your first contribution to this repo too :tada:

Issue #538, opened 2018-02-26
Labels and Selectors - a common practice within the templates


consideRatio:
#### Related issue: #615

---

I was about to make set labels and selectors for a StatefulSet I tried out, and wanted to mimic what had been done for other k8s object templates, but did not find a clear pattern.

So, I'd like to figure out a practice to follow within the guide.
- What labels do we declare?
- What are important labels within a selector?

```yaml
metadata:
  name: proxy
  labels:
    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
    component: proxy
    heritage: {{ .Release.Service }}
    release: {{ .Release.Name }}
spec:
  replicas: 1
  selector:
    matchLabels:
      component: proxy
      release: {{ .Release.Name }}
      heritage: {{ .Release.Service }}
```

I'd like to discuss the following labels:
- chart, app, component, heritage, release

### My take
- I recommend we don't use the `chart` label, or is there a benefit to including that?
- I recommend we select on few labels in a minimalistic manner, perhaps only component or component and release is needed? What can go wrong by including / not including various labels?

Perhaps something like below?

#### The labels we declare on controllers
- [x] chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
- [x] app: jupyterhub
- [x] component: hub
- [x] release: {{ .Release.Name }}
- [x] heritage: {{ .Release.Service }}

#### The important labels within a selector and a controllers pod template
- [ ] chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
- [x] app: jupyterhub
- [x] component: hub
- [x] release: {{ .Release.Name }}
- [ ] heritage: {{ .Release.Service }}

yuvipanda:
https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md is where I took my cues from, and I think we should stick to it. Note that I didn't see that page until after a few releases, so it isn't consistently applied.

I'm happy for us to standardize this! I think we should include all the 5 ones that helm recommends in all our components. There's nothing wrong with changing them, but I think sticking to the standards helps other tooling / people who are used to helm charts to identify / write additional scripts that deal with z2jh.

consideRatio:
Ah! What about label selection in a `selector:`? It would be overkill to select based on every label. What about going for component + release?

yuvipanda:
[...]

PR #711, opened 2018-06-06
remove chart from jupyterhub.labels

minrk:
follow-up to #625 and related to discussion in #538

Using the chart version in a label causes chart version upgrades to relaunch everything with this label regardless of other changes, which we don't want, at least on most things (especially hub and proxy).

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -92,13 +92,17 @@
 {{- /*
   jupyterhub.commonLabels:
     Foundation for "jupyterhub.labels".
-    Provides labels: app, release, (chart and heritage).
+    Provides labels: app, release, (heritage).
 */}}
 {{- define "jupyterhub.commonLabels" -}}
 app: {{ .appLabel | default (include "jupyterhub.appLabel" .) }}
 release: {{ .Release.Name }}
 {{- if not .matchLabels }}
+{{- /*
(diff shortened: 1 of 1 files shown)


consideRatio:
Ah nice catch!!

I researched this and found an issue ([helm#1390](https://github.com/kubernetes/helm/issues/1390)) about it.

### Summary
* Labels on a Deployment object can be helm upgraded without restarts etc.
* The pod template declared within a Deployment, when updated, will force a restart of the Pods the Deployment controls. This is the actual issue we have.

Helm's best practices is to label objects we create with the chart label, but if we want to avoid a silly restart of the pods created by the Deployment, we should not change the template specs and labels within the template spec. I looked at a random chart in the kubernetes/chart repo and found that they did not specify chart or heritage on the Deployment created pods.

Thinking about it, it does not make much sense to say that the heritage of a pod created by a deployment, created by Tiller, is supposed to have a `heritage: tiller`.

### Suggested solution
We use the `matchLabels` helper in the Deployment's pod template's labels field instead of the `labels` helper.

consideRatio:
I'd like to fix this for all the Deployments, Daemonsets and Jobs (object that has a template for a pod they will control). Is it okay if I close this PR and make a new PR to do this @minrk ?

minrk:
@consideRatio thanks! Closed in favor of #712

PR #723, opened 2018-06-13
Fix link to extending Jupyterhub page

alanderex:


doc/source/setup-jupyterhub.rst
```diff
@@ -164,6 +164,6 @@ Install JupyterHub
    entering any username and password combination will let you enter the hub.
 
 Congratulations! Now that you have JupyterHub running, you can
-`extend it <extending-jupyterhub.html>`_ in many ways. You can use a pre-built
+`extend it <extending-jupyterhub.rst>`_ in many ways. You can use a pre-built
```
minrk: The target in the built docs is html, not rst, so this change would break the link.

However, these links should be using rst refs, not file extensions, so this should be:


```rst
:ref:`extend it <extending-jupyterhub>`
```

There appear to be several that should be fixed in this doc. Would you like to take a stab? You can run `make linkcheck` to verify that links resolve.

choldgraf:
thanks for the fix @alanderex !

PR #733, opened 2018-06-19
bump nginx-ingress and kube-lego images

minrk:


consideRatio:
I have not tried this on my cluster, but I did not see any breaking changes that should affect us in the nginx-ingress nor in kube-lego.

PR #744, opened 2018-06-29
bump kubespawner version (actually commit hash)

jhamman:
This just bumps the version of kubespawner in the hub requirements file. Perhaps a proper release from kubespawner would be a better idea but in the short term, I'm working on propagating the changes in https://github.com/jupyterhub/kubespawner/pull/193 through to JupyterHub and Binder.


yuvipanda:
thisisfine.jpg

Thanks for making the PR, @jhamman. This is fine!

PR #752, opened 2018-07-06
Fixes #751 - A pod-culler bug

### LINKED ISSUE #751: Pod culler is using the wrong label to watch pods
Min noticed that the pod culler on mybinder.org was watching for pods labelled as:
```
watching for pods with label selector component=pod-culler,app=jupyterhub,release=prod in namespace prod
```
which means it was watching for itself being older than the max age and then deleting itself.

The label is set via an environment variable https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/34cf104582f964e7818faf2b595e063fb9209dc5/jupyterhub/templates/pod-culler/deployment.yaml#L34 which takes the value from `jupyterhub.podCullerSelector`. This should be set to `component=singleuser-server`.

consideRatio:
@betatim @minrk Sorry for causing this, this closes #751.

the helper `jupyterhub.matchLabels` will return labels based on the files parent directory (`component: pod-culler`), but that can be overridden if one passes the helper a scope containing a `.componentLabel`.

Note that the output will be:
`component=singleuser-server,app=jupyterhub,release=prod`
Instead of:
`component=pod-culler,app=jupyterhub,release=prod`

Using the `app` and `release` as selectors as well is done systematically on all pods according to helm chart best practices.

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -126,9 +126,11 @@ component: {{ include "jupyterhub.componentLabel" . }}
 
 {{- /*
   jupyterhub.podCullerSelector:
-    Used to by the pod-culler to select singleuser-server pods. 
-    It simply reformats "jupyterhub.matchLabels".
+    Used to by the pod-culler to select singleuser-server pods. It simply
+    reformats "jupyterhub.matchLabels" and sets the componentLabel value so
+    `component=singleuser-server` is outputted.
 */}}
 {{- define "jupyterhub.podCullerSelector" -}}
-{{ include "jupyterhub.matchLabels" . | replace ": " "=" | replace "\n" "," | quote }}
+{{- $_ := merge (dict "componentLabel" "singleuser-server") . -}}
+{{ include "jupyterhub.matchLabels" $_ | replace ": " "=" | replace "\n" "," | quote }}
 {{- end }}
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -7,36 +7,39 @@
   ## How helpers work
   Helm helper functions is a good way to avoid repeating something. They will
   generate some output based on one single dictionary of input that we call the
-  helpers scope.
+  helpers scope. When you are in helm, you access your current scope with a
+  single a single punctuation (.).
+  
+  When you ask a helper to render its content, one often forward the current
+  scope to the helper in order to allow it to access .Release.Name,
+  .Values.rbac.enabled and similar values.
 
-  Often the current scope (.) is passed to the helper as it give the helper
-  access to the current scope containing .Release.Name, .Values.rbac.enabled and
-  similar values.
-
+  #### Example - Passing the current scope
   {{ include "jupyterhub.commonLabels" . }}
 
-  Sometimes the helper function is passed something specific instead of the
-  current scope (.), that would make .Release.Name etc. inaccessible.
+  It would be possible to pass something specific instead of the current scope
+  (.), but that would make .Release.Name etc. inaccessible by the helper which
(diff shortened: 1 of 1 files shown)


jupyterhub/templates/_helpers.tpl
```diff
@@ -4,6 +4,41 @@
...
+  ## How helpers work
+  Helm helper functions is a good way to avoid repeating something. They will
+  generate some output based on one single dictionary of input that we call the
+  helpers scope.
+
+  Often the current scope (.) is passed to the helper as it give the helper
+  access to the current scope containing .Release.Name, .Values.rbac.enabled and
+  similar values.
+
+  {{ include "jupyterhub.commonLabels" . }}
```
betatim: This is an example of calling the helper called `jupyterhub.commonLabels` with scope `.`?

jupyterhub/templates/_helpers.tpl
```diff
@@ -4,6 +4,41 @@
...
+
+  Sometimes the helper function is passed something specific instead of the
+  current scope (.), that would make .Release.Name etc. inaccessible.
+
+  {{ include "demo.bananaPancakes" (dict "pancakes" 5 "bananas" 3) }}
+
+  To let a helper acces the current scope along with additional values we can
```
betatim: acces -> access

jupyterhub/templates/_helpers.tpl
```diff
@@ -4,6 +4,41 @@
...
+      matchLabels:
+        {{- $_ := merge (dict "appLabel" "kube-lego") . }}
+        {{- include "jupyterhub.matchLabels" $_ | nindent 6 }}
+
+  In this way, the code within the definition of `jupyterhub.matchLabels` will
+  be able to access .Release.Name and .appLabel.
+
+  NOTE:
+    The ordering of merge is crucial, the latter argument is copied and merged
+    into the former. So if you would swap the order you would influence the
```
betatim: A different way of saying this would be that an argument to the right overwrites an argument to the left of it? In this case if appLabel was also defined in the current scope we'd end up with the value from the current scope being available to the helper?

jupyterhub/templates/_helpers.tpl
```diff
@@ -126,9 +161,11 @@ component: {{ include "jupyterhub.componentLabel" . }}
 
 {{- /*
   jupyterhub.podCullerSelector:
-    Used to by the pod-culler to select singleuser-server pods. 
-    It simply reformats "jupyterhub.matchLabels".
+    Used to by the pod-culler to select singleuser-server pods. It simply
+    reformats "jupyterhub.matchLabels" and sets the componentLabel value so
+    `component=singleuser-server` is outputted.
```
betatim: Used to by the -> Used by the
is outputted -> is output

betatim:
Looks good to me. There are a few typos and some questions for my education.

When I read the comments before about scope I had assumed it was talking about things set in the template where the helper was being called. Somehow that confused me :)

jupyterhub/templates/_helpers.tpl
consideRatio: Yes I think so, but this will never be an issue I think, because I don't know how to add things to the current scope without doing it the other way around:

```
{{- $_ := merge . (dict "addedKey" "value1") }}
```

If one later execute the code below, I think you are correct in that the right dictionary (.) would override the left and make `addedKey` have `"value1"` instead of `"value2"`.

```
{{- $_ := merge (dict "addedKey" "value2") . }}
```

But doing...
```
{{- $addedKey := "something" }}
```
Will not be accessible through `.addedKey` but instead `$addedKey`.

consideRatio:
@betatim thank you for helping me process this! I really appreciate it!

betatim:
🎉 1

PR #758, opened 2018-07-10
[TO BE SPLIT UP] User scheduler, user placeholders, node pool taints

consideRatio:
# NOTE
This PR will be split up into pieces, see #841 

# This PR
While a cluster autoscaler which adds and removes nodes can improve cloud resource efficiency, some issues using it with this chart exists. This PR makes cluster autoscaling more robust and efficient to use.

## PR Status
- Writing Documentation
- Considering how to get it merged, breaking apart the PR into various pieces is at least partially plausible, but a lot of the things is entangled.

## Fixed and alleviated issues
- Scaling down is inefficient as new pods spread out on the nodes
  - Solved by having a custom scheduler schedule the users to pack tight instead of spread out
- Scaling down nodes often fails because a system pod like `kube-dns` or a chart pod like `hub` or `proxy` has been scheduled on them instead of staying together on one node
  - Solved by *tainting* dedicated user nodes and setting a *toleration* on the user pods that allows them to still schedule on the tainted nodes. See [the kubernetes documentation about taints and tolerations](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) and [a google cloud example on how to setup a tainted node pool](https://cloud.google.com/kubernetes-engine/docs/how-to/node-taints).
- Scaling up just in time forces users to wait several minutes
  - Alleviated by JupyterHub 0.9 allowing kubespawner to inform the waiting users on the spawning progress that contains messages about a triggered scale up.
  - Solved (requires k8s 1.11) with `user-placeholder` pods. You can configure that you want for example two placeholder pods, and they will trigger node scale up if they can't schedule just like a user, but when a user is arriving without available room for it to schedule the user-placeholder pods will make room for the real user. See [the kubernetes documentation on pod priority and preemption](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) to learn about the details.
- It is troublesome to test a cluster autoscaler setup
  - Alleviated by the introduction of `user-dummy` pods which mock real users but is easier to add / remove for testing purposes.

## Feature availability
K8s 1.11 is required for pod preemption and pod priority. [K8s 1.11 is at the moment I'm writing this not yet available on GKE](https://cloud.google.com/kubernetes-engine/release-notes), and Helm lacks support for k8s 1.11 and we need to await Helm 2.11. We can [help them out by by testing k8s 1.11 support](https://github.com/helm/helm/pull/4420).

# Test this PR
[...]

PR #759, opened 2018-07-10
CI bump: k8s 1.9.0 / helm 2.9.1 / minikube 0.28.0

consideRatio:
Bumping CI to use latest k8s/helm/minikube.

Currently #758 CI is failing to start because Helm 2.9.1 is required.

- [Minikube changelog](https://github.com/kubernetes/minikube/blob/v0.28.0/CHANGELOG.md)

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -35,7 +35,7 @@ done
 kubectl get nodes
 
 echo "installing helm"
-curl -ssL https://storage.googleapis.com/kubernetes-helm/helm-v2.7.2-linux-amd64.tar.gz \
+curl -ssL https://storage.googleapis.com/kubernetes-helm/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
   | tar -xz -C bin --strip-components 1 linux-amd64/helm
 chmod +x bin/helm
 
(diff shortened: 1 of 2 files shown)


clkao:
Is it time to create travis env matrix for multiple versions of k8s?

consideRatio:
@clkao perhaps but I suspect it is overkill because I run with the assumption that kubernetes api-server will behave similarly assuming you using an older API.

The solution to the build error seems to be to upgrade the build system to run on Ubuntu 16.04, but travis doesn't support that and after reading posts I've given up all hope on that is going to happen in a reasonable time.

I noticed that the jupyterhub/jupyterhub repo used CircleCI, something we can use for z2jh as well? I'm clueless about if you are paying for it, or how it would work etc. Anyone with clues about this? @clkao @minrk ?

consideRatio:
How should we go about making minikube execute on Ubuntu 16.04 instead of 14.04 where `systemd` is available and the issues we face above will go away?

consideRatio:
@clkao it could be useful to test 1.9 as well as 1.11, because helm could be smart about certain features by default if we are on k8s 1.11 but shut them down if we are below.

Do you have experience on creating such travis environment matrix?

clkao:
FTR matrix ci is in #762

PR #762, opened 2018-07-12
test on different versions of k8s

clkao:


betatim:
Looks like a good addition to me.

Do we need to special case one of the version of helm and only run chartpress/deploy for that version instead of all of them?

clkao:
@betatim re helm versions: i think it's not much of a blocker for people to upgrade if they run into issues, unlike upgrading k8s. so it should reasonable for us to just smoke with the latest stable helm release.

betatim:
I made a typo: meant kube versions :-(

What I meant was: with this setup we run three jobs instead of just one. My question was if this means we need to pick one of the versions and only for that build the docker images and publish the helm chart. Or is the deploy step of travis already setup so that it will only run once.

clkao:
@betatim right, the deploy stage seems to be separate from the tests expanded with env matrix: https://docs.travis-ci.com/user/build-stages/matrix-expansion/

If we are not using helm/minikube in deploy.sh, it should be safe to just leave it as is.  it is unclear to me in the travis doc how env matrix interacts with deploy script.

consideRatio:
@clkao @betatim thanks for the PR and discussion - I keep learning a lot and I love it!

consideRatio:
## About Travis build failure
I made a small adjustment before about to merge, got a build error, corrected it with another commit and saw the builds succeed. Then I decided to squash those two commits into one to avoid clutter. So I added @clkao's repo as a remote, squashed my two commits into one, and force updated @clkao's PR branch and merged.

Travis became upset though with...
`The command "eval git fetch origin +refs/pull/762/merge: " failed.`

betatim:
I think the advice is: don't do that :) The advanced features of `git` are powerful and in my experience almost always when you try to be clever and use them you end up creating more trouble for yourself. At least that is my personal learning from years of git'ing.

[The build](https://travis-ci.org/jupyterhub/zero-to-jupyterhub-k8s/builds/407904199) of master after the merge does not run the deploy stage  so I would propose we revert this PR as we have to have the automatic deploys working.

PR #763, opened 2018-07-12
Hub default deployment strategy should be Recreate

### LINKED ISSUE #754: pdb and hub upgrade failures
It seems `helm upgrade` no longer works for latest z2jh default install, as the default sqlite-pvc is mounted by the old hub rs during upgrades, so the new pod can not mount the pv.

### LINKED ISSUE #757: Hub persists in 'ContainerCreating' state
I am running the the jupyterhub version-`v0.7-4c22c3d` on GKE with tiller version `2.9.1` and GKE `1.9.7-gke.3`. I can run helm install but the hub persists in `ContainerCreating` state.

clkao:
Since the default storage backend is sqlite on pvc, we should be using the Recreate strategy as we bumped the deployment to apps/v1beta2. Otherwise upgrading will have new hub pod stuck in ContainerCreating state.

This is a regression caused by a84ad294. Detailed analysis can be found in #754.

Closes #754
Closes #757

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -9,6 +9,8 @@ spec:
   selector:
     matchLabels:
       {{- include "jupyterhub.matchLabels" . | nindent 6 }}
+  strategy:
+    {{- .Values.hub.deploymentStrategy | toYaml | trimSuffix "\n" | nindent 4 }}
   template:
     metadata:
       labels:
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -11,6 +11,11 @@ hub:
   nodeSelector: {}
   concurrentSpawnLimit: 64
   activeServerLimit:
+  deploymentStrategy:
+    # sqlite-pvc backed hub requires Recreate strategy to work
+    type: Recreate
+    # This is required for upgrading to work
+    rollingUpdate: null
   db:
     type: sqlite-pvc
     upgrade:


betatim:
Is there a test we can add to demonstrate that this works or add a test that would catch it if someone changes this away from "recreate" in a future PR?

clkao:
It seems the current tests are quite limited - only a single sh file ;) I can add a very basic `helm upgrade` and verify there is only one hub pod running after certain timeout.

for the long term e2e tests, is there preferred test library for z2jh?

clkao:
on a second thought, perhaps we should first test upgradability from previous major release.

betatim:
How to get started with having better tests and especially how to make step 0 which is setting up a lot of the infrastructure is hard. On the one hand I can see the argument "so far there are no good tests why block this PR on making the effort" on the other hand at some point we will have to do this otherwise we never get to a better place :-/ Maybe building some of the infrastructure and doing a little bit more in terms of testing is a compromise, then we can spread the load over several PRs.

pytest is used a lot in JupyterHub projects so I'd go with that.

betatim:
It seems like adding more tests before merging this PR is going to cause a lot of delay. Should we create a new PR for tests and merge this as is?

Or do you want to make a new PR using `StatefulSet` @consideRatio?

betatim:
Merged this as is as it means we don't need to manually intervene when deploying changes to mybinder.org. Let's make a new PR that fixes this for the long term.

PR #768, opened 2018-07-18
Make sure image puller works with long repo names

clkao:
Currently this fails since container should be alphanum.

```
prePuller:
  extraImages:
    cpu:
      name: custom.registry-server/repo/image
      tag: latest
```


consideRatio:
Excellent solution and catch @clkao!! :heart:

PR #770, opened 2018-07-18
Update authentication.rst

prkriz:
Overcoming [Caveats while setting up CILogon OAuthAuthenticator for JupyterHub](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/cilogon.py)

### DIFF
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -116,6 +116,18 @@ CILogon
           clientSecret: "an0ther1ongs3cretstr1ng"
           callbackUrl: "http://<your_jupyterhub_host>/hub/oauth_callback"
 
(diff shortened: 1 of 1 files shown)


choldgraf:
Thanks very much for the docs PR! I agree this is often a point of confusion for some people. I wonder if it'd be better to just link out to the oauthenticator docs where this issue is addressed (and if it isn't, we should add it to the docs there).

@minrk do you know if this issue is mentioned in oauthenticator? Currently it seems to exist only in the docstring of that module...

minrk:
It is mentioned [in oauthenticator](https://github.com/jupyterhub/oauthenticator/blob/75d8b9b5ee366e7094d28a4c569986f04e3ea802/oauthenticator/cilogon.py#L10), but could be more prominent. Thanks for the PR!

consideRatio:
Thank you @prkriz !!! :D :heart:

prkriz:
Happy to contribute!! :)

choldgraf:
wooo thanks!

PR #775, opened 2018-07-20
Update version of the helm chart to current release

### LINKED ISSUE #742: Default documentation version does not match current release
https://zero-to-jupyterhub.readthedocs.io/ defaults to the latest version (master). This is confusing when it doesn't match the latest ~jupyterhub~ zero-to-jupyterhub release, currently v0.6. Examples:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/722
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/741

rnestler:


betatim:
LGTM. @choldgraf or @jzf2101 when is the PEARC tutorial? Purely for my paranoia we should try and avoid making last minute changes to the docs that will throw up new kinds of things that people struggle with during a tutorial.

Otherwise: merge it!

jzf2101:
PEARC is Monday

consideRatio:
this is documentation only in the master/latest docs right?

the master and to be 0.7 helm chart release uses JH 0.9.*, while the 0.6 helm chart uses 0.8.1.

we have not kept this documentation up to date.

jzf2101:
I think @betatim and @choldgraf have checked that the current documentation works.  However, I think they said both the versions in the docs and the latest release have helm chart issues?

betatim:
>  However, I think they said both the versions in the docs and the latest release have helm chart issues?

On the JupyterHub deployments I administer we switched to using `master` of the helm chart which seems to work well. The choice was driven by #757 which was annoying for continuous deployments. However I don't follow the z2jh guide when setting up new deployments so I don't know if the docs work or not.

This PR just moves the docs to the latest released chart version. This should just work and be a safe move, I just wanted to double check with the PEAR tutorial people that this wouldn't catch them unawares. If you don't have time pre-tutorial to test this out I'd recommend we merge this as soon as your tutorial is over.

consideRatio:
LGTM! Note that this is just a documentation update that we should had done with the 0.6 release.

PR #777, opened 2018-07-22
allow image puller policy override

clkao:


consideRatio:
Thank you @clkao :heart: - lgtm!

PR #778, opened 2018-07-22
ignore bin/ downloaded by ci/install.sh

clkao:


consideRatio:
Excellent! Thanks for all the work you put in @clkao :heart: !
I'm closing in favor of the same change implemented in #758 where I also add the `kubeval` binary.

clkao:
@consideRatio oh! if the branch is still WIP and could take some time before it is reviewed, perhaps we can just make a change here to ignore ci/bin and get this merged first.

consideRatio:
Figured I would save myself a rebase but I agree with your point!!

PR #789, opened 2018-07-25
Change travis setup to only deploy once

betatim:
Follow up #762 

Since adding a build matrix testing several kubernetes versions we are
deploying twice, once for each kube version. This changes our deploy
setup so it only deploys once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ deploy:
   skip_cleanup: true
   script: ./ci/deploy.sh
   on:
+    env: KUBE_VERSION=1.10.0
     branch:
     - master
     - v0.7-dev


consideRatio:
@betatim Thank you sorry for causing some headache!

From [this travis documentation](https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A) it seems like we could solve this if we do...

```yaml
deploy:
  provider: script
  skip_cleanup: true
  script: ./ci/deploy.sh
  on:
    branch:
      - master
      - v0.7-dev
    condition: "$KUBE_VERSION = 1.9.4"
```

consideRatio:
Yikes... Perhaps I can blame the jetlag I'm having, I thought you simply reduced the env-matrix to one version but you are attempting to do the same thing i suggested but in a much nicer way.

Let's try it!

PR #792, opened 2018-07-25
Travis: Please deploy once and only once

consideRatio:


consideRatio:
Got it working @betatim !

For one of the builds...
image

For the other...
image

betatim:
Nice!

consideRatio:
@betatim are you sure? From discussion in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/789#issuecomment-407675267 I drew the conclusion that it did not work, thats why I made this PR.

betatim:
Not sure, though it would only work if we used `jobs`, not the setup we currently use. So this is more about "there seem to be 10 different ways to achieve same outcome" than anything else :)

PR #793, opened 2018-07-25
Support extra resources for single user

### LINKED ISSUE #795: Should allow spawning single-user container with extra resources like nvidia.com/gpu
In our case, the GPU is a necessary resource, but the current version doesn't support requesting GPU or other non-built-in resources.

I made a patch to solve this issue, see details in my pr: #793, it was tested and running on our production environment now.

zacharyzhao:
Add extra resource support for k8s-hub.

closes #795


minrk:
Thanks!

consideRatio:
Thank you @zacharyzhao ! :D

PR #794, opened 2018-07-25
hub: correctly delete named servers

rokroskar:
This PR handles the deletion of named servers in the culling script. The change also affects the main jupyterhub repo, e.g. https://github.com/jupyterhub/jupyterhub/blob/master/examples/cull-idle/cull_idle_servers.py#L190

### DIFF
--- a/images/hub/cull_idle_servers.py
+++ b/images/hub/cull_idle_servers.py
@@ -168,10 +168,16 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
                 log_name, format_td(age), format_td(inactive))
             return False
 
+        if server_name:
+            # culling a named server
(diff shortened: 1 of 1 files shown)


betatim:
Nice. Can you reset the two files from the helm-chart that got committed?

We once discussed where to keep "the" version of the culler script because there are several copies of it spreading (and getting out of sync). Does someone remember what the decision was?

rokroskar:
@betatim oops sorry about that. Fixed.

consideRatio:
Woohoooo!! Thank you @rokroskar for the PR and @betatim and @minrk for reviewing it!

minrk:
@betatim I think the plan is to make it a standalone package, but for now keeping the two in sync by hand.

Issue #675, opened 2018-05-03
Support SSL offloading to AWS ELB


pcfens:
Rather than terminating SSL in the application, we've traditionally terminated on the load balancer in AWS.

With one small change (and some perhaps creative use of the config file fed to helm), we're able to make this work. I'm happy to submit a PR, but I'm not sure how the maintainers would like to structure it, so I'm starting here.

Our sanitized config.yaml file looks like
```yaml
---
hub:
  image:
    name: wmit/jupyterhub-k8s-hub-casauth
    tag: v0.6
  extraConfig: |
    import jhub_cas_authenticator.cas_auth
proxy:
  secretToken: REDACTED
  https:
    enabled: true
    type: ACM
    hosts: jupyterhub.example.edu
  service:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:us-east-1:1234567891011:certificate/uuid"
      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "tcp"
      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "https"
      service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '3600'
auth:
  type: custom
  custom:
    className: jhub_cas_authenticator.cas_auth.CASAuthenticator
    config:
      cas_login_url: https://cas.example.edu/cas/login
      cas_service_url: https://jupyterhub.example.edu/login
      cas_service_validate_url: https://cas.example.edu/cas/p3/serviceValidate # Requires CAS 5
  admin:
    users:
      - user1
```

Helm creates a service with an exposed HTTPS port, while the additional annotations ensure that it's a TCP listener (so that websockets continue to work). The annotations also attach an SSL certificate that's managed outside of k8s.

To actually make things work, I change the HTTPS target port in the proxy-public service to match the HTTP target port (`kubectl --namespace=kube-system edit svc proxy-public`).

Thoughts on how we might add this to the existing chart?

PR #796, opened 2018-07-25
Support offloading SSL to the proxy service

pcfens:
Rather than always having different target ports for the HTTP and HTTPS frontends, setting proxy.https.type to offload will tell the service to route ports 443 and 80 both to port 80 on the backend.

For this to work, annotations need to be added to proxy.service.annotations to attach certificates to the load balancer.

### DIFF
--- a/doc/source/security.md
+++ b/doc/source/security.md
@@ -77,6 +77,36 @@ If you have your own HTTPS certificates & want to use those instead of the autom
 2. Apply the config changes by running helm upgrade ....
 3. Wait for about a minute, now your hub should be HTTPS enabled!
 
+### Off-loading SSL to a Load Balancer
+
+In some environments with a trusted network, you may want to terminate SSL at a
+load balancer. If https is enabled, and `proxy.https.type` is set to `offload`,
+the HTTP and HTTPS front ends target the HTTP port from JupyterHub.
+
+The HTTPS listener on the load balancer will need to be configured based on the
+provider. If you're using AWS and a certificate provided by their certificate
+manager, your config.yml might look something like:
+
+```yaml
+proxy:
+  https:
+    enabled: true
+    type: offload
+  service:
+    annotations:
+      # Certificate ARN
+      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:us-east-1:1234567891011:certificate/uuid"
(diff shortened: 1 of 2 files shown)


betatim:
Is this a common enough use case that we want to take on supporting it?

If yes, can you add some documentation in the HTTPS section of the guide on when and how you'd use this feature?

pcfens:
I don't know how widely it'd be used, but I know it's a common use case for us, especially when testing (the Let's Encrypt duplicate certificate rate limit keeps us at 5 iterations per week otherwise).

The approach I took here is designed to have as small an impact as possible, which should also keep maintenance to a minimum going forward.

I've updated the documentation in the latest push too. The original issue (#675) has a complete example if it helps figure out next steps.

minrk:
Thanks! This looks good to me now. Well documented.

consideRatio:
@pcfens @betatim @minrk - I have never understood why `proxy.https.type: offload` was helpful, but at this point in time I'm blocked in another PR (#1811) without understanding this better.

I look at the provided example and wonder what would happen if...

```diff
 proxy:
   https:
-    enabled: true
-    type: offload
+    enabled: false
   service:
     annotations:
       # Certificate ARN
       service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:us-east-1:1234567891011:certificate/uuid"
       # The protocol to use on the backend, we use TCP since we're using websockets
       service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "tcp"
       # Which ports should use SSL
-      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "https"
+      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "http"
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '3600'
```

Would that work? I really want to figure out if we can make use of this AWS LB without having a dummy 443 port alongside a 80 port.

pcfens:
Hi @consideRatio-

In the example you provide a load balancer would be created, but it would only listen on port 80, but the annotation tells AWS to attach a certificate to it anyway (https on port 80 I believe, unless Amazon has some logic on their end). The original intent of this PR was to support offloading SSL termination with the smallest amount of disruption to the how things previously worked. I'll comment on the other PR with some information that might be helpful there.

PR #799, opened 2018-07-26
Clean-up tips section

### LINKED ISSUE #798: Code blocks in the "Tips and command snippets" section don't render properly
The code blocks in the "Tips and command snippets" section render as indented, but otherwise normal, text; not as code. See https://zero-to-jupyterhub.readthedocs.io/en/latest/tips.html
It looks like this is because of missing double colons `::`, or a `.. code-block:: bash` directive as used elsewhere.

(The potentially most problematic issue is that `--` for long options renders as an n-dash, instead of two dashes, making it not possible to copy-paste the occasional line.)

Funnily enough, when viewed on GitHub directly, it renders better: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/tips.rst . I guess GitHub's ReST parser is quite lenient.

I have not found any other pages that have this problem (but I haven't gone through all of them yet).

Since this is a relatively straightforward change, I'd be happy to put in a PR for this.

evertrol:
- Provide a bash code-block directive for code blocks.
- Remove single underscores surrounding text: these were presumably
  meant for emphasis (as in Markdown), but incorrect for ReST and not
  necessary.
- Apply :kbd: role for TAB mention
- Remove trailing whitespace

Fixes #798

### DIFF
--- a/doc/source/tips.rst
+++ b/doc/source/tips.rst
@@ -11,14 +11,14 @@ to be added to this page, please feel free to make a PR!
 
 Kubernetes has a helper script that allows you to auto-complete commands
 and references to objects when using ``kubectl``. This lets you
-``<TAB>``-complete and saves a lot of time.
+:kbd:`TAB`-complete and saves a lot of time.
 
 `Here are the instructions to install kubectl auto-completion <https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion>`_.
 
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me! Thanks for taking the time to make our docs better!

consideRatio:
Thank you @evertrol !!!! :heart:

PR #813, opened 2018-08-02
Add missing local development setup steps to CONTRIBUTING.md

cmd-ntrf:
Helm installation was missing, now step 2, and the virtual environment was not activated (now step 6.2).


choldgraf:
very helpful! I made some tiny edits but looks great to me! thanks!

PR #846, opened 2018-08-15
[WIP] CI Tests

consideRatio:
## Whats in the PR
- [x] Verify that a singleuser server can spawn to a running state
- [x] Single command (`ci/run`) to run integration tests *both locally and on travis*
  - [x] Downloads binaries in-repo
  - [x] Lints and verifies Helm templates
  - [x] Lints and verifies the rendered output against k8s API schema
  - [x] Setups a minikube cluster using virtualhost
  - [x] Uses chartpress to build chart and images
  - [x] Installs chart on cluster and runs tests
- [x] Single command (`ci/cleanup`) to remove all in-repo binaries and configuration folders such as `.kube/`, `.minikube/` and `.helm/`.

## TODOs
- [ ] Tackle documented FIXMEs
- [ ] Make sure that we can still use the deploy script
- [ ] Create checks for `--rm` and that our $PWD is in the repo
- [ ] Name files that we should `source` `.env` and don't set shell options within them
- [ ] Use `helm version` to really make sure tiller is up and running before we go
- [ ] Rebase and make the commits easier to follow


jupyterhub/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: jupyterhub
-version: v0.7-dev
+version: v0.7-2940c96a
```
betatim: Is this a side effect of running `chartpress` and should be undo? Not sure.

jupyterhub/values.yaml
```diff
@@ -41,7 +41,7 @@ hub:
   extraVolumeMounts: []
   image:
     name: jupyterhub/k8s-hub
-    tag: generated-by-chartpress
+    tag: 'ac54cf18'
```
betatim: Undo these changes that come from running `chartpress`

betatim:
Another huge PR! Both in number of lines changed but also in terms of new functionality :) Also all those nice comments!

I had a quick read through and right now think this must make things better than they were.

Because it moves around a lot of code I think we won't be able to find all the parts where it is flakey without merging it and using it. Then fixing as we go along. We should get at least one more pair of eyes to look at it and vote 👍  or 👎 though.

.travis.yml
```diff
@@ -7,18 +7,13 @@ services:
 - docker
 install:
 - pip install --no-cache-dir -r dev-requirements.txt
-- . ci/minikube.env
-- ./ci/install.sh
```
betatim: What was your thinking behind moving everything into `script` instead of running the install/setup things here? Might be worth writing a comment for future generations.

ci/0-setup-env
```diff
@@ -0,0 +1,41 @@
...
+
+# 4.
+export KUBE_VERSION=1.10.0
+export MINIKUBE_VERSION=0.28.2
+export HELM_VERSION=2.9.1
+export KUBEVAL_VERSION=0.7.1
+
+# Since this script will be sourced, revert to previous shell options
+# FIXME: we are just guessing what the previous options were now
+set +eux
```
betatim: I'd run this script without changing to `set -eux` at the top so that we don't have to guess when we revert things at the bottom here. I'd prefer that trade-off to guessing.

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+  echo "waiting for tiller"
+  kubectl --namespace=kube-system rollout status --watch=true deployment/tiller-deploy
+  
+  # FIXME: It would be better to have the "waiting for tiller" at a later stage
+  # where we need to utilize tiller, like in the test that will install the
+  # chart. I think we need to wait for till in another way than rollout status,
+  # because apparently I still end up tiller being unavailable when I run the
+  # helm upgrade command after having waited.
+fi
+set -e
```
betatim: why this at the end?

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+  minikube config set vm-driver none
+  minikube config set WantNoneDriverWarning false
+  export CHANGE_MINIKUBE_NONE_USER=true
+
+  minikube addons disable dashboard
+  SUDO='sudo -E'
+else
+  # LOCAL
+  if (timeout 3 bash -c 'kubectl get pod -n kube-system --selector app=helm,name=tiller | grep 1/1'); then
+    echo "minikube cluster was already setup"
```
betatim: I'm confused by this. The message makes me think that this bit of code had tried to setup the minikube cluster but then decided it didn't have to. However there is no code here that sets up a cluster.

Just a language thing? "minikube cluster configured and up and running"?

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+  # WARNING: kubeadm requires systemd. systemd isn't available for Ubuntu 14.04
+  # (trusty) VMs as used by Travis and CircleCI. It is available on Ubuntu 16.04
+  # (xenial) or 18.04 (bionic) VMs though.
+    $(echo $SUDO) $(which minikube) start
+fi
+
+# IF LOCAL ENVIRONMENT
+if set +u && [[ -z "${CICD}" ]] && set -u; then
+  echo "update minikube context"
+  # minikube update-context
```
betatim: Why not run this? If we want to tell people to run this locally (I think we should!) then we need to make it very hard to accidentally run the following `kubectl` commands on anything but a minikube cluster (or some other cluster the user specifically configured. The number of times I've had my kube context set incorrectly and nearly deployed/wiped something from the wrong cluster is too big for me to publicly say.

So I think we should have some simple checks that protect the tired, frustrated and/or stressed user who is running this to debug stuff from accidentally running these on anything but a minikube.

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+fi;
+
+echo "waiting for kubernetes"
+JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'
+until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do
+  sleep 1
+done
+kubectl get nodes
+
+set +e # Allow resources to pre-exist by tolerating errors a while
```
betatim: "errors a while" -> "errors for a while"

ci/README.md
```diff
@@ -0,0 +1,35 @@
...
+# yet setup.
+
+
+# Clean up of the in-repo folders .kube/, .minikube/ and .helm/ and deletion of
+# the minikube cluster.
+ci/cleanup
+```
+
+## About the vagrant files
+They are currently not up to date and were a previous way to do some CI.
```
betatim: Let's delete the unused files.

ci/cleanup
```diff
@@ -0,0 +1,22 @@
...
+# -e : Exit immediately if a command exits with a non-zero status.
+# -u : Treat unset variables as an error when substituting.
+# -x : Print commands and their arguments as they are executed.
+set -eux
+
+# 1.
+minikube delete
+
+# 2.
+rm -rf .kube/ .minikube/ .helm/
```
betatim: we need to make sure we are absolutely in the right directory before we run this. Maybe test that `.kube` etc exist and that some other directories and files we know should be in the repo at a certain place are exactly where we think they are so this script can be sure it is not about to `rm -rf` something the user did want to keep :)

tests/test_spawn.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Makes heavy use of JupyterHub's API:
+# http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/docs/rest-api.yml
+
+def test_api(request_data):
+    print("asking for the hub's version")
+    r = requests.get(request_data['hub_url'])
+    r.raise_for_status()
+    print(r.json())
+    return r.json()
```
betatim: why return something from the test function?

tests/test_spawn.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_api_info(request_data):
+    print("asking for the hub information")
+    r = requests.get(request_data['hub_url'] + '/info', headers=request_data['headers'])
+    r.raise_for_status()
+    print(r.json())
+    return r.json()
+
+def test_api_create_user(request_data):
```
betatim: maybe move all the `test_api_...` functions into a new `test_api.py` file and rename them `test_create_user` instead of `test_api_create_user`?

jupyterhub/Chart.yaml
consideRatio: Yepp it is! I included the changes by mistake to a commit!

ci/0-setup-env
consideRatio: Ah excellent input!

Is using .env a common way to indicate that? I figure if it contains environment variable export using it in any other way doesn't make sense!

ci/3-setup-k8s
consideRatio: Ah lets make the language change!

ci/3-setup-k8s
consideRatio: Yes I agree!
Let's add some fail safes, it will be easy because I've already tried to make this a sandbox.

Sandbox steps:
- [x] Use $KUBECONFIG, $HELM_HOME, and $MINIKUBE_HOME with in-repo .kube/ .minikube/ .helm/
- [ ] Run checks that `which kubectl` `which minikube` and `which helm` points to the binary in repo.

ci/4-setup-cluster-env
```diff
@@ -0,0 +1,29 @@
...
+# Set shell options
+# -e : Exit immediately if a command exits with a non-zero status.
+# -u : Treat unset variables as an error when substituting.
+# -x : Print commands and their arguments as they are executed.
+set -ux
+
+# 1.
+if set +u && [[ -z "${CICD}" ]] && set -u; then
+  # LOCAL
+  eval $(minikube docker-env)
```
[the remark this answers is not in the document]
consideRatio: I don't know if it will hurt to run on Travis, but I don't think we need to do it as there is a difference between running locally and on Travis. Locally we use a VM while setting up minikube: `--vm-driver=virtualbox` while on Travis we use: `--vm-driver=none`.

ci/6-run-tests
```diff
@@ -0,0 +1,45 @@
...
+  --namespace jh \
+  --values ci/6-tests-config.yaml \
+  ./jupyterhub/
+
+echo "waiting for servers to become responsive"
+until curl --fail --silent --show-error $HUB_API_URL; do
+  # FIXME:
+  # At least Travis can get stuck like this...
+  # +curl --fail --silent --show-error http://10.20.0.27:31212/hub/api
+  # curl: (52) Empty reply from server
```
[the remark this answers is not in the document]
consideRatio: The curl command can get stuck taking way to log to finish. This causes the inside of the loop executing way to seldom. That isn't great as the goal of the loop the way I see it, is to give some feedback of whats happening while we wait.

ci/README.md
consideRatio: Deal

tests/test_spawn.py
consideRatio: Good point :P

tests/test_spawn.py
consideRatio: Sure! It is the first time i setup to use pytest myself and have lots to learn.

There is a lot of state relating to these tests. Should I ensure all tests will return the cluster to the original state? If i do so I'll need to make very few and very large tests though.

If i don't ensure state is restored in between tests, I need to keep track of the execution order. I learned that the test functions will be called in the order I've declared them within the file, but what about having multiple files, in what order will they execute? Will google this soon

consideRatio:
@betatim Thank you for reviewing this so and giving me all this great feedback!

I'm working hard on reducing the PR sizes, I have realized a lot about the challenges larges PRs can do, I will do better!

ci/1-setup-binaries
```diff
@@ -0,0 +1,71 @@
...
+  echo "installing kubeval"
+  curl -sSLo bin/kubeval-${KUBEVAL_VERSION}.tar.gz https://github.com/garethr/kubeval/releases/download/${KUBEVAL_VERSION}/kubeval-linux-amd64.tar.gz
+  tar --extract --file bin/kubeval-${KUBEVAL_VERSION}.tar.gz --directory bin
+  rm bin/kubeval-${KUBEVAL_VERSION}.tar.gz
+  mv bin/kubeval bin/kubeval-${KUBEVAL_VERSION}
+fi
+cp bin/kubeval-${KUBEVAL_VERSION} bin/kubeval
+
+# IF LOCAL ENVIRONMENT
+if set +u && [[ -z "${CICD}" ]] && set -u; then
```
minrk: Instead of unsetting and resetting `-u`, this can be `${CICD:-}`, which explicitly defaults to an empty string if undefined.

ci/1-setup-binaries
```diff
@@ -0,0 +1,71 @@
...
+  echo "setting up kubectl and minikube autocomplete"
+  source <(kubectl completion bash)
+  source <(minikube completion bash)
```
minrk: We also probably shouldn't be setting up autocompletion in a script for CI. That's up to users' own environments, so let's remove this block.

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+set -eux
+
+# Setup Kubernetes
+# ----------------
+
+# Don't use the default profile: this helps to make it clear whats what while
+# inspecting virtualbox's VMs.
+minikube profile z2jh
+
+if set +u && [[ ! -z "${CICD}" ]] && set -u; then
```
minrk: Same `${CICD:-}` here

ci/3-setup-k8s
```diff
@@ -0,0 +1,100 @@
...
+minikube config set ShowBootstrapperDeprecationNotification false
+
+if [[ ${KUBE_VERSION} == 1.8* ]] || [[ ${KUBE_VERSION} == 1.9* ]] || [[ ${KUBE_VERSION} == 1.10.0* ]]; then
+  # WARNING: localkube not supported beyond Kubernetes 1.10.0
+    $(echo $SUDO) $(which minikube) start --bootstrapper=localkube --extra-config=apiserver.Authorization.Mode=RBAC
+else
+  # WARNING: kubeadm requires systemd. systemd isn't available for Ubuntu 14.04
+  # (trusty) VMs as used by Travis and CircleCI. It is available on Ubuntu 16.04
+  # (xenial) or 18.04 (bionic) VMs though.
+    $(echo $SUDO) $(which minikube) start
```
minrk: This can just be $SUDO

minrk:
I'm not sure about the design of the same sequence of scripts that do one thing when CICD=true and a different thing when CICD is undefined. Instead, I think it makes more sense to have dedicated scripts for doing the steps and have the run script for travis that does install & run.

For instance, I wouldn't expect or intend home users to run the minikube/helm-setup scripts ever, because they will likely do a bunch of things that users don't want (they certainly do things that I wouldn't want on my system, like installing and setting up a duplicate minikube). Instead, we should target the tests to be runnable for home installs, starting from the "I have minikube and helm already, let's run the tests" point. If we want to include "setup minikube/helm" for home steps, they should be a dedicated, standalone script, I think, wholly independent of what we do on Travis.

consideRatio:
@minrk thanks for the review, I'm especially happy about the way to avoid setting/unsetting -u, it was so ugly.

> For instance, I wouldn't expect or intend home users to run the minikube/helm-setup scripts ever, because they will likely do a bunch of things that users don't want (they certainly do things that I wouldn't want on my system, like installing and setting up a duplicate minikube). Instead, we should target the tests to be runnable for home installs, starting from the "I have minikube and helm already, let's run the tests" point. If we want to include "setup minikube/helm" for home steps, they should be a dedicated, standalone script, I think, wholly independent of what we do on Travis.

These are steps that we could make optional or extract from the actual testing steps:
- [ ] Make the install of kubectl/minikube/helm binaries to the in-repo folder optional
- [ ] Make the in-repo configuration of kubectl/minikube/helm binaries to the in-repo folder optional
- [ ] Make the minikube start etc optional

I really like utilizing the in-repo configuration of kubectl/minikube/helm though, as it will avoid dangers of running kubectl commands on the wrong cluster.

jupyterhub/values.yaml
consideRatio: - [x] Resolved

ci/3-setup-k8s
consideRatio: It will not matter, unless we add something afterwards.

ci/3-setup-k8s
consideRatio: - [x] Resolved

ci/3-setup-k8s
consideRatio: I ran into issues with the minikube update-context command, I dont think it played well with a custom .kube/config location or that I configured the minikube's profile, which will influence things i don't fully know but I think one of these things are the name of the minikube VM created for example, to avoid clashes with other VMs

minikube update-context is described to...

```
	Use:   "update-context",
	Short: "Verify the IP address of the running cluster in kubeconfig.",
	Long: `Retrieves the IP address of the running cluster, checks it
			with IP in kubeconfig, and corrects kubeconfig if incorrect.`,
```

ci/3-setup-k8s
minrk: Rather than disabling `-e` here, catching actual errors where we know they can occur may be better, e.g.

    kubectl get pod || echo "no pod yet..."

consideRatio: @minrk thank you a bunch for introducing a lot of useful techniques!

ci/0-setup-env
betatim: Not sure if there is a standard way, I often see `.env` or `something.env` as files that contain environment variables and nothing else.

Naming the file would help with knowing that it should be sourced (and exports variables) without having to look inside the file.

ci/3-setup-k8s
betatim: Ah, because on L77 we do `set +e`?

ci/4-setup-cluster-env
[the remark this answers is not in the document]
betatim: Ok, if on travis minikube has access to the same dockerd that chartpress uses to build images then this isn't needed. Thanks for teaching me something about the inner workings of minikube :)

ci/3-setup-k8s
consideRatio: Ah! Yes, they drifted apart when I patched the kubernetes dashboard in between as well, a comment would be suitable now.

tests/test_spawn.py
betatim: Ok. I hadn't considered that the functions depend on each other. Extrapolating from myself I would say that test functions that depend on each other isn't something we should do as no one would expect that to be the case. (At least setup in a vanilla way like this)

There are helpers in pytest to setup/undo state for a test function and reuse that tooling across many test functions. Maybe something to tackle in a new PR though. That way others with a lot of pytest experience can also get involved. Let's see what others think.

betatim:
After reading MIn's comment above about separating CI/CD things from helping users setup things I think I agree with him. A few of my comments about having to make sure things don't accidentally delete things or what sets environment variables etc are motivated by me wanting to know what is potentially a dangerous script to run on my local machine. We could also remove a lot of if statements by having one set of scripts purely for travis and then a second set of instructions/scripts for humans.

My ideal state to reach would be that locally I once setup my minikube and then just run `pytest` from the repo. No arguments, no scripts needed.

minrk:
> These are steps that we could make optional or extract from the actual testing steps:

Yes, these should be made very clear, with what inputs (environment state) are required:

1. install the hub on a cluster
  - inputs:
    - cluster is running and currently selected
    - helm is installed
    - more?
2. run the tests
  - kubernetes is configured
  - assumes Hub URL is available as env
  - more?

Setting up the cluster should be a *wholly separate* step that is only run as part of the tests on CI, because it's quite costly. For developers, this would be run *very* rarely.

We have this

minrk:
Oops, clicked "Close & Comment" instead of delete comment.

ci/4-setup-cluster-env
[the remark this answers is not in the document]
consideRatio: Thats it! Thanks for teaching me about docker haha lets add more comments!

minrk:
> My ideal state to reach would be that locally I once setup my minikube and then just run pytest from the repo. No arguments, no scripts needed.

We could get there with pytest fixtures. We would have to decide which state pytest assumes we are in to start:

1. hub is up and running (fixtures just connect), or
2. helm is ready and hub can be started (fixture installs the hub chart itself)

The first is simpler and offers much faster turnaround for local development, so we can start there.

ci/cleanup
minrk: In fact, I think there's no need to delete files in a cleanup script. We already have that with `git clean`, so we can leave our own cleanup as deleting the test cluster, since that's the only thing being created that wouldn't normally get cleaned up.

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+@pytest.fixture(scope='function')
+def delete_user(request_data):
+    yield
+    r = requests.delete(request_data['hub_url'] + f"/users/{request_data['username']}", headers=request_data['headers'])
+    r.raise_for_status()
+    print(r.status_code)
+    
+@pytest.fixture(scope='function')
+def request_data():
+    token = '0cc05feaefeeb29179e924ffc6d3886ffacf0d1a28ab225f5c210436ffc5cfd5'
```
minrk: Rather than hardcoding this, let's retrieve it from the config.yaml

tests/test_spawn.py
minrk: We can improve this with some useful pytest fixtures, likely some module- and session-scope fixtures. Relying on test order is something one generally tries to avoid, since it makes isolating tests and failures difficult. That can be improved in another PR, though.

consideRatio: I would also love for the test to be independent and be runable without any state to consider. This is not testing of Python code itself that can be isolated so easily though, it is testing of a kubernetes cluster behavior when running a few webrequests to a JupyterHub server running in the cluster.

Yes it would be **great to have someone more experienced** look at it later. I have never setup pytest myself before, my **experienced** friend @heyhugo got me started yesterday! :D

tools/contributors-requirements.txt
```diff
@@ -0,0 +1,6 @@
+# These are pip requirements for contributors.py. They were extracted from
```
minrk: Let's remove this file. Having more files to maintain is more costly than five seconds.

dev-requirements.txt
```diff
@@ -4,11 +4,7 @@
 # - can publish the built Helm chart (--publish)
 chartpress==0.2.1
 
-# yamllint is important for local development, CI
-# - used within tools/templates/lint-and-validate.py
+# yamllint and pytest are important for local development, CI
+pytest==3.7.1
```
minrk: Do not pin these so tightly, it causes problems for people's environments and prevents our tests from catching compatibility issues. Let's use `>=` here, and pin to something like:

```
pytest>=3.3
requests>=2
```

tests/conftest.py
consideRatio: Ah thanks yes I forgot to add a # FIXME to this quick workaround

tools/contributors-requirements.txt
consideRatio: Deal!

dev-requirements.txt
consideRatio: Ah okay!

consideRatio:
@betatim @minrk :heart: :heart: :heart: for your reviewing effort! I learn so much I love it!

consideRatio:
# Related
A typical failure that we have seen with curl requests in testing are the following. Note that it has systematically happened for the first job in the travis CI build containing two jobs. Are they interfering with eachother?

```
Events:
  Type    Reason                 Age   From                                                      Message
  ----    ------                 ----  ----                                                      -------
  Normal  Scheduled              12s   default-scheduler                                         Successfully assigned user-scheduler-58449bfd8b-77czk to travis-job-3b9b103a-4bb6-48a0-a7d0-04e914ec8047
  Normal  SuccessfulMountVolume  12s   kubelet, travis-job-3b9b103a-4bb6-48a0-a7d0-04e914ec8047  MountVolume.SetUp succeeded for volume "user-scheduler-token-gxqjp"
  Normal  Pulling                11s   kubelet, travis-job-3b9b103a-4bb6-48a0-a7d0-04e914ec8047  pulling image "gcr.io/google_containers/kube-scheduler-amd64:v1.11.2"
+sleep 10
+curl --fail -s http://10.20.0.101:31212/hub/api
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

consideRatio:
stale / outdated

Issue #848, opened 2018-08-16
helm upgrade fail: selector does not match template labels


ryanlovett:
Output from travis:
```
Update Complete. ⎈ Happy Helming!⎈ 
INFO:root:Executing helm upgrade --install --wait staging jupyterhub/jupyterhub --namespace staging --version v0.6 -f hub/config.yaml -f hub/secrets/staging.yaml -f /tmp/tag-4759189.yaml --timeout 3600
Error: UPGRADE FAILED: Deployment.apps "proxy" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"component":"proxy", "heritage":"Tiller", "name":"proxy", "release":"staging"}: `selector` does not match template `labels`
Traceback (most recent call last):
  File "./deploy.py", line 140, in <module>
    main()
  File "./deploy.py", line 138, in main
    deploy(args.release)
  File "./deploy.py", line 109, in deploy
    '--timeout', '3600'
  File "./deploy.py", line 31, in helm
    return subprocess.check_call([arg0] + list(args), **kwargs)
  File "/opt/python/3.6.3/lib/python3.6/subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['helm', 'upgrade', '--install', '--wait', 'staging', 'jupyterhub/jupyterhub', '--namespace', 'staging', '--version', 'v0.6', '-f', 'hub/config.yaml', '-f', 'hub/secrets/staging.yaml', '-f', '/tmp/tag-4759189.yaml', '--timeout', '3600']' returned non-zero exit status 1.
```

This is on Azure with a kubernetes cluster deployed via acs-engine.

chart v0.6
helm 2.9.1

ryanlovett:
If I retry the upgrade after deleting the failed configmap, I get
```
Error: UPGRADE FAILED: no Secret with the name "manual-tls-proxy-staging" found
```
which is #466.

ryanlovett:
I'm attempting to upgrade after having defined
```
proxy:
  https:
    hosts:
      - my.host.name
    type: manual
    manual:
      key: ...
      cert: ...
```

ryanlovett:
I had to delete the release to stop this from happening. Normally I wouldn't do this on production, but the semester hasn't started yet.

consideRatio:
Thanks for reporting this @ryanlovett! 

[...]

PR #849, opened 2018-08-16
remove "v" prefix from versions

minrk:
"v" may be used as a git tag prefix for versions, but isn't part of the version string itself


betatim:
LGTM

consideRatio:
by looking at for example how ryan is installing the chart, it seems like we should provide a v0.6 to the --version param for helm upgrades

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/848

consideRatio:
ah we could certainly transition to doing it without v though. LGTM

PR #844, opened 2018-08-14
Improved template linting/validation and CI-integration

consideRatio:
linting-and-validation-in-ci

PR #851, opened 2018-08-16
Pulling images from external private image repositories

### LINKED ISSUE #871: singleUser.imagePullSecret is created only after the pre-pull
## Setup
- Configure Jupyterhub to pull from a private registry
- Set the imagepullSecret in the config 

## Contents of `config.yaml`
```
singleuser:
  image:
    name: <registry-url>/containers/data-analysis-notebook
    tag: latest
  imagePullSecret:
    username: <name>
    password: <password>
    registry: <registry-url>
```

## Outcome
- The pre pull will fail with "Access denied". 
- Pulling images when the hub is running works (since the secret is in place then)

## Workaround
Disabling the pre pull with 
`prePuller.hook.enable: false`
fixes this, but it should probably be handled differently to still allow pre pulling images.

consideRatio:
## TODO

- [x] Verify hook-pulling from private registries works as well.
- [x] Verify output is EXACTLY the same (as kubeclt create secret ...)  with a registry using symbols like `:/` in it as well

---

@AlexMorreale helped us use config.yaml to install a Kubernetes Secret which could be referenced by pods that want to pull from image registries that required some credentials - this is #801.

This PR complements Alex work by...
- Allowing for the email field to optionally be configured, which seem to be utilized by some registries but not all.
- Adds support for setting large JSON blobs in the password field, something that is required in order to pull from a private gcr.io registry from an external cluster.
- Ensures that the image puller DaemonSets, both `hook-` and `continuous-` have the same credentials to pull the images.

Fixes #871.

---

Note that with #844, this pr will ensure that the Helm template within `jupyterhub/templates/singleuser/image-credentials-secret.yaml` is valid and generated valid output which is a very important check for this PR.

Also note that the secret created with values in config.yaml, will have the *exact* same content as a secret created by kubectl's command if you provide a simple password or if you pass it a big JSON blob.
```
kubectl create secret docker-registry regcred --docker-server=<your-registry-server> --docker-username=<your-name> --docker-password=<your-pword> --docker-email=<your-email> 
```

See [the Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-in-the-cluster-that-holds-your-authorization-token) for more details.

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -191,21 +191,16 @@ component: {{ include "jupyterhub.componentLabel" . }}
 {{- define "jupyterhub.dockerconfigjson.yaml" -}}
 {{- with .Values.singleuser.imagePullSecret -}}
 {
-  {{- /* */ -}}
-  "auths":{
-    {{- .registry | default "https://index.docker.io/v1/" | quote }}:{
-      {{- /* */ -}}
-      "username": {{- .username | quote }},
-      {{- /* */ -}}
-      "password": {{- .password | quote }},
-      {{- if .email -}}
-      "email": {{- .email | quote }},
-      {{- end -}}
-      "auth": {{- (print .username ":" .password) | b64enc | quote -}}
+  "auths": {
+    {{ .registry | default "https://index.docker.io/v1/" | quote }}: {
+      "username": {{ .username | quote }},
+      "password": {{ .password | quote }},
+      {{- if .email }}
+      "email": {{ .email | quote }},
+      {{- end }}
+      "auth": {{ (print .username ":" .password) | b64enc | quote }}
     }
-    {{- /* */ -}}
   }
-  {{- /* */ -}}
 }
 {{- end }}
 {{- end }}
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -189,7 +189,7 @@ component: {{ include "jupyterhub.componentLabel" . }}
 {{- end }}
 
 {{- define "jupyterhub.dockerconfigjson.yaml" -}}
-{{- with . -}}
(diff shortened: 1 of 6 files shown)


jupyterhub/templates/_helpers.tpl
```diff
@@ -173,10 +173,39 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
+{{ include "jupyterhub.dockerconfigjson.yaml" . | b64enc }}
+{{- end }}
+
+{{- define "jupyterhub.dockerconfigjson.yaml" -}}
+{{- with . -}}
+{
+  {{- /* */ -}}
+  "auths":{
+    {{- .registry | default "https://index.docker.io/v1/" | quote }}:{
+      {{- /* */ -}}
```
minrk: What are all these empty comments doing?

consideRatio: I wanted to recreate the exact output of when using kubectl to create the secret in order to ensure things works. But that output was very compact, without any whitespace or newlines. The comments allow me to whitespace chomp the added spaces and newlines that makes this somewhat readable at all. Without them it became a blob of text i could not read.

We could make the output contain the extra whitepace though, my guess is that it would still work, im not 100% on it though so i figured it was best not to deviate at all.

consideRatio: sunmary: they make a blob without whitespace somewhat readable. if removed we deviate from generated output of kubectl doing the same thing, which may be fine but i dont know for sure

consideRatio: Deal! Thanks for the decisive input!

consideRatio:
TADAAA! :tada:

minrk:
Thanks!

AlexMorreale:
Woooo! Sorry i missed this I was on Vacation last week.

PR #853, opened 2018-08-17
cull.maxAge ignored bugfix

consideRatio:
The config entry cull.maxAge was previously never consumed by the
jupyterhub_config.py as it reads from its configmap where the entry
cull.max-age was never set. This affected maxAge based cullings on
JupyterHub 0.9+ to never cull due to exceeding a set maxAge.


consideRatio:
Lets go all out @betatim, 50 is way to high! How about 1 line?

PR #854, opened 2018-08-18
No more duplicates of puller pods

consideRatio:
Thanks to the new `before-hook-creation` deletion policy, we can avoid the `object already exist` error as Helm would simply delete and recreate the resource instead if it found a resource with the same name. This means that we can use a single name for the puller daemonset and won't ever orphan a puller daemonsets after a failed upgrade followed by a successful upgrade.

### DIFF
--- a/jupyterhub/templates/image-puller/job.yaml
+++ b/jupyterhub/templates/image-puller/job.yaml
@@ -15,7 +15,7 @@ metadata:
     hub.jupyter.org/deletable: "true"
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
     "helm.sh/hook-weight": "10"
 spec:
   template:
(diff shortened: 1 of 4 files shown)


jupyterhub/templates/image-puller/job.yaml
```diff
@@ -9,24 +9,24 @@ command.
...
 metadata:
-  name: hook-image-awaiter-{{ .Release.Time.Seconds }}
+  name: hook-image-awaiter
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
     hub.jupyter.org/deletable: "true"
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
minrk: What happens on hook-failed?

consideRatio: great point, this bothinh before and nothing when it is removed, this was a bug id say but helm dis not consider it very important.

only the last created Job resource the among hook resources will be deleted if this deletion policy is around on a Job resource. For a daemon set it does kot matter at all and with ko warning if another job etc failed in the dame helm upgrade etc.

removing it as it is dead code causing confusion  but i should perhaps add notes about this

consideRatio: I edited the reply above, it was not really readable before - will stop writing from mobile without reviewing my text haha sorry @minrk!

consideRatio:
Let's merge this! It is tested by me over the course of about 20+ helm upgrade command or so and it has worked like a charm. If my helm upgrade would fail for some reason and then after a rerun succeed there would not remain any image-pullers daemonset or pods left thanks to this PR! It can and have been a headache with such issues.

I'm hoping for a lot of trust in me getting this right even though we have not gotten all testing infrastructure in place. What do you say @minrk ?

minrk:
Awesome, thanks @consideRatio!

Issue #341, opened 2017-12-08
Move all parameter defaults for images into the jupyterhub_config.py


yuvipanda:
Right now, we have a bunch of parameter defaults that are in values.yaml, and a bunch that are in jupyterhub_config.py. This is quite confusing.

We should standardize on putting all the defaults in the jupyterhub_config.py. This lets us make the docker image a bit more self contained, so users who don't want to use the helm chart can still benefit from it. It would also allow us to possibly test the image without spinning up a full k8s cluster...

consideRatio:
**Action item:**
Identify what default values we have in values.yaml, that propegates to configmap, and then in turn ends up being consumed in images/hub/jupyterhub_config.py

minrk:
I'm not quite sure what the best solution to this is. Having defaults in values.yaml is super useful for seeing the structure/type of parameters (way more useful for humans than the json schema). Putting the defaults only in the jupyterhub_config.py means users can't really see what they are (maybe this is good?).

From a technical perspective, it does make sense - it makes Hub images resilient to small variations in the configmap/chart version. But from that perspective, the defaults should really reside only in the classes themselves, and only defaults that *deviate* from the class defaults should be defined in jupyterhub_config.py.

consideRatio:
Hmmm, there are JupyterHub's actual defaults, the defaults in values.yaml, propegated to configmap and into jupyterhub_config.py that can also have defaults.

I'd like:
* To avoid having two sets of defaults
* To make it easy for users to overview default values of the configuration that they can override, this will be within values.yaml

It is not very user friendly to inspect the values.yaml file in order to learn the default values of the Helm chart, but it is not too bad considering the complexity of the deployment in general.

---

All in all, I figure we should perhaps focus on setting the defaults in values.yaml after all anyhow.

yuvipanda:
I've spent more time thinking about this and seeing the approach that has worked in TLJH. 

1. There should exist a 'z2jh' module (similar to tljh module).
2. This module has a 'configurer' that takes a YAML config, and applies that to JupyterHub / KubeSpawner config as needed (https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/configurer.py)
3. The defaults are in values.yaml. This is *directly* passed in to the configurer, with no go templating in the configmap. This eliminates all the passing around with the configmap we have.
[...]

PR #856, opened 2018-08-18
avoid setting None in config values if unspecified

### LINKED ISSUE #912: hub/configmap - missing callback-url entry, i think
We are lacking a pass through here of `callback-url` in the hub's configmap, I think. At least such callback url is needed systematically with the other authenticators and is now attempted to be consumed in `jupyterhub_config.py` even though it does not exist in the configmap.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/f5dd68cf7e6ddec4579b1c8fa87b56ddb56a59ce/jupyterhub/templates/hub/configmap.yaml#L72-L76

minrk:
If a value is unspecified in the configmap, avoid passing it along to the config object.

Setting a value as None is not the same as not setting it at all, and setting it as None is often invalid and a source of bugs, especially when using a custom Hub image.

This causes issues like https://github.com/jupyterhub/kubespawner/issues/187

Because this is a big bulk replacement and vulnerable to typos, I'd like to land this *after* we release 0.7.

In general, I think we have a problem where each new option requires specifying it in three places:

- values.yaml (camelCase)
- configmap (css-case)
- jupyterhub_config (snake_case)

I think we can probably simplify this in the future by creating the configmap with the right shape in the first place,
so the configmap contains a dict that can be loaded directly into jupyterhub config.

Then only the more complex transformations (e.g. the storage config) would need to reside in jupyterhub_config.py, and most new options would only need to be defined in two places, not three.

closes jupyterhub/kubespawner#187
closes #912


consideRatio:
Excellent!

---

:100: if we can reduce the steps from config.yaml -> hub-config configmap -> jupyterhub_config.py -> kubespawner configuration.

The configmap is simply a way for the jupyterhub_config.py to read content.
Config.yaml is made to allow any helm resource to read content and will be broadly consumed.
The jupyterhub_config.py -> kubespawner configuration could perhaps also be simplified by passing it a batch of settings instead of one by one with various checks in beforehand.

The easiest first improvement would probably be simplifying the configmap to hold a single "data" entry or similar though. Should be doable I figure, even though i bet it will involve a lot of fiddling around with toYaml, toJson, or even b64enc steps or similar.

images/hub/jupyterhub_config.py
```diff
@@ -22,70 +22,63 @@
...
 # the hub should listen on all interfaces, so the proxy can access it
 c.JupyterHub.hub_ip = '0.0.0.0'
 
-c.KubeSpawner.common_labels = get_config('kubespawner.common-labels')
+set_config_if_not_none(c.KubeSpawner, 'common_labels', 'kubespawner.common-labels')
 
 c.KubeSpawner.namespace = os.environ.get('POD_NAMESPACE', 'default')
 
-c.KubeSpawner.start_timeout = get_config('singleuser.start-timeout')
+set_config_if_not_none(c.KubeSpawner, 'start_timeout', 'singleuser.start-timeout')
```
consideRatio: start_timeout added below

images/hub/jupyterhub_config.py
```diff
@@ -22,70 +22,63 @@
...
-c.KubeSpawner.extra_labels = get_config('singleuser.extra-labels', {})
+for trait, cfg_key in (
+    ('start_timeout', 'start-timeout'),
+    ('image_pull_policy', 'image-pull-policy'),
+    ('image_pull_secrets', 'image-pull-secret-name'),
+    ('events_enabled', 'events'),
+    ('extra_labels', 'extra-labels'),
+    ('extra_annotations', 'extra-annotations'),
+    ('uid', 'uid'),
+    ('fs_gid', 'fs-gid'),
```
consideRatio: Just a note: We are required by kubespawner to set uid/fs_gid atm I think, perhaps it was image_pull_policy also, a few traitlets are special.

Perhaps we should add some default values on them in kubespawner, allowing values provided to a hub image to still work which yuvi suggested would be good.

images/hub/jupyterhub_config.py
```diff
@@ -120,87 +113,106 @@
...
+        set_config_if_not_none(c.GitLabOAuthenticator, trait, 'auth.gitlab.' + cfg_key)
 elif auth_type == 'mediawiki':
     c.JupyterHub.authenticator_class = 'oauthenticator.mediawiki.MWOAuthenticator'
-    c.MWOAuthenticator.client_id = get_config('auth.mediawiki.client-id')
-    c.MWOAuthenticator.client_secret = get_config('auth.mediawiki.client-secret')
-    c.MWOAuthenticator.index_url = get_config('auth.mediawiki.index-url')
+    for trait, cfg_key in (
+        ('client_id', 'client-id'),
+        ('client_secret', 'client-secret'),
+        ('oauth_callback_url', 'callback-url'),
```
consideRatio: Is this a bugfix in this authenticator configuration? There are no deleted entry about oauth_callback_url.

Perhaps we need to fix the configmap as well if so

images/hub/jupyterhub_config.py
```diff
@@ -254,18 +264,19 @@ def generate_user_name(spawner):
...
 
 c.JupyterHub.services = []
 
 if get_config('cull.enabled', False):
-    cull_timeout = get_config('cull.timeout')
-    cull_every = get_config('cull.every')
-    cull_concurrency = get_config('cull.concurrency')
+    cull_timeout = get_config('cull.timeout', 3600)
+    cull_every = get_config('cull.every', 600)
+    cull_concurrency = get_config('cull.concurrency', 10)
```
consideRatio: Should we perhaps remove the default entry in the values file while we add the default here? If not, this default will be ignored forever assuming we are running the helm chart.

consideRatio: related: #341

consideRatio: Perhaps we should do this systematically when fixing #341 though, it would reduce risks of me messing up with rebasing a lot of work already done.

images/hub/jupyterhub_config.py
consideRatio: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/f5dd68cf7e6ddec4579b1c8fa87b56ddb56a59ce/jupyterhub/templates/hub/configmap.yaml#L72-L76

Yepp, so either we should not set `oauth_callback_url` here, or update the configmap to accept such entry as well.

images/hub/jupyterhub_config.py
minrk: Yes, that should be a separate PR. This one is just fixing a bug when things were undefined in the configmap.

images/hub/jupyterhub_config.py
minrk: Fortunately, this is why this PR exists! Failing to set oauth callback URL will have no negative consequences, because it's only set if the value is provided. This is exactly the kind of bug we are now resilient to after this PR.

But yes, it is a bug that the oauth callback URL isn't exposed for this authenticator.

images/hub/jupyterhub_config.py
consideRatio: @minrk remove this duplication of setting start_timeout which is already done on line 59, the rest LGTM.

minrk:
@consideRatio thanks for the review! I think I addressed the remaining points.

PR #859, opened 2018-08-19
CHANGELOG update for beta: upgrade instructions

consideRatio:


CHANGELOG.md
```diff
@@ -1,13 +1,166 @@
...
+#### Troubleshooting
+
+If things fail, you could try the following before installing the chart.
+
+```sh
+RELEASE_NAME=<YOUR-RELEASE-NAME>
+
+# WARNING: Deletes everything installed by the Helm chart
+# NOTE: This does not include user pods or user storage PVCs as they have been
+#       indirectly created by KubeSpawner
```
minrk: Maybe add a note that this *does* include all Hub state (user lists, tokens, etc.)

CHANGELOG.md
```diff
@@ -1,13 +1,166 @@
...
+If you are running v0.5 of the chart, you should upgrade to v0.6 first before
+upgrading to `0.7.0-beta.1`. You can find out what version you are using by
+running `helm list`.
+
+Follow the steps below to upgrade to the new version.
+
+#### 1. Ensure the hub's user data isn't lost
+
+```sh
+# The script will is a saftey measure and patches your hub's PersistentVolume
```
betatim: delete "will"

CHANGELOG.md
```diff
@@ -1,13 +1,166 @@
...
+RELEASE_NAME=<YOUR-RELEASE-NAME>
+
+# Give yourself an overview of this release's revisions
+helm history $RELEASE_NAME
+
+# Check if you have multiple revisions in a DEPLOYED status (a bug), or if you
+# have old PENDING_UPGRADES or FAILED revisions (may be problematic).
+helm history $RELEASE_NAME | grep --extended-regexp "DEPLOYED|FAILED|PENDING_UPGRADE"
+
+# If you have multiple revisions in DEPLOYED status, you this script will clean
```
betatim: "you this script" -> "this script"

CHANGELOG.md
```diff
@@ -1,13 +1,166 @@
...
+helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
+helm repo update
+
+# NOTE: We need the --force flag to allow recreation of resources that can't be
+#       upgraded to the new state by a patch.
+helm upgrade $RELEASE_NAME jupyterhub/jupyterhub --install \
+    --force \
+    --version=0.7.0-beta.1 \
+    --namespace=$NAMESPACE \
+    --values config-dev.yaml \
```
betatim: We use `config.yaml` in most of the rest of the docs, use that here as well?

CHANGELOG.md
consideRatio: We need to clarify something somehow, but actually it should not influence the hub state etc, because in step 1 I patch the PV bound to the hub-db-dir PVC to use the reclaimPolicy of "Retain". Meaning; the PVC is deleted, the PV is not deleted but retained and will be rebound later, I think.

I have not verified this, so its based on theory, but I have verified that the hub-db-dir PV isnt garbage collected when it is unbound because the PVC is deleted.

CHANGELOG.md
consideRatio: Ooops absolutely

CHANGELOG.md
consideRatio: Learning! See #860

consideRatio:
@betatim @minrk thanks for the feedback. I learned and experimentally verified things related to backup of PVs to avoid loosing any previous data.

We now have quite a long upgrade instruction with many steps, even though some are optional. I think this will actually stand a good chance of letting most users succeed! If the selective cleanup of faulty Helm state fails, they could resort to the dramatic `kubectl delete namespace <namespace>` or `helm delete <release> --purge` without loosing hub or user storage thanks to the instructions on how to `Retain` PVs when PVCs are deleted along with instructions on how to make them go from `Released` to `Available`.

PR #877, opened 2018-08-27
CHANGELOG.md contained duplicate instructions

consideRatio:


consideRatio:
Hmmmm, we are triggering two builds on travis, one called PR and one is called PUSH - whats going on here... To be resolved.

PR #882, opened 2018-08-29
Add extra annotations option

### LINKED ISSUE #810: Annotations on single user notebook
I'm trying to add an annotation to all notebooks which are created, however I can't seem to find where to configure this in the chart config. The hub, proxy and ingress all have annotation options, but the singleuser section doesn't.

Am I missing something or has this not been implemented yet?

_NB if I could do this dynamically based on username that would be even better. But that may be a bigger ask and invoke KubeSpawner._

jacobtomlinson:
Kubespawner allows you to specify extra annotations for the singleuser notebook pods however there is no way to set this from the helm chart.

This PR adds some extra config to pass these values through.

Fixes #810

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -57,6 +57,8 @@ c.KubeSpawner.image_pull_secrets = get_config('singleuser.image-pull-secret-name
 
 c.KubeSpawner.events_enabled = get_config('singleuser.events', False)
 
+c.KubeSpawner.extra_annotations = get_config('singleuser.extra-annotations', {})
+
 c.KubeSpawner.extra_labels = get_config('singleuser.extra-labels', {})
 
 c.KubeSpawner.uid = get_config('singleuser.uid')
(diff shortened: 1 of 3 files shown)


consideRatio:
Excellent thanks for the PR @jacobtomlinson !
- great that you added a stub in the values.yaml file
- configmap entry looks good

This will still not note make the spawned users annotated though, kubespawner must be configured as well! Could you add an entry in the images/hub/jupyterhub_config.py to do so?

consideRatio:
@jacobtomlinson LGTM! Could you verify that it works for you as well?

jacobtomlinson:
I can confirm that it works! I added a "hello world" tag to a vanilla z2jh deployment built from this branch.

```
$ kubectl describe po jupyter-test
Name:         jupyter-test
Namespace:    z2jh-test
Node:         <redacted>
Start Time:   Wed, 29 Aug 2018 12:03:36 +0100
Labels:       app=jupyterhub
              chart=jupyterhub-0.7-dev
              component=singleuser-server
              heritage=jupyterhub
              hub.jupyter.org/network-access-hub=true
              release=z2jh-test
Annotations:  hello=world
              hub.jupyter.org/username=test
```

consideRatio:
Wieee! Thanks awesome @jacobtomlinson !

PR #884, opened 2018-08-29
Instantly redirect to the progress page

consideRatio:
@minrk pointed out that the slow redirection to the progress page is
related to the hub is waiting for the spawning to complete before it
redirects, as that may happen very quickly sometimes. But, on
kubernetes and clouds, we won't be able to insta-spawn a pod as we can
insta-spawn a process, so always redirecting to the progress page right
away makes sense.


consideRatio:
@minrk should we do this pre / post 0.7.0?

consideRatio:
Certainly, I was not super confident on the stability though. Tested it once and got a redirect loop, but that was probably related to something else. Tested it again later and got it working without trouble.

Happy to have it merged :D

PR #888, opened 2018-08-30
Refreshed singleuser-sample

consideRatio:
The singleuser-sample image that we provide needed a refresh.

- I updated the README.md file next to it
- I removed an additional install of jupyterhub (docker-stacks images already has it)
- I added git and nbgitpuller
- The compressed image size went from about 230 to 260.

### DIFF
--- a/images/singleuser-sample/Dockerfile
+++ b/images/singleuser-sample/Dockerfile
@@ -6,4 +6,18 @@ FROM jupyter/base-notebook:8ccdfc1da8d5
 # The jupyter/docker-stacks images contains jupyterhub, jupyterlab and the
 # jupyterlab-hub extension already.
 
+# Example install of git and nbgitpuller.
+# NOTE: git is already available in the jupyter/minimal-notebook image.
+USER root
+RUN apt-get update && apt-get install --yes \
+    git \
+ && rm -rf /var/lib/apt/lists/*
+USER $NB_USER
+
+RUN pip install nbgitpuller && \
+    jupyter serverextension enable --py nbgitpuller --sys-prefix
+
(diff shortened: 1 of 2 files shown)


images/singleuser-sample/Dockerfile
```diff
@@ -1,11 +1,23 @@
...
-# conda/pip/apt install additional packages here, if desired.
+# The jupyter/docker-stacks images contains jupyterhub, jupyterlab and the
+# jupyterlab-hub extension already.
+
+# Example install of git and nbgitpuller.
+# NOTE: git is already available in the jupyter/minimal-notebook image.
+USER root
+RUN apt-get update && apt-get install --yes \
+    git \
+ && rm -rf /var/lib/apt/lists/*
```
Carreau: hum, astonished there is not apt-get command for that...

consideRatio: Yepp! Perhaps there is, I'm confused still about the things that is done by various `clean` commands. I've been afraid to use `apt-get` within a Dockerfile due to unknown cleanup procedures to make sure the image size does not explode as I thought it did when comparing. I had made a mistake of comparing actual size with Docker Hub's *compressed* size field though.

Since I needed to use `apt-get` now, I looked things up further and learned that this was the best practice on a ubuntu/debian based image. Apparently this is all the cleanup you need, and it is also important to have `apt-get update` run on the same line as `apt-get install` as if you don't, the `apt-get update` layer could be cached while `apt-get install` could be updated to get some new version that would not be found as `apt-get update` was not rerun.

Carreau: Thanks, I was roughly aware of all these issue, and still astonished there is no simpler way to fix those.

PR #890, opened 2018-08-30
[Part 1 of #758] Remove pod-culler remnants

consideRatio:
As of JupyterHub 0.9, we ended up with some dead code no longer in use.


chartpress.yaml
```diff
@@ -18,5 +18,3 @@ charts:
         valuesPath: singleuser.image
         buildArgs:
           JUPYTERHUB_VERSION: 0.9.2
-      pod-culler:
-        valuesPath: cull.podCuller.image
```
consideRatio: No longer building the image, should be very safe.

jupyterhub/templates/_helpers.tpl
```diff
@@ -162,18 +162,6 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
-    Used to by the pod-culler to select singleuser-server pods. It simply
-    reformats "jupyterhub.matchLabels" and sets the componentLabel value so
-    `component=singleuser-server` is output.
-*/}}
-{{- define "jupyterhub.podCullerSelector" -}}
-{{- $_ := merge (dict "componentLabel" "singleuser-server") . -}}
-{{ include "jupyterhub.matchLabels" $_ | replace ": " "=" | replace "\n" "," | quote }}
-{{- end }}
-
-
```
consideRatio: This helper was only used by the deleted pod-culler deployment

images/hub/jupyterhub_config.py
```diff
@@ -277,12 +277,8 @@ def generate_user_name(spawner):
     if get_config('cull.users'):
         cull_cmd.append('--cull-users')
 
-    # FIXME: remove version check when we require jupyterhub 0.9 in the chart
-    # that will also mean we can remove the podCuller image
-    import jupyterhub
-    from distutils.version import LooseVersion as V
     cull_max_age = get_config('cull.max-age')
-    if cull_max_age and V(jupyterhub.__version__) >= V('0.9'):
+    if cull_max_age:
```
consideRatio: This will have the same truth table no matter what, if we assume jupyterhub.__version__ to always be >= 0.9, which we do.

jupyterhub/templates/pod-culler/rbac.yaml
```diff
@@ -1,42 +0,0 @@
-{{- if and .Values.cull.enabled .Values.cull.maxAge }}
-{{- if .Values.rbac.enabled -}}
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: pod-culler
-  labels:
-    {{- include "jupyterhub.labels" . | nindent 4 }}
----
```
consideRatio: This serviceAccount, and associated role bound to it, was only used by the pod-culler deployment.

jupyterhub/templates/_helpers.tpl
consideRatio: The service running in the hub pod will now take care of the culling, and the hub pod has the credentials required do to so already.

jupyterhub/templates/pod-culler/deployment.yaml
```diff
@@ -1,39 +0,0 @@
-{{- if and .Values.cull.enabled .Values.cull.maxAge -}}
```
consideRatio: OH! Note that this culler ran alongside the jupyterhub in-hub-pod service culler since JH 0.9, we had two of them before! I wonder if that caused some issues?

ping @minrk @betatim - in case you have had some unexpected culling behaviors.

jupyterhub/values.yaml
```diff
@@ -235,10 +235,6 @@ cull:
   every: 600
   concurrency: 10
   maxAge: 0
-  podCuller:
-    image:
-      name: jupyterhub/k8s-pod-culler
-      tag: generated-by-chartpress
```
consideRatio: These values were only consumed by the deleted deployment.

PR #891, opened 2018-08-30
[Part 2 of #758] user-scheduler added

### LINKED ISSUE #542: Make 'pack' schedulerStrategy work properly

consideRatio:
# About
The user-scheduler is great for use with a cluster autoscaler and bad 
without it. With the user-scheduler optionally enabled users will be 
packed to fill up one node at the time. This in turn tends to create 
totally unused nodes that are easy for a cluster autoscaler to scale 
down the unused node. Without a cluster autoscaler, it is just a waste 
of resources to get yourself unused nodes though, so only enable the 
user-scheduler if you have a cluster autoscaler.

NOTE: This works without k8s 1.11. Only *user-placeholder* utilizing *pod priority* and *pod preemption* require k8s 1.11.

## A code overview - useful for anyone reviewing this code
**IMPORTANT:** Everything inside this PR is *disabled by default*, and can be optionally enabled by config.yaml by passing `scheduling.userScheduler.enabled: true`. This means that the PR is very unintrusive and could perhaps be merged quite quickly for quicker feedback from curious users.

A Kubernetes Deployment called *user-scheduler* is added. This Deployment ensures one (or more) replica of a pod is running, and this pods runs a container based on an image and associated binary provided by Kubernetes called *kube-scheduler*. *kube-scheduler* is configured to run with different than-default settings found in a configmap, so anything it schedules will be scheduled to pack tight on a node that already consumes a lot of resources.

Our manually configured kube-scheduler will only schedule pods requesting to be scheduled by it. That is why this PR modifies the jupyterhub_config.py wherein kubespawner is configured to set the pods *spec.schedulerName* field to match the our specific scheduler and not the default one named *default*.

The RBAC requirements for the scheduler are the same as the ClusterRole called `system:kube-scheduler` but needs to be augmented slightly to access the configmap with the configuration we provide. We end up with a ServiceAccount coupled with two ClusterRoles (`system:kube-scheduler`, and the one with small additions) through two ClusterRoleBindings.

I'm also making the `singleuser.schedulerStrategy` Helm chart configuration meaningless. Note that there is still an entry about it in schema.yaml that sais it is now deprecated and that the user-scheduler should be used instead.

---

Fixes #542

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -150,9 +150,6 @@ data:
   singleuser.service-account-name: {{ .Values.singleuser.serviceAccountName | quote }}
   {{- end }}
   singleuser.node-selector: {{ toJson .Values.singleuser.nodeSelector | quote }}
-  {{- if .Values.singleuser.schedulerStrategy }}
-  singleuser.scheduler-strategy: {{ .Values.singleuser.schedulerStrategy | quote }}
-  {{- end }}
   singleuser.storage.type: {{ .Values.singleuser.storage.type | quote }}
   singleuser.storage.home_mount_path: {{ .Values.singleuser.storage.homeMountPath | quote }}
   singleuser.storage.extra-volumes: {{ toJson .Values.singleuser.storage.extraVolumes | quote }}
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -213,6 +213,9 @@ data:
     {{- end }}
   {{- end }}
 
+  {{- if .Values.scheduling.userScheduler.enabled }}
+  singleuser.scheduler-name: "user-scheduler"
+  {{- end }}
 
   {{- /* KubeSpawner */}}
   kubespawner.common-labels: |
--- a/jupyterhub/templates/scheduling/user-scheduler/_helpers.tpl
+++ b/jupyterhub/templates/scheduling/user-scheduler/_helpers.tpl
@@ -1,0 +1,12 @@
+{{- /*
+Renders the kube-scheduler's image based on .Values.scheduling.userScheduler.name and
+optionally on .Values.scheduling.userScheduler.tag. The default tag is set to the clusters
+kubernetes version.
+*/}}
+{{- define "jupyterhub.scheduler.image" -}}
+{{- $name := .Values.scheduling.userScheduler.image.name -}}
+{{- $valuesVersion := .Values.scheduling.userScheduler.image.tag -}}
+{{- $clusterVersion := (split "-" .Capabilities.KubeVersion.GitVersion)._0 -}}
+{{- $tag := $valuesVersion | default $clusterVersion -}}
+{{ $name }}:{{ $tag }}
+{{- end }}
--- a/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
@@ -1,0 +1,73 @@
+{{- if .Values.scheduling.userScheduler.enabled -}}
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: user-scheduler
+  labels:
+    {{- include "jupyterhub.labels" . | nindent 4 }}
+data:
+  policy.cfg: |
+    {
+      "kind": "Policy",
+      "apiVersion": "v1",
+      "predicates": [
+        { "name": "PodFitsResources" },
+        { "name": "HostName" },
+        { "name": "PodFitsHostPorts" },
+        { "name": "MatchNodeSelector" },
+        { "name": "NoDiskConflict" },
+        { "name": "PodToleratesNodeTaints" },
+        { "name": "MaxEBSVolumeCount" },
+        { "name": "MaxGCEPDVolumeCount" },
+        { "name": "MaxAzureDiskVolumeCount" },
+        { "name": "CheckVolumeBinding" },
+        { "name": "NoVolumeZoneConflict" },
+        { "name": "MatchInterPodAffinity" }
+      ],
+      "priorities": [
+        { "name": "NodePreferAvoidPodsPriority",  "weight": 10000 },
+        { "name": "NodeAffinityPriority",         "weight": 1000 },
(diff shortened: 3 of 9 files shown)


jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+          image: {{ include "jupyterhub.scheduler.image" . }}
+          command:
+            - /usr/local/bin/kube-scheduler
+            - --scheduler-name=user-scheduler
+            - --policy-configmap=user-scheduler
+            - --policy-configmap-namespace={{ .Release.Namespace }}
+            - --lock-object-name=user-scheduler
+            - --lock-object-namespace={{ .Release.Namespace }}
+            - --leader-elect-resource-lock=configmaps
+            - --v=100
```
consideRatio: The scheduler is configured to support leader election, not that it matters much, but it works anyhow. The scheduler is also contained within the jupyterhub namespace.

I should perhaps lower the verbosity =D

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -213,6 +210,9 @@ data:
     {{- end }}
   {{- end }}
 
+  {{- if .Values.scheduling.userScheduler.enabled }}
+  singleuser.scheduler-name: "user-scheduler"
+  {{- end }}
```
consideRatio: This is the way we pass through userScheduler enabled to be true or false pretty much. One could argue that it should be declared as true/false instead and let the jupyterhub_config.py figure out what scheduler-name to set based on that.

jupyterhub/templates/scheduling/user-scheduler/_helpers.tpl
```diff
@@ -0,0 +1,12 @@
...
+optionally on .Values.scheduling.userScheduler.tag. The default tag is set to the clusters
+kubernetes version.
+*/}}
+{{- define "jupyterhub.scheduler.image" -}}
+{{- $name := .Values.scheduling.userScheduler.image.name -}}
+{{- $valuesVersion := .Values.scheduling.userScheduler.image.tag -}}
+{{- $clusterVersion := (split "-" .Capabilities.KubeVersion.GitVersion)._0 -}}
+{{- $tag := $valuesVersion | default $clusterVersion -}}
+{{ $name }}:{{ $tag }}
+{{- end }}
```
consideRatio: **This is currently not in use.**
- [ ] Make a note about this not being in use yet, but could be if we want to match the k8s version.

jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
```diff
@@ -0,0 +1,73 @@
...
+      "priorities": [
+        { "name": "NodePreferAvoidPodsPriority",  "weight": 10000 },
+        { "name": "NodeAffinityPriority",         "weight": 1000 },
+        { "name": "InterPodAffinityPriority",     "weight": 100 },
+        { "name": "MostRequestedPriority",        "weight": 50 }
+      ],
+      "hardPodAffinitySymmetricWeight" : 100,
+      "alwaysCheckAllPredicates" : true
+    }
+{{- end }}
```
consideRatio: These are the actual configuration that makes the user-scheduler special and different from the default scheduler.

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
consideRatio: Note that this way of declaring what configmap to use was deprecated, without an option available in beta-state (k8s 1.11).

We could go for the new fancy way of doing it, involving a new type of configuration resource pointing to the configmap instead, but then we would end up dependent on k8s 1.11+ while we currently are not (even though we use a kube-scheduler binary tagged 1.11.2)

minrk: Let's stick with the deprecated way for now. We don't want to be too aggressive with our base versions.

minrk: Can you add a TODO comment in the code to indicate this for when we do eventually  require 1.11?

minrk:
This is really great, and I'm looking forward to using it on Binder! Can we add some documentation for how/when to use it? Then I think we can ship it and begin testing.

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
consideRatio: - [ ] Add a TODO comment in code to indicate this for when we do eventually require

consideRatio:
@minrk thanks for looking through this! About documentation, yes sure we should add that! There are many PRs coming with related settings, where the docs should probably be written with them in mind. Some documentation is already provided in the schema.yaml, which will be used to generate the "configuration reference" in the helm chart.

Perhaps we can settle with that until we have all the pieces related to scheduling in place, allowing us to then write a more cohesive usage example? I'm thinking of the future PRs to come with user-placeholders & pod-priority, affinity configuration options, and updated documentation about autoscaling in general.

minrk:
sounds like a good plan

PR #892, opened 2018-08-30
correct error in documentation regarding singleuser.image.pullPolicy

samuelmanzer:
In the file `jupyterhub/templates/hub/configmap.yaml`, the `imagePullPolicy` for the `singleuser` pods is set with the line:

```
singleuser.image-pull-policy: {{ .Values.singleuser.image.pullPolicy | quote }}
```

I have corrected the docs to indicate this.

Thank you for creating this very useful project!

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -336,17 +336,6 @@ properties:
               - "null"
             description: |
               Note that this field is referred to as *requests* by the Kubernetes API.
-      imagePullPolicy:
-        type: string
-        enum:
-          - IfNotPresent
-          - Always
-          - Never
-        description: |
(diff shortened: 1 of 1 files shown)


consideRatio:
@samuelmanzer wieeee thanks for your PR! You are spot on, apparently we have confused ourselves by not being systematic. I found that we set `hub.imagePullPolicy`, but `singleuser.image.pullPolicy` and `proxy.chp.image.pullPolicy`.

I wonder if we should also change `hub.imagePullPolicy` to `hub.image.pullPolicy` hmm.. that would be a breaking change for anyone having configured this though so I won't suggest we do this atm.

Thank you :D :heart:

PR #896, opened 2018-09-03
remove `-network-policy` from network-policy names

minrk:
it's redundant to ask for `kubectl get networkpolicy hub-network-policy`.

In general: don't put resource type in resource names


consideRatio:
Hmmmm it is quite often that the k8s 1.10 build fails... I wonder if it is simply because it is the FIRST build rather than it is the k8s 1.10 build.

PR #898, opened 2018-09-03
Fix changelog / docs for 0.7.0

consideRatio:


CHANGELOG.md
```diff
@@ -40,7 +39,7 @@ manual cleanup of PVs when you want to stop spending money for the storage.
 NAMESPACE=<YOUR-NAMESPACE>
 
 # Ensure the hub's and users' data isn't lost
-hub_and_user_pvs=($(kubectl get persistentvolume | grep -E "$NAMESPACE/(hub-db-dir|claim-)" | awk '{print $1}'))
+hub_and_user_pvs=($(kubectl get persistentvolumeclaim --no-headers --namespace $NAMESPACE | awk '{print $3}'))
```
consideRatio: The script looked at the PV's available outside the cluster as well. With this change, we are only looking at the PVC's referenced by PVC's in the current namespace.

PR #899, opened 2018-09-03
contributors.py performance by caching

consideRatio:
Thanks @minrk for the caching solution.

I also added some notes and reduced numbers of requests made by using the `since` keyword for the API where it was made available. I learned that PyGitHub did not perfectly correspond with the actual GitHub API.

### DIFF
--- a/tools/contributors.py
+++ b/tools/contributors.py
@@ -25,39 +25,51 @@ variable. For example,
 Note that if you put a space before your command, it does not
 get stored in your bash history (by default)!. Look up `HISTCONTROL`
 to learn more about this feature of shells.
+
+IMPORTANT:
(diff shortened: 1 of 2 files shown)


tools/contributors.py
```diff
@@ -25,56 +25,74 @@
...
     def include(date):
-        return from_date < date.replace(tzinfo=dateutil.tz.tzlocal()) < to_date
+        return since < date
+
     repo = gh.get_repo(repo)
-    issues = repo.get_issues(state='all')
+    
+    # get all issues created or updated since given date
+    issues = repo.get_issues(state='all', since=since)
     pulls = repo.get_pulls(state='closed')
```
consideRatio: get_pulls did not support the `since` keyword.

PR #900, opened 2018-09-03
pass on upgrade steps in changelog

minrk:
- Remove recommendation to change the global PV default. This is not recommended as part of the upgrade process.
- explain why we upgrade with `--force`


minrk:
@consideRatio thanks! I did have one more thing, but it's pushed now.

PR #910, opened 2018-09-06
Support singleuser extraContainer

### LINKED ISSUE #713: Add support for singleuser extra containers
There's an option to add singleuser extra containers in the Jupyterhub documentation, here: https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L764

It doesn't seem to be supported by the chart (although I may be missing something...). Is there any chance this can be added?

My use case is that I want to run a postgres DB in the same pod as the Jupyter container, for better or for worse.

### LINKED ISSUE #830: Supporting launching companion Pods
We're starting work on a new project for cybersecurity education building on the [TryCybSI](http://try.cybersecurity.ieee.org/trycybsi/) environment.  The basic premise is that students are able to spin up a multi-pod sandbox environment to be able to walk through various security concepts.  The primary environment will be Jupyter, but we need to be able to launch companion pods with each notebook environment.  For example, in the [ArpSpoof](http://try.cybersecurity.ieee.org/trycybsi/help/arpspoof) case we'd need the "hacker" environment 
 (Jupyter) alongside "victim" (client, maybe Jupyter) and "server" (webserver) environments. These would need to be separate Pods with a common network/security policy, since it wouldn't be interesting to do this on locahost (i.e., a single pod). I've seen the `hub.extraContainers` configuration and mention of potentially allowing users to "[do additional things in their namespaces (like spawning other clusters)](https://github.com/jupyterhub/kubespawner/issues/76).  Is this type of scenario Z2JH can support (or would be interested in supporting, if we were to work on a way)?

chaoleili:
Support config singleuser extraContainer in helm chart.
If config extraContainer. image-puller will also pull extraContainer
image.

Fixes #713 
Fixes #830


jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -68,6 +68,15 @@ spec:
...
         {{- end }}
+        {{- range $k, $v := .Values.singleuser.extraContainers }}
+        - name: image-pull-singleuser-extra-container-{{ $k }}
+          image: {{ $v.image }}
+          imagePullPolicy: IfNotPresent
+          command:
+            - /bin/sh
+            - -c
+            - echo "Pulling complete"
+        {{- end }}
```
consideRatio: >
```yaml
image: {{ $v.image }}
```

I wonder about this. Can you add an demonstrative value for `singleuser.extraContainers` to the config we use to lint and verify consistency in the templates as well? It resides in tools/templates.

consideRatio: Note that you can use the linter yourself without needing to use the CI by running it from tools/templates/lint-and-validate.py (if i recall the path to it correctly)

images/hub/jupyterhub_config.py
```diff
@@ -129,6 +129,10 @@
 if init_containers:
     c.KubeSpawner.init_containers.extend(init_containers)
 
+extra_containers = get_config('singleuser.extra-containers')
+if extra_containers:
+    c.KubeSpawner.extra_containers.extend(extra_containers)
+
```
consideRatio: Thanks for positioning this code addition in a very logically consistent place within the config =D just below init-containers like in other places.

jupyterhub/values.yaml
```diff
@@ -162,6 +162,7 @@ singleuser:
   extraEnv: {}
   lifecycleHooks:
   initContainers:
+  extraContainers:
```
consideRatio: I think both `extraContainers` and `initContainers` should be declared to be `[]` (still Falsy) by default in order to avoid some coalesce warnings by Helm when Helm coalesces the provided value sources.

consideRatio: Feel free to add it to both.

consideRatio:
Fixes #713 #830.

Makes for user pods the same as #294 did for the hub pod.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
chaoleili: I have added a new lint file in tools/templates. `singleuser.extraContainers` can be used as the same with `singleuser.initContainers` or `hub.extraContainers` :)

tools/templates/lint-singleuser-extraContainers.yaml
```diff
@@ -0,0 +1,15 @@
...
+# https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#container-v1-core.
+#
+# For now, the extraContainers' image will be prepull in daemonset. It can also be removed
+# from prepull daemonset if necessary
+
+singleuser:
+  extraContainers:
+  - name: mock-extra-container-name
+    image: mock-extra-container-image-name
+    imagePullPolicy: IfNotPresent
```
consideRatio: This is exactly the kind of stub content I wanted, but could you add it to [`lint-and-validate-values.yaml`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/tools/templates/lint-and-validate-values.yaml) instead? This new file wont be automatically tested by the CI or the `lint-and-validate.py` script, only the one named [`lint-and-validate-values.yaml`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/tools/templates/lint-and-validate-values.yaml) is utilized.

tools/templates/lint-and-validate-values.yaml
```diff
@@ -176,6 +176,10 @@ singleuser:
   extraEnv: {}
   lifecycleHooks:
   initContainers:
+  extraContainers:
+  - name: mock-extra-container-name
+    image: mock-extra-container-image-name
+    imagePullPolicy: IfNotPresent
```
consideRatio: Can you copy and add it for the initContainers as well, and indent the name/image/imagePullPolicy two spaces?

Thank you for hanging in there!

chaoleili: Sorry the indent is my mistake, i'll fix right now and add initContainers as well.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -68,6 +68,15 @@ spec:
             - -c
             - echo "Pulling complete"
         {{- end }}
+        {{- range $k, $container := .Values.singleuser.extraContainers }}
+        - name: image-pull-singleuser-extra-container-{{ $k }}
+          image: {{ $container.image }}
```
consideRatio: Ah yes! This matches the kubernetes container model as it should, thanks for the helpful link in another commit!

tools/templates/lint-and-validate-values.yaml
consideRatio: Wooo this is awesome you are closing two issues and making a PR that I'll get very happy about merging thank you!

consideRatio:
Wieee! Let's see what the build sais now! I'll make a final read-through but I think this is good to go now!

consideRatio:
:tada: :star2: This was awesome @chaoleili :D

chaoleili:
:D Thanks for your review!  🙌

Chrisjw42:
Hi guys,

I've been trying to make use of the extraContainers attribute, but it seems that when I try to define other kubeSpawner paramerets, such as "port", it crashes my deployment every time. 

I am working under the assumption that any kubeSpawner parameter can be passed to the extraContainers attribute, and it would be passed through? I've been able to pass "command" and "arg" through to the container in question.

My main requirement is that I want to be able to specify docker parameters on these extracontainers, for example "-p 3000:3000" and other such docker arguments. Is this possible with extraContainers, or have I misunderstood it's purpose?

consideRatio:
I think i cant help you properly with this question, it is mostly a question about what it would mean in kubernetes to have an extra container within a pod i think.

having another container in a pod is often referred to having a sidecar container ive learned. Exposing ports to a container within a pod is very possible etc, but what it means and who is able to access them is another question.

Useful is to read about the ContainerSpec in the kubernetes API definition, that is what you would actually pass to extraContainers. If you search PodSpec, you can see the containers field within it and the type of info you would need to pass there.

I hope this can guide you further towards your goal.

PR #914, opened 2018-09-06
add pdb, nodeSelector for user-scheduler

minrk:
to allow controlling where it runs, let it be disrupted for scale-down.

Additionally, `minAvailable` for other pdbs is configurable, so that it can be set to 0 to allow disruptions for scale-down.

### DIFF
--- a/jupyterhub/templates/hub/pdb.yaml
+++ b/jupyterhub/templates/hub/pdb.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  minAvailable: {{ .Values.hub.pdb.minAvailable | default 1 }}
+  minAvailable: {{ .Values.hub.pdb.minAvailable }}
   selector:
     matchLabels:
       {{- include "jupyterhub.matchLabels" . | nindent 6 }}
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -50,6 +50,7 @@ hub:
   imagePullPolicy: IfNotPresent
   pdb:
     enabled: true
+    minAvailable: 1
   networkPolicy:
     enabled: false
     egress:
@@ -98,6 +99,7 @@ proxy:
   nodeSelector: {}
   pdb:
     enabled: true
+    minAvailable: 1
   https:
     enabled: true
     type: letsencrypt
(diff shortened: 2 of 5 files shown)


jupyterhub/values.yaml
```diff
@@ -50,12 +50,13 @@ hub:
...
     enabled: true
+    minAvailable: 1
   networkPolicy:
     enabled: false
     egress:
       - to:
-          - ipBlock:
-              cidr: 0.0.0.0/0
+        - ipBlock:
+            cidr: 0.0.0.0/0
```
consideRatio: # TL, DR:
I dislike list indentation matters! Do not spend time thinking on what I write, and try to forgive me for posting it even though I say it is a pointless read ^^ Just feels too bad not to click send after considering this for too long in the past and now.

---

I'm unhappy about options of indentation here, its like choosing between two bad options...

- The change you made is more readable.
- The one we had is the one that matches the linter and what we do throughout the project. If one got to choose and stick with one, this is the one to avoid some issues such as needing to consider what is provided, a list or a dict when using the toYaml function and indenting the content.

---

I think the linter is currently set to `whatever` regarding this.

jupyterhub/values.yaml
```diff
@@ -50,12 +50,13 @@ hub:
   imagePullPolicy: IfNotPresent
   pdb:
     enabled: true
+    minAvailable: 1
```
consideRatio: Perhaps we should add a note that a change here will require a `--force` update on help. You can verify by attempting to `kubectl edit pdb hub` and modify the value. If it is okay then we are good, but if not lets add a note warning about the need for a `--force` on next upgrade.

For the other pdb's as well if so.

jupyterhub/templates/hub/pdb.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  minAvailable: 1
+  minAvailable: {{ .Values.hub.pdb.minAvailable | default 1 }}
```
consideRatio: We are duplicating the default values again and making one pointless, in this case the one here in the template. Let's stick with having them only in values.yaml.

Same for the user-scheduler and proxy pdb.

jupyterhub/values.yaml
minrk: I didn't mean to make this change, forget it was here. This layout looks like a typo, but I understand it now.

PR #915, opened 2018-09-06
Remove deprecated GIT_ environment variables

### LINKED ISSUE #743: Unset GIT_AUTHOR_NAME?
The GIT_AUTHOR_NAME environment variable is set every time I start a terminal in Jupyterlab, and I cannot figure out where it is set, or how to unset it permanently. I would be interested in not having this variable set for any user on the deployment if possible, because it overrides the user name in .gitconfig, and it does not get full/correct names. Why is this variable set for users, and is there any way of disabling this behavior? Thanks!

I am currently using jupyterhub-v0.7-e2757f0.

yuvipanda:
nbgitpuller seems to work without these. I left them in
for an ancient version of Berkeley's DataHub, from which
z2jh was generalized.

Fixes #743

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -243,27 +243,6 @@ set_config_if_not_none(c.OAuthenticator, 'scope', 'auth.scopes')
 
 c.Authenticator.enable_auth_state = get_config('auth.state.enabled', False)
 
-def generate_user_email(spawner):
-    """
-    Used as the EMAIL environment variable
-    """
-    return '{username}@{domain}'.format(
-        username=spawner.user.name, domain=email_domain
-    )
-
-def generate_user_name(spawner):
-    """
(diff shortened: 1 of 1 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -243,27 +243,6 @@
...
-    )
-
-def generate_user_name(spawner):
-    """
-    Used as GIT_AUTHOR_NAME and GIT_COMMITTER_NAME environment variables
-    """
-    return spawner.user.name
-
-c.KubeSpawner.environment = {
-    'EMAIL': generate_user_email,
```
consideRatio: I wonder if someone is using the `EMAIL` env. var for other purposes. Perhaps we should make an entry to the CHANGELOG about to make sure they know whats to expect in between 0.7 and 0.8.

consideRatio: #916

PR #924, opened 2018-09-11
[Part 6 of #758] Storage labels: configurable extras

consideRatio:
## To review
- [**Storage labels: configurable extras**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/924/commits/a95873b17309c5ca71642b68b34fa6a3698e61e1)

## TODO:
- [x] Change to storage.extraLabels

---

## TODO
- [x] Delete storage-kind label, it is not needed with other labels in place such as `component: hub` on the hub-db-dir PVC and `component: singleuser-storage` on the user PVCs.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -65,6 +65,7 @@ for trait, cfg_key in (
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
 
 c.KubeSpawner.image_spec = get_config('singleuser.image-spec')
 # Configure dynamically provisioning pvc
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -171,6 +171,7 @@ singleuser:
   serviceAccountName:
   storage:
     type: dynamic
+    extraLabels: {}
     extraVolumes: []
     extraVolumeMounts: []
     static:
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -208,6 +208,12 @@ data:
     {{- range $key, $value := .Values.singleuser.extraLabels }}
     {{ $key | quote }}: {{ $value | quote }}
     {{- end }}
+  {{- if .Values.singleuser.storage.extraLabels }}
+  singleuser.storage-extra-labels: |
+    {{- range $key, $value := .Values.singleuser.storage.extraLabels }}
+    {{ $key | quote }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
   {{- if .Values.singleuser.extraEnv }}
   singleuser.extra-env: |
     {{- range $key, $value := .Values.singleuser.extraEnv }}
--- a/tools/templates/lint-and-validate-values.yaml
+++ b/tools/templates/lint-and-validate-values.yaml
@@ -172,7 +172,6 @@ singleuser:
                 - 169.254.169.254/32
   events: true
   extraLabels: {}
-  storageExtraLabels: {}
   extraEnv: {}
   lifecycleHooks:
   initContainers:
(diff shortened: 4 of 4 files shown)


jupyterhub/templates/hub/pvc.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: hub-db-dir
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
+    hub.jupyter.org/storage-kind: "core"
```
consideRatio: Note that this can be applied on the PVC without trouble. You may not change the selector field in the spec of the PVC though.

minrk: What does this storage-kind label do?

The extra labels makes sense, but I don't see what is accomplished by the 'storage-kind' label.

consideRatio: The main benefit as I see it, is that having such labels allows us to write easier instructions for various tasks in the future, or simply allow someone to maintain a jupyterhub better thanks to `--selector`.

What would you do if you wanted to delete all PVCs, but NOT the hub-db-dir ? Well enter regexp and scripting currently, but if you have a label on the hub-db-dir you can exclude that with a `--selector`.

So, it is on other words a harmless but potentially useful thing, perhaps it can be useful to us if we support someone solve something regarding this as well.

minrk: If there's a problem that user volumes aren't well labeled, we can address that by making sure they have a good label (is component not an adequate standard label that exists for this purpose?), but I think the kind=core/user distinction is being applied too broadly, and the core label isn't useful here, especially since it's being applied to exactly one already-distinguishible volume.

images/hub/jupyterhub_config.py
```diff
@@ -67,7 +65,12 @@
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
+c.KubeSpawner.storage_extra_labels.update({
+    "hub.jupyter.org/storage-kind": "user"
```
minrk: If there's a default label volumes should have, I think this is an issue for KubeSpawner *itself*, not this chart.

jupyterhub/templates/hub/pvc.yaml
consideRatio: Ah, the ideas of this were introduced before some other labeling improvements, now I see we have the following labels, which i find to be enough:

```yaml
  # on user PVC
  labels:
    app: jupyterhub
    chart: jupyterhub-0.7.0
    component: singleuser-storage
    heritage: jupyterhub
    release: jhub

  # on hub-db-dir PVC
  labels:
    app: jupyterhub
    chart: jupyterhub-0.7.0
    component: hub
    heritage: Tiller
    release: jhub
```

I'll delete the storage-kind stuff.

images/hub/jupyterhub_config.py
consideRatio: Deleting this label as well!

images/hub/jupyterhub_config.py
```diff
@@ -65,6 +65,7 @@
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
```
consideRatio: I deleted another update statement here adding a "storage-kind" label. I squashed the commit after i did a rebase, hmmm... The git history became cleaner but the review got harder. Should I add commits instead?

consideRatio: I also deleted a line in the hub's PVC.

jupyterhub/values.yaml
```diff
@@ -161,6 +161,7 @@ singleuser:
   events: true
   extraAnnotations: {}
   extraLabels: {}
+  storageExtraLabels: {}
```
minrk: Shouldn't this be blow as `storage.extraLabels`?

PR #925, opened 2018-09-11
[Part 7 of #758] Added `singleuser.tolerations`, for node taints

consideRatio:
## To review

- [**Added `singleuser.tolerations`, for node taints**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/925/commits/f243103627235c6a786bcbd1ec64698576bfb368)

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -221,6 +221,9 @@ data:
     {{- end }}
   {{- end }}
 
+  singleuser.tolerations: |
+    {{- include "jupyterhub.userTolerations" . | nindent 4 }}
+
   {{- if .Values.scheduling.userScheduler.enabled }}
   singleuser.scheduler-name: "{{ .Release.Name }}-user-scheduler"
   {{- end }}
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -208,6 +208,12 @@ data:
     {{- range $key, $value := .Values.singleuser.extraLabels }}
     {{ $key | quote }}: {{ $value | quote }}
     {{- end }}
+  {{- if .Values.singleuser.storage.extraLabels }}
+  singleuser.storage-extra-labels: |
+    {{- range $key, $value := .Values.singleuser.storage.extraLabels }}
+    {{ $key | quote }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
   {{- if .Values.singleuser.extraEnv }}
   singleuser.extra-env: |
     {{- range $key, $value := .Values.singleuser.extraEnv }}
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -36,6 +36,8 @@ spec:
         {{- /* Changes here will cause the DaemonSet to restart the pods. */}}
         {{- include "jupyterhub.matchLabels" . | nindent 8 }}
     spec:
+      tolerations:
+        {{- include "jupyterhub.userTolerations" . | nindent 8 }}
       terminationGracePeriodSeconds: 0
       automountServiceAccountToken: false
       {{- if .Values.singleuser.imagePullSecret.enabled }}
--- a/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
+++ b/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
@@ -1,0 +1,17 @@
+{{- /*
+  jupyterhub.userTolerations
+    Lists the tolerations for node taints that the user pods should have
+*/}}
+{{- define "jupyterhub.userTolerations" -}}
+- key: hub.jupyter.org_dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
+- key: hub.jupyter.org/dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
+{{- if .Values.singleuser.extraTolerations }}
+{{- .Values.singleuser.extraTolerations | toYaml | trimSuffix "\n" | nindent 0 }}
+{{- end }}
+{{- end }}
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -65,6 +65,7 @@ for trait, cfg_key in (
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
 
 c.KubeSpawner.image_spec = get_config('singleuser.image-spec')
 # Configure dynamically provisioning pvc
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -68,6 +68,7 @@ for trait, cfg_key in (
 c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
 
 c.KubeSpawner.image_spec = get_config('singleuser.image-spec')
+c.KubeSpawner.tolerations.extend(get_config('singleuser.tolerations', []))
 # Configure dynamically provisioning pvc
 storage_type = get_config('singleuser.storage.type')
 if storage_type == 'dynamic':
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -460,6 +460,22 @@ properties:
         description: |
           Deprecated and no longer does anything. Use the user-scheduler instead
           in order to accomplish a good packing of the user pods.
+      extraTolerations:
+        type: list
+        description: |
+          Tolerations allow a pod to be scheduled on nodes with taints. These
+          are additional tolerations other than the user pods and core pods
+          default ones `hub.jupyter.org/dedicated=user:NoSchedule` or
+          `hub.jupyter.org/dedicated=core:NoSchedule`. Note that a duplicate set
+          of tolerations exist where `/` is replaced with `_` as the Google
+          cloud does not support the character `/` yet in the toleration.
+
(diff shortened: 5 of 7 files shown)


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -46,6 +46,11 @@ spec:
                     - key: release
                       operator: In
                       values: [{{ .Release.Name | quote }}]
+      {{- $_ := merge (dict "podKind" "core") . }}
+      {{- $dummy := include "jupyterhub.prepareScope" $_ }}
+      {{- if $_.hasTolerations }}
+      {{- include "jupyterhub.tolerations" $_ | nindent 6 }}
+      {{- end }}
```
consideRatio: prepareScope prepopulates $_ with various stuff, and then the `jupyterhub.tolerations` template utilizes that. The $_ will be reused by a future `jupyterhub.affinity` template so it makes sense to prepare the scope once instead of twice.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -0,0 +1,47 @@
...
+{{- if eq .podKind "core" -}}
+{{- else if eq .podKind "user" -}}
+- key: hub.jupyter.org_dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
+- key: hub.jupyter.org/dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
```
consideRatio: These are the actual tolerations that are added by default

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -0,0 +1,47 @@
...
+{{- if .Values.singleuser.extraTolerations -}}
+{{- .Values.singleuser.extraTolerations | toYaml | trimSuffix "\n" | nindent 0 }}
+{{- end }}
```
consideRatio: And they are augmented here with additional tolerations configurable by the chart consumer.

minrk:
I think we need to be careful with these helpers. I am finding that they are making it harder to understand and work on the chart. I genuinely can't tell what tolerations are being added to the 'core' pods in this PR, which have all this extra labelling and prepareScope logic. What are the resulting tolerations on each pod?

In particular, I'm interested in what this patch would look like if we:

- didn't have the prepareScope or tolerations helpers
- had no `pod-kind` label

I think we should try to minimize new helpers in the future, only adding them where they really are solving a problem. If we had a simple helper that was a macro for "tolerate-user-node-taint", with no other logic dependent on the parent scope, I think this might be simpler to write and understand.

consideRatio:
Yes I certainly agree, it is a mess to grasp this code. The big crux is to have something DRY used by both the kubespawner configured in jupyterhub_config.py as well as resources created by helm templates.

The key solution to avoiding complexity would be to have the kubespawner not be configured by jupyterhub_config.py, but by a configmap or similar directly which it observed.

---

This PR could be simpler by not having the preparescope etc, but the future PR with affinities and presets for configuring affinities.

---

In order for configurable tolerations to be useful at the same time as supporting for example the image pullers, they must be configured with the same tolerations also, and hence we have the helm templates + jupyterhub_config.py configuration from the same source challenge. Also, by having presets the challenge becomes a bit harder.

consideRatio:
> I think we should try to minimize new helpers in the future, only adding them where they really are solving a problem. If we had a simple helper that was a macro for "tolerate-user-node-taint", with no other logic dependent on the parent scope, I think this might be simpler to write and understand.

:+1:

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -0,0 +1,47 @@
...
+{{- $dummy := set . "hasTolerations" .tolerations }}
+{{- end }}
+
+
+
+{{- /*
+  jupyterhub.tolerations
+    Refactors the setting of the user pods tolerations field.
+*/}}
+{{- define "jupyterhub.tolerations" -}}
```
minrk: I think this helper is at least one too far. The define/end is more than the code it's deduplicating by being reused.

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -23,6 +23,11 @@ spec:
       {{- if .Values.rbac.enabled }}
       serviceAccountName: user-scheduler
       {{- end }}
+      {{- $_ := merge (dict "podKind" "core") . }}
```
minrk: These lines ` {{- $_ := merge (dict "podKind" "core") . }}` are I think the hardest to follow, and make contributing to/understanding the chart very difficult. I genuinely don't know what this means or how to use it, so I can't really evaluate whether this is the right thing to do and if/when we might need to do the same or similar thing in other templates.

There's now a lot of hidden logic that is several layers deep.

If, instead, this were a single:

```yaml
{{- include "jupyterhub.tolerations" "core" | nindent 6 }}
```

I feel like that would be simpler and easier to grok as "include the tolerations for a core pod". Is that achievable?

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -0,0 +1,47 @@
...
+    below.
+*/}}
+{{- define "jupyterhub.prepareScope" -}}
+{{- /* Update the copied scope */}}
+{{- $dummy := set . "component" (include "jupyterhub.componentLabel" .) }}
+
+{{- /* Fetch relevant information */}}
+{{- $dummy := set . "tolerationsList" (include "jupyterhub.tolerationsList" .) }}
+{{- $dummy := set . "tolerations" (include "jupyterhub.tolerations" .) }}
+{{- $dummy := set . "hasTolerations" .tolerations }}
```
minrk: Let's try to do this without prepareScope. I think helpers that setup the state for other helpers, rather than being self-contained is something we should try to avoid.

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
consideRatio: Perhaps we can reduce the complexity slightly, I dare not prove a negative, but I have found it very challenging. The merge thing could be simplified by using one or more `set` statements at least.

The reason for the merge, is that you can only pass one object to a helper, and the helper needs a lof of stuff that exists within the default scope referred to as `.` --- By augmenting the default scope, we would change a state globally, which also isnt great. What I have done here is to create a new variable called $_ that creates a new scope by creating a new dictionary, and merging the current scope to it.

https://gitlab.com/charts/gitlab/blob/master/doc/development/README.md#passing-variables-between-control-structures

---

```
{{- include "jupyterhub.tolerations" "core" | nindent 6 }}
```

This helper would not be aware of what `.Release.Name` was if it isnt passed the `.` scope or something that got the `.` scope merged with it. If in the scope of the tolerations helper one would reference `.` it would get `"core"` only. This is sometimes enough for certain helpers, but tend not to be for a lot of helpers.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
consideRatio: If you would add

```yaml
tolerations:
```
without any entries within it, you get invalid YAML. To know if you should add the `tolerations:` entry or not, you must first figure out if you have content within it.

`tolerationsList` is the content for the `tolerations`. It would have been a better name with `tolerationsContent`.

A lot of the struggles I've faced relates to "should i add this or not, do i have content for the field or not?", the prepareScope helper takes care of a lot of those questions.

I have tried to make do with less helpers and It is really hard... =/ I could try more and more, but I'm currently out of ideas on how to solve the challenges. I'll try address things better in a single point to avoid spreading discussion.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
consideRatio: Will attempt to address concerns at a single discussion entry, not in a review comment.

consideRatio:
The pod-kind labels are not needed for any logic anymore, they were needed for #930.

I got confused thinking they were needed at this stage.

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -221,12 +221,8 @@ data:
     {{- end }}
   {{- end }}
 
-  {{- $_ := merge (dict "podKind" "user") . }}
-  {{- include "jupyterhub.prepareScope" $_ }}
-  {{- if $_.hasTolerations }}
   singleuser.tolerations: |
-    {{- $_.tolerationsList | nindent 4 }}
-  {{- end }}
+    {{- include "jupyterhub.userTolerations" | nindent 4 }}
```
consideRatio: We can assume userTolerations has content, we cant with the affinity later though...

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -45,11 +45,6 @@ spec:
                     - key: release
                       operator: In
                       values: [{{ .Release.Name | quote }}]
-      {{- $_ := merge (dict "podKind" "core") . }}
-      {{- $dummy := include "jupyterhub.prepareScope" $_ }}
-      {{- if $_.hasTolerations }}
-      {{- include "jupyterhub.tolerations" $_ | nindent 6 }}
-      {{- end }}
```
consideRatio: There was a hatch for tolerations on core nodes, but it was not adding anything by default anyhow.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -36,11 +36,8 @@ spec:
         {{- /* Changes here will cause the DaemonSet to restart the pods. */}}
         {{- include "jupyterhub.matchLabels" . | nindent 8 }}
     spec:
-      {{- $_ := merge (dict "podKind" "user" "daemonSet" true) . }}
-      {{- $dummy := include "jupyterhub.prepareScope" $_ }}
-      {{- if $_.hasTolerations }}
-      {{- include "jupyterhub.tolerations" $_ | nindent 6 }}
-      {{- end }}
+      tolerations:
+        {{- include "jupyterhub.userTolerations" . | nindent 8 }}
```
consideRatio: The image puller also needs the same tolerations

PR #926, opened 2018-09-11
[Part 8 of #758] Added singleuser affinity configuration

consideRatio:
## To review

- [**Added `singleuser.` node/pod/podAnti affinity**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/926/commits/a3fbb8f788a87bdb2fceb4da5b9e3fd2e105414f)

### DIFF
--- a/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
+++ b/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
@@ -1,0 +1,17 @@
+{{- /*
+  jupyterhub.userTolerations
+    Lists the tolerations for node taints that the user pods should have
+*/}}
+{{- define "jupyterhub.userTolerations" -}}
+- key: hub.jupyter.org_dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
+- key: hub.jupyter.org/dedicated
+  operator: Equal
+  value: user
+  effect: NoSchedule
+{{- if .Values.singleuser.extraTolerations }}
+{{- .Values.singleuser.extraTolerations | toYaml | trimSuffix "\n" | nindent 0 }}
+{{- end }}
+{{- end }}
--- a/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
+++ b/jupyterhub/templates/scheduling/_scheduling-helpers.tpl
@@ -15,3 +15,41 @@
 {{- .Values.singleuser.extraTolerations | toYaml | trimSuffix "\n" | nindent 0 }}
 {{- end }}
 {{- end }}
+
+
+
+{{- define "jupyterhub.userNodeAffinityRequired" -}}
+{{- if .Values.singleuser.extraNodeAffinity.required -}}
+{{ .Values.singleuser.extraNodeAffinity.required | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
+
+{{- define "jupyterhub.userNodeAffinityPreferred" -}}
+{{- if .Values.singleuser.extraNodeAffinity.preferred -}}
+{{ .Values.singleuser.extraNodeAffinity.preferred | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
+
+{{- define "jupyterhub.userPodAffinityRequired" -}}
+{{- if .Values.singleuser.extraPodAffinity.required -}}
+{{ .Values.singleuser.extraPodAffinity.required | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
+
+{{- define "jupyterhub.userPodAffinityPreferred" -}}
+{{- if .Values.singleuser.extraPodAffinity.preferred -}}
+{{ .Values.singleuser.extraPodAffinity.preferred | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
+
+{{- define "jupyterhub.userPodAntiAffinityRequired" -}}
+{{- if .Values.singleuser.extraPodAntiAffinity.required -}}
+{{ .Values.singleuser.extraPodAntiAffinity.required | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
+
+{{- define "jupyterhub.userPodAntiAffinityPreferred" -}}
+{{- if .Values.singleuser.extraPodAntiAffinity.preferred -}}
+{{ .Values.singleuser.extraPodAntiAffinity.preferred | toYaml | trimSuffix "\n" }}
+{{- end }}
+{{- end }}
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -143,6 +143,16 @@ auth:
 
 singleuser:
   extraTolerations: []
+  nodeSelector: {}
+  extraNodeAffinity:
+    required: []
+    preferred: []
+  extraPodAffinity:
+    required: []
+    preferred: []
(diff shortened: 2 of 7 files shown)


jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -14,6 +14,17 @@
...
+{{- $dummy := set . "nodeAffinityRequired" (include "jupyterhub.nodeAffinityRequired" .) }}
+{{- $dummy := set . "podAffinityRequired" (include "jupyterhub.podAffinityRequired" .) }}
+{{- $dummy := set . "podAntiAffinityRequired" (include "jupyterhub.podAntiAffinityRequired" .) }}
+{{- $dummy := set . "nodeAffinityPreferred" (include "jupyterhub.nodeAffinityPreferred" .) }}
+{{- $dummy := set . "podAffinityPreferred" (include "jupyterhub.podAffinityPreferred" .) }}
+{{- $dummy := set . "podAntiAffinityPreferred" (include "jupyterhub.podAntiAffinityPreferred" .) }}
+{{- $dummy := set . "hasNodeAffinity" (or .nodeAffinityRequired .nodeAffinityPreferred) }}
+{{- $dummy := set . "hasPodAffinity" (or .podAffinityRequired .podAffinityPreferred) }}
+{{- $dummy := set . "hasPodAntiAffinity" (or .podAntiAffinityRequired .podAntiAffinityPreferred) }}
+{{- $dummy := set . "hasAffinity" (or .hasNodeAffinity .hasPodAffinity .hasPodAntiAffinity) }}
```
consideRatio: I need a lot of helpers, some that contains the output at a level where I can augment it with other default output from affinity presets in future PRs.

jupyterhub/values.yaml
```diff
@@ -143,6 +143,16 @@ auth:
 
 singleuser:
   extraTolerations: []
+  nodeSelector: {}
+  extraNodeAffinity:
```
minrk: Instead of exposing six different configurables, can we expose the Kubernetes `.affinity` structure directly? I think that would be a much better and better-documented experience.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -45,3 +56,103 @@ tolerations:
...
+{{- end }}
+{{- end }}
+{{- end }}
+
+
+
+{{- /*
+  input: podKind
+*/}}
+{{- define "jupyterhub.affinity" -}}
```
minrk: This helper seems to be defined but unused

jupyterhub/values.yaml
```diff
@@ -142,6 +142,17 @@ auth:
...
+    required: []
+    preferred: []
+  extraPodAffinity:
+    required: []
+    preferred: []
+  extraPodAntiAffinity:
+    required: []
+    preferred: []
```
minrk: Can we make this expose the Kubernetes `affinity` structure directly, instead of exposing six different configurables? It seems that would be a lot simpler, both for us to maintain and for users to understand. In general, let's try *very hard* to avoid using anything but Kubernetes' own names and structures for things.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
```diff
@@ -0,0 +1,158 @@
...
+{{- $dummy := set . "nodeAffinityRequired" (include "jupyterhub.nodeAffinityRequired" .) }}
+{{- $dummy := set . "podAffinityRequired" (include "jupyterhub.podAffinityRequired" .) }}
+{{- $dummy := set . "podAntiAffinityRequired" (include "jupyterhub.podAntiAffinityRequired" .) }}
+{{- $dummy := set . "nodeAffinityPreferred" (include "jupyterhub.nodeAffinityPreferred" .) }}
+{{- $dummy := set . "podAffinityPreferred" (include "jupyterhub.podAffinityPreferred" .) }}
+{{- $dummy := set . "podAntiAffinityPreferred" (include "jupyterhub.podAntiAffinityPreferred" .) }}
+{{- $dummy := set . "hasNodeAffinity" (or .nodeAffinityRequired .nodeAffinityPreferred) }}
+{{- $dummy := set . "hasPodAffinity" (or .podAffinityRequired .podAffinityPreferred) }}
+{{- $dummy := set . "hasPodAntiAffinity" (or .podAntiAffinityRequired .podAntiAffinityPreferred) }}
+{{- $dummy := set . "hasAffinity" (or .hasNodeAffinity .hasPodAffinity .hasPodAntiAffinity) }}
```
minrk: All these helpers seem to include exactly one thing and be used in exactly one place. It seems like there are too many layers of abstraction, and this would be simplified with much fewer helpers.

jupyterhub/values.yaml
consideRatio: We want the affinity to be utilized by the image puller as well, which is configured directly from a heml template. But we only want the daemonset to use the affinity **partially**, for example the daemonset should not care about `affinity.podAffinity` or but should care about `affinity.nodeAffinity.requiredDuring...`.

Attempting to extract pieces from a big dict passed to a value in config.yaml is not plausible even though I can imagine someone have made some super complicated trick to accomplish.

consideRatio: Then doing so would also rule out having a configurable preset of `require` `prefer` `ignore` matching node affinity for user / core pods with user / core pools for example.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
consideRatio: In this PR some of them may be underutilized to motivate their existence, but they were refactored in this way to reduce complexity in the end. It is a recurring issue with complexity, and I deeply agree. I have struggled a lot with the Helm charts to make things work in a manner that isn't breaking DRY in a big way or simply working around challenges with scope.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
consideRatio: Yes you are right, this PR only contained affinities required for configuring the user-related affinities, and this helper was not needed by the daemonset which only requires a subset of the affinity helpers introduced.

jupyterhub/templates/scheduling/_scheduling-helpers.tpl
minrk: DRY isn't the most important thing in general. Every design choice is a trade-off, so we have to answer the question: is the redundancy/complexity of having some duplicated logic more of a problem than the complexity of added abstractions needed to avoid repetition. In this case, as far as I can tell, the duplication is less costly than the abstraction.

jupyterhub/values.yaml
minrk: We're talking about one level of a dict `affinity.nodeAffinity`, aren't we? Isn't that less complex than what's happening here, where we are mapping several levels of a dict to different things?

jupyterhub/values.yaml
consideRatio: So I have considered this *in depth* now, and the conclusion is no, not with Helm 2, in the values section. If we do, we wont be able to augment our configurable presets within the templates.

But, in kubespawner, that is configured through python where we have great capabilities of working with the stuff coming from the configmap, we could combine them again and set one big affinity traitlet on kubespawner.

I was confused about this until recently. In summary, this PR must be like this, but the previous one for kubespawner did not have to be that way =/

PR #927, opened 2018-09-11
[Part 9 of #758] Preset: `scheduling.nodeAffinity.matchNodePurpose`

consideRatio:
## To review

- [**Preset: `scheduling.nodeAffinity.matchNodePurpose`**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/927/commits/972af4086517c92389ef7fc5c35326ed57fd3833)

--- 

## About

This preset makes pods of `user | core` kind either `prefer` `require` or `ignore` a scheduling on a node with a matching label for `hub.jupyter.org/node-purpose: user | core`.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -65,6 +65,7 @@ for trait, cfg_key in (
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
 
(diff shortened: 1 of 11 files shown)


consideRatio:
This PR is very bugged... Apparently everything is up to date even though i try force push crap content.

consideRatio:
>  Rewriting your Git commit history while performing rebases alters the space-time continuum, which means that commits may not be represented the way you expect them to in the GitHub

> However, rest assured that nothing is broken when you see things outside of a chronological order!

- https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

All good then! Just make sure to ignore the commit order as viewed in GitHub.

PR #928, opened 2018-09-11
[Part 10 of #758] Added `scheduling.podPriority`

consideRatio:
## To review

- [**Added `scheduling.podPriority` option**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/928/commits/e774b7dd271118fd5ee2d060fa8b1fccd7bfa0ab)

---

## TODO
- [x] Make `{{- if .Values.scheduling.podPriority }}` be `{{- if .Values.scheduling.podPriority.enabled }}` - thanks @minrk

## About

Pod priority requires k8s 1.11 and is a foundational functionality in order for us to use user-placeholder pods to create a headroom on nodes. This is also referred to as over provisioning. They can with a lower priority be evicted when a real user needs a place to be scheduled but cant fit on a node with them there, but at the same time when they are left unscheduled, will cause a node scaleup if an autoscaler is enabled.

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -224,6 +224,31 @@ data:
   singleuser.tolerations: |
     {{- include "jupyterhub.userTolerations" . | nindent 4 }}
 
+  {{- if include "jupyterhub.userNodeAffinityRequired" . }}
+  singleuser.node-affinity-required: |
+    {{- include "jupyterhub.userNodeAffinityRequired" . | nindent 4 }}
+  {{- end }}
+  {{- if include "jupyterhub.userNodeAffinityPreferred" . }}
+  singleuser.node-affinity-preferred: |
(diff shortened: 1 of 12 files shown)


jupyterhub/templates/hub/configmap.yaml
```diff
@@ -255,6 +255,9 @@ data:
   {{- if .Values.scheduling.userScheduler.enabled }}
   singleuser.scheduler-name: "{{ .Release.Name }}-user-scheduler"
   {{- end }}
+  {{- if .Values.scheduling.podPriority }}
```
minrk: .podPriority.enabled?

consideRatio: Woo this have been truthy during testing for sure already. I think it triggered a silent failure on the kubernetes API server that simply stripped it away when no such priorityClass was found. Normally it sets the `priority` field after reading the priorityClassName, and you cannot modify that in retrospect without recreating the pod.

PR #929, opened 2018-09-11
[Part 11 of #758]  Added `scheduling.userPlaceholder`

consideRatio:
## To review

- [**Added `scheduling.userPlaceholder` option**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/929/commits/ef102f8a65e0833bcfbc5ed827976b1c9721ef82)
- [**Refactor: relax linting and simplify code**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/929/commits/bf9a72d3f81e4508ad93c3e781171f895672adcd)

---

## About

Read about them in schema.yaml please, whats written here propagates to the *configuration reference* in the guide.

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -222,11 +222,7 @@ requests:
   {{- end }}
 {{- end }}
 
-{{- if and $r $l }}
-{{- println }}
-{{- end }}
-
-{{- if $l -}}
+{{- if $l }}
 limits:
   {{- if $l1 }}
   cpu: {{ .Values.singleuser.cpu.limit }}
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -192,3 +192,52 @@ component: {{ include "jupyterhub.componentLabel" . }}
 }
 {{- end }}
 {{- end }}
+
+
+{{- /*
+  jupyterhub.resources:
+    The resource request of a singleuser.
+*/}}
+{{- define "jupyterhub.resources" -}}
+{{- $r1 := .Values.singleuser.cpu.guarantee -}}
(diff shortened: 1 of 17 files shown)


jupyterhub/templates/_helpers.tpl
```diff
@@ -192,3 +192,62 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
+  {{ $key | quote }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{- /*
+A custom resource request.
+*/}}
+{{- define "jupyterhub.resources" -}}
```
minrk: I think this is a helpful rule of thumb: never make a helper that's only used once. This appears to only be used in one place. Even if you think it *might* be used elsewhere, the time to refactor is when you want to reuse it, not before.

consideRatio: :+1: 

It will be used in the user-dummy as well. Needing to make these helpers have made me cringe in general. I wonder if k-sonnet allows for better handling of combining dict values provided by the user than helm.

jupyterhub/templates/_helpers.tpl
```diff
@@ -192,3 +192,52 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
+  {{- if $r3 }}
+  {{- range $key, $value := .Values.singleuser.extraResource.guarantees }}
+  {{ $key | quote }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
+{{- end }}
+
+{{- if and $r $l }}
+{{- println }}
+{{- end }}
```
consideRatio: It is valid, so lets not make the stuff more complicated.

consideRatio: This is now removed in the added commit to this PR.

PR #931, opened 2018-09-11
[Part 13 of #758] Added `scheduling.userDummy`

consideRatio:
## Warning

Needs rebase before merge. The actual content of this PR is actually only the following commit.

- [**Added `scheduling.userDummy` option**](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/931/commits/6da28d6908ddcc656174f291aa954341112a1463)

---

## About

Read about it in schema.yaml.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -65,6 +65,10 @@ for trait, cfg_key in (
     ('node_selector', 'node-selector'),
 ):
     set_config_if_not_none(c.KubeSpawner, trait, 'singleuser.' + cfg_key)
+c.KubeSpawner.storage_extra_labels = get_config('singleuser.storage-extra-labels', {})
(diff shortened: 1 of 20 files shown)


consideRatio:
It is to mimic a real user as compared to a placeholder that has lower priority. If we as developers or a person setting up autoscaling want to verify eviction logic works as intended, I think we should have this, if not sure lets delete it.

This could allow for quite simple automated tests that takes very short time to execute.

minrk:
From my reading of #929, placeholders can be used to do the same thing. Is that not true?

minrk:
I think it makes the most sense to move this to a documentation/example rather than a utility we deploy.

consideRatio:
@minrk I think should be available as a zero replica controller, or dropped without regret entirely and only documented as a closed PR. Closing!

PR #936, opened 2018-09-13
allow specifying scheduler resources in values.yaml

minrk:


jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -45,8 +45,5 @@ spec:
             httpGet:
               path: /healthz
               port: 10251
-          resources:
-            requests:
-              cpu: 50m
-              memory: 512Mi
+          resources: {{ toJson .Values.scheduling.userScheduler.resources }}
```
consideRatio: The following would work:
1. Update `tools/templates/yamllint-config.yaml` `commas.min-spaces-after` to `0` from `1`. That helps as the current output is:
   ```
             resources: {"requests":{"cpu":"50m","memory":"256Mi"}}
   ```

2. Change to the following in the template:
   ```yaml
             resources:
               {{- .Values.scheduling.userScheduler.resources | toYaml | trimSuffix "\n" | nindent 12 }}
   ```

---

We have used the second approach for resources so far, let's stick with it for now.

minrk: Thanks! I think we should indeed relax the linter quite a bit

consideRatio: I agree =)

PR #938, opened 2018-09-15
user-scheduler: image locality priority added

consideRatio:
This may be useful for mybinder.org's deployment. I've add a small image locality consideration into the mix for how the user-scheduler decides where a user pod should schedule. It will still prioritize the most requested node though.

In practice I suspect it may be of minor importance, but I imagine that when a few nodes are almost full, it may be that these are both at the same requested resource score (from 0-10), and then image locality can come into play to decide the final reason for scheduling.

---

I've also ensure that the weights are causing the final node-rating to be decided in order from the most important thing to the least important thing. For example, `InterPodAffinityPriority` always wins over `MostRequestedPriority` and that in turn always win over `ImageLocalityPriority`.

---

Setting `alwaysCheckAllPredicates` to false again. It was only set to true for my own debugging / learning purposes initially. It will improve the schedulers performance to have to false without any drawbacks though.

---

Removing the pod-kind label as well, as I figure we won't add those systematically.

### DIFF
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -15,7 +15,6 @@ spec:
       labels:
         {{- /* Changes here will cause the Deployment to restart the pods. */}}
         {{- include "jupyterhub.matchLabels" . | nindent 8 }}
-        hub.jupyter.org/pod-kind: "core"
       annotations:
         # This lets us autorestart when the configmap changes!
         checksum/config-map: {{ include (print $.Template.BasePath "/scheduling/user-scheduler/configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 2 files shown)


jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -15,7 +15,6 @@ spec:
       labels:
         {{- /* Changes here will cause the Deployment to restart the pods. */}}
         {{- include "jupyterhub.matchLabels" . | nindent 8 }}
-        hub.jupyter.org/pod-kind: "core"
```
minrk: The scheduler is a core pod, no?

consideRatio: Yes, but as these labels never were utilized, I removed them

PR #939, opened 2018-09-16
[Final part of #758] jupyterhub_config.py organized for readability

consideRatio:
## UPDATE
Closed, see: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/939#discussion_r219268087

## To review
Review only `jupyterhub_config.py: default values are set in values.yaml` and the ones after it within this PR. NOTE: Ordering of commits becamse wacky because of rebase operations made GitHub unhappy, it is only in their web UI though.

## TODO
- [x] Cast `concurrent_spawn_limit` to an int.

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -6,6 +6,28 @@ from kubernetes import client
 from z2jh import get_config, get_secret, set_config_if_not_none
 from jupyterhub.utils import url_path_join
 
+# About traitlets
+# ------------------------------------------------------------------------------
+# In order to understand how this configuration works, it is useful to read the
+# traitlets documentation: https://traitlets.readthedocs.io/en/stable/config.html
+#
+# A Configuration object (traitlets.config.Config) is made available as `c` in
+# this file.
+#
+### Lazy evaluation and default value initialization
+# If you would print c.KubeSpawner.common_labels now within this file, you would
+# get...
+# <traitlets.config.loader.LazyConfigValue object at 0x7f1c53071a20>
+#
+# But if you would run the following...
+# c.KubeSpawner.common_labels.update({'heritage': 'something-new'})
+#
+# You would be fine, and end up with...
+# {'app': 'jupyterhub', 'heritage': 'something-new'}
+#
+# Where these labels had defaults set in c.KubeSpawner.common_labels but you
+# overrode the second labels value.
+
 # Network related
 # ------------------------------------------------------------------------------
 # Configure JupyterHub to use the curl backend for making HTTP requests,
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -6,18 +6,20 @@ from kubernetes import client
 from z2jh import get_config, get_secret, set_config_if_not_none
 from jupyterhub.utils import url_path_join
 
+# Network related
+# ------------------------------------------------------------------------------
 # Configure JupyterHub to use the curl backend for making HTTP requests,
 # rather than the pure-python implementations. The default one starts
 # being too slow to make a large number of requests to the proxy API
 # at the rate required.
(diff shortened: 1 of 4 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -29,119 +53,130 @@
...
-    ('concurrent_spawn_limit', 'concurrent-spawn-limit'),
-    # Max number of consecutive failures before the Hub restarts itself
-    # requires jupyterhub 0.9.2
-    ('consecutive_failure_limit', 'consecutive-failure-limit'),
-    # Max number of servers to be running at one time
-    ('active_server_limit', 'active-server-limit'),
-):
-    set_config_if_not_none(c.JupyterHub, trait, 'hub.' + cfg_key)
+# Max number of servers that can be spawning at any one time
+c.JupyterHub.concurrent_spawn_limit = get_config('hub.concurrent-spawn-limit')
```
minrk: This needs to not revert all of the `set_config_if_not_none` calls. Those were all needed to make this robust to missing values in the configmap.

consideRatio: I've considered this, the `set_config_if_not_none` call remains for all non-blank defaulted `string` values still, but not for `bools` or `ints` where we have a defined default in `values.yaml` already, which we have done for all such values.

Before I made the commit I deleted a comment on each line get_config() is called (except the auth section for each specific provider), I verified there was a set default in values.yaml for each of them.

Speaking of `ints`, `concurrent_spawn_limit` should be casted to an `int`.

minrk: I don't think that's accurate. The point of the `set_config_if_not_none` is that the exact image version and chart versions can drift a bit, e.g. when using a custom hub image. If a value is unspecified in values.yaml or the configmap for any reason, it should remain unspecified in the jupyterhub configuration, regardless of type. This was the point of #856, which this PR mostly reverts, restoring the issues #856 fixed. So as much as possible, *no* values should be assumed present in the configmap.

consideRatio: ## TL, DR
I conclude I agree with you ;D I will ditch this PR and extract some parts that are still relevant as a new PR. The parts of relevance are only few since we have #941 coming, but for example the casting of some value to an int.

## I need to think this through a bit more, I'll think aloud!
* The image used by pods, or in general the values provided by the user: as they are values provided by the user, they are out of our control and could diverge from the Chart. We cannot support someone running JupyterHub 0.1 with our chart though.
* Helm chart templates and Helm chart default values: they are declared in the chart, and will always go together.

* Default values can be found in many places, which to me is not desirable. We have had an can have a default value in:
  - values.yaml (in repo)
  - jupyterhub_config.py (in repo)
  - jupyterhub, kubespawner or other configurable component (outside repo, outside control)

It would be nice to have things centralized if possible, it would force us to use `values.yaml` over `jupyterhub_config.py` as plenty of values to be consumed in the `jupyterhub_config.py` also needs to be consumed by helm templates, but only by having them in `values.yaml` they can propegate to both the Helm templates and `jupyterhub_config.py`. Conclusion, if centralizing defaults within the repo, do it in `values.yaml`.

We may decide to honor the default values declared in components like jupyterhub or kubespawners traitlets if the user does not want to override it. To honor kubespawners defaults one must not provide a default at all in the chart, neither in `values.yaml` or `jupyterhub_config.py`.

By googling a bit I just learned that if a user overrides a value like this in `config.yaml`:
```yaml
# this will unset singleuser:
# it will not hold the null value
# it will simply disappear
singleuser: null
```

With this background, I read your statement as we should honor all the kubespawners defaults if no default is provided to us, and since it is possible to override the default values to not even exist, I think it can make sense because then the user have a way to utilize the components default rather than the charts default.

So, we can allow the user to choose:
1. Use the chart defaults
2. Use the components defaults (by setting the value `null`)
3. Use its own values

PR #941, opened 2018-09-18
WIP: passthrough values.yaml

minrk:
opened for discussion of #341

passthrough values as-is as a single configmap entry, rather than specifying a mapping of values -> configmap -> jupyterhub config

secret fields are omitted from the configmap and only secret fields are included in the secret

both secret and configmap have the same structure, the only difference is in the mutually-exclusive filtering of which fields are included.

As it stands, the only manual maintenance required of the configmap should be when new secrets are added, because they must be excluded explicitly from the configmap.

The fiddly bits here are that we need to specifically locate which fields are secret to exclude them from the configmap and include them in the secret. Fortunately, secret values change a lot less often than other values.

An alternative is to restructure values.yaml so that there is a top-level `secrets` key. This would be a lot more convenient for us, since the splitting logic would be clear and stable, but it would be less convenient for users, who would need to have:

```yaml
secrets:
  proxy:
    secretToken: ...
proxy:
  otherConfig
```

It would make much less of a difference for users who are maintaining a best-practices two-file `{config|secret}.yaml` deployment, since the top-level `secrets` key would only mean an increase of indentation by one level in the `secret.yaml` file.

I haven't implemented the hub-side of the configuration, but the yaml->hub config transform now entirely belongs in the hub image's z2jh.py.


consideRatio:
:heart:

Excellent use of `omit`, great start to something.

The values.yaml file will come along even in Helm 3 I figure, having the secrets on a separate path could make it easier for users to extract them to a separate file as well, I like that.

yuvipanda:
I love it!

I think we should err on the side of 'makes more sense for users' and keep the structure of explicitly listing secret paths, rather than exposing that internal difference to the users.

yuvipanda:
I'm curious how we can do this in short PRs that can be merged easily.

minrk:
> I'm curious how we can do this in short PRs that can be merged easily.

I think the shortest first one is to do this passthrough and update jupyterhub_config.py to take config from values layout instead of the configmap layout.

Then we can do the `tljh.py` configurator refactor to clean it up / make it more testable in a later step.

images/hub/jupyterhub_config.py
```diff
@@ -50,42 +76,106 @@
...
+    tag = get_config("singleuser.image.tag")
+    if tag:
+        image = "{}:{}".format(image, tag)
+
+    c.KubeSpawner.image_spec = image
+
+if get_config('singleuser.imagePullSecret.enabled'):
+    c.KubeSpawner.image_pull_secrets = 'singleuser-image-credentials'
+
+# scheduling:
```
minrk: Because there is no longer any logic transforming .values before it arrives in the image,
the affinity helpers are no longer used for 'real' users, only the user-placeholder. The logic is the same, though, and much easier to follow/test here in Python.

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -5,312 +5,18 @@ metadata:
...
+{{- $_ := set $values.proxy "https" (omit $values.proxy.https "manual") -}}
+{{- $_ := set $values "hub" (omit $values.hub "cookieSecret") -}}
+{{- /* make a copy of values.auth to avoid modifying the original */ -}}
+{{- $_ := set $values "auth" (merge dict .Values.auth) -}}
+{{- $_ := set $values.auth "state" (omit $values.auth.state "cryptoKey") -}}
+{{- range $key, $service := $values.hub.services -}}
+  {{- if $service.apiToken -}}
+    {{- $_ := set $values.hub.services $key (omit $service "apiToken") -}}
+  {{- end -}}
+{{- end }}
```
consideRatio: Remove all right whitespace chomping here, if not you will have invalid output because everything will glue together on one line.

My practice throughout so far is to have right whitespace chomping on all statements until the first output in a .yaml file or helper, then only left whitespace chomping.

In this PR TODO:
- From line 8-19, delete the right whitespace chomping.

For me TODO:
- Make a templates styleguide

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -5,312 +5,18 @@ metadata:
...
+  values.yaml: |
+    {{- $values | toYaml | nindent 4 }}
```
consideRatio: We are always using  `| trimSuffix "\n"` in the current templates after `toYaml` to avoid an extra new line, we could relax the linter here and stop doing so, but this isn't only a linting issue, but mostly is.

The new line generated, and the indentation from the `nindent` following it, is generated content, any whitespace chomping left you attempt to do will make you end up with that indentation. I recall that has been an issue, that is why I was frustrated enough to add the ugly practice of `trimSuffix "\n"` after `toYaml` statements. I'm not confident this is still an issue.

For this PR TODO:
- Pipe `toYaml` into `trimSuffix "\n"`

For us TODO in future:
- See if we can clean up our code from needing to use `trimSuffix "\n"` by ignoring the linter without getting into trouble with the actual validity of generated yaml.

jupyterhub/templates/hub/secret.yaml
```diff
@@ -10,11 +10,15 @@ data:
...
   auth.state.crypto-key: {{ (required "Encryption key is required for auth state to be persisted!" .Values.auth.state.cryptoKey) | b64enc | quote }}
   {{- end }}
+  {{- $values := dict "hub" dict }}
+  {{- /* pluck only needed secret values, preserving values.yaml structure */ -}}
+  {{- $_ := set $values.hub "services" dict -}}
+  {{- range $key, $service := .Values.hub.services -}}
+    {{- if $service.apiToken -}}
+      {{- $_ := set $values.hub.services $key (pick $service "apiToken") -}}
+    {{- end -}}
+  {{- end }}
```
consideRatio: See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/941/files#r219466675

jupyterhub/templates/hub/configmap.yaml
minrk: > Remove all right whitespace chomping here, if not you will have invalid output because everything will glue together on one line.

This is not true. The whitespace chomping here does nothing other than ensure that all of these lines that produce no output at all actually produce no output.

> Line 18 cant have left whitespace chomping either, it is actual content.

This is also not true. None of these chomped lines produce output.

> Since you are not whitespace chomping left on line 18, this line will be indented now forcing similar indentation to happen on neighboring YAML content.

I think this is also inaccurate. These lines are pure logic (they produce no output). In general, indenting logic *should* be done, not avoided, to improve readability. Only where complex whitespace chomping causes a problem (i.e. not in this case) should it be avoided.

> Make a templates styleguide

That would be helpful, but more important, I think, is to be less strict about these things. Strict enforcement of style without an easy auto-formatter only creates an unnecessarily hostile contribution environment.

jupyterhub/templates/hub/secret.yaml
minrk: Indeed, looks like there's nothing to change here.

jupyterhub/templates/hub/configmap.yaml
minrk: Since the linter doesn't complain, I don't think there's anything to do here. If the linter did complain, it's the linter that's the problem, but since this is both the more desirable code and producing no lint errors, there's nothing to do, right?

jupyterhub/templates/hub/configmap.yaml
consideRatio: Whitespace chomping left / right influences whitespace content left and right, nothing about the actual output of the template itself. Assignments `{{ $something := ... }} won't generate any output though, neither will the helm template comments `{{/* */}}`.

Ah I get it now, it was only assignments within the range. I got to thinking you were generating content in the `range` on line 17-19, but you were just iterating assignments. All generated content is formatted properly on line 22.

The enforcing to not generate output, works with only left whitespace chomping though, which is what I tend to do. Right whitespace chomping can mess things up, and is mainly done for linting pleasing. If you would had generated content in that range loop though, that content would have ended up right after `data:`.

We could go with "always left whitespace chomp, never right" as a rule. Together with usage of nindent I think that may be all we need to generate valid output

consideRatio: Sorry for the misguided conclusions @minrk

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -110,6 +110,8 @@ spec:
           env:
             - name: PYTHONUNBUFFERED
               value: "1"
+            - name: HELM_RELEASE_NAME
+              value: {{ .Release.Name | quote }}
```
consideRatio: Perhaps we should pass the [built in objects](https://docs.helm.sh/chart_template_guide/#built-in-objects) as well in the configmap?

consideRatio: We could pass:
- Capabilities, Chart and Release
But not:
- Files and Template

jupyterhub/templates/hub/configmap.yaml
minrk: Right chomping removed, though it makes no difference in the output.

jupyterhub/templates/hub/secret.yaml
minrk: right-chomps removed.

minrk:
ok, I think this works now. The hub configmap is now a simple passthrough of values.yaml, except:

- only the relevant top-level sections are included: auth, cull, hub, scheduling, singleuser, scheduling (notably: excludes proxy, prePuller which prevents proxy config changes from relaunching the hub)
- secrets are excluded via `omit`:
  - `hub.cookieSecret`
  - `hub.services.*.apiToken`
  - `singleuser.imagePullSecret.password`

The hub secret is mostly unchanged, except the structure of `hub.services.*.apiToken` is passed through exactly as values.yaml via `pick` (the inverse of `omit`), rather than a single-purpose structure. The advantage going forward is that the config doesn't care if values came through the secret or not, they show up in the same location.

Sticky bits I ran into:

- the affinity 'helpers' are no longer used for users. This is now implemented in Python (better, I think), but the helpers can't be removed since they are still used for the user-placeholder. This is making me wonder if the user-placeholder ought to be somehow a feature of KubeSpawner instead of the helm chart.
- while the sprig [dict docs](https://github.com/Masterminds/sprig/blob/master/docs/dicts.md) suggest that merge does a deep merge, it apparently doesn't do a deep *copy* so modifying the filtered $values in-place can result in modifying the main .Values. if we're not careful. I think I was careful enough to get it right on the few filtered-out values at the moment.

I'm not confident that I ported every transform without typos, but it was mostly a big find-replace. I took the opportunity to simplify the many cases where `some_value` is set from `someValue` (formerly `some-value`) by adding a `camelCaseify` function. This means that most configuration is iterating through lists of traits and loading them from values.yaml if present and not-null.

jupyterhub/templates/hub/deployment.yaml
minrk: I initially tried this, but discovered that we actually *shouldn't* do that, at least for release and chart. Release is different on every deploy (it includes timestamps) and a change to this configmap triggers a Hub relaunch, which we don't want. So instead, I included the only value that is actually used via an environment variable. Plus, I can't think of any other subsets of this that we should be using in the deployment, and if it turns out there's another item we need, we can add it in the future.

We could include release & chart in a *different* configmap that doesn't trigger relaunch, if we find we need more than one key out of them.

jupyterhub/templates/hub/configmap.yaml
consideRatio: I got a bad conscience about these comments, I have been too nitty, the linting of rendered output influenced me in a bad way. I'll fix it!

```
$ humctl patch human erik --patch '{"spec": {"nittyness": 0.1}}'

# Human          # Status
erik-61v8zlh     Terminating
erik-c6l1z9q     Pending
```

```
$ humctl get humans

# Human          # Status
erik-c6l1z9q     Running
```

consideRatio:
@minrk regarding the modification about @values in place, you can make a copy first if you want. Just ensure you put a new new dictionary to the left of a merge.

```
{{- $copy := merge dict .Values }}
```

consideRatio:
Excellent work! This has been a complex additional step for all PRs and anyone wanting to grasp things. Now while it will still be an extra step, it does not require much manual intervention.

minrk:
@consideRatio thanks for the review! I think this change should be ready to review and test.

> you can make a copy first if you want. Just ensure you put a new new dictionary to the left of a merge. {{- $copy := merge dict .Values }}

If you look, you'll see that this is what the code already does (or did), and it's insufficient. What I discovered is that this only does a *one level* copy. It's a deep merge, but if there are no collisions, modifying `$copy.auth.cryptoKey` without also doing `set $copy "auth" (merge dict .Values.auth)` will modify the original `.Values.auth` because `$copy.auth is .Values.auth`  (in Python parlance).

images/hub/z2jh.py
```diff
@@ -3,32 +3,71 @@
...
+from collections import Mapping
+from functools import lru_cache
 import os
-import sys
+
 import yaml
 
+
+# memoize so we only load config once
+@lru_cache()
```
consideRatio: Ah sweet! 💯

consideRatio:
## Breaking change
Current master works, but my setup fail with this these changes: the hub crashes when starting up with the following error:

```
[W 2018-10-04 11:38:56.936 JupyterHub configurable:168] Config option `consecutive_failure_limit` not recognized by `JupyterHub`.
[I 2018-10-04 11:38:56.937 JupyterHub app:1667] Using Authenticator: builtins.CustomAuthenticator
[I 2018-10-04 11:38:56.937 JupyterHub app:1667] Using Spawner: kubespawner.spawner.KubeSpawner
[I 2018-10-04 11:38:56.938 JupyterHub app:1010] Loading cookie_secret from env[JPY_COOKIE_SECRET]
[W 2018-10-04 11:38:56.968 JupyterHub app:1129] JupyterHub.hub_connect_port is deprecated as of 0.9. Use JupyterHub.hub_connect_url to fully specify the URL for connecting to the Hub.
[E 2018-10-04 11:38:56.970 JupyterHub app:1952]
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1949, in launch_instance_async
        await self.initialize(argv)
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1678, in initialize
        await self.init_users()
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1146, in init_users
        if self.authenticator.enable_auth_state:
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 556, in __get__
        return self.get(obj, cls)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 535, in get
        value = self._validate(obj, dynamic_default())
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 675, in _authenticator_default
        return self.authenticator_class(parent=self, db=self.db)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/config/configurable.py", line 84, in __init__
        self.config = config
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 585, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 574, in set
        obj._notify_trait(self.name, old_value, new_value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 1139, in _notify_trait
        type='change',
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 1176, in notify_change
        c(change)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 819, in compatible_observer
        return func(self, change)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/config/configurable.py", line 186, in _config_changed
        self._load_config(change.new, traits=traits, section_names=section_names)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/config/configurable.py", line 153, in _load_config
        setattr(self, name, deepcopy(config_value))
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 585, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 559, in set
        new_value = self._validate(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 591, in _validate
        value = self.validate(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 2054, in validate
        self.error(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 625, in error
        raise TraitError(e)
    traitlets.traitlets.TraitError: The 'client_id' trait of a CustomAuthenticator instance must be a unicode string, but a value of None <class 'NoneType'> was specified.
```

## Why?

It relates to this line of code:
```
# inside my `hub.extraConfig` within config.yaml
# I'm reading 
c.GenericOAuthenticator.client_id = get_config('custom.client-id')
```

And how I provide a value for this `custom.client-id` entry.

My configmap is set like this:
```yaml
hub:
  extraConfigMap:
    client-id: asdf-asdf-asdf-asdf-asdfasdf
```

**Conclusion:** `get_config('custom.client-id')` returned None, so the configmap didnt have `custom.client-id` as an entry any more.

To debug this better, I stashed two configmaps, one generated by using the master branch and one based on your PR:

## Master branch's configmap
```yaml
apiVersion: v1
data:
  auth.admin.access: "false"
  auth.admin.users: |
    - erik.sundell
  auth.scopes: |
    - User.Read
    - offline_access
  auth.state.enabled: "true"
  auth.type: dummy
  custom.callback-url: '"https://example.com/hub/oauth_callback"'
  custom.client-id: '"redacted"'
  custom.client-secret: '"redacted"'
  custom.hosted-domain: '"example.com"'
  custom.login-service: '"Example"'
  debug.enabled: "false"
  hub.allow-named-servers: "false"
  hub.base_url: /
  hub.concurrent-spawn-limit: "64"
  hub.consecutive-failure-limit: "5"
  hub.db_url: sqlite:///jupyterhub.sqlite
  hub.extra-config.main.py: "redacted"
  kubespawner.common-labels: |
    app: jupyterhub
    release: jhub
    chart: jupyterhub-0.8-5e4d24e
    heritage: jupyterhub
  singleuser.cloud-metadata: |
    enabled: false
    ip: 169.254.169.254
  singleuser.cmd: jupyterhub-singleuser
  singleuser.cpu.limit: "7"
  singleuser.default-url: /lab
  singleuser.events: "true"
  singleuser.extra-labels: |
  singleuser.fs-gid: "100"
  singleuser.image-pull-policy: IfNotPresent
  singleuser.image-spec: redacted
  singleuser.memory.guarantee: 1G
  singleuser.memory.limit: 4G
  singleuser.network-tools.image.name: consideratio/k8s-network-tools
  singleuser.network-tools.image.tag: cc865bd
  singleuser.node-affinity-required: |
    - matchExpressions:
      - key: hub.jupyter.org/node-purpose
        operator: In
        values: [user]
  singleuser.node-selector: '{}'
  singleuser.scheduler-name: jhub-user-scheduler
  singleuser.start-timeout: "900"
  singleuser.storage.capacity: 10Gi
  singleuser.storage.dynamic.pvc-name-template: claim-{username}{servername}
  singleuser.storage.dynamic.storage-access-modes: '[ReadWriteOnce]'
  singleuser.storage.dynamic.volume-name-template: volume-{username}{servername}
  singleuser.storage.extra-volume-mounts: '[]'
  singleuser.storage.extra-volumes: '[]'
  singleuser.storage.home_mount_path: /home/jovyan
  singleuser.storage.type: dynamic
  singleuser.tolerations: |
    - key: hub.jupyter.org_dedicated
      operator: Equal
      value: user
      effect: NoSchedule
    - key: hub.jupyter.org/dedicated
      operator: Equal
      value: user
      effect: NoSchedule
  singleuser.uid: "1000"
  token.cookie_secret: redacted
kind: ConfigMap
metadata:
  creationTimestamp: 2018-08-29T22:37:41Z
  labels:
    app: jupyterhub
    chart: jupyterhub-0.8-5e4d24e
    component: hub
    heritage: Tiller
    release: jhub
  name: hub-config
  namespace: jhub
  resourceVersion: "7955873"
  selfLink: /api/v1/namespaces/jhub/configmaps/hub-config
  uid: redacted
```

## This PR's configmap
```yaml
apiVersion: v1
data:
  values.yaml: |
    auth:
      admin:
        access: false
        users:
        - erik.sundell
      dummy:
        password: null
      ldap:
        dn:
          search: {}
          user: {}
        user: {}
      scopes:
      - User.Read
      - offline_access
      state:
        enabled: true
      type: dummy
      whitelist:
        users: null
    cull:
      concurrency: 10
      enabled: false
      every: 600
      maxAge: 0
      timeout: 28800
      users: false
    hub:
      activeServerLimit: null
      allowNamedServers: false
      annotations:
        prometheus.io/path: /hub/metrics
        prometheus.io/scrape: "true"
      baseUrl: /
      concurrentSpawnLimit: 64
      consecutiveFailureLimit: 5
      db:
        pvc:
          accessModes:
          - ReadWriteOnce
          annotations: {}
          selector: {}
          storage: 1Gi
          storageClassName: null
          subPath: null
        type: sqlite-pvc
        upgrade: null
        url: null
      deploymentStrategy:
        rollingUpdate: null
        type: Recreate
      extraConfig:
        main: "redacted"
      extraConfigMap:
        callback-url: https://example.com/hub/oauth_callback
        client-id: redacted
        client-secret: redacted
        hosted-domain: example.com
        login-service: Example
      extraContainers: []
      extraEnv:
        JUPYTERHUB_API_TOKEN: redacted
        JUPYTERHUB_CRYPT_KEY: redacted
        OAUTH2_AUTHORIZE_URL: https://login.microsoftonline.com/redacted/oauth2/v2.0/authorize
        OAUTH2_TOKEN_URL: https://login.microsoftonline.com/redacted/oauth2/v2.0/token
        OAUTH2_USERNAME_KEY: userPrincipalName
      extraVolumeMounts:
      - mountPath: /etc/gitlab/
        name: gitlab
      extraVolumes:
      - name: gitlab
        secret:
          secretName: gitlab
      fsGid: 1000
      image:
        name: consideratio/k8s-hub
        tag: bcd6266
      imagePullPolicy: IfNotPresent
      labels: {}
      networkPolicy:
        egress:
        - to:
          - ipBlock:
              cidr: 0.0.0.0/0
        enabled: false
      nodeSelector: {}
      pdb:
        enabled: true
        minAvailable: 1
      publicURL: null
      resources:
        requests:
          cpu: 50m
          memory: 512Mi
      service:
        ports:
          nodePort: null
        type: ClusterIP
      services: {}
      uid: 1000
    scheduling:
      corePods:
        nodeAffinity:
          matchNodePurpose: require
      podPriority:
        enabled: false
      userPlaceholder:
        enabled: true
        replicas: 0
      userPods:
        nodeAffinity:
          matchNodePurpose: require
      userScheduler:
        enabled: true
        image:
          name: gcr.io/google_containers/kube-scheduler-amd64
          tag: v1.11.2
        logLevel: 4
        nodeSelector: {}
        pdb:
          enabled: true
          minAvailable: 1
        replicas: 1
        resources:
          requests:
            cpu: 50m
            memory: 256Mi
    singleuser:
      cloudMetadata:
        enabled: false
        ip: 169.254.169.254
      cmd: jupyterhub-singleuser
      cpu:
        guarantee: 0
        limit: 7
      defaultUrl: /lab
      events: true
      extraAnnotations: {}
      extraContainers: []
      extraEnv: {}
      extraLabels: {}
      extraNodeAffinity:
        preferred: []
        required: []
      extraPodAffinity:
        preferred: []
        required: []
      extraPodAntiAffinity:
        preferred: []
        required: []
      extraResource:
        guarantees: {}
        limits: {}
      extraTolerations: []
      fsGid: 100
      image:
        name: redacted
        pullPolicy: IfNotPresent
        tag: redacted
      imagePullSecret:
        email: null
        enabled: false
        registry: null
        username: null
      initContainers: []
      lifecycleHooks: null
      memory:
        guarantee: 1G
        limit: 4G
      networkPolicy:
        egress:
        - to:
          - ipBlock:
              cidr: 0.0.0.0/0
              except:
              - 169.254.169.254/32
        enabled: false
      networkTools:
        image:
          name: consideratio/k8s-network-tools
          tag: 8fc0199
      nodeSelector: {}
      serviceAccountName: null
      startTimeout: 900
      storage:
        capacity: 10Gi
        dynamic:
          pvcNameTemplate: claim-{username}{servername}
          storageAccessModes:
          - ReadWriteOnce
          storageClass: null
          volumeNameTemplate: volume-{username}{servername}
        extraLabels: {}
        extraVolumeMounts: []
        extraVolumes: []
        homeMountPath: /home/jovyan
        static:
          pvcName: null
          subPath: '{username}'
        type: dynamic
      uid: 1000
kind: ConfigMap
metadata:
  creationTimestamp: 2018-08-29T22:37:41Z
  labels:
    app: jupyterhub
    chart: jupyterhub-0.8-762f25c
    component: hub
    heritage: Tiller
    release: jhub
  name: hub-config
  namespace: jhub
  resourceVersion: "7955223"
  selfLink: /api/v1/namespaces/jhub/configmaps/hub-config
  uid: redacted
```

consideRatio:
## Secrets in the configmap
I'm also passing secrets through extraEnv and extraConfigMap.

Those in extraEnv is now exposed with this PR, but not in master.
Those in extraConfigMap were and remains exposed.

minrk:
@consideRatio thanks for testing! These were both easy to fix:

- omit extraEnv/extraConfig from the values.hub passthrough
- restore the original implementation of extraConfigMap. This no longer makes a lot of sense, since values.hub is itself passed through as-is, but there's no need to break backward-compatibility unnecessarily. extraConfigMap is a user->user communication channel (BinderHub uses it), so I think there's no harm in leaving it alone.

Allowing users to pass extra secrets is something to consider, but not related to this PR, I think.

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -21,3 +21,6 @@ data:
 {{- $_ := set $values.singleuser "imagePullSecret" (omit .Values.singleuser.imagePullSecret "password") }}
   values.yaml: |
     {{- $values | toYaml | nindent 4 }}
+{{- range $key, $value := .Values.hub.extraConfigMap }}
+  custom.{{ $key }}: {{ toJson $value | quote }}
+{{- end }}
```
consideRatio: This is the outcome of the new configmap.

```
apiVersion: v1
data:
  custom.callback-url: '"https://example.com/hub/oauth_callback"'
  custom.client-id: '"redacted"'
  custom.client-secret: '"redacted"'
  custom.hosted-domain: '"example.com"'
  custom.login-service: '"Example"'
  values.yaml: |
    auth:
      admin:
        access: false
        users:
# ...
```

I guess that the new get_config function looks in values.yaml, because I still got the same error

minrk: ah, right. Need to preserve `.get_config`.

jupyterhub/templates/NOTES.txt
```diff
@@ -18,3 +18,14 @@ Note that this is still an alpha release! If you have questions, feel free to
   1. Read the guide at https://z2jh.jupyter.org
   2. Chat with us at https://gitter.im/jupyterhub/jupyterhub
   3. File issues at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues
+
+{{- if .Values.hub.extraConfigMap }}
+
+DEPRECATION: hub.extraConfigMap is deprecated in jupyterhub chart 0.7.
+Use top-level `custom` instead:
```
consideRatio: It is not deprecated in 0.7.0

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -19,8 +19,7 @@ data:
 {{- /* copy values.singleuser */ -}}
 {{- $_ := set $values "singleuser" (omit .Values.singleuser "imagePullSecret") }}
 {{- $_ := set $values.singleuser "imagePullSecret" (omit .Values.singleuser.imagePullSecret "password") }}
+{{- /* preserve behavior of deprecated hub.extraConfigMap */ -}}
+{{- $_ := set $values "custom" (merge .Values.hub.extraConfigMap $values.custom) }}
```
consideRatio: Hmmm, this is so complicated but i think this will happen:

1. You will update `.Values.hub.extraConfigMap` (the leftmost dict)
2. `$values.custom` will get lower priority than `.Values.hub.extraConfigMap`

I suggest:

```
{{- $_ := set $values "custom" (merge $values.custom .Values.hub.extraConfigMap) }}
# or, this may work as well
{{- $_ := merge $values.custom .Values.hub.extraConfigMap }}
```

consideRatio:
```
# this worked fine, NOTES.txt is a beauty as well
{{- $_ := merge $values.custom .Values.hub.extraConfigMap }}
```

minrk: Good catch, I was thinking of Python priority, where the last item wins, not first. Fixed.

images/hub/jupyterhub_config.py
```diff
@@ -50,42 +76,105 @@
```
consideRatio: Line 75 will no longer function because this was removed from the configmap:

```
  {{- /* KubeSpawner */}}
  kubespawner.common-labels: |
    {{- $_ := merge (dict "heritageLabel" "jupyterhub") . }}
    {{- include "jupyterhub.commonLabels" $_ | nindent 4 }}
```

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -5,312 +5,21 @@ metadata:
...
-  {{- if .Values.hub.cookieSecret }}
-  token.cookie_secret: {{ .Values.hub.cookieSecret | quote }}
-  {{- end }}
-
-  {{- /* Other */}}
-  debug.enabled: {{ .Values.debug.enabled | quote }}
-  {{- range $key, $value := .Values.hub.extraConfigMap }}
-  custom.{{ $key }}: {{ toJson $value | quote }}
-  {{- end }}
+{{- $values := pick .Values "auth" "cull" "custom" "hub" "scheduling" "singleuser" }}
```
consideRatio: pick `"debug"` as well

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -5,312 +5,21 @@ metadata:
...
+{{- range $key, $service := $values.hub.services }}
+  {{- if $service.apiToken }}
+    {{- $_ := set $values.hub.services $key (omit $service "apiToken") }}
+  {{- end }}
+{{- end }}
+{{- /* copy values.singleuser */ -}}
+{{- $_ := set $values "singleuser" (omit .Values.singleuser "imagePullSecret") }}
+{{- $_ := set $values.singleuser "imagePullSecret" (omit .Values.singleuser.imagePullSecret "password") }}
+{{- /* preserve behavior of deprecated hub.extraConfigMap */ -}}
+{{- $_ := set $values "custom" (merge $values.custom .Values.hub.extraConfigMap) }}
```
consideRatio: Make it:
```
{{- /* preserve behavior of deprecated hub.extraConfigMap */ -}}
{{- $_ := merge $values.custom .Values.hub.extraConfigMap }}
```

jupyterhub/templates/NOTES.txt
```diff
@@ -18,3 +18,14 @@ Note that this is still an alpha release! If you have questions, feel free to
...
+
+---
+custom:
+{{- (merge .Values.custom .Values.hub.extraConfigMap) | toYaml | nindent 2}}
```
consideRatio: this will modify .Values.custom, do:

{{- merge (dict) .Values.custom .Values.hub.extraConfigMap | toYaml | nindent 2 }}

jupyterhub/templates/hub/configmap.yaml
minrk: No, that would be wrong.

consideRatio: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ef4b0f757ac3314e2679b16a66373a372df960ae/images/hub/jupyterhub_config.py#L333-L335

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ef4b0f757ac3314e2679b16a66373a372df960ae/jupyterhub/templates/hub/configmap.yaml#L312-L313

consideRatio: @minrk good idea I think would be simpler to maintain

jupyterhub/templates/NOTES.txt
consideRatio: (but 0.8.0)

images/hub/jupyterhub_config.py
minrk: I'm trying to figure out the right thing to do here, and I see two options:

1. implement the correct labels in Python (needs Chart/Release inputs, but we need to subset them, since they contain timestamps/hash versions, which shouldn't be included in the configmap lest they trigger relaunch of the Hub)
2. preserve this common-labels passthrough as-is

My inclination is to reimplement in Python, with filtering out things like Chart.version and Release.timestamp to avoid unnecessary churn. What do you think?

consideRatio: @minrk +1 to reimplement in Python and scrap the helper. This is one of these helpers that is over the top. It is defined, and used twice. See below.

**Definition:**
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ef4b0f757ac3314e2679b16a66373a372df960ae/jupyterhub/templates/_helpers.tpl#L130-L142

**Usage 1:**
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ef4b0f757ac3314e2679b16a66373a372df960ae/jupyterhub/templates/_helpers.tpl#L145-L152

**Usage 2:**
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ef4b0f757ac3314e2679b16a66373a372df960ae/jupyterhub/templates/hub/configmap.yaml#L259-L262

jupyterhub/templates/hub/configmap.yaml
```diff
@@ -21,5 +21,8 @@ data:
 {{- $_ := set $values.singleuser "imagePullSecret" (omit .Values.singleuser.imagePullSecret "password") }}
 {{- /* preserve behavior of deprecated hub.extraConfigMap */ -}}
 {{- $_ := set $values "custom" (merge dict $values.custom .Values.hub.extraConfigMap) }}
+{{- /* passthrough subset of Chart / Release */ }}
```
consideRatio: This line
```
{{- /* passthrough subset of Chart / Release */}}
```

Remove one single space in the end ^^

images/hub/jupyterhub_config.py
minrk: This is fixed now, I believe. I do think we should revamp the helpers (#964), but save that for a later single-purpose PR.

consideRatio:
## Ideas from discussion etc summarized
- Have a configmap that does not trigger a restart of hub/proxy pod
- Make it easier to pass things not through a configmap but through a secret, perhaps we can have a `customSecret` field, rather than `custom` field for the configmap passthrough.

consideRatio:
@minrk if you run `tools/templates/lint-and-validate.py` you catch these kinds of errors as well, allowing you to iterate quicker

minrk:
arg, .Chart isn't a dict, it's a special 'Chart.metadata' type that we can't treat as a dict. That's annoying. Should be working now.

consideRatio:
@minrk this L-Awesome-TM ! Thanks for the effort of fixing this once properly. No longer WIP / Ready for merge?

minrk:
> Ready for merge?

As far as I know, yes!

images/hub/z2jh.py
```diff
@@ -3,32 +3,71 @@
...
-        return default
-
-def get_secret(key, default=None):
-    """Get a secret from /etc/jupyterhub/secret"""
-    path = os.path.join('/etc/jupyterhub/secret', key)
-    try:
-        with open(path) as f:
-            return f.read().strip()
-    except FileNotFoundError:
-        return default
```
consideRatio: You removed `get_secret`?

minrk: Yes, get_secret no longer makes sense

images/hub/jupyterhub_config.py
```diff
@@ -290,27 +371,25 @@
...
 for name, service in get_config('hub.services', {}).items():
-    api_token = get_secret('services.token.%s' % name)
     # jupyterhub.services is a list of dicts, but
     # in the helm chart it is a dict of dicts for easier merged-config
     service.setdefault('name', name)
+    # handle camelCase->snake_case of api_token
+    api_token = service.pop('apiToken', None)
     if api_token:
         service['api_token'] = api_token
     c.JupyterHub.services.append(service)
```
consideRatio: @minrk I don't see how the `api_token` propegates to here without `get_secret` that utilized a path where `hub-secret` is mounted, which is different from where the `hub-config` is mounted.

images/hub/z2jh.py
consideRatio: Ah now it makes sense that it no longer makes sense to me as well =)

images/hub/jupyterhub_config.py
consideRatio: Figured it out, you are merging the two mounted directories, nice i like it!

images/hub/jupyterhub_config.py
```diff
@@ -46,46 +72,125 @@
...
+        c.KubeSpawner.node_affinity_preferred.append(
+            dict(
+                weight=100,
+                preference=node_selector,
+            ),
+        )
+    elif match_node_purpose == 'require':
+        c.KubeSpawner.node_afinity_required.append(node_selector)
+    else:
+        raise ValueError("Unrecognized value for matchNodePurpose: %r" % match_node_purpose)
```
consideRatio: I have documented `prefer`, `require` and `ignore`.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/1f672e250ead51a8f43f2fc8d7cecf02e6bcdaaa/jupyterhub/schema.yaml#L706-L718

minrk: Fixed.

PR #947, opened 2018-09-24
HTTPS - only require hosts for setup with letsencrypt

### LINKED ISSUE #874: Offload HTTPS not working
When I try to follow the directions here: <https://zero-to-jupyterhub.readthedocs.io/en/latest/security.html#off-loading-ssl-to-a-load-balancer>, the load balancer does not correctly configure the SSL listener, and the security group doesn't have the port 443 open either.  I don't know if this is a problem with kubernetes or the helm chart. (I'm using the v7.0-beta.2 helm chart)

## Contents of `config.yaml`
```yml
proxy:
  https:
    enabled: true
    type: offload
  service:
    annotations:
      # Certificate ARN
      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "XXX"
      # Which ports should use SSL - originally tried 'https', but changed to 443 based on information at
      # https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#aws
      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
      service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '3600'
```

consideRatio:
Fixes #874 

This has not been verified by me though. Thanks @cilquirm for figuring out how to solve this! :heart:

### DIFF
--- a/jupyterhub/templates/proxy/secret.yaml
+++ b/jupyterhub/templates/proxy/secret.yaml
@@ -1,4 +1,4 @@
-{{- $manualHTTPS := (and .Values.proxy.https.enabled (eq .Values.proxy.https.type "manual")) }}
+{{- $manualHTTPS := and .Values.proxy.https.enabled (eq .Values.proxy.https.type "manual") -}}
 {{- if $manualHTTPS -}}
 apiVersion: v1
 kind: Secret
--- a/jupyterhub/templates/proxy/secret.yaml
+++ b/jupyterhub/templates/proxy/secret.yaml
@@ -3,7 +3,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: manual-tls-proxy-{{ .Release.Name }}
+  name: proxy-manual-tls
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 type: kubernetes.io/tls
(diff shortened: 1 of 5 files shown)


jupyterhub/templates/proxy/secret.yaml
```diff
@@ -1,9 +1,9 @@
-{{- $manualHTTPS := (and .Values.proxy.https.enabled (eq .Values.proxy.https.type "manual")) }}
+{{- $manualHTTPS := and .Values.proxy.https.enabled (eq .Values.proxy.https.type "manual") -}}
 {{- if $manualHTTPS -}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: manual-tls-proxy-{{ .Release.Name }}
+  name: proxy-manual-tls
```
consideRatio: A convention for tls secrets is to use `example-com-tls` - hence the name change

consideRatio:
I just self reviewed this after some time passed, lgtm.

consideRatio:
I have ordered some books about open source software, I think I will read what I've come to learn. PR's get easier to review if you only change the bare minimum required for the change to happen. I feel like I've been doing too much non essential things in my PRs as it pleased me, such as in this one: I renamed a secret, and I reordered two if statements.

PR #951, opened 2018-09-25
small fix in docs - aythentication.rst

kalaytan:
Incorrect location of dnAttribute: 'cn'. 

According to images/hub/jupyterhub_config.py
    `set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn_user_dn_attribute', 'auth.ldap.dn.user.dn-attribute')`
dn-attribute should be under dn.user.


choldgraf:
nice catch! thanks for the patch 🎉

Issue #945, opened 2018-09-21
Using imagePullSecrets for hub


clancychilds:
From what I can see in the charts, there is a imagePullSecrets implemented for singleuser images, but not for the hub itself.
I was hoping to make my own hub image based off of juptyerhub/k8s-hub:0.7.0 and then pull it down from a private gcr.io registry. I have this working for my singleuser image using _jsonkey in the config,yaml, but the same functionality isn't in place for hub images.

consideRatio:
true, quick question, is the private registry in your same google cloud project as you have your cluster? then it works right away without setup

clancychilds:
It would be at some point, but I'm trying to make sure that it also is able to run on minikube, on premise k8s and elsewhere.
I'm actually pulling from the same docker registry as singleuser, so I'm looking to see if I can just inject imagePullSecrets.name.singleuser-image-credentials into the hub Pod spec right now. Might try to make a PR later to "do it right".

clancychilds:
This has been fixed and merged.

consideRatio:
Thank you @clancychilds ! :heart:

PR #952, opened 2018-09-25
Allow for imagePullSecret on hub image

clancychilds:
This makes it easier to provide credentials to a private docker repo for the hub image. This is mainly a cut-and-paste from the same feature for the singleuser repos. 
This addresses issue #945

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -38,6 +38,89 @@ properties:
 
           See the [Kubernetes docs](https://kubernetes.io/docs/concepts/containers/images/#updating-images)
           for more info on what the values mean.
+      imagePullSecret:
+        type: object
+        description: |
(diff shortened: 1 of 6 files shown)


consideRatio:
Thank you @clancychilds for taking the time to contribute :D :heart: Makes me so happy to see!

consideRatio:
Sorry for the slow review turnaround btw, your PR looks fine though!

I was thinking that perhaps we should try to refactor the documentation to avoid too much repetition, but it didn't feel obvious. So, if we want to change it, lets do it another PR if it makes sense to you and others as well. Whats written in the schema.yaml will end up in the [Configuration reference](https://zero-to-jupyterhub-with-kubernetes.readthedocs.io/en/latest/reference.html#helm-chart-configuration-reference).

Thank you @clancychilds and congratulations to your first contribution! :D

PR #956, opened 2018-09-26
install jupyterhub 0.9.4

minrk:
fixes 0.8->0.9 upgrade

we may want to do an 0.7.1 chart release with this, since it greatly improves the upgrade experience.


consideRatio:
Yes, I agree, we want a 0.7.1 release with this, should we cherry pick this on top of 0.7.0 or similar or just release 0.7.1 with all the added things since 0.7.0?

Issue #959, opened 2018-10-01
"Secure Helm" step is broken with Helm 2.10


manics:
@consideRatio pointed out in Gitter that Helm 2.10 has a bug which means The `Secure Helm` step on https://zero-to-jupyterhub.readthedocs.io/en/stable/setup-helm.html breaks Helm. The doc currently says `Make sure you have at least version 2.9.1`. Should this be updated to 2.11? Or just add a note to avoid 2.10?

consideRatio:
@manics I think we should go with recommending 2.11, as that is also required for k8s 1.11 which is out on various cloud providers - even though GKE is behind :'(

Thanks for helping out with this @manics !!! :heart:

PR #960, opened 2018-10-02
Change minimum helm version to 2.11.0

manics:
2.10 has a bug that means the "secure helm" breaks Helm.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/959


consideRatio:
Since this is just documentation this is certainly fine, one could enforce it in Chart.yaml as well, hmmm... Should we go for that as well soon?

Thanks for the PR and all the great work you do!

PR #962, opened 2018-10-08
extraImages is a dict

minrk:
not a list.

Having a list here results in a (harmless, but misleading) warning when users define prePuller.extraImages:

> `warning: destination for extraImages is a table. Ignoring non-table value []`


consideRatio:
Yepp, I just confirmed in the advanced docs about the supposed usage:

```yaml
prePuller:
   extraImages:
     ubuntu-xenial:
       name: ubuntu
       tag: 16.04
       policy: IfNotPresent
```

PR #967, opened 2018-10-10
Update default kube-lego version

### LINKED ISSUE #963: kube-lego doesn't renew certificates
There seems to be a bug in kube-lego 0.1.6 (the default in the chart) that prevents it from actually renewing the certificate (see [jetstack/kube-lego#338](https://github.com/jetstack/kube-lego/issues/338)).

As a temporary workaround you can set the version of kube-lego in your `values.yaml`:
```
lego:
    image:
      tag: 0.1.7
```

Deploying this now and will make a PR if that goes well.

Has anyone else experienced this?

cc @gedankenstuecke

betatim:
Fixes #963

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -98,7 +98,7 @@ proxy:
   lego:
     image:
       name: jetstack/kube-lego
-      tag: 0.1.6
+      tag: 0.1.7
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks @betatim !! :heart:

PR #969, opened 2018-10-11
better output of lint errors

minrk:
rather than showing a `CalledProcessError` traceback, exit with a simple error message


minrk:
Thanks!

PR #970, opened 2018-10-11
ensure values.hub.services dict is a copy

minrk:
avoids modifying values.hub.services in-place


jupyterhub/templates/hub/configmap.yaml
```diff
@@ -11,6 +11,7 @@ data:
 {{- $_ := set $values "auth" (merge dict .Values.auth) }}
 {{- $_ := set $values.auth "state" (omit $values.auth.state "cryptoKey") }}
 {{- $_ := set $values "hub" (omit $values.hub "cookieSecret" "extraEnv" "extraConfigMap") -}}
+{{- $_ := set $values.hub "services" dict }}
 {{- range $key, $service := $values.hub.services }}
```
consideRatio: You will now iterate over an empty dictionary

consideRatio: It will be fine if we write...
```
{{- range $key, $service := .Values.hub.services }}
```
... I think. $key is a value type, and $service is a dict, but when we call `(omit $service "apiToken")` later, we will make a shallow copy I think.

minrk: Good catch. Fixed.

consideRatio:
Grrrr why does the 1.10 job version tend to behave bad and timeout? https://travis-ci.org/jupyterhub/zero-to-jupyterhub-k8s/jobs/440111876

PR #971, opened 2018-10-11
publish chart prior to running tests

minrk:
tests shouldn't prevent or delay publication of the chart


consideRatio:
Not so confident about this, but LGTM from what I can tell - perhaps we could merge this first, and then merge #970 to verify things works as they should.

PR #975, opened 2018-10-12
Add example for customizing login_service

moschlar:
Add example for customizing auth.custom.config.login_service


consideRatio:
Thank you @moschlar for this PR! :heart: I remember figuring this out by trying things out at some point =D

PR #979, opened 2018-10-16
Some fixes after passthrough values

bitnik:
After https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/941 some configuration params are not loaded correctly from values.yaml.


consideRatio:
Thank you @bitnik !! :heart:

PR #983, opened 2018-10-18
updating theme

choldgraf:
Updates the theme and adds circleci building for previews of docs


choldgraf:
here's the built docs: https://2-87849371-gh.circle-artifacts.com/0/html/index.html

PR #988, opened 2018-10-20
Support glob-matching ingress

tmshn:
GKE's default ingress (ingress-gce) matches path by glob-pattern.
Therefore, path `/` doesn't match `/hub` etc.

To make this to work properly, I need to patch the path to `/*`.

Here, I introduced `pathSuffix` option to support these situation.


minrk:
Makes sense

meeseeksmachine:

https://discourse.jupyter.org/t/binder-redirect-to-jupyterhub-returns-403-forbidden/8236/4

PR #977, opened 2018-10-15
Update link

willingc:
The https://z2jh.jupyter.org was failing to redirect correctly. I'm changing the link to https://zero-to-jupyterhub.readthedocs.io/en/stable/ . Feel free to troubleshoot the jupyter link if preferred. For now, I am just going to merge this and get it working again.

PR #989, opened 2018-10-22
Point links back to z2jh.jupyter.org

yuvipanda:
Partially reverts jupyterhub/zero-to-jupyterhub-k8s#977

It works now, with https://docs.readthedocs.io/en/latest/alternate_domains.html.
This gives us direct serving from z2jh.jupyter.org and 
HTTPS. 

Leaves the switch from pointing to latest to stable in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,12 @@
 **This is under active development and subject to change.**
 
 This repo contains resources, such as **Helm charts** and the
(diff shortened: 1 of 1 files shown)


willingc:
Are you sure you want to do a revert or update to link to stable?

yuvipanda:
@willingc good point! I hadn't seen that change in the PR. Will fix to point back to stable!

yuvipanda:
@willingc done!

PR #996, opened 2018-10-26
Conceal some auth values in hub-config

tmshn:
There might be `clientSecret` or `password` in values from `auth` field.
They should not be exposed in plain ConfigMap but should be concealed in Secret.
I made it so.

Example:

`values.yaml`

```yaml
auth:
  type: github
  github:
    clientId: xxx
    clientSecret: yyy
    callbackUrl: https://example.com/hub/oauth_callback
```

Rendered (before)

```yaml
# Source: jupyterhub/templates/hub/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: hub-config
  labels:
    # ...snip...
data:
  values.yaml: |
      # ...snip...
      github:
        callbackUrl: https://example.com/hub/oauth_callback
        clientId: xxx
        clientSecret: yyy
      type: github
      # ...snip...
```

Rendered (after)

```yaml
# Source: jupyterhub/templates/hub/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: hub-config
  labels:
    # ...snip...
data:
  values.yaml: |
      # ...snip...
      github:
        callbackUrl: https://example.com/hub/oauth_callback
        clientId: xxx
      type: github
      # ...snip...
```

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -10,6 +10,13 @@ data:
 {{- /* make a copy of values.auth to avoid modifying the original */ -}}
 {{- $_ := set $values "auth" (merge dict .Values.auth) }}
 {{- $_ := set $values.auth "state" (omit $values.auth.state "cryptoKey") }}
+{{- range $key, $auth := .Values.auth }}
+  {{- if typeIs "map[string]interface {}" $auth }}
+    {{- if (or $auth.clientSecret $auth.password) }}
+      {{- $_ := set $values.auth $key (omit $auth "clientSecret" "password") }}
+    {{- end }}
+  {{- end }}
+{{- end }}
 {{- $_ := set $values "hub" (omit $values.hub "cookieSecret" "extraEnv" "extraConfigMap") -}}
 {{- $_ := set $values.hub "services" dict }}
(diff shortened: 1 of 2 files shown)


tmshn:
Another may-be secret is `db.url`, which may contain password. However I manage this by creating extra secret to hold db password and set it as environment variable.

Not this

```yaml
  db:
    type: postgres
    url: postgresql+psycopg2://jovyan:secretpasswd@...
```

but this:

```yaml
  db:
    type: postgres
    url: postgresql+psycopg2://jovyan@...
  extraEnv:
    - name: PGPASSWORD
      valueFrom:
        secretKeyRef:
          name: hub-postgres
          key: password
```

So instead of putting entire `db.url` into Secret, should we create another secret to hold only password and set it as an environment variable for user?

tmshn:
@yuvipanda Thank you for the advice. I updated to the latest master.

tmshn:
I  opened a new pr about this: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/996#issuecomment-434295589

👉  https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1004

consideRatio:
Wieeee thank you @tmshn !!! :D

PR #1002, opened 2018-11-01
set consecutive failure limit on Spawner

minrk:
it was mapped to the wrong config, and thus had no effect

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -53,9 +53,6 @@ else:
 for trait, cfg_key in (
     # Max number of servers that can be spawning at any one time
     ('concurrent_spawn_limit', None),
-    # Max number of consecutive failures before the Hub restarts itself
-    # requires jupyterhub 0.9.2
-    ('consecutive_failure_limit', None),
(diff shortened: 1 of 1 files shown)


images/hub/jupyterhub_config.py
```diff
@@ -92,6 +89,14 @@ def camelCaseify(s):
 
 c.KubeSpawner.namespace = os.environ.get('POD_NAMESPACE', 'default')
 
+# Max number of consecutive failures before the Hub restarts itself
+# requires jupyterhub 0.9.2
+set_config_if_not_none(
+    c.Spawner,
+    'consecutive_failure_limit',
+    'hub.consecutiveFailureLimit',
+)
```
betatim: It took me a few attempts to see the difference: the old code set `c.JupyterHub.consecutive_failure_limit` and the new code sets `c.Spawner.consecutive_failure_limit`. Is that right?

betatim:
What is the z2jh testing setup like (I've not looked at it for a while)? Could we add a test for this as it has been an important thing to help keep mybinder.org up and running.

PR #1012, opened 2018-11-08
Bump PriorityClass k8s API to beta from alpha

consideRatio:
Bumping the API to beta resolved an issue I got during a Helm
upgrade of the chart. I previously ran into the following error:

```
Update Complete. ⎈ Happy Helming!⎈
Error: UPGRADE FAILED: unable to recognize "": no matches for kind "PriorityClass" in version "scheduling.k8s.io/v1alpha1"
```

The API bump resolved it though.

### DIFF
--- a/jupyterhub/templates/scheduling/priorityclass.yaml
+++ b/jupyterhub/templates/scheduling/priorityclass.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.scheduling.podPriority.enabled }}
-apiVersion: scheduling.k8s.io/v1alpha1
+apiVersion: scheduling.k8s.io/v1beta1
 kind: PriorityClass
 metadata:
   name: {{ .Release.Name }}-default-priority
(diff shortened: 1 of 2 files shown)


consideRatio:
@hans-permana, I recall you experienced this as well at some point, is that correct?

consideRatio:
@yuvipanda  @minrk you have not enabled podPriority on mybinder.org or berkely yet right? Get this first.

hans-permana:
@consideRatio I remember this error message. I got around this by explicitly stating the `cluster-version` to 1.10

consideRatio:
@hans-permana I think that was because only 1.10 with alpha features had this enabled, and only then this resource was found. For me though, I kept running k8s 1.11.2-gke.9 before and after, and got it working with the change in this PR. So I assume my issue was different than yours then.

Thanks for getting back to me about this!

consideRatio:
@yuvipanda @minrk before you do deploys to mybinder.org / berkeley - this may be needed on the latest GKE version, even though it worked on k8s 1.10 alpha it may not work on the GKE 1.11 version without this. I thought the API would support old stuff though, so I'm confused.

minrk:
@consideRatio thanks!

Now that 1.11 is GA, I think we should get mybinder.org to deploy 1.11 and enable this. Once it's verified sound in production, I think we should make a release of the jupyterhub chart.

PR #1025, opened 2018-11-15
Clarify documentation on nbgitpuller trigger

### LINKED ISSUE #1024: Unclear documentation on user-environment-nbgitpuller
It looks like something got lost in translation in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/866/commits/1659a67fb5c5a7d24c5a56518863acea02f1f8c4 , or a copy-paste mistake; possibly both: the [second sentence on nbgitpuller](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/866/commits/1659a67fb5c5a7d24c5a56518863acea02f1f8c4#diff-ae7d3197b7e61ee3c0f22a03c39cc495R237) is unclear, with a nonsensical sentence and bad links. ([Section](https://zero-to-jupyterhub.readthedocs.io/en/stable/user-environment.html#using-nbgitpuller-to-synchronize-a-folder) in the live documentation.)

I'm guessing it should something like "Note that this syncing can also be triggered by users following a link like <link>", with a working example link. But I can't tell. Perhaps @consideRatio can clarify?

Otherwise I'd suggest to remove the sentence to avoid confusion or puzzlement.

evertrol:
Rephrase sentence; closes #1024


consideRatio:
There were multiple errors in the URL actually, this is one that i tried and was functional:
```
# fixed ? -> / at one spot
# fixed spelling of "materials"
# using original repo instead of a forked repo
https://your-domain.com/hub/user-redirect/git-pull?repo=https://github.com/data-8/materials-fa18
```

Otherwise LGTM!

evertrol:
Whoops; good thing the URL mistakes were pointed out, because I (still) had a `?` instead of `/` and the wrong domain in there as well (bad reading instead of copy-pasting the example from a comment in #1024). 
Do please double check the URL is now correct with the amended commit before merging.

PR #1031, opened 2018-11-23
Make sure priority class is created before other resources

betatim:
I think not having these annotations for helm to make sure the priority class gets created first is the reason for the recent mybinder.org deploy to fail: https://travis-ci.org/jupyterhub/mybinder.org-deploy/builds/458381755


consideRatio:
(This should only have influenced the user-placeholder-pods being created before this was created)

PR #1037, opened 2018-11-25
Fix ingress.tls's default value

consideRatio:
`.Values.ingress.tls` should default to a `[]` instead of `{}` to avoid:
```
2018/11/25 21:56:36 warning: cannot overwrite table with non table for tls (map[])
```

... when the following valid configuration is provided:

```yaml
ingress:
  tls:
    - secretName: hub-neurips-mybinder-org-tls
      hosts:
        - hub.neurips.mybinder.org
```


betatim:
LGTM

Issue #402, opened 2018-01-12
Document how to enable user-selectable Docker images


choldgraf:
## UPDATED

Add documentation on how to utilize KubeSpawners profile_list traitlet implemented here: https://github.com/jupyterhub/kubespawner/pull/137

---

See https://gitter.im/jupyterhub/jupyterhub?at=5a59236e6117191e615b6729 for a recent question about this, it is possible to set up JupyterHub to let users select a Docker Image. We should document how to do this.

mrocklin:
This would be of high value to me personally.  I'd also welcome pointers to dev documentation, examples, source code, etc. that might help me on my way.

sylus:
I'd be curious as well as following the existing examples via kubeflow didn't quite give me enough of the info necessary to proceed.

Thanks a bunch!

cam72cam:
I'd also like to see how to do this :)  Happy to be a guinea pig!

clkao:
With https://github.com/jupyterhub/kubespawner/pull/137 the profile list is finally built-in!

vilhelmen:
I'm using `profile_list`, but it keeps selecting the default image in the `config.yml` file specified under `singleuser.image`, but when I don't specify `singleuser.image`, the image puller hangs. Ideas? I'm having difficulty finding much documentation on `profile_list` with z2jh

consideRatio:
```yaml
singleuser:
  profileList:
    - display_name: "Default: Shared, 8 CPU cores"
      description: "Your code will run on a shared machine with CPU only."
      default: True
    - display_name: "Personal, 4 CPU cores & 26GB RAM, 1 NVIDIA Tesla K80 GPU"
      description: "Your code will run a personal machine with a GPU."
      kubespawner_override:
        extra_resource_limits:
          nvidia.com/gpu: "1"
```

manics:
That YAML looks a lot nicer than the JSON in `hub.extraConfig` 😀

consideRatio:
Yepp it does! Thanks @vilhelmen for making it possible!

yuvipanda:
I think we should have some sort of narrative descriptive documentation for this as well.

ablekh:
@consideRatio What do you mean by _Personal_ (vs. _Shared_) machine in your example above? Do you imply by Personal an exclusive use of a K8s host by a specific user or something else? Please clarify ...

consideRatio:
@ablekh ah, I revised my example!

[...]

PR #1039, opened 2018-11-27
Enable KubeSpawner.profile_list direct assign

vilhelmen:
Teaches the config file to look for `singleuser.profileList` and map that to `c.KubeSpawner.profile_list`

I don't think the image awaiter is aware of these containers, but that's something I look to address later.

Semi-related to #402

### DIFF
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -60,6 +60,15 @@ spec:
             - /bin/sh
             - -c
             - echo "Pulling complete"
+        {{- range $k, $container := .Values.singleuser.profileList }}
+        - name: image-pull-singleuser-profilelist-{{ $k }}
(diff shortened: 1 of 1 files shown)


vilhelmen:
Added the images to the awaiter after some guessing. It will need to be changed from `image_spec` to `image` on the next KubeSpawner release, as master is using `image`.

consideRatio:
@minrk @vilhelmen this will crash deployments that does not specify an image_spec override i think, since the range loop will run, but the container.override.image_spec wont be found.

Sorry for late review!

vilhelmen:
Oh no I made things worse! I've only just started messing with the helm chart, so I'm not that great with it yet. There may be another issue I hadn't remembered until now.

When I was trying to get the profile list working, I noticed that I still HAD to declare a singleuser image or else the image puller would hang. I'm not sure if that was because the image declaration was necessary, or because it had nothing to pull and locked up.

PR #1043, opened 2018-11-30
Make it easier to build hub image with git commit hash.

clkao:
Try to install from git if JUPYTERHUB_VERSION starts with git, such as:

    git+https://github.com/jupyterhub/jupyterhub.git@bbc2847530ccdb0f40a48f8e561415ea808a6561

Also switch to python3-pycurl so libcurl4-openssl-dev won't conflict with npm, which
is required by jupyterhub from git for lessc to work.


joshbode:
Can we try to figure out how to support both?
This will break TLS for Google OAuth...

PR #1045, opened 2018-12-02
Allow singleuser servers to communicate with hub with NetworkPolicy is enabled

### LINKED ISSUE #1044: z2jh not working if NetworkPolicy enabled
With latest z2jh, launched user pod can not communicate with hub when NetworkPolicy is enabled.

This is caused by f20f3e1 [removing the singleuser extraLabel](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/f20f3e16a222b2e6b2bc7de78760e08c76f1593b#diff-96a1791deadeccb98a6f334d195c73a7L207) during the value pass-through change. the `hub.jupyter.org/network-access-hub` label is used as podselector in the network policies installed.

A quick fix would be adding the following the the default values.yaml:
```
singleuser:
  extraLabels:
```

But this means for installations not enabling NetworkPolicy, the user pods will also have this label, which has no effect but might confuse people looking at pod yaml.  the [original implementation](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/546/files#diff-96a1791deadeccb98a6f334d195c73a7R174) only sets this label when NetworkPolicy is enabled.

clkao:
Fixes #1044

This is a regression since #941.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -182,7 +182,8 @@ singleuser:
                 - 169.254.169.254/32
   events: true
   extraAnnotations: {}
-  extraLabels: {}
+  extraLabels:
+    hub.jupyter.org/network-access-hub: "true"
   extraEnv: {}
   lifecycleHooks:
   initContainers: []


jupyterhub/values.yaml
```diff
@@ -182,7 +182,8 @@ singleuser:
                 - 169.254.169.254/32
   events: true
   extraAnnotations: {}
-  extraLabels: {}
+  extraLabels:
+    hub.jupyter.org/network-access-hub: "true"
```
consideRatio: Should we append this in `jupyterhub_config.py` instead perhaps? If not, utilizing the `extraLabels` option in the helm chart as a user would break the network policy function.

clkao: oh, i thought the value files are merged recursively? so unless the user specifies:

```
singleuser:
  extraLabels: {}
```

or

```
singleuser:
  extraLabels:
    hub.jupyter.org/network-access-hub: null
```

the label should always be there.  but correct me if i am wrong.

consideRatio: Ah yes you are right!

consideRatio: But the first example would also work i think, then singleuser.extraLabels, being an empty dict, would merge with a populated dict.

PR #1048, opened 2018-12-04
restore deprecated extraConfig string

### LINKED ISSUE #1009: Breaking change: hub.extraConfig no longer works as a simple string
In v0.7, the following works:

hub:
  extraConfig: |
    print("hi", flush=True)

In 0.8, this fails with:

```
      File "/srv/jupyterhub_config.py", line 424, in <module>
       for key, config_py in sorted(get_config('hub.extraConfig', {}).items()):
   AttributeError: 'str' object has no attribute 'items'
```

We *should* be backwards compatible, since it's a small check.

minrk:
with warning and everything

closes #1009

### DIFF
--- a/images/hub/jupyterhub_config.py
+++ b/images/hub/jupyterhub_config.py
@@ -1,5 +1,6 @@
 import os
 import re
+import sys
 
 from tornado.httpclient import AsyncHTTPClient
 from kubernetes import client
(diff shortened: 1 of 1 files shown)


betatim:
LGTM. Does this work for you @yuvipanda?

yuvipanda:
I like it!

PR #1050, opened 2018-12-05
kubespawner 0.10

minrk:
- use KubeSpawner.image instead of deprecated .image_spec
- fix pre-puller with profiles when images are not defined

I didn't add backward-compatibility support for `.image_spec` in profileList prepuller since we haven't made a release with support for profileList yet and the consequence is only that profile images won't be pre-pulled if they use the deprecated name. They will still work.


consideRatio:
LGTM - feel free to merge if you consider it mature in your own mind as well.

minrk:
Thanks!

Issue #1055, opened 2018-12-10
Latest build (84ee483) crashes hub on startup


consideRatio:
About time to sleep, just wanted to quickly say that my deployment it crashing like this if upgrading from the earlier to the later of these:

release | date | status
-|-|-
jupyterhub-0.8-84ee483 | Mon, 10 Dec 2018 10:41:10 +0000 | crashing
jupyterhub-0.8-ccc1e6b | Tue, 04 Dec 2018 16:56:30 +0000 | OK

/cc: @manics @minrk 

```
[E 2018-12-10 22:57:16.748 JupyterHub app:1958]
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1955, in launch_instance_async
        await self.initialize(argv)
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1689, in initialize
        await self.init_spawners()
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 1478, in init_spawners
        spawner = user.spawners[name]
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 125, in __getitem__
        self[key] = self.spawner_factory(key)
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 221, in _new_spawner
        spawner = spawner_class(**spawn_kwargs)
      File "/usr/local/lib/python3.6/dist-packages/kubespawner/spawner.py", line 104, in __init__
        super().__init__(*args, **kwargs)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/config/configurable.py", line 84, in __init__
        self.config = config
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 585, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 574, in set
        obj._notify_trait(self.name, old_value, new_value)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 1139, in _notify_trait
        type='change',
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 1176, in notify_change
        c(change)
      File "/usr/local/lib/python3.6/dist-packages/traitlets/traitlets.py", line 819, in compatible_observer
        return func(self, change)
[...]

PR #1053, opened 2018-12-06
Deploy with network policy on travis

manics:
Commits up to and including c84dd3bdc471f737744d6122b7a0b817b4038a0e refactor the travis scripts to make it easier to have different configurations. I've kept the current travis `ENV=version` matrix but it might make more sense to rename it something like `SCENARIOS`?

Commit 73503d63abab76d02d52c1797da0acd362b9cad0 deploys Z2JH with network policies on K8S 1.12.3.

# Nice to haves
- [x] Test the network policy is enforced

E.g. With `minikube-netpol.yaml` the following test could be run inside the singleuser image:
```
wget -q www.ebi.ac.uk && echo Success || echo "Failed: Couldn't fetch allowed domain"
wget -q -t1 -T2 www.bbc.co.uk && echo "Failed Blocked domain was allowed" || echo Success
```
Any suggestions on how to start a singleuser server and run the above?

Note @consideRatio I've belatedly realised this will conflict with https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/846

PR #1057, opened 2018-12-11
kubespawner 0.10.1

minrk:
fixes deprecation of hub_connect_ip, which caused launch failures in #1053, #1055


minrk:
Thanks @consideRatio!

PR #1068, opened 2018-12-15
Simplify steps for setting up Kubernetes cluster

yuvipanda:
- The following syntax doesn't work anywhere other than
  the google cloud console:

  A=5

  echo $A

  If you want it to, it should be `export A=5`. However,
  it is only useful if we re-use them later - but we do
  not need to, and we don't. Having them on a separate line
  makes it easier to edit rather than having them in the
  middle of a command - so I've moved them all towards the
  end of the command.
- ZONE is only ever required when you create your cluster,
  so don't explicitly set the zone as a requirement to create
  cluster.

### DIFF
--- a/doc/source/google/step-zero-gcp.rst
+++ b/doc/source/google/step-zero-gcp.rst
@@ -78,7 +78,7 @@ your google cloud account.
    * ``--machine-type`` specifies the amount of CPU and RAM in each node within
      this default node pool. There is a `variety of types
      <https://cloud.google.com/compute/docs/machine-types>`_ to choose from.
-     You can pick something from this `this list
+     You can pick something from `this list
      <https://cloud.google.com/compute/docs/regions-zones/regions-zones#available>`_.
      that is not too far away from your users.
    
--- a/doc/source/google/step-zero-gcp.rst
+++ b/doc/source/google/step-zero-gcp.rst
@@ -55,23 +55,7 @@ your google cloud account.
 
             gcloud components install kubectl
 
-4. Decide and configure to use a specific data center.
-
-   From a terminal, enter:
-
-   .. code-block:: bash
-
-      # Enter a zone
-      ZONE=us-east1-c
-
-      gcloud config set compute/zone $ZONE
-   
-   * ZONE specifies which data center to use. Pick something `from
-     this list
-     <https://cloud.google.com/compute/docs/regions-zones/regions-zones#available>`_.
-     that is not too far away from your users.
-
-5. Create a managed Kubernetes cluster and a default node pool.
+4. Create a managed Kubernetes cluster and a default node pool.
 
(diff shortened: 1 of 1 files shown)


doc/source/google/step-zero-gcp.rst
```diff
@@ -80,41 +64,49 @@ your google cloud account.
...
+        --node-labels hub.jupyter.org/node-purpose=core \
+        <CLUSTERNAME>
       
+   * Replace `<CLUSTERNAME>` with a name that can be used to refer to this cluster
+     in the future.
+
    * ``--machine-type`` specifies the amount of CPU and RAM in each node within
      this default node pool. There is a `variety of types
      <https://cloud.google.com/compute/docs/machine-types>`_ to choose from.
+     You can pick something from this `this list
```
choldgraf:
```suggestion
     You can pick something from `this list
```

consideRatio:
For to-be production clusters, I would recommend some settings that cannot be changed at a later time. But sadly GKE has not made their CLI mirror their Web UI so well sadly, and some things cannot be done with the CLI afaik. I also don't know what we should do in the docs that makes sense.

1. One can only make a regional GKE cluster as recommended practice (with three master servers) with only one zone in active use (chosen in advanced section). This is not vital, its only nice to avoid downtime of the kubernetes-api during random times when they make updates without notice.
  image
  **In the advanced section:**
  image

1. Having ip-alias enabled is recommended and has no drawback afaik.
  image

doc/source/google/step-zero-gcp.rst
```diff
@@ -55,23 +55,7 @@ your google cloud account.
...
-      ZONE=us-east1-c
-
-      gcloud config set compute/zone $ZONE
-   
-   * ZONE specifies which data center to use. Pick something `from
-     this list
-     <https://cloud.google.com/compute/docs/regions-zones/regions-zones#available>`_.
-     that is not too far away from your users.
-
-5. Create a managed Kubernetes cluster and a default node pool.
```
consideRatio: It would be used if we create another node pool, or do various gcloud commands in the future, but it is a "prepare for future" part that adds complexity and yepp let's reduce the complexity.

yuvipanda: Yeah, I think the 'wow' factor of 'I ran all these commands and they just worked' is very important. Most folks don't need to create multiple pools ever, so we should focus on the default use case.

yuvipanda:
Amended @choldgraf's suggestion

@consideRatio you're totally right re A=5 syntax. I've no idea why it didn't work when I first tried it...

I personally prefer the method in this PR, since it doesn't require explaining how variables work in bash to people when they do this. The only thing I feel strongly about is ZONE, since that adds an extra layer of complexity most people don't need to know.

PR #1073, opened 2018-12-18
MRG after 0.8.0 - Tune image-pullers pull policies

consideRatio:
If the image pull policy is left out, it will act as "IfNotPresent"
almost always. The difference is that the default is to make an
adjustment if the tag is "latest", then it will become "Always".

So, hardcoding it makes little sense, and this would help us if we have
a hook-image-puller and we make an upgrade using an image with "latest" tag because we are actively developing the image... Hmmm...

Well, perhaps by submitting this commit in a PR, we can conclude whats
best to do about it.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -110,7 +110,9 @@ spec:
             {{- end }}
           resources:
             {{- .Values.hub.resources | toYaml | trimSuffix "\n" | nindent 12 }}
-          imagePullPolicy: {{ .Values.hub.imagePullPolicy }}
+          {{- with .Values.hub.image.pullPolicy }}
+          imagePullPolicy: {{ . }}
+          {{- end }}
           env:
             - name: PYTHONUNBUFFERED
               value: "1"
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -55,7 +55,9 @@ spec:
       initContainers:
         - name: image-pull-singleuser
           image: {{ .Values.singleuser.image.name }}:{{ .Values.singleuser.image.tag }}
-          imagePullPolicy: IfNotPresent
+          {{- with .Values.singleuser.image.pullPolicy }}
+          imagePullPolicy: {{ . }}
+          {{- end }}
           command:
             - /bin/sh
             - -c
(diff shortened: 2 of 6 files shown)


jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -83,7 +80,6 @@ spec:
         {{- range $k, $v := .Values.prePuller.extraImages }}
         - name: image-pull-{{ $k }}
           image: {{ $v.name }}:{{ $v.tag }}
-          imagePullPolicy: {{ $v.policy | default "IfNotPresent" }}
```
minrk: This is deleting an option from the configuration. Maybe this should be conditional `if $v.policy` and only set it if defined?

consideRatio: Yes in the case of extraImages that would make sense. But we must not provide a truthy default value in `values.yaml` if we create such conditionals - and we won't for `extraImages`.

Perhaps we should remove the default pullPolicies in `values.yaml` and create conditionals throughout of declaring the imagePullPolicy, as the omission is an option in itself.

consideRatio: To formulate things easier:

Do you agree that defining `"IfNotPresent"` as default value in `values.yaml` is bad, as the default of omission is what's really what is expected?

yuvipanda: @consideRatio yeah, if it is set in config, we should set it. If not, we should omit it.

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -110,7 +110,9 @@ spec:
             {{- end }}
           resources:
             {{- .Values.hub.resources | toYaml | trimSuffix "\n" | nindent 12 }}
-          imagePullPolicy: {{ .Values.hub.imagePullPolicy }}
+          {{- with .Values.hub.image.pullPolicy }}
+          imagePullPolicy: {{ . }}
```
minrk: Rather than explicitly setting this as null, it makes more sense to me to have 

```yaml
{{- if .Values.hub.image.pullPolicy }}
imagePullPolicy: {{ .Values.hub.image.pullPolicy }}
{{- end }}
```

that way the field is actually omitted rather than affirmatively set to null. Does that make sense?

consideRatio: Helm's with is also a conditional, it will not render at all if the provided value is falsy. I'm torn between doing it this way and using ifs. Helm's official templates utilizes the `with` way while I just like you got confused about it the first time I saw it.

minrk: Gotcha. Seems confusing, but always learning. Thanks!

PR #1076, opened 2018-12-19
Remove setting node labels in gcloud container create

yuvipanda:
This is only needed if you are setting up multiple node
pools - which isn't a common case. Even then, you can create
an additional nodepool for your core then - and you most likely
want to, since it should be sized differently. Removing it
makes the initial setup page fully explainable.


consideRatio:
The core label isnt something of importance even in production. So sure! Lets reduce complexity

PR #1079, opened 2018-12-20
Reference docs for prePuller

### LINKED ISSUE #1064: Update config reference with prePuller config keys
I noticed that we don't have any prePuller configuration keys in the reference page:

http://z2jh.jupyter.org/en/stable/reference.html

Is there a different that this information exists? I know we mention it here: http://z2jh.jupyter.org/en/stable/advanced.html?highlight=prepuller but it should probably be in the "catch-all" reference page too

consideRatio:
Closes #1064


yuvipanda:
Thanks :)

choldgraf:
wooo!

PR #1082, opened 2018-12-20
make linkcheck used to fix links

consideRatio:


yuvipanda:
Ty for making this happen!

PR #1084, opened 2018-12-21
Explicitly set namespace on all resources

pcfens:
By explicitly setting the namespace on all resources, rendered templates can be applied by tools other than Tiller without also requiring special command line flags.

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -2,6 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: hub-config
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 data:
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: hub
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
@@ -59,7 +60,7 @@ spec:
       {{- if .Values.hub.imagePullSecret.enabled }}
       imagePullSecrets:
         - name: hub-image-credentials
-      {{- end }}  
+      {{- end }}
       containers:
         {{- if .Values.hub.extraContainers }}
         {{- .Values.hub.extraContainers | toYaml | trimSuffix "\n" | nindent 8 }}
--- a/jupyterhub/templates/hub/image-credentials-secret.yaml
+++ b/jupyterhub/templates/hub/image-credentials-secret.yaml
@@ -3,6 +3,7 @@ kind: Secret
 apiVersion: v1
 metadata:
   name: hub-image-credentials
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- $_ := merge (dict "componentSuffix" "-image-credentials") . }}
     {{- include "jupyterhub.labels" $_ | nindent 4 }}
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -3,6 +3,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: hub
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
--- a/jupyterhub/templates/hub/pdb.yaml
+++ b/jupyterhub/templates/hub/pdb.yaml
@@ -3,6 +3,7 @@ apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
 metadata:
   name: hub
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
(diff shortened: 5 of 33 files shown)


minrk:
I don't think I've seen a chart do this before. Can you point to some examples of charts doing this? I'd like for us to follow best practices in https://github.com/helm/charts as much as possible, and I don't see this over there.

pcfens:
It's not an incredibly common pattern since most templates are applied exclusively by Tiller. Perhaps the most widely used chart that provides a good example is the [Istio chart](https://github.com/istio/istio/tree/master/install/kubernetes/helm) (though their recommended install path is using `helm template...`)

The [best practices page](https://docs.helm.sh/chart_best_practices/) doesn't really address explicitly setting namespaces, though some of the examples on the page [about charts](https://github.com/helm/helm/blob/master/docs/charts.md) includes the namespace in their [example template](https://github.com/helm/helm/blob/master/docs/charts.md#template-files). The few issues around GitHub that I could find on the subject seem to be inconclusive too.

We're trying to manage everything in our cluster using the gitops pattern (with auto-deploy), so pre-rendering helm charts outside of tiller simplifies the process.

yuvipanda:
@pcfens I use the pattern too sometimes with `helm template`, but you can pass a --namespace flag to it too. Can you tell us what tool you are using to render these instead?

consideRatio:
@pcfens did you find a sustainable solution?

Based on the discussion, it seems like that it is possible to generate templates with `helm template --namespace <mynamespace>`, and I also think it is possible when they have been generated to install them in a certain namespace by adding `--namespace <mynamespace>` for the `kubectl apply` or `kubectl create` command.

I think this repo needs an even stronger reason to break the common practice of not hardcoding a certain namespace, so for now I'll go ahead and close this PR. I'll be around for continued discussion!

tirumerla:
Hi @consideRatio - I'm stuck due to this same issue with multiple approaches forward. I was following this discussion and wanted to get your opinion on what is the best approach.

**Issue:**

```
jupyterhub   19s         Warning   FailedCreate              statefulset/user-placeholder           create Pod user-placeholder-0 in StatefulSet user-placeholder failed error: pods "user-placeholder-0" is forbidden: exceeded quota: jupyterhub-resourcequota, requested: limits.cpu=2,limits.memory=4G,requests.cpu=1,requests.memory=2G, used: limits.cpu=810m,limits.memory=1290Mi,requests.cpu=250m,requests.memory=640Mi, limited: limits.cpu=1,limits.memory=1536Mi,requests.cpu=300m,requests.memory=768Mi
```
As you can see user-placeholder pods are getting deployed in the `jupyterhub` namespace which has minimal resources compared to `jupyter-users` which is where the `single-user` pods are getting deployed. Since i'm not providing any resources to user-placeholder it is taking the resources given to single user pods.

What's the best approach here?

- Specify resources to user-placeholder in my `config.yaml` within the limit of `jupyterhub` namespace, so that there won't be issue of resourcequota and when single user pods gets deployed it will be deployed in `jupyter-users` namespace as configured -- Hope there won't be any issues with this.

or

- Change namespace to deploy user-placeholder pods in `jupyter-users` namespace since `single-user` pods will be deployed in that namespace. 

If changing the namespace is the right approach, what's the best way to do that since `helm install .... -n jupyterhub` will deploy all the pods in `jupyterhub` namespace.

 

**Context:**

We have two namespaces `jupyterhub` and `jupyter-users`.

- `jupyterhub` namespace has `hub`, `proxy`, `image-puller`, `user-scheduler` pods running.
- `jupyter-users` namespace has all the `single-user` pods. 

This isolation was done based on our network policies. 

I'm using `c.KubeSpawner.namespace = 'jupyter-users'` in my `config.yaml` to have this isolation between namespaces. 

**Deploy command:** 
```
helm upgrade --install jhub jupyterhub/jupyterhub --version 0.10.2 -n jupyterhub --values config.yaml
```
`jupyterhub` namespace has resourcequota enabled meaning any pod that gets deployed within the namespace needs to be within the range of the resourcequota provided.

**Configuration:**

- Snippet from my `config.yaml`

```
singleuser:
    image:
        name: "xxxxxxxxx/jupyter-singleuser-z2jh"
        tag: "1.0"
        pullPolicy: Always
    # Change these values according to your machine
    memory:
        limit: 4.0G
        guarantee: 2.0G
    cpu:
        limit: 2.0
        guarantee: 1.0

scheduling:
    podPriority:
        enabled: true
    userPlaceholder:
        # Specify three dummy user pods will be used as placeholders
        replicas: 3
    # https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#scheduling-corepods-nodeaffinity
    corePods:
        nodeAffinity:
            # matchNodePurpose valid options:
            # - ignore
            # - prefer (the default)
            # - require
            matchNodePurpose: require
    # https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#scheduling-userpods-nodeaffinity
    userPods:
        nodeAffinity:
            # matchNodePurpose valid options:
            # - ignore
            # - prefer (the default)
            # - require
            matchNodePurpose: require
    userScheduler:
        enabled: true
        resources:
            requests:
                cpu: 50m
                memory: 128Mi
            limits:
                cpu: 200m
                memory: 256Mi
        containerSecurityContext:
            runAsUser: 65534  # nobody user
            runAsGroup: 65534 # nobody group
            readOnlyRootFilesystem: true 
            capabilities: 
                add: ["NET_BIND_SERVICE", "NET_ADMIN"]
                drop: 
                    - ALL
```

Appreciate your help. Let me know if you have any questions.

consideRatio:
Post questions in jupyter.discourse.org please, you can ping me there!

PR #1087, opened 2018-12-26
JupyterLab by default

consideRatio:


consideRatio:
Okay, happy to close this PR - should we prepare for the coming change perhaps?

Administrators may indeed end up getting a change they did not prepare for when making this upgrade. Perhaps we should then write in `NOTES.txt` an encouragement that the following configuration should be added to `config.yaml` in order to preserve classic behavior as a preparation for the future versions of the chart when the default behavior will change?

```yaml
# Let users arrive to the classical user interface (not JupyterLab)
singleuser:
  cmd: jupyterhub-singleuser
  defaultUrl: /

# # The future default: let users arrive to JupyterLab
# singleuser:
#   cmd: jupyter-labhub
#   defaultUrl: /lab
```

Besides this aspect of preparing the z2jh users (jupyterhub administrators) for the change in default behavior, is there something else you think we should await @yuvipanda?

yuvipanda:
@consideRatio I think we can switch by default to jupyter-labhub, but set defaultURL to `/tree`. This makes it so that classic notebook is still default, but switching to lab only requires one change (default URL) rather than two.

What I *think* we should do is:

1. Ship a notebook extension that puts a 'try JupyterLab' button in the tree view / notebook view
2. Encourage admins to switch to it, and mention it'll be default in the future.

I think this will let both users and admins know it's coming. I'd also like to see some real world educational usage of JupyterLab (which doesn't exist currently) before we make it the default.

betatim:
Another thing that could give a quantitative input to the "when to switch" decision is to look at popular notebook extensions used with JupyterHub and whether there are Jupyter lab equivalents for them that work well.

choldgraf:
Is there any way to throw a warning with a particular helm config in a way that admins would notice? Seems to me the best path would be to add something like a deprecation warning in the next z2jh release, then in the one after, switch the default

minrk:
Switching to JupyterLab should already be only one change (command -> jupyter-labhub). It shouldn't need both defaultUrl and command, since the change in command ought to define what `/` redirects to.

However, the non-default combination @yuvipanda mentioned has the benefit that jupyterlab has the hub extensions ready to go for users who manually navigate to `/lab` even when it's not the default app. However, it has the disadvantage that images that lack jupyterlab won't start at all without custom configuration.

At some point, it will probably become appropriate for `jupyterhub-singleuser` to become aware of jupyterlab and make sure both are set up properly without needing to change the entrypoint.

consideRatio:
I marked this PR with a 0.8 milestone to indicate I want something more done about this situation before release, but I'm not currently actionable.

### My current technical understanding summarized
- Changing the `c.Spawner.cmd` to `jupyter-labhub` allows the `/lab` prefix to work, but not `/` to redirect to `/lab` by default (?)
- Changing the `c.Spawer.cmd` to `/lab` setups a redirect from `/` to `/lab` instead of `/tree` (?)

### Various inputs summarized
1. @betatim Suggested a way to decide among other options. By looking at a Venn diagram with two circles representing a notebook domain and a lab domain, and that we can evaluate when to switch based on the sizes of the _notebook only_ / _both notebook and lab_ / _lab only_ areas.
2. @choldgraf Asked how we could provide info to z2jh administrators making an upgrade, and I'm thinking of:
   - Our Helm chart's `NOTES.txt` that is written out during `helm upgrade` in the end of a successful upgrade. Note that this can be made to warn only those matching a certain criteria, for example not using `jupyter-labhub` as their `singleuser.cmd` already along with `singleuser.defaultUrl` as `/lab`
3. @minrk Suggested that:
   - `jupyter-labhub` _should but doesn't_ (?) cause `/lab` to be the `defaultUrl`.
   - Using `jupyter-labhub` as default `cmd` causes us to require jupyterlab to be installed. Note that it is on our default image and on all docker-stacks images though.
   - `jupyterhub-singleuser` (is this `start-singleuser.sh`?) should be made jupyterlab aware? It will start `jupyter-labhub` if we set an environment variable called `JUPYTER_LAB_ENABLE` if i recall correctly.

### Suggested options as I understand them summarized
1. @consideRatio Suggested to make JupyterLab the default.
2. @yuvipanda Suggested:
   - to use `jupyter-labhub` as `c.Spawner.cmd` ...
   - while not changing the `defaultUrl` to `/lab` ...
   - make a simple extension for `jupyter notebook` showing a simple button to go to `/lab`
   - install this extension in our default singleuser image
   - (loose interpretation) perhaps also installing it to [jupyter/docker-stacks's base-notebook](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile).

### Stuff to resolve in my view
- [ ] @minrk @yuvipanda I bet I didn't get this correctly, can you help me correct my interpretation?
- [ ] Decide on actions required before `0.8` release

minrk:
I was wrong about jupyter-labhub making jupyterlab the default. It's still two commands, so you need to set both cmd and default_url. I opened https://github.com/jupyterlab/jupyterlab/pull/5865 to make what I thought was the case actually the case, but that will be the future.

I don't think there's anything to change for 0.8 on this front.

manics:
Following the discussion on https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/562#issuecomment-453985206
Can `c.Spawner.cmd` default to the images CMD? IIUC you could then enable jupyterlab by default by setting `JUPYTER_ENABLE_LAB`, and if you had a image that's not based on docker-stacks it would ignore the environment variable and continue to work with it's own command?

consideRatio:
I would very much like to see JupyterLab become the default at least in the next z2jh release, if there is anything we need to do in this release for that to happen I'd be greatly motivated to get it done.

minrk:
I think allowing the image's default command to come through is probably a reasonable change to make for 0.9 (this is how DockerSpawner behaves). I don't think there's anything we should change here for 0.8.

clkao:
One options is to document and recommend using `JUPYTER_ENABLE_LAB` is profile's override. With that you can actually configure two launch options for the same image, one launching lab and one launching classic. This should be flexible enough that we default to image's default command as @minrk suggested.

manics:
What does [`jupyterhub-singleuser`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/84e5a3b0fe0bae356c851a01b52bb2ca312c271c/jupyterhub/values.yaml#L204) do that isn't done through the default docker stacks `CMD` which should take `JUPYTER_ENABLE_LAB` into account?

https://github.com/jupyter/docker-stacks/blob/814ef10d64fb79570b4eea7a3dfae02669a01dec/base-notebook/Dockerfile#L122

meeseeksmachine:

https://discourse.jupyter.org/t/binder-jupyterhub-activity-round-up-week-1/2704/1

consideRatio:
I want to clean up some PRs, closing this micro-pr that is easy to fix later and start fresh with a new discussion.

Issue #1097, opened 2019-01-08
Bug with the pass-through for profile lists


choldgraf:
I was getting a multi-profile jupyterhub set up in order to write some docs for #1085 and I think I ran into a bug with the profile list pass through.

I'm using the latest version of the helm chart (`--version 0.8-17d72b0`). If I use this configuration:

```

hub:
  extraConfig: |
    c.KubeSpawner.profile_list = [
      {
        "display_name": "Minimal environment",
        "description": "Super minimal",
        "defualt": True,
      },
      {
        "display_name": "Datascience environment",
        "description": "If you want the additional bells and whistles: Python, R, and Julia.",
        "kubespawner_override": {'image': "jupyter/datascience-notebook:2343e33dec46"}
      }]
```

then it works fine and I get a list of profiles. However, if I use this config:

```yaml
singleuser:
  profileList:
    - display_name: "Minimal environment"
      description: "To avoid too much bells and whistles: Python."
      default: true
    - display_name: "Datascience environment"
      description: "If you want the additional bells and whistles: Python, R, and Julia."
      kubespawner_override:
        image: jupyter/datascience-notebook:2343e33dec46
```

Then I get the following error:

```
Error: UPGRADE FAILED: render error in "jupyterhub/templates/image-puller/daemonset.yaml": template: jupyterhub/templates/image-puller/daemonset.yaml:9:4: executing "jupyterhub/templates/image-puller/daemonset.yaml" at <include "jupyterhub....>: error calling include: template: jupyterhub/templates/image-puller/_daemonset-helper.yaml:64:25: executing "jupyterhub.imagePuller.daemonset" at <$container.kubespawn...>: can't evaluate field image in type interface {}
```

It seems like something isn't being interpolated properly? Anyone have ideas?

consideRatio:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/17d72b04719139ceaa41519278de8fb394063c93/jupyterhub/templates/image-puller/_daemonset-helper.yaml#L63-L64

Look at the second line in the Helm template for daemonset that is about to configure pre-pulling of images listed under `kubespawner_override`, it assumes there will be a dict kind object named `kubespawner_override`, and that this will have a field called `image`. But, in the first entry does not contain it at all.

This is why it fails. It is indeed a proper bug that should be fixed. I'll do it tomorrow.

PR #1098, opened 2019-01-08
profile list instructions

### LINKED ISSUE #1085: profileList only mentioned in the configuration reference (schema.yaml)
We should add some narrative surrounding the use of profileList in the documentation, and perhaps not only have it in the reference section without mentioning it anywhere else.

(Idea by @yuvipanda, and I'm writing it down as an issue to not loose track of it)

choldgraf:
This adds a short section on configuring the profile list for users. It currently
uses the passthrough configuration, though I think I ran into a bug (#1097).

closes #1085 closes #1083

### DIFF
--- a/doc/source/user-environment.rst
+++ b/doc/source/user-environment.rst
@@ -65,15 +65,19 @@ image containing useful tools and libraries for datascience, complete these step
 
 2. Apply the changes by following the directions listed in
    `apply the changes`_.
-   
-   
+
+
    .. note::
-   
+
       If you have configured *prePuller.hook.enabled*, all the nodes in your
       cluster will pull the image before the the hub is upgraded to let users
       use the image. The image pulling may take several minutes to complete,
       depending on the size of the image.
 
+.. note::
+
+   If you'd like users to select an environment from **multiple docker images**,
+   see :ref:`multiple-profiles`.
 
 
 .. _jupyterlab-by-default:
@@ -266,7 +270,6 @@ using this tool.
    tool in production.
 
 
-
 .. setup-conda-envs:
 
 Allow users to create their own ``conda`` environments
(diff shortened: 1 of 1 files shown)


doc/source/user-environment.rst
```diff
@@ -293,6 +296,103 @@ across sessions. To resolve this, take the following steps:
...
+Docker images, to select the hardware on which they want their jobs to run,
+or to configure default interfaces such as Jupyter Lab vs. RStudio.
+
+Each configuration is a set of options for `Kubespawner <https://github.com/jupyterhub/kubespawner>`_,
+which defines how Kubernetes should launch a new user server pod. Any
+configuration options passed to the `profileList` configuration will
+overwrite the defaults in Kubespawner (or any configuration you've
+added elsewhere in your helm chart).
+
+Profiles are stored under ``singluser: profileList:``, and are defined as
```
consideRatio:
```suggestion
Profiles are stored under ``singluser.profileList``, and are defined as
```

doc/source/user-environment.rst
```diff
@@ -293,6 +296,103 @@ across sessions. To resolve this, take the following steps:
...
+         kubespawner_override:
+           image: jupyter/datascience-notebook:2343e33dec46
+       - display_name: "Spark environment"
+         description: "The Jupyter Stacks spark image!"
+         kubespawner_override:
+           image: jupyter/all-spark-notebook:2343e33dec46
+
+This allows users to select from three profiles, each with their own
+environment (defined by each Docker image in the configuration above). This is
+equivalent to the following configuration using `extraCongig` to directly define
```
consideRatio:
```suggestion
equivalent to the following configuration using `hub.extraConfig` to directly define
```

doc/source/user-environment.rst
```diff
@@ -293,6 +296,103 @@ across sessions. To resolve this, take the following steps:
...
+configuration for KubeSpawner:
+
+.. code-block:: yaml
+
+   singleuser:
+     image:
+       name: jupyter/minimal-notebook
+       tag: 2343e33dec46
+   hub:
+     extraConfig: |
```
consideRatio: While it is still supported to write a string directly to `hub.extraConfig`,  tend to do it like this instead now days. I avoid some warnings about conversion etc. that bugged me as the default value of `hub.extraConfig` is a `{}` and overwriting a dict with a string results in a warning if I recall correctly.

```yaml
hub:
  extraConfig:
    myProfiles: |
      ...
```

@minrk / @yuvipanda got input on this? Examplify with dict or string?

choldgraf: I'm happy with whatever you all think is the best practice to recommend to people (we could also remove this section if we're confident that the passthrough works now)

choldgraf:
I just gave another shot at the passThrough config for profileList using the latest-published helm chart, and it works! So I think we can now merge this without feeling like we're misleading people :-)

re: @consideRatio 's comments about the `extraConfig` stuff above - I'm also fine just removing that addendum, since the pass-through seems to be working now!

consideRatio:
@choldgraf thanks for the work on this!!

choldgraf:
oops - accidental close - latest push removes that "extraConfig" section at the end. This is good to go from my end

PR #1106, opened 2019-01-15
adding information about helm chart versions etc

choldgraf:
Once https://github.com/jupyterhub/helm-chart/pull/94 lands there will be a nice little table with jupyterhub versions listed per helm chart...this adds in links to that table in a few places in Z2JH where we think people might be looking.


consideRatio:
Excellent LGTM!

Issue #1109, opened 2019-01-22
Configure existing volume for User's pod


tregin:
Hello

Although this is not a real issue I wanted to ask whether there is a possibility to configure a volume/storage that would be available/mounted automatically to all Jupyter hub pods  so everyone would have access the resources stored there?
Would it be even possible to have this volume mounted as an additional volume to the existing one which is created automatically just for the user?

Any suggestion on how to this could be solved are welcome! :)
Thank you very much

```

**Comment by manics** [comment]
You'll have to create a shared PersistentVolume and PersistentVolumeClaim as @consideRatio said. Then you can mount it into your singleuser servers like this:
```
singleuser:
  storage:
    extraVolumes:
    - name: jupyterhub-sharedscratch-rw
      persistentVolumeClaim:
        claimName: jupyterhub-sharedscratch-nfs-volume
    extraVolumeMounts:
    - name: jupyterhub-sharedscratch-rw
      mountPath: /home/jovyan/shared
```

tregin:
Thank you @manics!
After setting up the PV and PVC this was the missing piece!

PR #1110, opened 2019-01-22
Document example of extraVolumes

manics:
See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1109

### DIFF
--- a/doc/source/user-storage.md
+++ b/doc/source/user-storage.md
@@ -226,7 +226,8 @@ with caution!
 If you already have a `PersistentVolume` and `PersistentVolumeClaim` created
 outside of JupyterHub you can mount them inside the user pods.
 For example, if you have a shared `PersistentVolumeClaim` called
-`jupyterhub-shared-volume` you could mount it in all user pods:
+`jupyterhub-shared-volume` you could mount it as `/home/shared` in all user
+pods:
 
 ```yaml
 singleuser:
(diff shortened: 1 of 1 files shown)


doc/source/user-storage.md
```diff
@@ -220,3 +220,25 @@ kubectl --namespace=<your-namespace> delete pvc <pvc-name>
...
+```yaml
+singleuser:
+  storage:
+    extraVolumes:
+      - name: jupyterhub-shared
+        persistentVolumeClaim:
+          claimName: jupyterhub-shared-volume
+    extraVolumeMounts:
+      - name: jupyterhub-shared
+        mountPath: /home/jovyan/shared
```
consideRatio: @manics :heart: a documentation PR! I'm growing more and more fond of documentation as I have consumed so much lately.

---

I think this mount may fail in some way, as the `mountPath` is within another mount's path. If I recall correctly what I've read but not tested myself. Can one nest volumeMounts?

Perhaps we could also reference [k8s docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#volumemount-v1-core) for volumeMounts and/or provide a note about `subPath` for entries in `extraVolumeMounts` while we are at it.

consideRatio: If it is possible to nest volumes like this, it LGTM!

manics: Good point, on my system there aren't any per-user volumes. How does `/home/shared` sound?

consideRatio: Sounds great @manics !

manics: Updated!

manics:
Looks like travis failed with a transient error: `curl: (7) Failed to connect to storage.googleapis.com port 443: Connection timed out`, can you retrigger the job?

PR #1004, opened 2018-11-02
Configure liveness/readinessProbe

tmshn:
I configured liveness/readinessProbe for both proxy and hub.

With this, they will be more robust against failures.

In addition to that, when deployed on GKE with Ingress, health check is configured properly (GKE Ingress's health check is configured automatically by following readinessProbe, or against `/` if not set).

## Ref:

- https://github.com/jupyterhub/configurable-http-proxy/pull/181
- https://github.com/jupyterhub/jupyterhub/pull/2257

⚠️  I believe this works, but not tested yet. I shall take time soon.

PR #1112, opened 2019-01-23
add livenessProbe and readinessProbe to proxy

hnykda:
This is necessary for GCE ingress to work, because it pickups healthchecks from readinessProbe. If there is none, then it takes `/`, which in case of proxy-public returns `302` (which is not sufficient). `/hub/login` always returns 200, so this works.


hnykda:
Ahh, I see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1004 now :man_facepalming:

PR #1113, opened 2019-01-23
Hub Prometheus annotations needs to include baseUrl

manics:
If jupyterhub is run under a prefix e.g. `hub.baseUrl: /jupyter` this prefix must be included in the Prometheus scrape path. Since Values can't be interpolated I've added a new value `hub.service.annotatePrometheus`. An alternative would be to not add a new value and to unconditionally include the prometheus annotations.

### DIFF
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -5,6 +5,10 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
   annotations:
+    {{- if .Values.hub.service.annotatePrometheus }}
+    prometheus.io/scrape: "true"
+    prometheus.io/path: {{ .Values.hub.baseUrl }}hub/metrics
+    {{- end }}
     {{- if .Values.hub.service.annotations }}
     {{- .Values.hub.service.annotations | toYaml | nindent 4 }}
     {{- end }}
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -5,8 +5,10 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
   annotations:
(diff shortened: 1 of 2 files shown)


consideRatio:
LGTM!

jupyterhub/templates/hub/service.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
   annotations:
+    {{- if .Values.hub.service.annotatePrometheus }}
+    prometheus.io/scrape: "true"
+    prometheus.io/path: {{ .Values.hub.baseUrl }}hub/metrics
```
minrk: Is there a way we can avoid the boolean flag? e.g. by checking for the keys in hub.service.annotations?

consideRatio: Oh btw: I realize that this could fail if both are false now, then we will end up with a blank `annotations:`.

@minrk we cannot concatinate `.Values.hub.baseUrl` with hub/metrics within `values.yaml`. So, if we don't want a boolean flag but instead have it enabled no matter what without the option, we would need to merge the `hub.service.annotations` dict with a dynamically created dict containing the dynamic value of `prometheus.io/path: {{ .Values.hub.baseUrl }}hub/metrics`.

What makes you wish to avoid the boolean flag?

minrk: I mean something like:

```yaml
prometheus.io/path: {{ default .Values.hub.service.annotations['prometheus.io/path'] .Values.hub.baseUrl + hub/metrics }}}
```

(that's not quite the right syntax, but I can't test it right now)

manics: @minrk I've got something working with the latest commit.

minrk: Perfect, thanks!

consideRatio: Nice solution!

Issue #1054, opened 2018-12-10
Releasing 0.8


minrk:
It's time to start getting ready to release v0.8 of the chart. Most of the new functionality is tested and working well in various deployments, including mybinder.org, so it's time to be making a release.

The main thing in 0.8 is the new scheduling-related work, including the user-scheduler and user-placeholders. This needs lots of new documentation to be written before we can make a release.

Things to do:

- [x] document the new scheduling options (#503, being closed by: #1074)
- [x] fill out the changelog
- [x] make sure other new features and changes are documented
- [x] run upgrade tests from 0.7 (@choldgraf made tests with Helm 2.11 and 2.12 and the result were :tada:) #1140

edit from @choldgraf:

- [x] Wait for feedback from the v0.8-beta.1 and
- [x] Release the stable v0.8

consideRatio:
I'd like to see a transition to JupyterLab as default for 0.8 as well - thoughts on this?

I'll create a PR.

choldgraf:
Just a note that we are almost done with the milestone issues for this one! All that is left is updating the changelog + making sure new features are documented.

What are the new features that we need to check if they're documented or not?

Wooo 0.8!

consideRatio:
## Our docs on how to make a release
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CONTRIBUTING.md#releasing-a-new-version-of-the-helm-chart

jspc:
Hi @minrk, @consideRatio, can I help in anyway to get this over the line and out?

choldgraf:
@consideRatio I can help with this this week as well, @jspc would you like to try the upgrade steps above and I can work on something else, maybe finalizing the changelog?

consideRatio:
Have we documented new features and stuff? A lot the autoscaling aspects resides in: https://z2jh.jupyter.org/en/latest/optimization.html

- User Scheduler
- User Placeholders
- Pod Priority
- matchNodePurpose

choldgraf:
Another friendly ping that it'd be really helpful if people could help fill out the changelog that I've started here! https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1114

minrk:
working on the changelog today. Let's do 0.8b1 tomorrow and final next week or so if it goes smoothly.

@consideRatio I think the autoscaling and scheduler docs are well covered now. I'll look through for other new things that need updating.

[...]

Issue #841, opened 2018-08-14
Already developed - Extracting smaller PRs from #758


consideRatio:
# About #758 
I worked for a long time to make improvements wherever I saw them, but that ended me with a too big PR with no focus. I'll close this issue when I have extracted everything of use from that PR. Below are various parts I want to extract from the PR.

## Developer PRs
- [x] **Improved linting and validation + CI integration** - #844 
  A test that can run before chartpress builds images etc, allowing us to spot faulty Helm templates and if they render in a way that makes them invalid k8s resources. Also some less important linting is made on the helm templates and the rendered output.
  1. Helm template linting with `helm lint`
  2. Helm template verification with `helm template`
  3. Rendered output linting with `yamllint`
  4. Rendered output validation against the k8s API schema with `kubeval`
- [x] **Improved CI tests** - #846 
  Simplifies local development and ensures a user is at least spawned correctly before signing off on the PR.

## New feature PRs
- [x] **Configurable image pull secrets** - #851
  @AlexMorreale helped us use config.yaml to install a Kubernetes Secret which could be referenced by pods that want to pull from image registries that required some credentials - this is #801.
    - #851 complements Alex work by allowing for the email field to also be configured optionally, which seem to be utilized by some registries but not all.
    - #851 also adds support for setting large JSON blobs in the password field, something that is required in order to pull from a private gcr.io registry from an external cluster.
    - #851 also ensures that the image puller DaemonSets have the same credentials to pull the images.
- [x] **User scheduler** - #891
  Want to make your autoscheduler work efficiently? Then you should schedule pods to pack tight instead of spread out. The user scheduler accomplishes this.

  Note: the `schedulerStrategy: pack` that meant to accomplish this, but it did not work as intended and only became a minor improvement. The configuration option will be deprecated with the introduction of the user scheduler.
- [x] **Allow setting of storage labels** - #924 
  This allows labels on the PVCs to be configured. This can be useful if you want to delete, patch etc all user PVCs or similar, as it can help with the kubectl selection part.
- [x] **Tolerations for node taints** - #925
  Want to forbid anything from executing on a node except for a user pod, to ensure it can scale down without fail when the users have left? Then with this PR in place, you can taint that node with `hub.jupyter.org/dedicated=user:NoSchedule` and the user pods will tolerate scheduling there anyhow. There is also a similar thing you can do but for the core pods.

[...]

PR #1114, opened 2019-01-23
[need help] updating changelog for v0.8

choldgraf:
Since we are ready to release v0.8 (#1054) this PR updates the changelog for new features.

**It needs help from the jupyterhub team to fill in missing pieces!**. Please do this either via "suggested edits" in the GitHub UI, or by pushing directly to my branch.

I've marked areas where we need more information with "TODO".

Once we get it filled out, we can merge and get a move-on with the new version!

cc @jupyterhub/jupyterhubteam @jupyterhub/binder-team 

# Things I could use help with

- Documenting any upgrade tips / warnings for people coming from v0.7
- Running "upgrade tests" from v0.7 (not sure what this means but @minrk has it in #1054)
- Documenting any breaking changes
- Pushing a v0.8 image to Dockerhub (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CONTRIBUTING.md#push-built-images-to-dockerhub--bump-version)


CHANGELOG.md
```diff
@@ -5,13 +5,84 @@ Releases are named after famous [Cricket](https://en.wikipedia.org/wiki/Cricket)
 
 
 
-## [Unreleased](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/0.7.0..master)
+## [Unreleased](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/0.8.0..master)
 
 ### Breaking changes
```
manics:
```suggestion
- Github organisation OAuth: `auth.github.org_whitelist` has been renamed to `auth.github.orgWhitelist`

```
See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1118. This is a potential security hole for anyone who doesn't realise this since if no org is whitelisted any GitHub user can login

choldgraf: woot! thanks for the suggestion!

manics:
`hub.extraConfig`: not clear to me whether the old format is deprecated:
- https://zero-to-jupyterhub.readthedocs.io/en/latest/advanced.html?highlight=extraconfig#hub-extraconfig says the old and new formats are supported
- https://zero-to-jupyterhub.readthedocs.io/en/latest/reference.html?highlight=extraconfig#custom only has the new format
- If you use the old format the hub logs say
```
hub.extraConfig should be a dict of strings,
but found a single string instead.

The keys can be anything identifying the
block of extra configuration.

Try this instead:

    hub:
      extraConfig:
        myConfig: |
```

consideRatio:
The new format allows for multiple extraConfig entries as a dict can be merged by Helm, while a string cant.

So, let's add a deprecation note in the changelog?

CHANGELOG.md
```diff
@@ -5,13 +5,85 @@ Releases are named after famous [Cricket](https://en.wikipedia.org/wiki/Cricket)
...
+
+TODO: Instructions
+
+#### Troubleshooting
+
+TODO: Instructions
+
+### New Features
+
+#### TODO: ITEM 1
```
consideRatio: - Better autoscaling support, as can be read about in http://z2jh.jupyter.org/en/latest/optimization.html

CHANGELOG.md
```diff
@@ -5,13 +5,85 @@ Releases are named after famous [Cricket](https://en.wikipedia.org/wiki/Cricket)
...
+
+This release contains JupyterHub version 0.9.4. It requires Kubernetes >= 1.11 and Helm >= 2.11.0.
+See [the Helm Chart repository](https://github.com/jupyterhub/helm-chart#versions-coupled-to-each-chart-release) for
+a list of relevant dependencies for all Helm Chart versions.
+
+It contains new features, additional configuration options, and bug fixes.
+
+### Upgrading from 0.7
+
+TODO: Instructions
```
consideRatio: I think using `--force` should be recommended at all times, I don't remember if it is required but I think it may be.

choldgraf:
I added a few more features using @consideRatio's excellent summary from #841. I also added the contributor list. Edits / additions / subtractions / reorgs are welcome!

minrk:
@consideRatio 

> So, let's add a deprecation note in the changelog?

There is a deprecation note in the changelog when this was deprecated (0.6). What's changed in 0.8 is that users are now notified when they use the deprecated configuration. I updated the deprecation note in the image to clarify that the config is loaded and a note in the changelog that the deprecation warnings and validation errors are a new feature.

minrk:
I added a few more notes for things I found, and some upgrade and troubleshooting steps. I think we're in pretty good shape for making a beta release tomorrow.

CHANGELOG.md
```diff
@@ -20,14 +20,37 @@ It contains new features, additional configuration options, and bug fixes.
 
 ### Upgrading from 0.7
 
+To upgrade your cluster:
+
+1. backup any persistent volumes and previous configuration, to be safe
```
consideRatio: I find it quite extreme to recommend backing up of PDs, do we really want to suggest this? It is a tricky thing to do.

choldgraf: I feel like it's always safe to say "we recommend backing up XXX" as long as you don't imply that you **must** back up PDs. We don't want to create unnecessary work for people, but we also don't wanna be on the hook if something goes wrong and a person gets mad at us because we never mentioned it :-)

minrk: We should clarify that this is only helm-managed PVs (just hub state), not user PVs which are not managed by the chart, and thus won't be deleted by an upgrade.

doc/source/google/step-zero-gcp.rst
```diff
@@ -102,14 +102,54 @@ your google cloud account.
...
+      --node-taints hub.jupyter.org_dedicated=user:NoSchedule
+
+  .. note::
+
+    Consider adding the ``--preemptible`` flag to reduce the cost
+    significantly. You can `compare the prices here
+    <https://cloud.google.com/compute/docs/machine-types>`_. See
+    the `preemptible node documentation
+    <https://cloud.google.com/compute/docs/instances/preemptible>`_ for more
+    information.
```
consideRatio: Related to this, is an issue that is not yet resolved I think: https://github.com/jupyterhub/kubespawner/issues/223

consideRatio: In summary, I recall that whenever a user is evicted, as they are from preemptible node pools that run out of their usage time for some reason, they will not be able to login again without a hub-pod restart.

minrk: I don't think that's something we should look at for this release. Do you think this means we should not recommend looking at preemptible nodes for now?

minrk: commented out the recommendation for now

consideRatio: Yeah, I would not recommend them until one of us have successfully utilized them with users that authenticate regularly etc. I ended up with trouble doing it.

consideRatio:
@minrk this is excellent!!

This LGTM!

minrk:
OK! Let's hit the button and make the beta. This PR updates the version in the docs to install the beta, so just building and publishing the tag should do it.

minrk:
ok, published 0.8.0-beta.1 (I forgot that helm doesn't accept Python semver-style 0.8.0b1)

PR #1117, opened 2019-01-24
Minor fixes in step-zero-gcp.rst

lesteve:
* Typo
* move available regions list to the right bullet point describing `--zone`


betatim:
Thanks a lot for helping out and making the docs better!

PR #1120, opened 2019-01-25
Refresh README.md

consideRatio:
I felt the README.md were in need of some attention.


choldgraf:
really nice! Only one comment (I think that "chart" is a bit too generic so added "helm" back in there). What do you think?

consideRatio:
Excellent I agree @choldgraf ! Is the image okay to use? I picked it from a slide deck without proper understanding of its origin.

choldgraf:
Fine w/ me, I'm +1 on merge - let's leave it open a day or two so others can chime in since it's a relatively prominent file

choldgraf:
🚀 thanks @consideRatio !!!

PR #1122, opened 2019-01-27
Bump k8s/helm requirement to 1.11 / 2.11

consideRatio:
The policy of helms chart repository is to support two versions of k8s
backwards. Since this chart relies on 1.11 for various optional features
for this release, that are still quite essential for the release, I
think we should declare this version require k8s 1.11 now that k8s 1.13
is out. Regarding the Helm version, 2.11 is the first version supporting
k8s 1.11.


consideRatio:
I did not add a test for k8s 1.11, as I ran into an issue, see #1123.

manics:
Looks good to me! I've got a fix for crictl in 1.11: https://github.com/consideRatio/zero-to-jupyterhub-k8s/pull/6

consideRatio:
@manics :heart: excellent work! Thank you so much helping out with this!

PR #1129, opened 2019-01-30
Update setup-jupyterhub.rst

alanjcastonguay:
`describe` or `--output=wide`, not both.

The command as-presented doesn't work. It sounds like the author wanted the output of `describe`, so removed `--output=wide`.


yuvipanda:
Thank you for your PR, @alanjcastonguay! Hope to see more <3 :)

Issue #1125, opened 2019-01-28
Where (and if) to put documentation for the following


hnykda:
Hello,

the last week, I used this awesome stack in my team and it works great. Nevertheless, there were some non-trivial findings/steps I had to do and I am unsure if it's worth mentioning in the docs or no (because the docs are good in how they are concise). I can't take all the credit, my colleagues helped me, but that's irrelevant for the contribution.

1. `GCE Ingress`. First of all, #1004  must be used. Additionally, I wasn't able to make ingress work if I didn't added the default backend, so this:
```
{{- if .Values.ingress.enabled -}}
apiVersion: extensions/v1beta1
kind: Ingress
...
spec:
  backend:
    serviceName: proxy-public
    servicePort: 80
  rules:
    {{- range $host := .Values.ingress.hosts }}
    - host: {{ $host | quote }}
      http:
        paths:
          - path: {{ $.Values.hub.baseUrl }}{{ $.Values.ingress.pathSuffix }}
            backend:
              serviceName: proxy-public
              servicePort: 80
    {{- end }}
...
```
and in config then:

```
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.global-static-ip-name: juphub-static-ip
    kubernetes.io/ingress.class: gce
    kubernetes.io/ingress.allow-http: "false"
  hosts:
  - ...hostname...
  pathSuffix: "*"
  tls:
  - hosts:
    - ...hostname...
    secretName: jupyterhub-tls
proxy:
  https:
    enabled: true
    type: offload
  service:
    type: NodePort
```
2. profile list is so awesome feature that it would deserve some more examples, the below is mine. A sad thing is that [kubespawner](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py) documentation is broken ([nothing shows up in the readthedocs](https://github.com/jupyterhub/kubespawner/issues/291)) and one must dive into the code to find corresponding settings.
 
```
singleuser:
  image:
    name: ...image...
    tag: ...tag...
  profileList:
  # there is some DRY violations below
  - display_name: data tools  [Version 0.1.1]
    description: 'Some our stuff'
    default: true
  - display_name: data tools  [Version latest]
    description: 'Some our stuff'
    kubespawner_override:
      # image pull policy for latest tag is automatically always: 
      # https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L376
      image: ...image:latest...
[...]

PR #1136, opened 2019-02-02
GCE ingress improvements and documentation extension on github and profile

hnykda:
# Attempt to summarize by @consideRatio

### About using the GCE Ingress Controller

This PR adds documentation (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1136/commits/331f1cbf3045b837481fe50ccd588e4e919c3e74) and two lines of code (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1136/commits/87380cc99b1fee75e3f67eb94ef78aa35ba1f147) to use the GCE Ingress Controller to manage the optional ingress k8s resources that the z2jh helm chart can provide.

### Additional parts (LGTM)

This PR also clarifies setup of GitHub authentication (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1136/commits/60268f5581b4f25a7587fc5e65357502be7c15e3) and adds an example to use the kubespawner's profile list feature (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1136/commits/63fd107faffa30584a9fe0fb762121495585beba).

---

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1125

### DIFF
--- a/doc/source/advanced.md
+++ b/doc/source/advanced.md
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
 controller, [**kubernetes/ingress-nginx**](https://github.com/kubernetes/ingress-nginx).
 Note that Nginx maintains two additional ingress controllers.
 For most use cases, we recommend the community maintained **kubernetes/ingress-nginx** since that
-is the ingress controller that the development team has the most experience using.
+is the ingress controller that the development team has the most experience using. See below for
+**GCE Ingress** settings.
+
+#### GCE Ingress and manual HTTPS
+To set up HTTPS with GCE Ingress, you must:
+1. offload the proxy, enable `https` (and ideally, change it from 
+ external `LoadBalancer` to close it from outside world - we have ingress for that)
+2. create TLS secrets same as for other Ingresses (e.g. [nginx](https://kubernetes.github.io/ingress-nginx/user-guide/tls/) )
+3. reference the secret in the ingress spec and add `pathSuffix: *`
+
+```yaml
+proxy:
+  https:
+    enabled: true
+    type: offload
+  service:
+    type: NodePort
+ingress:
+  enabled: true
+  annotations:
+    # kubernetes.io/ingress.global-static-ip-name: YOUR-STATIC-IP-NAME
+    kubernetes.io/ingress.class: gce
+    kubernetes.io/ingress.allow-http: "false"
+  hosts:
+  - example.com  # change it to you domain
+  pathSuffix: "*"
+  tls:
+  - hosts:
+    - example.com
+    secretName: jupyterhub-tls
+```
+*Note*: Ingress must have a default backend, it can be equivalent to the default jupyterhub path.
+
+The default timeout for a GLB backend is 30 seconds, which is not ideal for 
+websocket connections. Currently the timeout must be increased manually after the first deployment:
+```
+kubectl describe ingress $INGRESS_NAME  # find the backend name
+gcloud compute backend-services update k8s-be-99999--9999999999xxxxxx --global --timeout=86400
+```
 
 ### Ingress and Automatic HTTPS with kube-lego & Let's Encrypt
 
--- a/doc/source/authentication.rst
+++ b/doc/source/authentication.rst
@@ -76,8 +76,8 @@ GitHub organisations and users.
 For example, omitting the scope means members of an organisation must `set
 their membership to Public
 <https://help.github.com/articles/publicizing-or-hiding-organization-membership/>`_
-to login, whereas setting it to ``read:org`` may require approval of the
-application by a GitHub organisation admin.
+to login, whereas setting it to ``read:org`` (which is needed to let non-admin organization members
+sign in) may require approval of the application by a GitHub organisation admin. 
 Please see `this issue
 <https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/687>`_ for further
 information.
--- a/doc/source/user-environment.rst
+++ b/doc/source/user-environment.rst
@@ -313,7 +313,9 @@ Each configuration is a set of options for `Kubespawner <https://github.com/jupy
 which defines how Kubernetes should launch a new user server pod. Any
 configuration options passed to the `profileList` configuration will
 overwrite the defaults in Kubespawner (or any configuration you've
-added elsewhere in your helm chart).
+added elsewhere in your helm chart). All `kubespawner` options can be found in the source
+code of `kubespawner <https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L79>`_
+class. 
 
 Profiles are stored under ``singluser.profileList``, and are defined as
 a list of profiles with specific configuration options each. Here's an example:
@@ -333,8 +335,9 @@ a list of profiles with specific configuration options each. Here's an example:
 The above configuration will show a screen with information about this profile
 displayed when users start a new server.
 
-Here's an example with two profiles that lets users select the environment they
-wish to use.
+Here's an example with three profiles that lets users select the environment they
+wish to use, where environment can range from different image to spawning 
+on a different node with different requirements or limits.
 
 .. code-block:: yaml
 
@@ -355,6 +358,17 @@ wish to use.
          description: "The Jupyter Stacks spark image!"
          kubespawner_override:
            image: jupyter/all-spark-notebook:2343e33dec46
+       - display_name: "Spark environment on High Mem Node with taints"
+         description: "The Jupyter Stacks spark image on high mem node!"
+         kubespawner_override:
+           image: jupyter/all-spark-notebook:2343e33dec46
+           node_selector:
+             your_label: my_high_mem_node_specifier
+           tolerations:
+             - key: "resilent"
+               operator: "Equal"
+               value: "true"
+               effect: "NoSchedule"
 
 This allows users to select from three profiles, each with their own
 environment (defined by each Docker image in the configuration above).
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -12,6 +12,9 @@ metadata:
     {{- end }}
   {{- end }}
 spec:
+  backend:
+    serviceName: proxy-public
+    servicePort: 80
   rules:
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host | quote }}


consideRatio:
Thank you @hnykda! Seems great from a quick glance, I will try find time to review this properly, but I'm quite swamped right now.

On question lingers in my mind: is it possible to reference a servicePort by name rather than by its actual port number from the ingress? I have also some stuff to read up on regarding this.

hnykda:
Good question. I have never seen it in any examples, reference API isn't really exhaustive on this ([there is no type specified](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#ingressbackend-v1beta1-extensions)). So don't know.

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
+proxy:
+  https:
+    enabled: true
+    type: offload
+  service:
+    type: NodePort
+ingress:
+  enabled: true
+  annotations:
+    # kubernetes.io/ingress.global-static-ip-name: YOUR-STATIC-IP-NAME
```
minrk: Is this static ip annotation required? If so, it might make sense to mention the steps required to get it

hnykda: I don't think it is required, you should get some automatically assigned, don't you? (I tested that only with my reserved address though).

hnykda:
Thanks for testing! Did you wait at least ~5 minutes? It takes some time, I was also surprised how long it takes.

choldgraf:
hey all - friendly ping on this one...@minrk or @consideRatio did waiting a bit improve the original problems getting this to work?

hnykda:
Hey here :wave: . Is there anything I can do to get this merged? This with #1112 is all we need to get merged to use this deployment on GCP without issues. Right now, we must vendorize the charts :-( .

cc @consideRatio

doc/source/advanced.md
consideRatio: To be clear, the `YOUR-STATIC-IP-NAME` refer to the name of a reserved ip address rather than the actual IP?

Btw, to reserve IPs on GCP, see: https://cloud.google.com/compute/docs/ip-addresses/reserve-static-external-ip-address

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
+    kubernetes.io/ingress.allow-http: "false"
+  hosts:
+  - example.com  # change it to you domain
+  pathSuffix: "*"
+  tls:
+  - hosts:
+    - example.com
+    secretName: jupyterhub-tls
+```
+*Note*: Ingress must have a default backend, it can be equivalent to the default jupyterhub path.
```
consideRatio: The GCE Ingress Controller may require this, but I don't require this on my setup with the ingress-nginx-controller.

The key questions about this PR lies here I think.
1. In what way does the GCE ingress controller fail if we lack this default backend?
2. How did you learn about 1? Is there documentation about this?
3. Will adding a default backend influence the function of the ingress-nginx controller?

I'll investigate further, right now, about what I can learn about these things.

consideRatio:
```suggestion
*NOTE:* Kubernetes ingress resources managed by the GCE Ingress Controller must have a default backend, it can be the same as the default jupyterhub path.
```

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
+    type: NodePort
+ingress:
+  enabled: true
+  annotations:
+    # kubernetes.io/ingress.global-static-ip-name: YOUR-STATIC-IP-NAME
+    kubernetes.io/ingress.class: gce
+    kubernetes.io/ingress.allow-http: "false"
+  hosts:
+  - example.com  # change it to you domain
+  pathSuffix: "*"
```
consideRatio: I learned:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/03127a9e351ed3ffca99969ef6f9a2b115b46516/jupyterhub/templates/ingress.yaml#L14-L24

So, this will influence the rule of what to match, by saying: match everything. Hmmm, this is the default behavior for ingress-nginx I believe. I mean, I arrive at the destination even though i request `https://myhubdomain.com/hub/home` for example. Is this another GCE Ingress Controller vs nginx-ingress-controller deviation?

doc/source/advanced.md
consideRatio: Regarding the default backend: https://github.com/kubernetes/ingress-gce/blob/master/docs/faq/gce.md#why-do-i-need-a-default-backend

consideRatio: This may be useful:

```
gcloud compute http-health-checks list
```

consideRatio: https://github.com/kubernetes/ingress-gce/blob/master/docs/faq/gce.md#can-i-configure-gce-health-checks-through-the-ingress

There is a broken link that should point to this deleted documentation, because it was supposedly covered someplace in GCP now, https://github.com/kubernetes/ingress-gce/tree/e906f93959ede5a745ead5c7330a27e5f29a5e2c/examples/health-checks

consideRatio:
# Review summary
@hnykda thank you so much for your work on this!

I tried reading up to understand more details about using the GCE Ingress Controller to act on the Kubernetes Ingress resources, which are pointless without a controller to consider them. I've only used the z2jh helm chart's ingress resource along with the nginx-ingress controller and want this PR to make everything compatible with that as well.

The key questions I ended up with are:
- Why is an explicit default backend needed over the if-not-provided default backend?
- The if-not-provided default backend by the GCE Ingress Controller would give us a 404 response, isn't that to be preferred as a default routing endpoint also for us, if we explicitly need to define another default backend?
- Why do we need to adjust the rule to `/*` rather than `/` by using the Helm chart value `ingress.pathSuffix: "*"`?

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
 Note that Nginx maintains two additional ingress controllers.
 For most use cases, we recommend the community maintained **kubernetes/ingress-nginx** since that
-is the ingress controller that the development team has the most experience using.
+is the ingress controller that the development team has the most experience using. See below for
+**GCE Ingress** settings.
+
+#### GCE Ingress and manual HTTPS
+To set up HTTPS with GCE Ingress, you must:
+1. offload the proxy, enable `https` (and ideally, change it from 
+ external `LoadBalancer` to close it from outside world - we have ingress for that)
```
consideRatio: You are suggesting a change from `LoadBalancer` to `NodePort`. Do we need `NodePort` over `ClusterIP`?

Yikes it becomes quite complicated to figure things out with so many parts in play.

hnykda: I believe it is: https://github.com/kubernetes/kubernetes/issues/26508#issuecomment-222376962.

doc/source/advanced.md
hnykda: Exactly.

This could be omitted (hence the comment) but I guess that it is quite common usecase that people want to expose it.

hnykda:
Thanks for the great review, I totally understand it's a bit rough and may need more love. The differences between nginx-controller and gce-controller have been driving us crazy in different projects, not just here. GCE controller has sometimes weird behaviour and often lacks some basic features such as configurable healthchecks or http -> https redirect :roll_eyes: ...

I am feeling dumb, but I can't reliably answer your questions. Not that I didn't try, but I couldn't find out why that's happening. @dn0 any chance you may help us out here? 

I am happy to experiment to test some hypothesis, how could we scope it down?

----
P.S. To make our deployment websocket friendly, we added `backendconfig` for proxy:
```
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: proxy-public
spec:
  timeoutSec: 86401
  connectionDraining:
    drainingTimeoutSec: 60
```
but that's probably for another PR

doc/source/advanced.md
consideRatio: Ah - let's find a place to link to this for more details:
https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer#step_5_optional_configuring_a_static_ip_address

consideRatio:
@hnykda right now we both fail to answer but I'll try choose not to feel dumb about it ;D This requires plenty of knowledge about software that we have not developed (the gce ingress controller for example), lots of networking knowledge, kubernetes knowledge, helm knowledge, even nginx-ingress-controller as we want things to work for that as well, perhaps TLS termination stuff, this specific helm chart, etc... This is like biology if physics would be something like understanding of how a routing rule works!

hnykda:
Ok! Let me try not to change the default backend.

hnykda:
So I tried. I removed that default backend from the ingress template, deleted the ingress resource from the cluster and redeployed. 

This is what I end up with:
image
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/backends: '{"k8s-be-31963--9410637bf984f3ee":"Unknown","k8s-be-32567--9410637bf984f3ee":"Unknown"}'
    ingress.kubernetes.io/https-forwarding-rule: k8s-fws-jupyterhub-jupyterhub--9410637bf984f3ee
    ingress.kubernetes.io/https-target-proxy: k8s-tps-jupyterhub-jupyterhub--9410637bf984f3ee
    ingress.kubernetes.io/ssl-cert: k8s-ssl-ffd98441dea4ce05-9723e23028b310ec--9410637bf984f3ee
    ingress.kubernetes.io/url-map: k8s-um-jupyterhub-jupyterhub--9410637bf984f3ee
    kubernetes.io/ingress.allow-http: "false"
    kubernetes.io/ingress.class: gce
    kubernetes.io/ingress.global-static-ip-name: jupyterhub-staging
  creationTimestamp: "2019-04-19T14:16:08Z"
  generation: 1
  labels:
    app: jupyterhub
    chart: jupyterhub-0.9-03127a9
    component: ingress
    heritage: Tiller
    release: juphub
  name: jupyterhub
  namespace: jupyterhub
  resourceVersion: "21537937"
  selfLink: /apis/extensions/v1beta1/namespaces/jupyterhub/ingresses/jupyterhub
  uid: ad253d26-62ad-11e9-a7b7-42010a840038
spec:
  rules:
  - host: mydomain.com
    http:
      paths:
      - backend:
          serviceName: proxy-public
          servicePort: 80
        path: /*
  tls:
  - hosts:
    - mydomain.com
    secretName: jupyterhub-tls
status:
  loadBalancer:
    ingress:
    - ip: XXX
```
These are the healthchecks and LBs:
image

If I change the default backend healthcheck which is the failing one (because it tries the default `/healthz` which in case of jupyterhub just redirect to 404 `/hub/healthz`) to `/hub/login`, it starts to work even without the default backend. Sadly, this health check can be only changed manually in the GCP, not through an API. I am fine with mentioning this in the docs instead of changing it.

consideRatio:
Ah! Excellent investigation!

Then I think what we should do is:

1. Make this PR to document whats going on that we learned
3. Get #1246 merged, as a prerequisite for...
3. ... getting #1004 merged in a way that solves the issue!

hnykda:
Cool. Happy to continue, as you know, #1004 is something I am interested in as well... By "make this PR to document" you mean extending the documentation somehow? Happy to do that.

consideRatio:
@hnykda oh I meant kinda what this PR already is, but add some notes about what we have picked up.

Attempted summary of my current understanding
- Document that GCE Ingress Controller will make a health check apart from the internal k8s health checks, defaulting to /healthz, and that needs to return a OK status 200 code.
- Remove docs / code about the default backend requirement
- Clarify if we needed the `/` -> `/*` change coming from `ingress.pathSuffix` for use with the GCE Ingress Controller if we had no default backend, if that is the case, we may consider making it the default as I think it will be implicitly the same no matter what for the nginx-ingress-controller.
- Link to the "how to reserve an IP on GCP" somewhere with the description on how to configure the K8s ingress through config.yaml for use with the GCE Ingress Controller ( https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer#step_5_optional_configuring_a_static_ip_address )
-

hnykda:
OK. Can I just wait until we have the `/healthz` released (alpha is fine) and then I can test that `/*` and how it all plays without the default backend?

consideRatio:
@hnykda sounds great!

hnykda:
Any chance someone could ping me when it's done or is there a way how I can find myself?

consideRatio:
@hnykda I'll ping you!

dn0:
@hnykda You can use a check similar to this: https://github.com/helm/charts/blob/master/stable/drone/templates/ingress.yaml#L24 to make the ingress more universal.

doc/source/advanced.md
minrk: If I read correctly, the ingress controller hardcoding of `/healthz` is for the ingress's *default* backend, which isn't relevant to the Hub's health check endpoint.

consideRatio: Hmmm but the default backend will not respond if you have an ingress rule, right?

Hmmm.... This warps my brain.

doc/source/advanced.md
tmshn: Yes, they behaves differently. See my following pr: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/988

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
+  service:
+    type: NodePort
+ingress:
+  enabled: true
+  annotations:
+    # kubernetes.io/ingress.global-static-ip-name: YOUR-STATIC-IP-NAME
+    kubernetes.io/ingress.class: gce
+    kubernetes.io/ingress.allow-http: "false"
+  hosts:
+  - example.com  # change it to you domain
```
tmshn: super nits: typo `you` → `your`

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
...
 For most use cases, we recommend the community maintained **kubernetes/ingress-nginx** since that
-is the ingress controller that the development team has the most experience using.
+is the ingress controller that the development team has the most experience using. See below for
+**GCE Ingress** settings.
+
+#### GCE Ingress and manual HTTPS
+To set up HTTPS with GCE Ingress, you must:
+1. offload the proxy, enable `https` (and ideally, change it from 
+ external `LoadBalancer` to close it from outside world - we have ingress for that)
+2. create TLS secrets same as for other Ingresses (e.g. [nginx](https://kubernetes.github.io/ingress-nginx/user-guide/tls/) )
```
tmshn: IMO, it's better to note that you can use [Google manged certificates](https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs#setting_up_the_managed_certificate) or [pre-shared certs (not k8s resource)](https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-multi-ssl#specifying_certificates_for_your_ingress) instead of k8s secrets.

zxcGrace:
I updated my config.yaml following the instructions [https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/331f1cbf3045b837481fe50ccd588e4e919c3e74?diff=unified](url). I didn't put the my static ip name by the way. 

However, once I updated the config.yaml, I couldn't open the jupyterhub page (I waited for more than 5 minutes). When I opened the page on Safari, it showed "Safari can't open the page because the server where this page is located isn't responding". Can anyone help me out on that?

doc/source/advanced.md
```diff
@@ -39,7 +39,45 @@ We recommend the community-maintained [nginx-ingress](https://github.com/helm/ch
 controller, [**kubernetes/ingress-nginx**](https://github.com/kubernetes/ingress-nginx).
 Note that Nginx maintains two additional ingress controllers.
 For most use cases, we recommend the community maintained **kubernetes/ingress-nginx** since that
-is the ingress controller that the development team has the most experience using.
+is the ingress controller that the development team has the most experience using. See below for
+**GCE Ingress** settings.
+
+#### GCE Ingress and manual HTTPS
+To set up HTTPS with GCE Ingress, you must:
+1. offload the proxy, enable `https` (and ideally, change it from
```
tmshn: No need to enable https, it's handled at Ingress rather than `proxy`.

By setting `enabled: true, type: offload`, the `proxy` service

- opens port 443, but _ingress_ [never send request to it](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/eebc19c3c4a5e7d3da68b026c31ebd99d0f048b1/jupyterhub/templates/ingress.yaml#L23)
- is expected to terminate SSL, but GKE _service_ has no such feature

so it's useless.

ref: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/security.md#off-loading-ssl-to-a-load-balancer

<details>
<summary>Click here to see how `proxy` service works for each https type</summary>

Legend: `[*]` is where SSL termination happens.

## enabled: false

```
  svc: proxy                                          pod: proxy
  ┌────────────┐                                      ┌────────────┐
─ 80 ──────────┼───────────────────────────────────── 8000         │
  │            │                                      │            │
  └────────────┘                                      └────────────┘
```

## enabled: true, type: manual/secret

```
  svc: proxy                                          pod: proxy
  ┌────────────┐                                      ┌────────────┐
─ 80 ──────────┼───────────────────────────────────── 8000         │
─ 443 ─────────┼───────────────────────────────────── 8443─[*]→ redirect
  └────────────┘                                      └────────────┘
```

## enabled: true, type: offload

```
  svc: proxy                                          pod: proxy
  ┌────────────┐                                      ┌────────────┐
─ 80 ───────┬──┼───────────────────────────────────── 8000         │
─ 443 ──[*]─┘  │                                      │            │
  └────────────┘                                      └────────────┘
```

## enabled: true, type: auto

```
  svc: proxy                                          pod: proxy
  ┌────────────┐                                      ┌────────────┐
─ 80 ──────────┼─┐                                  ┌ 8000         │
─ 443 ─────────┼┐│ pod: autohttps    svc: autohttps │ │            │
  └────────────┘││ ┌────────────┐    ┌────────────┐ │ └────────────┘
                │└ 80 ───────┬──┼─── 8000 ────────┼─┘
                └─ 443 ──[*]─┘  │    │            │
                   └────────────┘    └────────────┘
```

</details>

hnykda:
@zxcGrace 5 minutes may not be enough, google recommends 10-20 (don't ask me why :roll_eyes: ).

doc/source/advanced.md
zxcGrace: Are you saying that I should remove **enabled: true, type: offload** in the **proxy** section and set up https following the directions below? 
ref: [https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/security.md#off-loading-ssl-to-a-load-balancer](url)

zxcGrace:
@hnykda  I waited for more than 30 min and it is still not working. My config.yaml file looks like this:

```
proxy:
  secretToken: MY-SECRET-TOKEN
  https:
    hosts:
        - MY-HOST-NAME
    letsencrypt:
      contactEmail: xczeng@ucdavis.edu
    service:
      type: NodePort
ingress:
  enabled: true
  annotations:
   # kubernetes.io/ingress.global-static-ip-name: YOUR-STATIC-IP-NAME
    kubernetes.io/ingress.class: gce
    kubernetes.io/ingress.allow-http: "false"
  hosts:
  - example.com  # change it to you domain
  pathSuffix: "*"
  tls:
  - hosts:
    - example.com
    secretName: jupyterhub-tls
```
Is there anything I did wrong?

doc/source/advanced.md
tmshn: If you want to use GCP's HTTPS load balancer,

> remove **enabled: true, type: offload** in the **proxy** section

Yes. (actually no problem to set that, but it's pointless)

> set up https following the directions below?

No, that direction is not for you. It's for who want to set up https using service of `type: LoadBalancer`, such as AWS ELB.

Note that k8s has 2 ways of representations of load balancers:

1. Ingress
    - e.g.: GCP's HTTP(S) LB
2. Service of `type: LoadBalancer`
    - e.g.: AWS's ELB/NLB, GCP's TCP LB

zxcGrace: Is there any specific instructions or any documentation that I can follow in order to build load balancer for k8s? Thanks for the help!

hnykda:
Well, did you change that `example.com` to your domain?

doc/source/advanced.md
hnykda: `build loadbalancer`? What do you mean by that? There is  a service of `type: LoadBalancer` or an `Ingress` resource, which can also be a loadbalancer (usually used for SSL purposes or to have single loadbalancer for many services).

zxcGrace:
I changed that to my domain and it was still not working. I wonder if it is the correct step to do in order to enable load balancer for the Google cloud platform.

hnykda:
And any reports from the ingress? What was it's status?

doc/source/advanced.md
zxcGrace: I mean that I want to let my JupyterHub site to have a loadbalancer. What should I do? I read the section **off-loading SSL to a load balancer** from this website [https://zero-to-jupyterhub.readthedocs.io/en/latest/security.html?highlight=load%20balance](url), but I am still confused on how to set up loadbalancer in the Google cloud platform. Any help would be appreciated.

zxcGrace:
No error reported when I updated my config file with the command 
`RELEASE=jhub
helm upgrade $RELEASE jupyterhub/jupyterhub \
  --version=0.8.0 \
  --values config.yaml`

The only thing went wrong was the site couldn't be opened, and I am not sure where to find reports from the ingress.

doc/source/advanced.md
hnykda: It's not necessary. E.g. if you are in a trusted network (e.g. at home) you don't need SSL (at least I wouldn't bother).


The reason for it for me was that we had SSL handled by Ingress, therefore there is no reason why to let SSL on for anything beyond Ingress. That's why it's off-loading. The `proxy` is by default a proper k8s `Service` resource with type `LoadBalancer`, if that's what you want. Then there is a k8s Ingress resource, which as different type of a LoadBalancer. It's a bit confusing.

hnykda:
"The site couldn't be opened" mean HTTP error 50X or 40X?

consideRatio:
Thank you for your work everyone. I've spent considerable amount of time reviewing and considering this PR already, and upon inspecting it again at this point in time, I consider it to be problematic to merge.

Some key points for my stance:
1. It involves multiple different changes, so while one change may make sense they all need to make sense.
2. It seem to introduce something that breaks k8s schema specification, such as `spec.backend` [for Ingress resources](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressspec-v1-networking-k8s-io).

My understanding summarized:
- It is my understanding that GCE's ingress controller doesn't match paths like `/hub` when `/` is specified as a rule, but only if `/*` is specified as a rule. This is a different behavior than other ingress controllers, which is by itself very problematic in my mind as they should unite behind what a Ingress resource means.
- If GCE's ingress controller still requires something like `/*` to be specified, it can be addressed using #988 in a quite hacky way.

Suggested action point:
If nothing else address the issues, I suggest opening a PR to allow the z2jh ingress resource be configured to add a `defaultBackend` entry to the ingress spec. That way, a wonky GCE ingress controller that requires `/*` rather than `/` can do its job still.

---

Thank you everyone for your work deliberating on this! :heart: Feel free to discuss changes we could make, but for now I'll go ahead and close this PR. I encourage opening an issue with a proposed change or a PR to configure the ingress resources `defaultBackend`.

PR #1137, opened 2019-02-04
Fix typo

manycoding:


manycoding:
My bad, it should be local machine.

PR #1142, opened 2019-02-06
Add support for existing secret for hub

koen92:
With this PR, Jupterhub can be used with an existing secret.

In this we, we don't need to specify secrets in the helm-config file, but we can refer to an existing secret directly.

Such an existing secret can then be managed by another tool like https://github.com/bitnami-labs/sealed-secrets.

This also fixed https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/440 in some way.


clkao:
Thanks for the PR! the template changes look good. can you also add `existingSecret` to both the reference doc and values.yaml comment?

Another issue is the current auto-secret manages auth related secrets. we'll also need docs to tell people to put secrets in the existing secret if it is being used.

consideRatio:
Hi @koen92 sorry for a slow input on this PR and thanks for presenting these ideas!!

Managing secrets is a challenge for sure. Hmmm... This would sidetrack the helm charts updates to the secret file that is mounted on the hub pod, so if various things stored there change, manual updates must be made. This would really lock down how the chart structures content in the K8s Secret resource.

My current understanding is that this kind of change could become unsustainable for the users of the functionality and may guide people to use this that perhaps shouldn't as it comes with these quite big drawbacks. It sure helps me better grasp the complexities of this domain though. Hmmm... I've struggled a long time and seen various other charts struggle with secret management. I've not come across any solution I think is great that does not lock into a certain toolchain.

Yikes I'd love to see a golden reference solution to look at! This is soo hard!

nicorikken:
I've updated the chart together with my colleague @koen92 . Now it features:
@clkao :
- A description in the reference documentation.
- A comment in the values.yml
- An example of the current secret in the reference documentation, and the note that it is up to the administrator to maintain compatibility.

@consideRatio  
- An explicit note in the reference documentation that it is up to the administrator to maintain compatibility.

I hope this non-breaking extension is now good enough to be merged, so we don't have to keep our own fork alive.

nicorikken:
Bump @consideRatio @clkao can you take a look at this?

sgloutnikov:
There is a PR #1426 that was recently merged which addresses existing pull secrets for `hub` and `singleuser`. Should this be now marked as resolved?

nicorikken:
@sgloutnikov thanks for reaching out. No, those are different secrets. The pullSecrets are secrets for pulling a container from a registry. This PR adresses the hub secret, and enables the use of a secret already present in the cluster, rather than rolling out a new one with the chart.
The pattern of ExistingSecrets is already in place in other Helm charts to facilitate solutions for secrets management. We already use our forked repo for that use-case, and like to contribute it upstream.
We still need a review though :wink:

manics:
> The pattern of ExistingSecrets is already in place in other Helm charts to facilitate solutions for secrets management.

@nicorikken to help with review could you point us to another chart (or charts) that you think are good examples?

nicorikken:
That pattern is spread all over the official charts repo: https://github.com/helm/charts/search?q=existingsecret&type=Code

nicorikken:
This PR:
- Has been open for 9 months
- Has been rebased 3 times
- Has received 3 :+1: 
- Only adds a feature common in other Helm charts, not breaking anything

Can it be merged, please?

consideRatio:
@nicorikken thanks for your work on this! :heart: :sunflower:

Sorry for the wait =/

koen92:
Thank you!

nicorikken:
Thanks!!

PR #1146, opened 2019-02-07
use chartpress 0.3

minrk:
will add chart version as prefix to image tags (0.8-abc123 instead of just abc123) making it easier to see which image builds go with which chart version


minrk:
oops, looks like something wrong in chartpress's updated tagging

minrk:
Tagging was fine, but the longstanding issues with not properly checking if images need to rebuilt came up with the updated tagging scheme. Went ahead and fixed that in chartpress 0.3.1

PR #1153, opened 2019-02-14
Better english

manycoding:


choldgraf:
really nice @manycoding , thanks for the improvements, I think this makes it clearer 🎉

PR #1155, opened 2019-02-18
Add a note on restart

manycoding:


choldgraf:
the travis error is on one of 3 builds, and isn't related to this one (it's a timeout)...thanks for another doc improvement @manycoding !

PR #1158, opened 2019-02-21
AWS EFS visibility

manycoding:


manycoding:
We don't want circle dependencies. Even in docs.

manycoding:
#1159

PR #1162, opened 2019-02-26
[WIP] use traefik for the proxy

### LINKED ISSUE #1226: Loss of state in chp container of proxy pod, forces us to restart the hub pod?
I have a nginx-ingress-controller running in my GKE cluster that together with cert-manager manages ingress aspects to my jupyterhub helm chart.

After an upgrade to the ingress-controller but without any change to hub pod, proxy pod, or autohttps pod, I think I caused major failure for my JupyterHub users even after the ingress controller was back online.

### Users experienced

- image (1)
- By refreshing the browsers, they experienced a 503 error page instead.

### Resolving the faulty state

Restarting the hub resolved the issue.

### Analysis

The proxy pod relies on a [chp proxy](https://github.com/jupyterhub/configurable-http-proxy) among other things. So I think that perhaps the CHP state got reset or invalid, and that the hub didn't know to update it.

Why was the CHP state lost or made invalid? I don't know... Hmmm... I don't have logs for this event =/

### Preliminary solution idea

Note, I lack understanding of the CHP project in general, my comprehension can be summarized as "I think it is something that will redirect traffic based on how it is configured by jupyterhub", but how it is configured etc by the jupyterhub I mostly only assume it is done through an API.

Perhaps we could make the state external to the CHP pod, for example in a k8s configmap, and if the autohttps pod restarts, its state will not need to be setup again, because it would simply resetup itself using the configmap. Something like that... Hmmm....

Step one is to pin the issue though.

### LINKED ISSUE #1364: Our proxy pod is a pet, but should be cattle
In kubernetes, there is an analogy that a pod should be treated as cattle as compared to as a pet. The idea is that a pet isn't interchangeable, while cattle is.

Our hub pod communicates with the proxy-api kubernetes service, that will redirect traffic to one of the available proxy pods. But, when our hub pod does so, it actively configures that one pod while it in reality should configure all proxy pods.

## Example issue scenario
Assume that there is not only a single proxy pod for some reason during a time interval. It could be that we want to have high availability (HA) and have made two be running at all time, or because we are making a helm chart upgrade that roll out a new proxy pod, or that the proxy pod crashed for some reason and a new started up.

The hub pod will speak with the `proxy-api` network service that will delegate traffic to one proxy pod, but not all. The hub will say to the proxy pod things like "Hey, when someone requests to access /user/erik they should go to 10.68.3.5!". The hub will also ask "Hey, what routes are you configured with already?", and if the hub concludes a route should be added or removed, it will speak up about that. But but but... The hub doesn't really know who it speaks with, it thinks it speaks with its single pet, but in reality it speaks with its cattle, and it does not try to make sure all cattle behave the same way but instead is focused on a single pet.

## Goals
- I'd like to see that a proxy that starts up, gets up to date directly somehow.
- I'd like to see that our solution allows multiple proxy pods to be kept up to date.

## Pain points
- We let the proxy pod await configuration by the hub pod.
- We communicate directly to a singular proxy pod to update it.

## Related I think
#1226 - I think the proxy pod restarted, and the hub were clueless and didn't update the proxy's state. After this there may have been automatic updates of the proxy pod implemented, so the issue will go away after a while due to this but will still occur briefly.

minrk:
Once this is ready, we'll be able to have multiple proxy replicas, enabling restart/upgrade without any downtime while the proxy restarts.

Tasks:

- [x] remove autohttps, internal ingress, kube-lego
- [x] start etcd (as single container in hub pod, using hub-db-dir for storage)
- [x] replace configurable-http-proxy with traefik
- [x] enable auth for etcd (use existing proxy.secretToken)
- [x] enable auth for traefik api (also use existing proxy.secretToken)
- [ ] test with multiple traefik replicas
- [ ] test automatic provisioning of certificates by meeting letsencrypt challenges
- [ ] test TLS-termination with use of provisioned certificates
- [ ] run hubtraf with https, multiple traefik replicas

cc @GeorgianaElena who's done all the great work implementing https://github.com/jupyterhub/traefik-proxy

closes #1226 
closes #1364


consideRatio:
@minrk @GeorgianaElena woooooo I'm excited about this!

minrk:
This is super close to working (it works for me). The biggest headache is getting authentication setup, because both etcd and traefik work pretty hard to prevent you launching them with simple passwords/token auth set up, which is what we want.

1. etcd requires launching insecurely and then enabling auth via insecure `etcdctl` calls (this appears to be working now), and
2. traefik uses Apache basic or digest auth and there doesn't appear to be a way to put the token into the template in the right format without requiring the user to specify them in their config file after calling `htpasswd` on them. This is not insurmountable, but it's pretty inconvenient. One hurdle is that these use base64-encoded digests and the sprig digest functions output hex (and base64-encoded hex is not the same as base64-encoded original bytes). The Apache Basic auth spec does define a 'plain' unencrypted format that isn't supported everywhere, and traefik is indeed one of the places that it's not supported.

I'm trying to avoid forcing the user to generate more secrets to set up one thing, but I may punt and say that you need to call `htpasswd` to paste it into the form traefik wants. That's pretty annoying, though. The alternative is to rely on network policies to protect the proxy (maybe not the best).

minrk:
I gave up trying to figure out how to enable traefik auth, given only a single token. That doesn't seem to be possible. Instead, users are just going to have to call `htpassword` to get the chart started. I really dislike this, but I don't see any way around it.

yuvipanda:
How about:

1. Make our own traefik image
2. CMD is a Python Script that does the things we need it to do. We can ship htpasswd here.
3. Then it execs to traefik

This should help avoid requiring users to run htpasswd on their own. What do you think?

meeseeksmachine:

https://discourse.jupyter.org/t/server-becomes-stale-in-a-matter-of-days-even-hours/1920/7

manics:
@consideRatio Unfortunately it sounds like Traefik can't save the ACME certificate as a K8S secret, it needs a separately deployed K-V store: https://github.com/containous/traefik/issues/2542

According to https://github.com/containous/traefik/issues/2542#issuecomment-455008630 the best suggestion if you don't want to setup your own K-V store it to use cert-manager.

consideRatio:
@manics thanks for this summary and insight! I read through the issue and comments and yeah it sure seems like they won't implement a good solution to automatically request certificates, but instead will rely on what exist now or delegate this to be done by cert-manager.

I'm conflicted on how to go about things, but I think I favor using cert manager in general at the moment.

consideRatio:
@GeorgianaElena @minrk and I discussed this work in person. @manics comment (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1162#issuecomment-528940013) above was also very relevant in this discussion.

We are aiming to transition to only support certificate provisioning through optional use of cert-manager, as compared to using kube-lego ACME stuff before, or using treafik's ACME stuff.

Somewhat less clear in my mind is discussion and thoughts on where and what kind of TLS termination using these provisioned certificates we would support though. Only through for example nginx-ingress etc or whatever custom method wrapping our deployment?

yuvipanda:
I'm curious re: cert-manager. Can traefik put the certs in the same KV store we already use to manage the routes? Creating a kubernetes secret requires access to k8s API, so we should avoid using it when possible, right? Would love to hear more about the discussion :)

Also <3 for continuing work on this!

manics:
https://docs.traefik.io/user-guide/kv-config/
```
Traefik supports several Key-value stores:

    Consul
    etcd
    ZooKeeper
    boltdb
```
I guess you could write a compatible interface to an existing KV store, though is it reliable enough for Traefik? Will it go down during upgrades?

betatim:
An advantage of not using traefik to deal with certs is that with cert-manager (or kubelego) we can provision certs for any ingress object in the cluster that don't go through the proxy. Storing the certs in etcd or consul feels like a way to tie yourself to traefik instead of playing nice with other k8s native tools. For zero2jupyterhub I think it is Ok to use kubernetes native tools/APIs/patterns :)

consideRatio:
## Use `toToml` on YAML values to generate `treafik.toml`
We currently use the helm chart's template values to configure Traefik by building up a string that becomes the Traefik TOML configuration file. This leads to a lot of conditionals etc. We could instead work with a YAML object that laters renders to a TOML string using the `toToml` helm directive.

If we are doing this, we would pass through a lot of configuration and avoid a lot of complexity by directly writing our configuration values to `traefik.toml`! For this to work all the way through, we need to integrate some of our other helm chart values into the YAML object that should render to the `traefik.toml` file through the `toToml` directive, but this is fine. I've done this already [in the mattermost-team-edition](https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-team-edition/templates/secret-config.yaml) for example where they configure their executable inside a docker image through their Helm chart.

GeorgianaElena:
Finally managed to switch from Etcd to Consul :confetti_ball: Also, it looks like the ET connection problem (communication between Etcd and Traefik took a long time to re-establish when Etcd pod restarted (along with the Hub)) seems to not be reproducing with Consul. @consideRatio, I would really appreciate if could confirm is not reproducing on your side either.

yuvipanda:
I need HTTPS for new hubs at UC Berkeley, and I don't wanna use Cert-Manager :D So I'm going to poke around with this...

sstarcher:
Looking forward to support for Traefik

meeseeksmachine:

https://discourse.jupyter.org/t/enabling-traefikproxy-in-z2jh/3244/2

johnsushant:
@minrk @yuvipanda Is there any chance of this getting picked up in the near future? We need an implementation of proxy for which we can have replicas of in order to HA it. Don't want to start writing manifests myself unless this is not being picked up.

GeorgianaElena:
I think there are a few things that changed since first opening this PR and also some questions we need to answer before moving forward with this.
### What changed
- We replaced the kube-lego + nginx with Traefik v2 for autohttps (more info about this here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1539)
- Traefik v2 
    - offers support for consul and etcd backends since version 2.2 (https://github.com/jupyterhub/traefik-proxy/issues/97#issuecomment-609904879)
    - introduced IngressRoute provider for managing the routing table (https://docs.traefik.io/providers/kubernetes-crd/)

### Questions
- Should we switch to Traefik V2 in the jupyterhub-traefik-proxy implementation now that it offers consul/etcd support and we already use it for autohttps?
  **Possible downside:** version 2 drops support for storing certificates in the key-value stores. Let's Encrypt in an HA setup is still possible, but only though third-party software, like CertManager or potentially by using Traefik Enterprise Edition. Refs:
        * https://github.com/containous/traefik/issues/5792
        * https://docs.traefik.io/providers/kubernetes-crd/#letsencrypt-support-with-the-custom-resource-definition-provider
        * https://community.containo.us/t/confusion-about-tls-certficates-stores-and-how-they-fit-together/3026/3
- Should we add a TraefikIngressRouteProxy to jupyterhub-traefik-proxy and use that in Z2JH instead of a key-value store TraefikProxy?
There are a couple of issues where pros and cons of this were discussed: 
   * https://github.com/jupyterhub/traefik-proxy/issues/85
   * https://github.com/jupyterhub/traefik-proxy/issues/88

   I'll try to summarize them here:
   **Pros:**
       - Kubernetes-native (routing table as CRD objects)
       - No need to manage any extra external services (etcd/consul)
       - High availability support

    **Cons:**
       - The highly available mode can't be used with Let's Encrypt unless Traefik enterprise version is used
       - Create services/endpoints to route to instead of pods

mofanke:
I have a question about this ,when use  traefik for the proxy as HA mode,  can i disiable check_routes  when jupyterhub started or other situations?

srggrs:
What's the status of this? My need is just with the simple toeml implementation, no containers, no consul or etcd. Just want to know if this is ready with simple toeml. Thanks

minrk:
@srggrs the traefik proxy is working great with toml config. This PR is only about integrating it as the default proxy in the helm chart. You can check out [the littlest jupyterhub](https://github.com/jupyterhub/the-littlest-jupyterhub) for an example of deploying traefik proxy with toml config on a single VM.

@mofanke you shouldn't disable check_routes, since that is still a recovery/reconciliation step in case of state changes on JupyterHub's side. You can increase the interval, though, if you think the issues are less likely to occur.

srggrs:
@minrk thank you! Sorry I meant if was ready for traefik v2, but I realised I'm in the wrong issue. I got lost in chasing issues...

lxm:
any progress here?

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -67,6 +70,37 @@ spec:
         {{- if .Values.hub.extraContainers }}
         {{- .Values.hub.extraContainers | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
+        - name: consul
```
chancez, line 73: What if we already run consul? Seems redundant to have this configured, especially since it doesn't have TLS, etc.

chancez:
I feel somewhat mixed about this. I don't really want to run another loadbalancer at all and I already have ingress controllers in my cluster which work just fine, is it possible to instead just use my existing ingress? How does that play into internal SSL?

meeseeksmachine:

https://discourse.jupyter.org/t/large-notebooks-save-fails-with-failed-to-fetch/9022/8

meeseeksmachine:

https://discourse.jupyter.org/t/replace-the-configurable-http-proxy-chp-with-the-traefik-proxy-as-a-cluster-ip-service-using-z2jk/12587/2

Vaibhav1919:
Hi everyone , Is there any update on this PR , i.e. is traefik supported in z2jh? Can i use this PR to create multiple replicas of proxy.

consideRatio:
Hey @minrk! In this PR I recall that we tried replacing autohttps (traefik) + proxy (chp) for a single traefik proxy deployment with X replicas, but concluded that they didn't support multiple replicas in conjunction with integration to acquire certs from Let's encrypt.

Due to this significant realization we arrived at part way through this PR, I'll close this PR as I see no clear path forward. We could go for traefik to replace chp and the jupyterhub proxy class representing traefik to replace the jupyterhub proxy class representing chp, but we wouldn't get the gains we hoped for. If we want to pursue this anyhow, I'd like to open a discussion in an issue at this point I think to overview the updated motivations more clearly.

minrk:
Yes, I think that's fine. I still think we should try to get traefik going, but it's so weird and frustrating that the two biggest selling points for us - native ACME support and multiple replicas, seem to be mutually exclusive!

PR #1163, opened 2019-02-26
Add pdb section to the schema for hub and proxy

jtpio:
Add a section to the doc about the Pod Disruption Budget for the hub and proxy pods.

Would it make sense to also add default values to show that it is enabled by default?

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -396,6 +396,23 @@ properties:
             type: object
             description: |
               Kuberentes annotations to apply to the hub service.
+      pdb:
+        type: object
+        description: |
+          Set the Pod Disruption Budget for the hub pod.
+
+          See [the Kubernetes
+          documentation](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/)
+          for more details about disruptions.
+        properties:
+          enabled:
+            type: boolean
+            description: |
+              Whether PodDisruptionBudget is enabled for the hub pod.
+          minAvailable:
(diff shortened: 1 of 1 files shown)


consideRatio:
Wohooo @jtpio thanks for adding this to `schema.yaml` that will render to the configuration reference in the z2jh.jupyter.org guide!

This LGTM as it is, I'm not sure myself about the default value question. The drawback is maintaining values.yaml with the text here and misleading anyone if we don't successfully do that.

jtpio:
> Wohooo @jtpio thanks for adding this to `schema.yaml` that will render to the configuration reference in the z2jh.jupyter.org guide! heart tada!

Yes that's the goal!

> The drawback is maintaining values.yaml with the text here and misleading anyone if we don't successfully do that.

Agreed. But at the same time the default value can lead to some surprises. For example we learned about the pdb "the hard way" when trying to replace nodes. Also what seems to be mentioned in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/394#issuecomment-463101609.

Maybe just having the pdb listed in the reference (like it is now without the default values) will actually be enough to make it more visible.

consideRatio:
Ah! You make the judgement call =)

I'm happy to merge anytime now that we have considered it.

jtpio:
Alright, let's keep it like this then!

consideRatio:
Another improvement! Wieee :D thanks @jtpio

jtpio:
Thanks @consideRatio!

PR #1168, opened 2019-02-27
Add CA

manycoding:
I am not quite sure how this `ref` works so I just put a usual link.


minrk:
Thanks! I added the necessary ref.

PR #1187, opened 2019-03-11
Add pdb section to schema.yaml for the user scheduler

jtpio:
Same as https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1163 but for the user scheduler.


jtpio:
@consideRatio yes a quick note about the `--force` flag would be a good addition to the docs too!

Issue #1208, opened 2019-03-29
Service name not hard-coded


Jdavid7x:
Hi,

We need to deploy your services with a different name. But now, i saw that is hard-coded on the template/service.yaml.

Is possible to have the name with some function like '_helpers.tpl/fullname'?

I will create a MR for that.

Regards,
JD.

consideRatio:
Thanks for being an active member of this project @Jdavid7x!

As this is a popular topic that has been discussed and considered in many places without clear resolution for a long time, I've decided to close all existing issues and PRs relating to it and refer to a central issue. In this central issue I've done my best to provide a clear and actionable path towards progress.

PR #1209, opened 2019-03-29
[resource prefixing] Service name with fullname function

Jdavid7x:
First commit of the feature to have the service name with the fullname

Related with the Issue #1208

Signed-off-by: Jdavid <jose.david@harbur.io>

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -82,11 +82,20 @@
     function properly. It is a way to override the default app label's value.
 */}}
 
+{{/*
+Create a fully qualified name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+
+{{- define "jhub.fullName" -}}
+{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- end }}
 
 {{- /*
   jupyterhub.appLabel:
     Used by "jupyterhub.labels".
 */}}
+
 {{- define "jupyterhub.appLabel" -}}
 {{ .Values.nameOverride | default .Chart.Name | trunc 63 | trimSuffix "-" }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


manics:
I haven't tested this and Travis is failing, but the idea makes sense, and follows most charts where the service is named after the chart. It changes the name of the service, so I think this should count as a breaking change.

Does anyone else have thoughts on whether the naming should be standardised to match other charts more fully? The template generated by `$ helm-2.12.2 create testchart` is
```
{{/* vim: set filetype=mustache: */}}
{{/*
Expand the name of the chart.
*/}}
{{- define "testchart.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}

{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "testchart.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}

{{/*
Create chart name and version as used by the chart label.
*/}}
{{- define "testchart.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```
and services are named `name: {{ include "testchart.fullname" . }}`

Jdavid7x:
From my point of view will be good if all is with the name standardised functions of helm instead of hard-coded inside the templates.

consideRatio:
Thanks for being an active member of this project @Jdavid7x!

As this is a popular topic that has been discussed and considered in many places without clear resolution for a long time, I've decided to close all existing issues and PRs relating to it and refer to a central issue. In this central issue I've done my best to provide a clear and actionable path towards progress.

PR #1213, opened 2019-04-02
This PR allows to use template call in singleuser values.yaml

mdrio:
This PR allows to use template call in singleuser values.yaml, enabling for example to mount volumes whose name depends on  .Release.Name

### DIFF
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -33,4 +33,4 @@ data:
 {{- $_ := set $values "Chart" (dict "Name" .Chart.Name "Version" .Chart.Version) }}
 {{- $_ := set $values "Release" (pick .Release "Name" "Namespace" "Service") }}
   values.yaml: |
-    {{- $values | toYaml | nindent 4 }}
+    {{- tpl ($values | toYaml | nindent 4) . }}


consideRatio:
Oh this sounds quite clever - could you make it concrete for me by providing an example of values you would pass ?

mdrio:
Here is an example  of a configmap defined in another chart (it is a hdfs HA which is deployed together with jupyterhub) that is mounted on the singleuser notebook,so that the user can access to the current hdfs configuration. 

singleuser:
    storage:
      extraVolumes:
         - name: hdfs-config
           configMap:
            name: "{{ template \"hdfs-k8s.config.fullname\" . }}" # fuilname computed runtime
      extraVolumeMounts:
        - name: hdfs-config
          mountPath: /opt/hadoop/etc/hadoop
          readOnly: true

minrk:
It might be a bit confusing that this templating is only available to the subset of values.yaml applied in the jupyterhub *image* as opposed to anywhere else in the chart. It isn't always obvious where that line is, but you're right that it's most of `singleuser` (not network policy, image credentials, or scheduling inputs such as resource requests, currently).

consideRatio:
Thanks for your PR and thoughts on this @mdrio! I think though it would add too much additional complexity to support something like this.

But, what to do then for your use case? I suggest to instead pass the configmap name to a environment variable to the hub, and then use `hub.extraConfig` to set `c.KubeSpawner.extra_volumes` and `c.KubeSpawner.extra_volume_mounts`.

```yaml
# example of relevant values to be passed to helm (config.yaml)
hub:
  extraEnv:
    CONFIGMAP_NAME: {{ include "hdfs-k8s.config.fullname" . | quote }}
  extraConfig:
    addVolume: |
      import os
      c.KubeSpawner.extra_volumes.append(...) # use os.environ["CONFIGMAP_NAME"]
      c.KubeSpawner.extra_volume_mounts.append(...)
```

PR #1217, opened 2019-04-03
Update user-environment.rst

manycoding:
It took me a while to figure how to actually use other conda envs since usual `conda activate env` doesn't work. This pr makes it more clear.


minrk:
Great!

PR #1231, opened 2019-04-11
Updated notes, pod status to Running

sgloutnikov:
Ready is a PodCondition type, but status returned from kubectl get pods is the phase, which should be Running.


betatim:
Thanks for the fix!

PR #1232, opened 2019-04-12
Added OAuth callback URL to keycloak OIDC example

sgloutnikov:
Had to specify an https callback URL for a keycloak OIDC auth, but didn't find how to do so in the docs. Found it discussed in the issues section with two options:

- OAUTH_CALLBACK_URL environment variables.
- oauth_callback_url in auth.custom.config.


minrk:
Thanks! This *should* be the default oauth callback URL, so it doesn't normally need to be set at all, but sometimes it does when the default cannot be correctly deduced.

sgloutnikov:
I think it was deducing the URL, but pointing to http, and I needed it to point to https. Maybe I missed some setting for that. I had proxy.https.enabled in my config.yaml though that looks like is unrelated.

Issue #1230, opened 2019-04-11
Upload Error Invalid response: 413


julienchastang:
I administer a JupyterHub running this project. I have a user running into the problem below when trying to upload a file.

Large file size warning: The file size is 34 MB. Do you still want to upload it?
When she selects "Upload", she gets the following message:
Upload Error: Invalid response: 413 Request Entity Too Large

I searched to find there were people reporting similar problems when trying to *save* a large notebook. I tried the solutions associated with that problem (e.g., `ingress.kubernetes.io/proxy-body-size: 64m`) but that did not solve the issue.

Ping @zonca 

## Contents of `config.yaml`
```
hub:
  cookieSecret: "xxx"


ingress:
  enabled: true
  annotations:
    kubernetes.io/tls-acme: "true"
  hosts:
    - jupyterhub.example.com
  tls:
      - hosts:
         - jupyterhub.example.com
        secretName: secret

proxy:
  secretToken: "xxx"

singleuser:
  startTimeout: 600
  memory:
    guarantee: 2G
    limit: 2G
  cpu:
    guarantee: 0.5
    limit: 0.75
  defaultUrl: "/lab"
  image:
    name: hub
    tag: ffffff
  lifecycleHooks:
    postStart:
      exec:
          command:
            - "sh"
            - "-c"
            - >
              cp /README_FIRST.ipynb /home/jovyan

hub:
  extraConfig: |-
    c.Spawner.cmd = ['jupyter-labhub']

auth:
  type: github
  github:
    clientId: "xxx"
    clientSecret: "xxx"
    callbackUrl: "https://jupyterhub.example.com:443/oauth_callback"
  admin:
    users:
      - admin
```

betatim:
Hi there 👋!

Can move this discussion to http://discourse.jupyter.org/ (where most of the people who hang out here also hang out, as well as more people). We are trying to streamline the [different discussion places](https://discourse.jupyter.org/t/a-proposal-for-jupyterhub-communications/505) and want to use the issues on GitHub repos for technical discussions on how to change the contents of the repo. More general discussions should go to discourse so that they are easier to find, better indexed by google and generally more accessible than being hidden in the bowls of a GitHub repository ;)

Thanks!

manics:
[...]

PR #1234, opened 2019-04-15
Suggest Discourse in issue template

manics:
See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1230#issuecomment-482305040
> Can move this discussion to http://discourse.jupyter.org/ (where most of the people who hang out here also hang out, as well as more people). We are trying to streamline the different discussion places and want to use the issues on GitHub repos for technical discussions on how to change the contents of the repo. More general discussions should go to discourse so that they are easier to find, better indexed by google and generally more accessible than being hidden in the bowls of a GitHub repository ;)

I thought it'd be useful to mention Discourse in the issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,16 +3,16 @@ Welcome to Zero to JupyterHub!
 
 Before filing an issue, please search through the issues and
 [Discourse forum](https://discourse.jupyter.org/) to see
-if your question has been discussed before. If you
-need more information after searching, feel
-free to message us on the gitter channel. Many
-JupyterHub community members watch the gitter channel
+if your question has been discussed before.
+
+If you have a general question about using this repository, or are
+looking for advice on using and configuring JupyterHub,
+please use the [Discourse forum](https://discourse.jupyter.org/).
+Many JupyterHub community members are members of the forum
 so you will have the benefit of other users' experience
 as well as the JupyterHub team.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,19 +1,20 @@
...
 
-Before filing an issue, please search through the issues to see
+Before filing an issue, please search through the issues and
+[Discourse forum](https://discourse.jupyter.org/) to see
 if your question has been discussed before. If you
 need more information after searching, feel
 free to message us on the gitter channel. Many
 JupyterHub community members watch the gitter channel
 so you will have the benefit of other users' experience
 as well as the JupyterHub team.
```
betatim: should we remove the mention of gitter from here or move it to after mentioning discourse? I think it would help with directing more people towards discourse (instead of gitter) which is what we want to achieve.

betatim:
Looks good to me! I left one comment about removing or moving the mention of gitter.

.github/ISSUE_TEMPLATE.md
manics: I've removed Gitter, and reworded the rest.

betatim:
Thanks a lot for helping out with keeping the gears of the contribution machine well oiled.

consideRatio:
@manics :heart: you are doing so much great stuff!!!

PR #1236, opened 2019-04-15
Change heading of multiple-profiles section

moschlar:
This doesn't really seem like a sub-item of add-files-to-home


minrk:
Thanks!

PR #1238, opened 2019-04-16
test appVersion in Chart.yaml

minrk:
instead of hardcoding the value in the test

### DIFF
--- a/tests/test_spawn.py
+++ b/tests/test_spawn.py
@@ -1,18 +1,27 @@
 import os
-import requests
-import pytest
 import subprocess
 import time
+
+import pytest
+import requests
+import yaml
 
 # Makes heavy use of JupyterHub's API:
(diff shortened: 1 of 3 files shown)


tests/test_spawn.py
```diff
@@ -7,12 +7,19 @@
 # Makes heavy use of JupyterHub's API:
 # http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/docs/rest-api.yml
 
+# load app version of chart
+here = os.path.dirname(os.path.abspath(__file__))
+chart = os.path.join(here, os.pardir, 'jupyterhub', 'Chart.yaml')
+
+with open(chartpress_yaml) as f:
```
consideRatio:
```
tests/test_spawn.py:14: in <module>
    with open(chartpress_yaml) as f:
E   NameError: name 'chartpress_yaml' is not defined
```

PR #1240, opened 2019-04-17
Supply runAsUser at the container level

tmc:
Without this the istio-init initContainer fails.

This is a similar issue as existed in this chart: https://github.com/helm/charts/commit/4daf2d02ed2532572872e8cf6eb4915e5a8af53a


consideRatio:
Thanks! It was great that you provided an example of another repo that did the same thing as well!

It is my understanding that this change would simply shift the requested security context to be container-scoped rather than pod-scoped. In other words, to narrow down the scope rather than broaden it. That should be perfectly fine!

LGTM!

PR #1246, opened 2019-04-18
Bump CHP image to 4.1.0 from 3.0.0

consideRatio:
Could help us get #1004 merged. I created the PR to trigger our CI pipeline, but I'm not able to conclude if it is safe to merge.

DockerHub: https://hub.docker.com/r/jupyterhub/configurable-http-proxy/tags
Changelog: https://github.com/jupyterhub/configurable-http-proxy/blob/master/CHANGELOG.md

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -85,7 +85,7 @@ proxy:
   chp:
     image:
       name: jupyterhub/configurable-http-proxy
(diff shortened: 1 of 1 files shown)


consideRatio:
Hmmm? What has happened to travis?

minrk:
Should be fine. The major version bump is only because of dropping support for old nodejs.

Issue #1023, opened 2018-11-14
config.yaml setting no longer works - possibly due to passthrough update? 


samuelmanzer:
Using this `pullPolicy` setting used to work to get pods to always pull the latest image version: 

```
singleuser:
   image:
     name: [my image]
     tag: latest
     pullPolicy: Always
```

However, it is now ignored, based on inspection of the spawned pods. Back when I submitted [this issue](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/892) the `pullPolicy` value was passed through like this:

```
singleuser.image-pull-policy: {{ .Values.singleuser.image.pullPolicy | quote }}
```

But I think [this PR](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/f20f3e16a222b2e6b2bc7de78760e08c76f1593b) may have changed how things are passed through. This is what now seems to work:

```
singleuser:
  imagePullPolicy: Always
```
Will do a PR for the docs when I have time - there may be other settings with this issue.

manycoding:
@consideRatio I tried several ways and nothing worked for me so far.

I have an image tag `latest` which I update periodically. To avoid unnecessary jhub deploys I set

```
singleuser:
  image:
    name: images.s.com/jupyterhub-app/arche-notebook
    tag: latest
    pullPolicy: "Always"
```

But continuous-image-puller hasn't pulled anything new after the first updated when I set latest.

I also tried:
```
singleuser:
  image:
    name: images.s.com/jupyterhub-app/arche-notebook
    tag: latest
  imagePullPolicy: Always
```

consideRatio:
@manycoding the imagePullPolicy on a pod will be considered only when the pod is starting up, so if it is already running it wont matter if there has arrived a new version of the image. If you add a new node to the cluster, a new continuous-image-puller pod will be added and hence a new pulling process start.

To influence both the hook (pre-upgrade) and continuous image-pullers, you should use the `singleuser.image.pullPolicy` value, at least as of latest chart release.

If we focus on the user pods which is what is really essential, you could try verify what policy they have been set with `kubectl get pod -o yaml jupyter-myuser | grep imagePullPolicy`

consideRatio:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/8ed2f8111b5575dc5df29afb114a8ee5906f9a96/images/hub/jupyterhub_config.py#L103

[...]

PR #1248, opened 2019-04-23
Set the image pull policy from Helm values

vmarkovtsev:
Addresses #1023

Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>


consideRatio:
Thanks for fixing this @vmarkovtsev !

PR #1254, opened 2019-04-26
removing extra buttons

choldgraf:
removes extra buttons from the docs theme


choldgraf:
the doc looks good and tests are happy so I'm gonna merge since it fixes a docs bug!

Issue #1206, opened 2019-03-28
development environment with kind


clkao:
https://github.com/kubernetes-sigs/kind is a way to spawn a kubernetes cluster with docker as the only requirement, and is REALLY fast. it's good for development environment and CI for multiple versions of kubernetes, and is particular useful if you need to test with [local kubernetes changes](https://github.com/kubernetes-sigs/kind#installation-and-usage).  This should be more favourable than a minikube setup for most people, and should get documented.

so, I've managed get fresh z2jh running within 3 minutes:

```
%  kind create cluster --name z2jh
Creating cluster "z2jh" ...
 ✓ Ensuring node image (kindest/node:v1.13.4) 🖼
 ✓ Preparing nodes 📦 
 ✓ Creating kubeadm config 📜 
 ✓ Starting control-plane 🕹️ 
Cluster creation complete. You can now use the cluster with:

export KUBECONFIG="$(kind get kubeconfig-path --name="z2jh")"
kubectl cluster-info

% export KUBECONFIG="$(kind get kubeconfig-path --name="z2jh")"
kubectl cluster-info

Kubernetes master is running at https://localhost:63027
KubeDNS is running at https://localhost:63027/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

% kubectl --namespace kube-system create sa tiller
% kubectl create clusterrolebinding tiller \
       --clusterrole cluster-admin \
       --serviceaccount=kube-system:tiller
% helm init --service-account tiller

% helm upgrade --wait --install --namespace=hub hub jupyterhub/ -f minikube-config.yaml 

% kubectl port-forward -n hub svc/proxy-public 8080:80
```

and viola, you've got z2jh running on localhost:8080

If you wish to use sqlite-pvc, install the local-path provisioner as the builtin host path provisioner doesn't set directory permission right:
```
kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml
```

and change minikube-config.yaml:

```
hub:
  db:
    type: sqlite-pvc
    pvc:
      storageClassName: local-path
```

For ingress (as I am testing keycloak auth which requires a proper domain name), I add the nginx-ingress:

```
helm install --namespace nginx-ingress -n nginx-ingress stable/nginx-ingress --set controller.hostNetwork=true
```

and edit minikube-config.yaml:

```
ingress:
  enabled: true
  hosts:
[...]

PR #1259, opened 2019-04-30
WIP: CI with kind

clkao:
Related to #1206


ci/test.sh
```diff
@@ -4,22 +4,33 @@ set -eux
...
+    IP=$(/sbin/ifconfig $iface | grep 'inet addr' | cut -d: -f2 | awk '{print $1}');
     if [ -n "$IP" ]; then
         echo "IP: $IP"
         break
     fi
 done
 if [ -z "$IP" ]; then
     echo "Failed to get IP, current interfaces:"
-    ifconfig -a
+    /sbin/ifconfig -a
```
minrk: was `/sbin/` not on PATH already? I think it had to be in order to be working.

ci/install-kind.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+set -ex
+
+mkdir -p bin
```
minrk: There's lots of duplication now between install-kind and install-minikube. 

I don't think we need to run tests in both minikube and kind. I think we should either switch to kind or not at all. Does that make sense? What is the value of maintaining and testing multiple bootstrap environments?

clkao: Right, that was also my idea to incrementally remove minikube once we are happy with kind.

However the unreleased kind 0.3.0 switched to the containerd CRI runtime, as suppose to using docker runtime.  I want to bring this up to see if people are comfortable when the CI runs solely on a non-typical kubernetes environment of underlying runtime.

ci/test.sh
clkao: it didn't seem to work in the local environment created by Vagrantfile

consideRatio:
@clkao :heart: thank you for working on this! I'm catching up with kind now!

I'm trying out to use kind as a local dev environment while learning to use it.

I don't see any mention of using `kind load docker-image` in this, yet the test succeed... Hmmmm.... I was forced to do `kind load docker-image jupyterhub/k8s-hub` in order to make the docker image built by chartpress to be available locally.

Is this because this PR didn't trigger chartpress to try rebuild the hub image for example or similar? Hmm... I'm not at all confident about this...

consideRatio:
Relevant: https://github.com/kind-ci/examples/blob/master/.travis.yml

ci/install-kind.sh
consideRatio: I'm 100% comfortable for z2jh CI to run with kind!

manics: I've just tried Kind locally (not using these scripts), it's really easy to get running on a Linux system. Main downside so far is I can't get NetworkPolicies to work.

manics:
In case anyone's interested I tried and failed to get NetworkPolicies working on Kind:
- https://github.com/kubernetes-sigs/kind/issues/842
- https://docs.projectcalico.org/v3.9/getting-started/kubernetes/installation/calico
```yaml
# https://github.com/kubernetes-sigs/kind/blob/v0.5.1/site/content/docs/user/kind-example-config.yaml
# kind create cluster --config kind-config.yaml
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
networking:
  disableDefaultCNI: true
```
```sh
kind create cluster --config kind-config.yaml

curl https://docs.projectcalico.org/v3.9/manifests/calico.yaml | sed -e "s?192.168.0.0/16?10.244.0.0/24?g" | kubectl apply -f -

kubectl wait --for=condition=Ready node/kind-control-plane --timeout=5m
```

ci/install-kind.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# nsenter is included on xenial
+
+# install socat (required by helm)
+sudo apt-get update && sudo apt-get install -y socat
```
consideRatio: It is? But this was not added to the minikube script? I tracked down this to debug why it took so much longer to run the jobs with kind compared to using minikube.

consideRatio:
@manics @clkao I got the network policies running functionally in #1422!

manics:
@consideRatio That's awesome! I'll check it out as soon as I get time.

PR #1261, opened 2019-04-30
WIP: multi-stage CI

### LINKED ISSUE #1260: multi-stage CI
While working on #1259, i found that every PR's env matrix is rebuilding all images using chartpress.  To make things run faster, we should probably use [multi-stage travis](https://docs.travis-ci.com/user/build-stages/) to do the following:

1. tools/templates/lint-and-validate.py
2. chartpress and publish
3. setup k8s env and run tests

Note that only stage 3 should do matrix expansion into multiple jobs for different k8s version/config. and it needs to run chartpress render without building the image, since all stage 3 jobs can pull images built by stage 2.

For stage 2, we might be able to pull images before the current COMMIT_SHA, and this should avoid doing unnecessary image builds if nothing changed in the image subdirectories.

clkao:
Fixes #1260


consideRatio:
Closed in favor of #1422

PR #1263, opened 2019-05-03
install jupyterhub 1.0 final

minrk:


consideRatio:
;D Wieeee!

consideRatio:
Eh... So I enthusiastically merged without looking at the tests, and one had failed due to a common timeout issue we have with the tests. Retesting after merge didn't work.

So I checked in on the outcome here, and all was indeed good: https://travis-ci.org/jupyterhub/zero-to-jupyterhub-k8s/builds/527846000

manics:
@consideRatio I noticed that! GitHub has an option to block merging of PRs without a passing test, you could enable that? Admins can still override.

PR #1272, opened 2019-05-10
enable user scheduler by default

minrk:
I think we've learned through testing of 0.8 that the user scheduler is a success and we can turn it on by default for users with the next release

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -229,7 +229,7 @@ singleuser:
 scheduling:
   userScheduler:
     enabled: true
-    replicas: 1
+    replicas: 2
     logLevel: 4
     image:
       name: gcr.io/google_containers/kube-scheduler-amd64
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -228,7 +228,7 @@ singleuser:
 
 scheduling:
   userScheduler:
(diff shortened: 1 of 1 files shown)


betatim:
Do we know of things that could trip up users? I don't have a good overview what would need to happen if you upgrade an existing cluster or what new things could go wrong during setup. If anyone can remember some adding them to the docs will save us time on support later.

jupyterhub/values.yaml
```diff
@@ -228,7 +228,7 @@ singleuser:
 
 scheduling:
   userScheduler:
-    enabled: false
+    enabled: true
     replicas: 1
```
```suggestion
    replicas: 2
```

I think we should also bump the replicas to two so one can be moved around without loosing availability to schedule pods, they are very cheap on resource requests.

consideRatio: Note that we do have a pdb eabled by default with 1 minAvailable, so right now that could bug people.

consideRatio: Yeah thinking more about this, introducing the user-scheduler by default should go hand in hand with this bump as it could result in unexpected creation of a pod that stops scale downs for core nodes or user nodes if they are not tainted to disallow the scheduler to show up there.

consideRatio: Commit this and merge? =)

PR #1276, opened 2019-05-12
Enable continuous prePuller by default

consideRatio:
Along with #1271 and #1272 I figure its time for this as well, which is very useful for autoscaling to function smoothly.


minrk:
I like this process of

1. adding off-by-default optimizations in a release,
2. learning that these behave well and represent best practice,
3. enabling them by default in a future release to reduce the amount of non-default configuration in a best-practices deployment

PR #1278, opened 2019-05-12
Clean up a line in the CI logs

consideRatio:
This value was previously quoted with ", but when `chartpress` parses this
YAML and writes it back in our CI steps, the following change arise and is
written to the logs which distracted me.

```diff
-    hstsIncludeSubdomains: "false"
+    hstsIncludeSubdomains: 'false'
```

So, by either making this value a boolean or a string with single quote
we get rid of this log entry coming from a `git diff`. I went with
single quotes for a minimalistic change.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -95,7 +95,7 @@ proxy:
       name: quay.io/kubernetes-ingress-controller/nginx-ingress-controller
       tag: 0.15.0
     proxyBodySize: 64m
-    hstsIncludeSubdomains: "false"
+    hstsIncludeSubdomains: 'false'
     resources: {}
   lego:
     image:


yuvipanda:
ty, @consideRatio! It needs to be a quoted value rather than a boolean, since configmap values must be strings.

consideRatio:
@yuvipanda ah but I think since the configmap in question adds them it would had worked with boolean values as well, but I like this as it is made clear that this will per parsed as a string in the end.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/1158a198160fb1a89276046b34cc8d44df0d8195/jupyterhub/templates/proxy/autohttps/configmap-nginx.yaml#L12

PR #1280, opened 2019-05-12
Disable becoming root inside hub and proxy containers

yuvipanda:
We never need anything to become the root user inside
the hub / proxy containers while they are running.
This removes the ability from them, making it harder
for any exploit in the hub / proxy to be used to gain
root on the underlying machine

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -55,8 +55,6 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.hub.fsGid }}
-        # Don't allow any process to execute as root inside the container
-        allowPrivilegeEscalation: false
       {{- if .Values.hub.imagePullSecret.enabled }}
       imagePullSecrets:
         - name: hub-image-credentials
@@ -116,6 +114,8 @@ spec:
           {{- end }}
           securityContext:
             runAsUser: {{ .Values.hub.uid }}
+            # Don't allow any process to execute as root inside the container
+            allowPrivilegeEscalation: false
           env:
             - name: PYTHONUNBUFFERED
               value: "1"
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -55,6 +55,8 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.hub.fsGid }}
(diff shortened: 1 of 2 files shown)


consideRatio:
Hmmm... is there a duplicate entry?

```
### 4/4 - kubeval
The document rendered-templates/jupyterhub/templates/ingress.yaml contains a valid Ingress
The document rendered-templates/jupyterhub/templates/hub/secret.yaml contains a valid Secret
The document rendered-templates/jupyterhub/templates/hub/pdb.yaml contains a valid PodDisruptionBudget
The document rendered-templates/jupyterhub/templates/hub/rbac.yaml contains a valid ServiceAccount
The document rendered-templates/jupyterhub/templates/hub/rbac.yaml contains a valid Role
The document rendered-templates/jupyterhub/templates/hub/rbac.yaml contains a valid RoleBinding
The document rendered-templates/jupyterhub/templates/hub/service.yaml contains a valid Service
The document rendered-templates/jupyterhub/templates/hub/deployment.yaml contains an invalid Deployment
---> allowPrivilegeEscalation: Additional property allowPrivilegeEscalation is not allowed
`kubeval rendered-templates/jupyterhub/templates/hub/deployment.yaml --kubernetes-version 1.11.0 --strict` exited with status 1
The command "./ci/travis-script.sh" exited with 1.
```

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -55,6 +55,8 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.hub.fsGid }}
+        # Don't allow any process to execute as root inside the container
+        allowPrivilegeEscalation: false
```
consideRatio: Ah, this is within the pod, it needs to be within the container I think.

See: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
           secret:
             secretName: {{ .Values.proxy.https.secret.name }}
       {{- end }}
+      securityContext:
+        # Don't allow any process to execute as root inside the container
+        allowPrivilegeEscalation: false
```
consideRatio: Same issue as above.

yuvipanda:
Good call, @consideRatio! I think I've fixed these, let's see.

PR #1288, opened 2019-05-18
Include jupyter-firstuseauthenticator.

danielballan:
It would be convenient to have this authenticator, which is nice model for
JupyterHubs used at workshops, available in the hub image by default.

If this change is acceptable, I would be happy to add mention of this option to
the documentation in this PR.

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -1,5 +1,6 @@
 # jupyterhub version is defined in chartpress.yaml
 jupyterhub-dummyauthenticator==0.3.1
+jupyterhub-firstuseauthenticator==0.12
 jupyterhub-tmpauthenticator==0.5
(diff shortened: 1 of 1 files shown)


consideRatio:
Ah hmm, I'm not confident on how the firstuseauthenticator functions, but will this persist well even in between hub pod restarts? If it is, then I'm up for having this available, if not I think it could be a bit confusing.

minrk:
Yeah, firstuseauthenticator should work fine. It creates a `passwords.dbm` file in the CWD, which will be persisted alongside jupyterhub.sqlite, jupyterhub_cookie_secret, etc.

danielballan:
Thanks!

consideRatio:
Wieeee @danielballan! :tada: :heart:

PR #1295, opened 2019-05-22
WIP: Allow manual definition of service accounts

### LINKED ISSUE #921: Use rbac.create convention ?
Based on my understanding, the templates use `rbac.enabled` to determine if RBAC resources will be created - e.g. https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/image-puller/rbac.yaml

According to https://docs.helm.sh/chart_best_practices/#rbac-resources-should-be-created-by-default, the best practice is to have `rbac.create` for the setting to create RBAC resources.

Other charts such as [nginx-ingress](https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/clusterrolebinding.yaml) do use `rbac.create`.

Would it be a good idea to use this convention ?

gsemet:
We are not allowed to create any service accounts in our cluster, so I need to be able to change the name of the SA used for hub/proxy/...

It tries to fix #921, but there may be additional work to do onit

TODO:
- [ ] Document the breaking change `rbac.enabled` to `rbac.create`
- [ ] Service accounts name does not depends on the the fullname as described in [Helm Best Practices](https://helm.sh/docs/chart_best_practices/#using-rbac-resources). That would also be a breaking change (the `hub` service account would be set to `myjhubreleasename-hub`). I think this is ok, that would ultimately result in being able to deploy several independent jupyterhub in the same namespace
- [ ] there is no `jupyterhub.fullname` template defined :(
- [ ] use default account if `serviceAccounts.*.create` is `false` and the name is not specified

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -43,10 +43,10 @@ properties:
         description: |
           Creates an image pull secret for you and makes the hub pod utilize
           it, allowing it to pull images from private image registries.
-          
+
           Using this configuration option automates the following steps that
           normally is required to pull from private image registries.
-          
+
           ```sh
           # you won't need to run this manually...
           kubectl create secret docker-registry hub-image-credentials \
@@ -108,7 +108,7 @@ properties:
 
               For gcr.io registries the password will be a big JSON blob for a
               Google cloud service account, it should look something like below.
-                            
+
               ```yaml
               password: |-
                 {
@@ -144,7 +144,7 @@ properties:
             type: string
             description: |
               The tag of the image to pull.
-              
+
               This is the value after the `:` in your full image name.
 
               ```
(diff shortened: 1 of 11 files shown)


gsemet:
Note: I do not have much time to spend on finalizing this story, sadly. I expect someone else to take over. I simply need this feature to have jupyterhub on our cluster.

pengye91:
I'm wondering if this PR has been taken over or not, this is actually a desired feature if we'd like to run notebooks for different service account.

consideRatio:
> I'm wondering if this PR has been taken over or not, this is actually a desired feature if we'd like to run notebooks for different service account.

This is something this PR wouldn't address, but you can adjust via `singleuser.serviceAccountName` for example, it would only allow you to replace the service accounts created by the Helm chart with custom service accounts.

---

@gsemet I'm sorry for your experience of not getting review attention in this PR. Personally I've been to conflicted about wishing to do it in the first place given that it introduces complexity etc, even though I'd like to align with helm chart documented practices.

Concrete action point for anyone wanting to champion this change:

1. Help clarify why this can be important today for a real world use case
2. Suggest changes to be made
3. Reach agreement if changes are to be made
4. Implement changes in a PR

PR #1304, opened 2019-06-04
Update chart version to 0.8.2 in the docs

jtpio:
`0.8.2` is the latest chart version so we might as well have it in the docs.


consideRatio:
Thanks @jtpio !! :tada:

Issue #1242, opened 2019-04-17
lifecycle_hooks can not be overridden in "kubespawner_override"


mhuttner:
``` config.yaml
singleuser:
  image:
    name: jupyter/minimal-notebook
    tag: 7d427e7a4dde
  profileList:
    - display_name: Minimal Python Environment
      description: "Only the basic notebook"
      default: true
    - display_name: Other Environment
      description: ""
      kubespawner_override:
        image:  jupyter/r-notebook:latest
        lifecycleHooks:
          postStart:
           exec:
             command: ["/bin/sh", "-c", "echo Hello from overridden handler > $HOME/message"]
  lifecycleHooks:
    postStart:
     exec:
       command: ["/bin/sh", "-c", "echo Hello from the postStart handler > $HOME/message"]
  storage:
    capacity: 5Gi
    dynamic:
      storageClass: nfs-client
  memory:
    limit: 8G
    guarantee: 1G
  cpu:
    limit: 4
    guarantee: .5
```
A configuration like this will always call the default lifecycle hook instead of the one in the profile.

consideRatio:
Ah... but if you do...

````yaml
  profileList:
    - display_name: Minimal Python Environment
      description: "Only the basic notebook"
      default: true
    - display_name: Other Environment
      description: ""
      kubespawner_override:
        image:  jupyter/r-notebook:latest
        lifecycle_hooks:
          postStart:
            exec:
              command: ["/bin/sh", "-c", "echo Hello from overridden handler > $HOME/message"]
  lifecycleHooks:
    postStart:
      exec:
        command: ["/bin/sh", "-c", "echo Hello from the postStart handler > $HOME/message"]
```

Yikes... This is veeeeeeeeeeeeeerrry messy.

There is a mix of:
1. The Z2JH Helm chart syntax for its values (`lifecycleHooks`)
2. kubespawner's configuration options (`lifecycle_hooks`)
3. Kubernetes pod specifications (`postStart`).

When you are writing things within `kubespawner_override` you are writing raw configuration for kubespawner, and must use the snake-case version.

consideRatio:
@clancychilds thank you so much for documenting this and @mhuttner for reporting this experience!

PR #1309, opened 2019-06-11
add explanation for lifecycle_hooks in kubespawner_override

clancychilds:
Adds some information about using lifecycle_hooks in kubespawner_override as mentioned by @consideRatio in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1242

(Would have saved me an hour of poking around!)


consideRatio:
Thank you @clancychilds for helping others get past this hurdle! :Heart: :tada:

PR #1322, opened 2019-07-03
Changing Azure Container Service to Azure Kubernetes Service

seanmck:
This PR updates the Azure setup article to use the current full name of AKS, Azure Kubernetes Service.


consideRatio:
THank you @seanmck !

PR #1327, opened 2019-07-07
Add py-spy to hub image

yuvipanda:
https://github.com/benfred/py-spy is a sampling profiler for Python
that works really solidly with minimum fuss, and has been extremely
helpful in debugging issues. It's also very useful in doing performance
analysis. Having this be present in the hub image by default will
make it easier for everyone to create profiles of the hub process
at loads they are having trouble with, and share it with the community
to see what is up.


minrk:
Nice!

consideRatio:
( The CI Pipeline for the PR failed, related to this being a in-repo-branch-pr, but the actual pipeline triggered on the merge to master ran as it should, so there should be a new helm chart release with this image bumped still )

PR #1337, opened 2019-07-25
Make proxy redirect to 443 instead of 8443

LucidNeko:
configurable-http-proxy redirects to '--port' which in this case is 8443 and unresolvable outside of the cluster.

The master branch now uses version 4.1.0 of configurable-http-proxy which added '--redirect-to' to solve this problem.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -59,11 +59,13 @@ spec:
             {{- if $manualHTTPS }}
             - --port=8443
             - --redirect-port=8000
+            - --redirect-to=443
             - --ssl-key=/etc/chp/tls/tls.key
             - --ssl-cert=/etc/chp/tls/tls.crt
             {{- else if $manualHTTPSwithsecret }}
             - --port=8443
             - --redirect-port=8000
+            - --redirect-to=443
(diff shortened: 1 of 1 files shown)


consideRatio:
I'm reading up about the [command line options for CHP](https://github.com/jupyterhub/configurable-http-proxy#command-line-options).

>  --port <n> (defaults to 8000)      Public-facing port of the proxy
>  --redirect-port <redirect-port>    Redirect HTTP requests on this port to the server on HTTPS
>  --redirect-to <port>               Redirect HTTP requests from --redirect-port to this port

I'm explaining to myself while thinking things through.

If we the CHP accepts HTTPS traffic and will make the TLS termination, traffic on the http:// port should be redirected to the https:// port. But, a redirect is simply a response to the client saying "hey visit this address", but if that address is the container local one as compared to the address to reach it from outside, which requires a passage of the k8s Service and the Pod, then it may not end up in the right spot.

Your change makes a lot of sense to me, I'm sorry you had to experience such a slow response, thank you for your contribution! :heart: :sunflower: @LucidNeko !

meeseeksmachine:

https://discourse.jupyter.org/t/manual-https-how-to-forward-from-http/2362/3

Issue #994, opened 2018-10-25
WIP: A deployment story - Using GPUs on GKE


consideRatio:
# GPU powered machine learning on GKE
I wanted the data science team I work with be able to utilize GPUs on GKE to do high performing ML from their laptops. This is what I needed to do, and this issue could later be turned in to documentation if you help me refine it =)

## 1. Make a docker image
I pushed my build to `consideratio/singleuser-gpu:v0.3.0` if you want to save some time and risk me being malicious.

```Dockerfile
# Dockerfile
# TAG: https://hub.docker.com/r/jupyter/datascience-notebook/tags/
FROM jupyter/datascience-notebook:f2889d7ae7d6

# GPU powered ML
# ----------------------------------------
RUN conda install -c conda-forge --yes --quiet \
    tensorflow-gpu \
    cudatoolkit=9.0 && \
    conda clean -tipsy && \
    fix-permissions $CONDA_DIR && \
    fix-permissions /home/$NB_USER

# drivers etc. installed by the nvidia-driver-installer daemonset
ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/nvidia/lib64
ENV PATH=${PATH}:/usr/local/nvidia/bin
```

```sh
TAG=v0.3.0 && docker build --tag my-docker-username/singleuser-gpu:${TAG} . && docker push my-docker-username/singleuser-gpu:${TAG}
```

## 2. Use a k8s 1.10+ GKE cluster
They have the `nvidia-gpu-device-plugin` daemonset running by default, needed for attaching the GPUs and a custom resource controller plugin enabled by default that will handle extra resource requests properly.

## 3. Configure some spawn options
It is easiest to allow the user choose among some profiles, as one may not always need a GPU node.

```python
      c.KubeSpawner.profile_list = [{
          'display_name': 'Default: Shared, 8 CPU cores',
          'description': 'Your code will run on a shared machine. This will already be running and you should be up and running in 30 seconds or less.',
          'default': True,
          'kubespawner_override': {
              'cpu_limit': '6',
              'mem_guarantee': '1Gi',
              'mem_limit': '4Gi',
          },
      }, {
          'display_name': 'Personal, 2 CPU cores, 1 NVIDIA Tesla K80 GPU',
          'description': 'Your code will run on a computer dedicated to you. No such computer is available ahead of time, so you need to wait 5-10 minutes before it is ready.',
          'kubespawner_override': {
[...]

PR #1358, opened 2019-08-16
Instructions for adding GPUs and increasing shared memory

tlkh:
This adds brief instructions to the docs for how to:

* allocate GPUs to users
* increase the shared memory (SHM) allocation for users, which is beneficial for deep learning workloads and even required for functions like PyTorch's DataLoader to run

I noticed #994, but I also think it's beneficial to start adding basic information and instructions into the documentation so users can have something to refer to as a starting point.

Any feedback or comments appreciated!

### DIFF
--- a/doc/source/user-resources.rst
+++ b/doc/source/user-resources.rst
@@ -73,6 +73,70 @@ This would limit your users to a maximum of .5 of a CPU (so 1/2 of a CPU core), 
 
    Remember to :ref:`apply the change <apply-config-changes>` after changing your ``config.yaml`` file!
 
+Set user GPU guarantees / limits
+--------------------------------
+
+It is possible to allocate GPUs to your user. This is useful for heavier
+workloads, such as deep learning, that can take advantage of GPUs.
+
+For example, to create a profile that allocates one NVIDIA GPU:
+
+    .. code-block:: yaml
+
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me.

Are there any downsides to the increased SHM or weird side effects it could have? If there is a  resource we could link to for "I changed my SHM size and now stuff is breaking" that could be useful.

Should we also link to https://cloud.google.com/kubernetes-engine/docs/how-to/gpus as GKE specific docs for getting GPUs setup on your node pool there? As we find them we could add links for other cloud providers as well.

tlkh:
@betatim good idea!

The only downside I'm aware of is that (and you're right that we should mention this):

* use of SHM counts towards the pod's memory limits
* if the SHM usage exceeds pod's memory limits the pod will be evicted

Otherwise, increasing SHM is recommended by NVIDIA and also is built into Kubeflow's notebook spawner for the same reasons I mentioned. 

I'll add in links for allocating GPUs on GKE, and the equivalents for AWS and Azure.

I'll amend the commit and push over.

tlkh:
Pushed the improvements!

betatim:
Thanks for helping improve the docs and the fast turn around!

Issue #1357, opened 2019-08-15
Readiness/Liveness setup challenge


consideRatio:
I recently merged #1004 that had successful CI tests but then reverted the merge in #1356 after concluding my upgrade failed.

I'm not confident on what goes on. My hub and proxy pod wasn't entering a Ready state after this PR was merged by me. So I'll investigate things further but for now I figure I'll revert the PR so it does not cause disruptions for others like me.

```
Events:
  Type     Reason     Age                From                                         Message
  ----     ------     ----               ----                                         -------
  Normal   Scheduled  78s                default-scheduler                            Successfully assigned jupyterhub/proxy-7d4dcc5c64-pqv9w to gke-ds-platform-core-4d566784-54l4
  Normal   Pulled     76s                kubelet, gke-ds-platform-core-4d566784-54l4  Container image "jupyterhub/configurable-http-proxy:4.1.0" already present on machine
  Normal   Created    76s                kubelet, gke-ds-platform-core-4d566784-54l4  Created container
  Normal   Started    76s                kubelet, gke-ds-platform-core-4d566784-54l4  Started container
  Warning  Unhealthy  10s (x7 over 70s)  kubelet, gke-ds-platform-core-4d566784-54l4  Readiness probe failed: Get http://10.72.0.84:8000/health: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

```
Events:
  Type     Reason     Age               From                                         Message
  ----     ------     ----              ----                                         -------
  Normal   Scheduled  31s               default-scheduler                            Successfully assigned jupyterhub/hub-b9dccf7f7-htrxq to gke-ds-platform-core-4d566784-lspf
  Normal   Pulled     29s               kubelet, gke-ds-platform-core-4d566784-lspf  Container image "jupyterhub/k8s-hub:0.9-dcde99a" already present on machine
  Normal   Created    29s               kubelet, gke-ds-platform-core-4d566784-lspf  Created container
  Normal   Started    29s               kubelet, gke-ds-platform-core-4d566784-lspf  Started container
  Warning  Unhealthy  6s (x3 over 26s)  kubelet, gke-ds-platform-core-4d566784-lspf  Liveness probe failed: Get http://10.72.2.185:8081/health: dial tcp 10.72.2.185:8081: connect: connection refused
[...]

PR #1356, opened 2019-08-15
Revert "Configure liveness/readinessProbe"

consideRatio:
Reverts jupyterhub/zero-to-jupyterhub-k8s#1004

My hub and proxy pod wasn't entering a Ready state after this PR was merged by me. So I'll investigate things further but for now I figure I'll revert the PR so it does not cause disruptions for others like me.

```
Events:
  Type     Reason     Age                From                                         Message
  ----     ------     ----               ----                                         -------
  Normal   Scheduled  78s                default-scheduler                            Successfully assigned jupyterhub/proxy-7d4dcc5c64-pqv9w to gke-ds-platform-core-4d566784-54l4
  Normal   Pulled     76s                kubelet, gke-ds-platform-core-4d566784-54l4  Container image "jupyterhub/configurable-http-proxy:4.1.0" already present on machine
  Normal   Created    76s                kubelet, gke-ds-platform-core-4d566784-54l4  Created container
  Normal   Started    76s                kubelet, gke-ds-platform-core-4d566784-54l4  Started container
  Warning  Unhealthy  10s (x7 over 70s)  kubelet, gke-ds-platform-core-4d566784-54l4  Readiness probe failed: Get http://10.72.0.84:8000/health: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

```
Events:
  Type     Reason     Age               From                                         Message
  ----     ------     ----              ----                                         -------
  Normal   Scheduled  31s               default-scheduler                            Successfully assigned jupyterhub/hub-b9dccf7f7-htrxq to gke-ds-platform-core-4d566784-lspf
  Normal   Pulled     29s               kubelet, gke-ds-platform-core-4d566784-lspf  Container image "jupyterhub/k8s-hub:0.9-dcde99a" already present on machine
  Normal   Created    29s               kubelet, gke-ds-platform-core-4d566784-lspf  Created container
  Normal   Started    29s               kubelet, gke-ds-platform-core-4d566784-lspf  Started container
  Warning  Unhealthy  6s (x3 over 26s)  kubelet, gke-ds-platform-core-4d566784-lspf  Liveness probe failed: Get http://10.72.2.185:8081/health: dial tcp 10.72.2.185:8081: connect: connection refused
  Warning  Unhealthy  1s (x3 over 21s)  kubelet, gke-ds-platform-core-4d566784-lspf  Readiness probe failed: Get http://10.72.2.185:8081/health: dial tcp 10.72.2.185:8081: connect: connection refused
```

PR #1361, opened 2019-08-16
[MRG] Readiness and Liveness probes re-added

consideRatio:
Manages parts of #1357 and lands #1004 finally that I merged and then reverted in #1356.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -181,9 +181,9 @@ spec:
               name: hub
           livenessProbe:
             httpGet:
-              path: /health
+              path: {{ .Values.hub.baseUrl }}hub/health
               port: hub
           readinessProbe:
             httpGet:
-              path: /health
+              path: {{ .Values.hub.baseUrl }}hub/health
               port: hub
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -179,3 +179,11 @@ spec:
           ports:
             - containerPort: 8081
               name: hub
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: hub
+          readinessProbe:
+            httpGet:
+              path: /health
+              port: hub
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -180,6 +180,7 @@ spec:
             - containerPort: 8081
               name: hub
           livenessProbe:
+            initialDelaySeconds: 30
             httpGet:
               path: {{ .Values.hub.baseUrl }}hub/health
               port: hub
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -179,11 +179,18 @@ spec:
           ports:
             - containerPort: 8081
               name: hub
-          livenessProbe:
-            initialDelaySeconds: 30
-            httpGet:
-              path: {{ .Values.hub.baseUrl }}hub/health
-              port: hub
+          ## livenessProbe notes:
+          ## We don't know how long hub database upgrades could take
+          ## so having a liveness probe could be a bit risky unless we put
+          ## a initialDelaySeconds value with long enough margin for that
+          ## to not be an issue. If it is too short, we could end up aborting
+          ## database upgrades midway or ending up in an infinite restart
+          ## loop.
+          # livenessProbe:
+          #   initialDelaySeconds: 30
+          #   httpGet:
+          #     path: {{ .Values.hub.baseUrl }}hub/health
+          #     port: hub
           readinessProbe:
             httpGet:
               path: {{ .Values.hub.baseUrl }}hub/health
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -101,3 +101,11 @@ spec:
               name: proxy-public
             - containerPort: 8001
               name: api
+          livenessProbe:
+            httpGet:
+              path: /_chp_healthz
+              port: proxy-public
+          readinessProbe:
+            httpGet:
+              path: /health # Wait until hub is ready (this is proxied to hub)
+              port: proxy-public
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -107,5 +107,5 @@ spec:
               port: proxy-public
           readinessProbe:
             httpGet:
-              path: /health # Wait until hub is ready (this is proxied to hub)
+              path: /_chp_healthz
               port: proxy-public


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -179,3 +179,12 @@ spec:
           ports:
             - containerPort: 8081
               name: hub
+          livenessProbe:
+            initialDelaySeconds: 30
```
betatim: We set this delay because sometimes it can take the hub process a while to become ready correct? Does this also mean the minimum restart time/dead time is 30s (because the pod will only be considered live if the check succeeds which won't happen until at least 30s after the pod starts). This would be a penalty for small hubs that tend to restart quickly, there are probably many more small hubs than big hubs and small hubs are operated by less experienced operators. All this is a hunch, so we need to discuss if these assumptions are true. If we think they are true should we make it work for the small hub case by default with a note for experts on how to increase the delay?

One thing I'd check is the case where a hub would need (say) 60s to get ready but after 30s it gets restarted because the liveness probe fails. Now when the pod has restarted will it need another 60s to complete getting ready or has the time gotten shorter? If the time gets shorter then eventually the hub will start up within the 30s initial delay we give. If the startup time remains constant the hub will never manage to get live and will constantly be restarted.

If the latter case is true I think we should find a different way to determine liveness, maybe by checking if the hub process is running or something.

Another alternative would be that we remove the liveness probe because in almost all cases an unhealthy hub will crash and exit -> the pod will be restarted. Not sure if this is true but if it is this might be a nice way out of having to determine "how big should the initial delay be".

manics:
Maybe something for a future PR. The CI test script has several workarounds to ensure the hub is ready, and to work around race conditions:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/154913a496b6f8fbaa20f3a3e641c548832ddafa/ci/test.sh#L2
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/154913a496b6f8fbaa20f3a3e641c548832ddafa/ci/test.sh#L50
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/154913a496b6f8fbaa20f3a3e641c548832ddafa/ci/test.sh#L54

Presumably all these sections can be removed and replaced with just one loop that checks the deployment status is ready?

betatim:
I left one big comment on the liveness probe for the hub. The other probes look good to me.

jupyterhub/templates/hub/deployment.yaml
manics: Does keeping the delay short but increasing the `failureThreshold` work? https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

manics: Does keeping the delay short but increasing the `failureThreshold` work? https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

consideRatio: > We set this delay because sometimes it can take the hub process a while to become ready correct?

Yepp!

> Does this also mean the minimum restart time/dead time is 30s (because the pod will only be considered live if the check succeeds which won't happen until at least 30s after the pod starts)

No the liveness probe adds another way to make the pod restart, but does not remove the ordinary way to crash and restart for that reason.

__About infinite restarts__
> One thing I'd check is the case where a hub would need (say) 60s to get ready but after 30s it gets restarted because the liveness probe fails. Now when the pod has restarted will it need another 60s to complete getting ready or has the time gotten shorter? If the time gets shorter then eventually the hub will start up within the 30s initial delay we give. If the startup time remains constant the hub will never manage to get live and will constantly be restarted.
>
>If the latter case is true I think we should find a different way to determine liveness, maybe by checking if the hub process is running or something.

I ended up noticing that there was a synchronous poll on previous user pods status, and that could fail and take 30 seconds to timeout. That caused liveness probe failures for me. This is why I reverted the original PR and now recreates it with this initial setup delay added. To be safe we could make it 60.

> Another alternative would be that we remove the liveness probe because in almost all cases an unhealthy hub will crash and exit -> the pod will be restarted.

With the current setup, we would crash on startups slower than 1 minute. If that happens, I would consider that a big failure, because the probe really match specifically to the hub container within the pod as compared to the full pod with potential init_containers etc.

Hmm... but... we have the hub upgrade situation, that should perhaps be allowed to take a while... Yeah I'm clueless on the time that may be required overall for that in worst case scenarios but I wouldn't want to mistakenly restart the hub pod during an upgrade.

I think due to the hub db upgrade complexity added, we should remove the livenessProbe of the hub, but keep the readinessProbe.

> (@manics) Does keeping the delay short but increasing the failureThreshold work? https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

It would also work to increase the delay until restart, but it would restart slower if the livenessProbe started failing during normal operation, and we mostly wanted to increase it during normal operation.

consideRatio:
@manics ah yes nice we could await for the hub and proxy to be ready according to k8s readinessProbes.

__Like this!__
```
$ kubectl wait pod --for=condition=Ready --selector "component in (hub, proxy)"
pod/hub-68854b5749-fn89h condition met
pod/proxy-6d7dbdbf47-sljz2 condition met
```

consideRatio:
The only downside of this change, as far as I know, is that we may delay successful responses by max 10 seconds to the hub, as the hub won't get traffic until it is considered ready, and that is polled every 10 seconds.

jupyterhub/templates/hub/deployment.yaml
betatim: Thanks. For some reason I can never keep liveness and readyness probes straight in my head. "Liveness probe' determines if a pod needs to be restarted or not. "Readyness probe' determines if a pod is ready to serve traffic (a `Service` will only route traffic to a ready pod).

What I often need to look up is if a pod can be ready without the liveness probe having run. The answer is that a pod can be ready and receiving traffic if the liveness probe hasn't run yet.

betatim:
I'll merge this now.

---

> Speaking of which... My hub pod shuts down, and then starts up. I wonder if I could instead make it a rolling update since I have an external DB.

I am not sure. Reading all your other comments/issues and your quest for HA hubs: I think that jupyterhub (the program) assumes it is the only one reading and writing to the database. Working on that assumption is (I think) what needs to happen before we can change the upgrade strategy and other HA related things. If the upgrade strategy wasn't `Recreate` you could have a situation where the new hub starts, reads from the DB, does some other stuff and then become ready. In the meantime (somewhere between the new hub reading from the DB and becoming ready) the old hub writes something to the DB. The new hub won't expect/be setup to handle this and things will become inconsistent (the state in the brain of the hub is different from the state in the DB).

manics:
Just had another thought, the probes are applied to the container so could you move the database upgrade to an initcontainer?

consideRatio:
@manics ooooh yes an excellent idea! Could you repost that idea as a new issue?

PR #1363, opened 2019-08-18
Fixed formatting error in links

tlkh:
This is a follow up to PR #1358 to fix a formatting issue with the links so that they will render properly. Apologies for the mistake in the previous PR.


betatim:
I didn't notice either :-/ thanks!

PR #1378, opened 2019-08-27
Proxy deployment: Change probes to https port

chicocvenancio:
When left on http and http to https redirection is configured the probes may be redirected to the wrong port.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -106,8 +106,8 @@ spec:
           livenessProbe:
             httpGet:
               path: /_chp_healthz
-              port: proxy-public
+              port: proxy-https
           readinessProbe:
             httpGet:
               path: /_chp_healthz
-              port: proxy-public
+              port: proxy-https
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -106,8 +106,16 @@ spec:
           livenessProbe:
             httpGet:
               path: /_chp_healthz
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks for this insight and PR!

There is quite a bit of complexity going on here... Hmm... I'm not confident at all yet about what should be done or if something should be done.

Questions raised in my mind:
- What is the proxy-https service pointing to? Is it https termination followed by the proxy-public service?
- I've learned that any response in the range of >=200 and <400 will be a OK sign. But, that they will follow the redirect if assuming it on the same host, I think.
- It is possible to specify if it should be a http or https request in the probe, that may be relevant to do as well if this PR would go through.
- What if you don't have internal https termination, but that is handled before arrival to the pod, then having this set would break something perhaps?

Did you experience the probes failing due to this somehow? I can imagine that...
- you got an additional redirect that was followed, but the bounce was unnecessary and could had been slimmed away
- the probe was redirected and not followed, leading to a potentially success without actually asking the _chp_healthz endpoint?
- something else happened?

consideRatio:
## Related documentation
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#probe-v1-core
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#httpgetaction-v1-core

Note how the livenessProbe and readinessProbe, being of the k8s type `probe` has the `httpGet` field of the k8s type `HTTPGetAction` that has a field called `scheme` that defaults to `"HTTP"` but could be changed to `"HTTPS"`. That may be required. I'm not that confident on this.

consideRatio:
LGTM! Thank you @chicocvenancio :heart:

PR #1380, opened 2019-08-29
Allow adding additional ingress rules in network policy

yuvipanda:
If you have a prometheus server in a different namespace,
explicit 'from' stanzas are needed to make sure it can
have access to your hub and user pods.

This change allows individual administrators to make
these changes.


consideRatio:
This LGTM! I like the change from extraRules to this!

Issue #1383, opened 2019-09-03
Allow customising the proxy image choices via the Helm chart


mz2:
The JupyterHub Helm chart at the moment allows for the hub and singleuser notebook server image to be configured. There is no (at least documented?) facility for doing the same for the proxy images. I mean these kinds of properties are missing for the images involved in the proxy service:

```
  image:
    name: ...
    tag: ...
  imagePullSecret:
    username: ...
    password: ...
    registry: ...
    enabled: ...
```

manics:
The proxy has multiple containers:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/54b884be1c165d30fe0a173411c770a37dd9eff8/jupyterhub/values.yaml#L87-L106

What do you need to configure?

mz2:
This is related to https://discourse.jupyter.org/t/intercepting-requests-to-jupyter-notebook-server/2046 – my plan was to alter the configurable-http-proxy containing image for my purposes (I'd like to intercept some specific endpoints), but to be honest I could use advice on that.

consideRatio:
This is now possible

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/5df3b123894bff4751684fdb91e7a719eda2268c/jupyterhub/templates/proxy/deployment.yaml#L51-L53

PR #1391, opened 2019-09-06
Adding optional imagePullSecrets for proxy Image

hdimitriou:
I noticed that 'hub' image gives you the opportunity to overwrite and use a custom along with docker registry credentials, so I did the same for proxy too


hdimitriou:
Don't think the failing tests were due to my PR, i just used master and the tests don't actually consider this case, using a custom image with credentials so it's basically as if my code is not used at all on the test

mz2:
To give some context, this PR concerns (for our purposes resolves) the issue I filed the other day: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1383 (me and @hdimitriou work on the same effort).

hdimitriou:
Please review this code again

consideRatio:
While this PR mirrors an existing chart behavior for the hub pod that I've been part implementing, I now favor another approach: could you make this PR like https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1426/files instead?

It feels like the most sustainable thing to support in the long run.

hdimitriou:
I arranged the code according to zero-to-jupyterhub-k8s/issues/1467 giving the users the ability to leverage existing imagePullSecrets

hdimitriou:
Ping @consideRatio

consideRatio:
@hdimitriou could you make this PR not add another k8s Secret resource, but instead just reference a self-created k8s secret, as done in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1426?

This chart is struggling with its own complexity, and adding another k8s Secret resource specifically for the proxy pod would add more than I think is justified.

hdimitriou:
On my PR I support both secret reference & proxy-specific definition. If I understand well you are asking me to remove the proxy-specific definition and only let the user define a secret set externally to the chart.

Isn't it worse to have inconsistency between hub that supports both methods and proxy, than the complexity of supporting a proxy-specific secret too? It would make more sense if we removed them all together, though I wonder if the effort is justified

consideRatio:
Ah... I thought we only had this for the singleuserimage currently. Deal let's add it.

I'd like to make some adjustments, to reuse templates instead of duplicate them etc. Do I have rights to add commits to this PR?

hdimitriou:
>  Do I have rights to add commits to this PR?
'Allow edits from maintainers.' is checked...If you still can't I can try adding you manually somehow

consideRatio:
Arrrgh travis doesn't show up as it should.

consideRatio:
@hdimitriou can you help me review the code I submitted?

hdimitriou:
Yes, I will review hopefully today

hdimitriou:
Sorry but it's currently overwhelming for me to check this out and review

hdimitriou:
Not really sure the failure is due to this PR, or due to master. Will trigger again in a couple of days hopefully

hdimitriou:
re-run

jfb74:
Is there any update on this? I am trying to deploy jupyterhub in an enterprise environment which requires a private docker image repository. Seems like this (proxy) and the scheduler still don't support imagePullSecrets.

consideRatio:
@hdimitriou first of all thank you so much for your work and activity in this project.

Considering @jfb74's question on workarounds to this, I found [this blog post](https://medium.com/titansoft-engineering/kubernetes-cluster-wide-access-to-private-container-registry-with-imagepullsecret-patcher-b8b8fb79f7e5). That got me thinking.

I think it is become quite complicated to duplicate the logic to create an image puller secret, and also to ask the users to specify separate image puller secrets multiple times over, for the hub pod, proxy pod, image puller pods, user-server-pods, autohttps pod, etc.

Due to this, what would make sense to do? Well I think to make the imagePullSecrets a chart-global configuration instead. Then we configure it once, and all pods get that imagePullSecrets field added, rather than just the hub and/or singleuser pods. I think this is the only sensible solution.

Since this has been open for quite some time, and has merges with master etc, and there is little code to reuse, I intent to open a new PR and add you (@hdimitriou) as a co-author of my commits to reflect your work towards this.

## Summary
- I intend to open a new PR meant to replace this PR with a different approach, and I will then attribute @hdimitriou as a co-author of all commits as a tribute to work done with this PR.
- The new approach is:
  1. Introduce a chart global configuration for imagePullSecrets.
  2. Deprecate the old imagePullSecrets configuration options for hub and singleuser
  3. Use the new configuration to provide imagePullSecrets credentials for all pods spawned in the Helm chart.

Is this approach okay to you @hdimitriou ?

consideRatio:
I've now opened #1794 and https://github.com/jupyterhub/kubespawner/pull/442 that together is to address this and is therefore closing this pr.

PR #1401, opened 2019-09-10
Recreate as CHP proxy pod's deployment strategy

consideRatio:
Using a rolling update by default on the proxy pod is a mistake by us,
because of the JupyterHub / CHP proxy interaction. JupyterHub assumes in
check_routes / add_route etc to be speaking to one specific CHP proxy
server, but there can be different ones responding if we make an upgrade
and the proxy pod is making a rolling upgrade.

For example, consider a hub pod making a recreate upgrade, and a proxy
pod makinga rolling upgrade. The new hub pod could for example get ready
before the proxy pod and start speaking with the old proxy pod and later
at a crucial point start speaking with the new pod. If you switch to
speaking with the new pod at the wrong time, you may end up with failure
to get responses from user pods that are verified to be around, and then
they are deleted.

So, this commit hope to fix a sneaky bug where user pods are deleted
during upgrades where the proxy pod is also updated!

---

Note that with the traefik proxy that would store state in a key value store, this may not be a problem, but we don't yet use traefik proxy.

PR #1404, opened 2019-09-10
Bugfix for proxy upgrade strategy pr #1401

consideRatio:
Upgrades from previous state to this would fail without this fix of the issue caused by removing the fix in this PR: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1401


betatim:
Thanks for the good comment! Hopefully next time around we remember why it is there (and that it can't be removed)

PR #1407, opened 2019-09-11
Mounting jupyterhub_config.py etc.

### LINKED ISSUE #1395: PR Discussion - mounting jupyterhub_config.py
# Mounting jupyterhub_config.py into the hub container
We have a configuration file that is built into the image. This makes for a more complicated and slower local development, CI builds, and provides no real benefit over writing it to a ConfigMap that we mount as far as I can tell.

## Relevant understanding
- ConfigMap and Secrets are limited to 1 MB in file size as that is what ETCD which is used by k8s is limited to. Our current configmap and secret for the hub is currently less than 30 kB together, and the jupyterhub_config.py is 18 kB, so we have a large margin here.

## PR work to be done (#1407)
- [x] Create or reuse a helm template .yaml file for a k8s ConfigMap resource, that reads the content of the jupyterhub_config.py and puts it into the `data.jupyterhub_config.py` key's value.
- [x] Mount it on the hub pod.
  - Don't override the full folder where you mount it by mounting the full ConfigMap and all its key's under `data` to a folder. Instead mount the specific key directly to `/etc/jupyterhub/jupyterhub_config.py`. To accomplish this, use subPath to select a specific key to mount and set the mountPath to a filename instead of a directory.
    ```yaml
              volumeMounts:
                  name: <config-volume-name-here>
                  subPath: jupyterhub_config.py
    ```
  - [x] Make sure to verify this works and the file can be read as it should, or if we need to also mount it as a specific user or similar.
- [x] Repeat the idea for z2jh.py and cull_idle_servers.py

consideRatio:
Closes #1395, where the motivation etc for the PR is described.

With this in place, we will not rebuild the images a lot less often, simplifying development and in the future where we have fixed a bug with chartpress, increasing the CI job speed.


consideRatio:
@betatim could you help with a review on this?

images/hub/Dockerfile
```diff
@@ -45,11 +45,6 @@ RUN PYCURL_SSL_LIBRARY=openssl pip3 install --no-cache-dir \
             echo jupyterhub==${JUPYTERHUB_VERSION}; \
           fi')
 
-ADD jupyterhub_config.py /srv/jupyterhub_config.py
-
-ADD z2jh.py /usr/local/lib/python3.6/dist-packages/z2jh.py
-ADD cull_idle_servers.py /usr/local/bin/cull_idle_servers.py
```
manics: Do these other two files need to be dynamically mounted instead of being embedded in the image?

consideRatio: Thanks @manics for helping out with reviewing this!

None of these three files need to be mounted as compared to be part of the image, but at least z2jh.py should go alongside like jupyterhub_config.py does as it is only used by jupyterhub_config.py and are greatly related.

But, I think all should be mounted instead of part of the image. It allows for:
1. Less need to rebuild this image in our CI job builds, it only needs to happen when we want to update the dependencies we install. This also increase local development speed significantly.
2. Whoever that use a custom image can keep in sync a lot easier as this person only needs to keep track of version changes as compared to also keeping track of any Helm chart configuration change that leads to a change in jupyterhub_config.py or similar.
3. This can help demonstrate a good practice on how to add a custom service in a sustainable way, where you would mount the additional source code files onto the container instead of rebuilding the hub image with the added service within it.

I think all of this leads to a better Developer Experience and less potential issues for helm chart users.

manics: That's arguably the case for `z2jh.py` though it feels wrong to be dynamically mounting a Python module into `{dist,site}-packages`.

In the case of `cull_idle_servers.py` isn't that tied to the JupyterHub API, so if for example the script was updated to make use of a new JupyterHub API feature then someone running an older hub image with a newer chart could end up with an incompatible `cull_idle_servers.py`?

consideRatio: Ah, yes I agree about mounting the module into `{dist,site}-packages`, it was what was done in the Docker image so I figured I'll simply follow that practice. I'm not yet python savvy enough to come up with a suggestion about where it should be located to be accessible from jupyterhub_config.py in a practical enough way. Got ideas?

consideRatio: Regarding `cull_idle_servers.py`, it has ties to the JupyterHub API:

```python
    c.JupyterHub.services.append({
        'name': 'cull-idle',
        'admin': True,
        'command': cull_cmd,
    })
```

Most of its API ties relates to the Helm chart I'd say, as the scripts is configured with various parameters passed to it through the Helm chart etc.

manics: I had a think about this, and I think we're missing instructions (or a set or requirements) for someone wanting to use a custom hub image.

Currently a custom hub image should have:
- a compatible version of jupyterhub
- compatible versions of any plugins such as Authenticators that are used
- the `cull_idle_servers.py` script
- the `z2jh.py` module or some other way of parsing the configuration properties
- Anything else? Does z2jh make any other assumptions?

In practice this is best done by extending `FROM` the official chart image

With this PR a custom hub image should have:
- a compatible version of jupyterhub
- compatible versions of any plugins such as Authenticators that are used
- no `cull_idle_servers.py` script as it'll be overwritten
- no `z2jh.py` module as it'll be overwritten
- Anything else? Does z2jh make any other assumptions?

This means you can have a custom hub image built from scratch.

Both these are reasonable (ignore for now the question of where the files are mounted), but if someone has a custom `cull_idle_servers.py` script this PR is arguably a breaking change (add to the release notes?).

consideRatio: Thanks @manics for your review and time spent thinking on this! I created https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1424 to represent this documentation!

consideRatio:
Without additional review dropping in by tomorrow I'll go ahead and merge this.

consideRatio:
Thanks Tim that's totally okay!

mrow4a:
@consideRatio which release of chart this will go into?

consideRatio:
@mrow4a it is now on the current master, anything there goes into the next release, which would be named 0.9.0. We are slow on making that =/

I recommend using current master, the README.md contains a link/badge to the latest published helm chart master commit.

mrow4a:
@consideRatio I see that z2jh.py is placed now in /usr/local/lib/python3.6/dist-packages/z2jh.py This location is correct for Ubuntu, but if someone is running another distribution, this path might not be correct e.g. it can be `/usr/local/lib/python3.6/site-packages/z2jh.py` instead.

consideRatio:
@mrow4a the file was previously hardcoded into that location during the Dockerfile build into an image, but now it is instead mounted onto the container running the image, allowing the image to remain the same even though we change something small.

If there is someone that makes a total remake of the image for some reason, and makes a case for this project to support such quite extreme customization, we could make it customizable. If that becomes the case for someone in the future, we could consider making it customizable.

mrow4a:
@consideRatio some organisations (ours) have organisation supported, secured and frequently patched docker images, which we would like to use as a base. The same for other critical components of the chart. 

Do you consider this an extreme customization, and using jupyterhub chart is something not suitable for this usecase? Is JupyterHub recommended only on Ubuntu (which we have absolutelly no control of and is a blackbox) ?

consideRatio:
Ah, now that I have one use case I think is very reasonable, it seems plausible to support that!

mrow4a:
@consideRatio any ideas how to make chart (thus ztjh.py) independent on the docker image, and be bound only to jupyterhub min version? Should it be part of jupyterhub_config.py itself? Thanks for hints, I can help pushing this forward.

consideRatio:
@mrow4a I think that makes a lot of sense, then we get rid of the "where to put it" issue all together etc.

The only concern I have, is to make it accessible for someone using extraConfig scripts as well. Will this code be accessible from there also then? I think so, but im not 100%.

```yaml
hub:
  extraConfig:
    aTestConfig: |
      # I'd like this function to be possible to access from here as well
      # if it was defined in the jupyterhub_config.py
      print(get_config)
      # I currently do:
      # from z2jh import get_config
```

mrow4a:
@consideRatio not sure if even this is needed, it is enough to put them in the same directory
```
$ mkdir pytest; cd pytest;
$ cat z2jh.py
def test():
	print("print from z2jh")

$ cat jupyterhub_config.py 
from z2jh import test

test()
$ python jupyterhub_config.py
print from z2jh
```

PR #1409, opened 2019-09-12
Allow user scheduler policy to be adjusted

yuvipanda:
Clusters with really big images should give much higher
priority to Image Locality, and this lets them tweak that


jupyterhub/values.yaml
```diff
@@ -271,6 +271,7 @@ scheduling:
     enabled: true
     replicas: 2
     logLevel: 4
+    policy: {}
```
```suggestion
    ## policy:
    ## Allows you to provide custom YAML/JSON to render into a JSON policy.cfg,
    ## a configuration file for the kube-scheduler binary.
    ## NOTE: The kube-scheduler binary in the kube-scheduler image we are
    ## currently using may be version bumped. It would for example happen if we
    ## increase the lowest supported k8s version for the helm chart. At this
    ## point, the provided policy.cfg may require a change along with that due
    ## to breaking changes in the kube-scheduler binary.
    policy: {}
```

jupyterhub/files/userscheduler-defaultpolicy.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+# Details: (cpu(10 * sum(requested) / capacity) + memory(10 * sum(requested) /
+# capacity)) / 2
+
+# ImageLocalityPriorityMap is a priority function that favors nodes that already
+# have requested pod container's images. It will detect whether the requested
+# images are present on a node, and then calculate a score ranging from 0 to 10
+# based on the total size of those images. - If none of the images are present,
+# this node will be given the lowest priority. - If some of the images are present
+# on a node, the larger their sizes' sum, the higher the node's priority.
```
consideRatio:
```suggestion
# on a node, the larger their sizes' sum, the higher the node's priority.

```

jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
```diff
@@ -6,83 +6,6 @@ metadata:
...
-        { "name": "NodePreferAvoidPodsPriority",  "weight": 161051 },
-        { "name": "NodeAffinityPriority",         "weight": 14641 },
-        { "name": "InterPodAffinityPriority",     "weight": 1331 },
-        { "name": "MostRequestedPriority",        "weight": 121 },
-        { "name": "ImageLocalityPriority",        "weight": 11}
-      ],
-      "hardPodAffinitySymmetricWeight" : 100,
-      "alwaysCheckAllPredicates" : false
-    }
+  {{ $defaultPolicy := .Files.Get "files/userscheduler-defaultpolicy.yaml " | fromYaml}}
```
consideRatio: - Chomp left to not cause a blank line after data:
- remove a trailing white space in the path
- nitpick helm formatting by adding a space

```suggestion
  {{- $defaultPolicy := .Files.Get "files/userscheduler-defaultpolicy.yaml" | fromYaml }}
```

yuvipanda:
Thanks, @consideRatio! I've accepted the changes :)

consideRatio:
@yuvipanda wiee!

I had one remaining concern about the default value not being used if the provided value was empty but not null, but that went up to smoke by looking at [this documentation](https://github.com/Masterminds/sprig/blob/master/docs/defaults.md) so this LGTM!

Issue #1414, opened 2019-09-18
PR Discussion - Optimizing auto scaling through self destruction


consideRatio:
@yuvipanda suggested some optimizations relating to the user-placeholder pods. I like these ideas so I'll elaborate on them here. To understand the optimizations, we need to understand the issues we currently have first, we can do that in a thought experiment!

# Thought experiment
1. An autoscaling k8s cluster has nodes that can fit 10 users and there are 5 user placeholder pods. Currently the first and only node has 7 users and 3 placeholders, and 2 placeholders are pending.
2. A second node is added as there were pending pods, and is now starting up slowly.
3. The first new user arrives, a placeholder is pre-empted on the only available node to make room for the new user. If the user would been able to schedule without pre-empting a placeholder pod, the pre-emption would not have happened.
4. The second node becomes ready for pod scheduling, and is starting to pull images thanks to the continuous pre-puller daemonset pod that quickly scheduled there. Currently 8 users and 2 placeholders are on the first node, and the second node has 3 placeholders on it.
5. The second new user arrives, and is scheduled on the node with most resources used which has room, which becomes the second node. This node has not yet pulled all the images though so the user need to wait. The second node now has 1 user and 3 placeholders on it.

   **Optimization opportunity (OO#1):** Force scheduling of user pods on nodes with image locality.

6. The image pre-pulling on the second node by the continuous pre-puller daemonset pod completes.
7. The third new user arrives, and is scheduled on the second node that now has 2 users and 3 placeholders on it.

   **Optimization Opportunity (OO#2):** Force scheduling on the most truly busy node where user-placeholder pods shouldn't count. Auto scaling down a node will become easier.

# Analysis
1. The user-placeholder pods help trigger auto scaling of nodes ahead of time by ending up in a pending state after getting pre-empted by user pods filling up the available nodes. Well done user-placeholder pods and thanks k8s for the pod priority mechanism! We made scale up ahead of time!
2. When the new node has been become available for scheduling, it will no longer pre-empt user-placeholder pods in order to schedule as it can simply schedule on the new node. The kube-scheduler will not evict unless it is required to in order to schedule the pod. The problem is that this node still isn't ready for the user pods as they want their images pulled on it already.

# Solution ideas
## Required node labels
We could maintain pre-emption of placeholder pods until image locality is available indirectly. We could ourselves make sure to set node labels that the user pods require.

[...]

PR #1416, opened 2019-09-18
Set anti-affinity between user and placeholder pods

yuvipanda:
We ideally want *all* our placeholder pods to be in the newest
autoscaled node. This way, user pods are packed in as tightly
as possible, instead of nodes 'wasting' some resources on
user placeholder pods that won't get pre-empted. Part of
accomplishing this is:
  1. Make sure the placeholder pods all hang out together.
     This is accomplished using the custom user scheduler
     right now.
  2. Try and make sure user pods & placeholder pods do not
     mix. This anti affinity term helps with that, although
     there also needs to be active descheduling.

Ref #1414

### DIFF
--- a/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
+++ b/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
@@ -36,8 +36,30 @@ spec:
       tolerations:
         {{- include "jupyterhub.userTolerations" . | nindent 8 }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
-      {{- if include "jupyterhub.userAffinity" . }}
       affinity:
+        # We ideally want *all* our placeholder pods to be in the newest
+        # autoscaled node. This way, user pods are packed in as tightly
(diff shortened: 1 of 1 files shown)


yuvipanda:
Linting fails with:

```
### 3/4 - yamllint
rendered-templates/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
  108:9     error    duplication of key "podAntiAffinity" in mapping  (key-duplicates)

`yamllint -c yamllint-config.yaml rendered-templates` exited with status 1
```

Currently we assume that user pods & placeholder pods' affinities need to be exactly the same, which is no longer true if we add this anti-affinity just to the placeholder pod. We still need to merge in other affinities from user pod though (such as node affinities), so I'm guessing this has to be some kinda helm merging trickery.

yuvipanda:
Screen Shot 2019-09-16 at 11 01 16 AM

You can see the drastic effect of adding pod anti-affinities and then deleting all user placeholder pods. They were earlier distributed across multiple nodes, and won't be pre-empted until next scale up event. Instead, they are now all on the newest node!

PR #1417, opened 2019-09-19
Add proxy.service.loadBalancerSourceRanges

### LINKED ISSUE #1253: Restrict proxy access to certain IP addresses
We're running JupyterHub on Azure Kubernetes Service using the z2jh Helm chart and would like to restrict inbound traffic to company IP addresses only. As far as I know this would have to be specified in the NetworkPolicy for the proxy service. Currently this cannot be customised using the values.yaml.

Is there any way for me to achieve this while still using the z2jh Helm chart, or would I have to create my own?

itssimon:
Fixes #1253

### DIFF
--- a/doc/source/security.md
+++ b/doc/source/security.md
@@ -264,3 +264,20 @@ singleuser:
 See the [Kubernetes
 documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
 for further information on defining policies.
+
+## Restricting Load Balancer Access
+
+By default any IP address can access your JupyterHub deployment through the load balancer service.
+In case you want to restrict which IP addresses are allowed to access the load balancer, you can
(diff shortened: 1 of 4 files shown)


GergelyKalmar:
Haha, just implemented this too. However, I also added some documentation, see https://github.com/wbrp/zero-to-jupyterhub-k8s/commit/4eb035fd1f8379453722d12b4679a87a66c13bbc, perhaps you can take those over too. I think we don't need the default value because it will default to this anyways when nothing is provided.

GergelyKalmar:
Forgot the toYaml conversion though in service.yaml, so that line I wouldn't take over.

itssimon:
Closing this in favour of #1418. Thanks @GergelyKalmar!

PR #1418, opened 2019-09-19
Add load balancer access restrictions

GergelyKalmar:
An extended implementation of pull request #1417.


GergelyKalmar:
Hm, this test looks flaky. It seems to fail even when I'm removing my changes. Am I missing something?

GergelyKalmar:
@consideRatio Would you mind reviewing / merging sometime this suggestion for resolving #1253? Thank you very much!

itssimon:
@consideRatio any chance you could merge this please?

betatim:
Looks good to me.

One question for my education: do you know how this interacts with using an nginx-ingress? In that case all the proxy related services are of type `ClusterIP`. Maybe a comment in the docs that this won't (I think) work if you are using a different service type.

Another thing to double check/agree on is if other config options in this chart prefer empty defaults to signal that something isn't configured instead of setting it to `0.0.0.0/0`. I somehow prefer that but happy to go with what ever the rest of the chart does.

itssimon:
@betatim the loadBalancerSourceRanges has no effect if the service type is not LoadBalancer. Probably worth mentioning that in the docs, even though the name already suggests it.

For reference: https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/#restrict-access-for-loadbalancer-service

GergelyKalmar:
Thank you @betatim and @consideRatio for the feedback!

I thought about not introducing a default value but wouldn't that produce some surprising behavior? In particular, if someone sets this configuration value and later removes it in their values file, will that not leave the service and the loadBalancerSourceRanges unchanged without the default?

As a user I would prefer that in cases when the value is not set in the values file the configuration reverts to the original default, that is, being open. Or perhaps I'm misunderstanding how upgrades work?

manics:
Helm and K8s _should_ be fully declarative, try `helm template ... ` and/or `kubectl get ... -o yaml` to spit out the YAML so you can verify for yourself

consideRatio:
I didnt fully understand your question, but to conclude my understanding and opinion briefly from a mobile:

Ive had past issues with the service objects when using helm and changing their type. These issues related to removing previously rendered fields like nodePort on a servicetype that is changed to clusterIP service type.

To avoid potential ssuea like that, i suggest that our helm templates, especially on this service object with fields related to its type that can change, only explicitlly let our helm template render those fields if they serve a purpose. As the defualt is to accept all sourceRanges, then there is no need to add 0.0.0.0/0 as a explicit default but instead leave it as a implicit default by omitting it. Because it isnt a required field, and does that by default, right?

I lack full understanding and my caution to render something additional than needed is mostly a way to avoid issues that could pop up when changing the service type to a clusterip - something we may do when starting to use cert-manager for example

consideRatio:
So what i suggest to change is: Add one more condition to render the loadBalancerSourcesRanges relating to actually having a non-empty list provided, and make the default an empty list

GergelyKalmar:
@consideRatio Okay, I'll probably have some time on the weekend, can try to explore this further and check if it works as intended.

consideRatio:
Thanks @GergelyKalmar, this looks good to me! Could you rebase on master to resolve the merge conflicts?

GergelyKalmar:
@consideRatio Well this did take a little longer than planned but I got to it eventually...

consideRatio:
Thank you for your work on this @GergelyKalmar ! :heart: :tada:

PR #1426, opened 2019-09-27
Added the ability to use an existing/shared imge pull secret

LaurentGoderre:
This enables using pull secrets in a way that is more consistent with other helm charts.


minrk:
Thanks!

PR #1453, opened 2019-10-19
Use helm 2.15.1

consideRatio:
Helm 2.15.0 stopped using extensions/v1beta1 for tiller, which was
required for installation on k8s 1.16 as we have testing for. So now we
don't have to maintain a separate Helm version for k8s 1.16 tests.


consideRatio:
I debugged this locally, and conclude that the error on running `helm upgrade` to install something fails, but if you run the command again it works... I did not find any error in the logs, and `--debug` didn't help me much.

consideRatio:
Fixed in Helm 2.15.1 I think:
https://github.com/helm/helm/commit/41a697ed10ca21ec0d97e4b3bf7726e14c4957cc

PR #1454, opened 2019-10-22
Bump python-kubernetes to 9.0.*

clkao:
It seems the V1PodSpec in 8.0 does not support enableServiceLinks in case we want to disable it.

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -5,7 +5,7 @@ jupyterhub-tmpauthenticator==0.6
 jupyterhub-ltiauthenticator==0.3
 jupyterhub-ldapauthenticator==1.2.2
 jupyterhub-kubespawner==0.10.1
-kubernetes==8.0.*
+kubernetes==9.0.*
 nullauthenticator==1.0
 oauthenticator==0.8.2
 pymysql==0.9.2


consideRatio:
Thanks for the PR! This LGTM.

According to the python-kubernetes compatebility matrix, version 8 match with k8s 1.12's k8s API, 9 with 1.13, and 10 with 1.14. Do we perhaps want to bump it to version 10? What is your take on that @clkao ?

https://github.com/kubernetes-client/python#compatibility

manics:
K8s 1.16 is out which means the oldest supported version is 1.14 which justifies moving to 10. On the other hand there are always deployments which lag behind so this would force a K8S upgrade.

Are there any other benefits of moving to 10?

consideRatio:
> On the other hand there are always deployments which lag behind so this would force a K8S upgrade.

Hmmm, i think the usage of certain python-kubernetes is equivalent to the usage of certain `kubectl` versions. As you can see from my own `kubectl`, I work with a very new version but with an older cluster, and that is fine, so I think it does not force any upgrades.

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-17T17:16:09Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-gke.24", GitCommit:"2ce02ef1754a457ba464ab87dba9090d90cf0468", GitTreeState:"clean", BuildDate:"2019-08-12T22:05:28Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

betatim:
The GKE cluster for mybinder.org is still on 1.13 but could be upgraded to 1.14. It seems like that is the newest version you can go to (for the k8s master). I'd vote for only increasing the minimum required version of something if we need to in order to fix a bug or get a feature that is needed vs updating because we can as that causes work for people with no benefit (which tends to make people unhappy ;) )

PR #1470, opened 2019-11-04
Fix single-user storage section

Nelvarion:
pvcNameTemplate and volumeNameTemplate under the singleuser.storage section we're missing quotation marks around the values, causing Helm to ignore the entire section.


consideRatio:
Thanks for your PR @Nelvarion, I tried to verify if this made any difference as I didn't think so, and I conclude it doesn't from using `helm template` on a isolated micro example.

If you experience this section to be ignored, I don't believe this is the reason, it could be for another reason though. Perhaps you could open up an issue and describe what you have observed?

PR #1472, opened 2019-11-09
Switching to the Pandas Sphinx theme

choldgraf:
This switches the theme to the Pandas Sphinx theme, the theme we're now using for the [team compass documentation](https://jupyterhub-team-compass.readthedocs.io). It should make the documentation a little easier to follow and search through.

As a result of the switch, I did a few things worth noting:

* A few titles have been slightly altered, mostly to reduce some visual redundancy and to avoid super long titles
* The docs pages have been organized into folders and sub-folders, corresponding to each major section. This will change some of the URLs.

Here's how it looks:

https://683-87849371-gh.circle-artifacts.com/0/html/index.html

What do folks think?


consideRatio:
This looks awesome to me! :D Nice!

betatim:
Can we setup redirects for the URLs that have stopped existing? There are a lot of tutorials, docs, blog posts, search engine results, etc out there that link directly into this guide. It would be great to not have people arrive in "nothing here any more, sorry confused traveller" land. However I don't know how to do this in RTD :-/

choldgraf:
Yeah - I am not sure how to do re-directs either unless we manually set up pages to do this. I think these are our options:

* Manually create a redirect page for each moved page in this PR
* Re-organize this PR back to a flat list of docs pages (we can achieve the same user-facing structure with a totally flat list of `.rst` files, it'll just be more confusing to maintain)
* Ignore the broken links as a one-time change

betatim:
https://docs.readthedocs.io/en/stable/user-defined-redirects.html#page-redirects should be what we need. Unfortunately they don't read the redirects from a file but you need to click buttons in a UI :(

As a user I get (very) unhappy when projects break the external links to their docs. I'm already confused while trying to learn something new and now the blog post/search result I found and hoped would solve my problem links to a none existent thing. Because I am new, inexperienced or unfamiliar with the project I don't know how to guess which new page is the replacement for the old page. To avoid angry Tim I'd be happy to help with setting up the redirects :)

choldgraf:
ok - well if we think that the added benefit of more nicely-organized docs outweighs the annoyingness of redirects, then maybe we can have a redirect session together? :-)

betatim:
If you have a list of the pages you moved and where they moved to (or a pattern of changes) we can bash out a bunch of redirects.

choldgraf:
I think that the thing to do is to look at the `index.rst` page in each of the new sub-folders. That will have a list of that subfolder's pages, which are the pages that we have moved as a part of this PR.

I don't think it should take too long, though we'll probably need to merge this PR first before trying it out (as I don't know how else to test that the redirect works). I am happy to spend some time making redirects this week if folks are 👍

choldgraf:
anything that I should do to move this forward?

betatim:
I think we should merge this when we know someone/a few people have time to install the redirects "right after" or we install them first.

What we need is a list of the old and new path for each page (or instructions on how to read that off from the `index.rst` in a sub-folder) and a few people who have admin rights to the project on RTD (I don't :( ).

consideRatio:
@betatim I added you as a maintainer on RTD!

choldgraf:
Hey all - I just went ahead and made a bunch of redirects on the RTD site, I think all of the files should be covered:

image

I tested the redirecting on a sample page and it worked as expected, so I just added rules for each of the changed pages.

Now, I just realized that I might have done something stupid because the new pages don't yet exist, but I *think* we won't get redirected to non-existent pages if the old page exists. That said, we should probably merge this sooner than later if the redirection was the main blocker.

consideRatio:
@choldgraf you are amazing!!!

:heart: :sunflower: :tada:

choldgraf:
yay! I think the new theme is a real improvement!

betatim:
Nice work! Big thank you for doing the tedious work of generating the list of which URLs to redirect and clicking all the buttons to implement them!!

Issue #1464, opened 2019-10-29
Putting z2jh.py next to jupyterhub_config.py in the hub image


consideRatio:
@mrow4a came with excellent input and a concrete suggestion in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1407#issuecomment-538936442 on how to make the ability to customize the hub image more reliable. I like this proposal, it just needs to be implemented.

You are very welcome to contribute with the implementation as well @mrow4a! :sunflower:

mrow4a:
Yes, I will have a look next days.

mkjpryor-stfc:
+1

Putting `z2jh.py` in the `WORKDIR` is important when the hub image in use does not use the exact same OS and Python version as the reference implementation. At the moment, if you are not using Python 3.6 with `dist-packages` under `/usr/local/lib`, `z2jh.py` cannot be found!

PR #1478, opened 2019-11-12
[1464] Move z2jh.py to python and distribution agnostic path

mrow4a:
Path `/usr/local/lib/python3.6/dist-packages` binds chart to Ubuntu-like distribution and Python 3.6 docker images.

Currently, configuration script is placed in `/etc/jupyterhub/jupyterhub_config.py` and JupyterHub execs it.

```
jupyterhub
--config /etc/jupyterhub/jupyterhub_config.py
...
```

Thus, it is not enough that `z2jh.py` is in `/etc/jupyterhub/` as it is not from where the python config is executed. Thus, the optimal solution could be to add the script to `PYTHONPATH` to container spec.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1464
@consideRatio @mkjpryor-stfc

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -6,6 +6,10 @@ from tornado.httpclient import AsyncHTTPClient
 from kubernetes import client
 from jupyterhub.utils import url_path_join
 
+# Make sure that modules placed in the same directory as the jupyterhub config are added to the pythonpath
+configuration_directory = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, configuration_directory)
+
 from z2jh import get_config, set_config_if_not_none
 
 # Configure JupyterHub to use the curl backend for making HTTP requests,
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -106,7 +106,7 @@ spec:
             - mountPath: /etc/jupyterhub/jupyterhub_config.py
               subPath: jupyterhub_config.py
               name: config
-            - mountPath: /usr/local/lib/python3.6/dist-packages/z2jh.py
+            - mountPath: /etc/jupyterhub/z2jh.py
               subPath: z2jh.py
               name: config
             - mountPath: /etc/jupyterhub/cull_idle_servers.py


manics:
This potentially overrides a `PYTHONPATH` set by the user.

Since this is mostly internal to the image how about insert `/etc/jupyterhub/` to `sys.path`, e.g. in `jupyterhub_config.py`
```
import sys
sys.path.insert(0, '/etc/jupyterhub')
from z2jh import get_config, set_config_if_not_none
```
(I haven't tested this)

mrow4a:
@manics how would you define PYTHONPATH for the container before env is set for the container?  

If you set PYTHONPATH before container starts (as part of container ENV definition like in this PR), all other operations in entrypoint will just append to that PYTHONPATH. If you have some custom PYTHONPATH, you probably would do it as part of entrypoint (which will append to PYTHONPATH), but firstly defined is PYTHONPATH container env. Am I right?

manics:
It could be set with `hub.extraEnv`: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/5eb48bc3b0c2dcdc8cf7387fc5a53638aa2e0f2a/jupyterhub/templates/hub/deployment.yaml#L181

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -106,7 +106,7 @@ spec:
             - mountPath: /etc/jupyterhub/jupyterhub_config.py
               subPath: jupyterhub_config.py
               name: config
-            - mountPath: /usr/local/lib/python3.6/dist-packages/z2jh.py
+            - mountPath: /etc/jupyterhub/z2jh/z2jh.py
```
consideRatio: Let's put in directly in /etc/jupyterhub alongside the cull script and jupyterhub_config.py etc.

```suggestion
            - mountPath: /etc/jupyterhub/z2jh.py
```

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -136,6 +136,8 @@ spec:
             # Don't allow any process to execute as root inside the container
             allowPrivilegeEscalation: false
           env:
+            - name: PYTHONPATH
+              value: /etc/jupyterhub/z2jh
```
consideRatio:
```suggestion
              value: /etc/jupyterhub
```

manics:
Expanding on my rationale: I think of `PYTHONPATH` as a last resort for users when there's no other way of configuring Python other than completely rebuilding the image, so where possible I think we should keep this option open to them (in all projects, not just this one).

mrow4a:
@consideRatio @manics I took an approach of `sys.insert`. To me it looks clean - all modules belonging to the configuration script should be in the directory where configuration script is.

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -6,6 +6,10 @@
 from kubernetes import client
 from jupyterhub.utils import url_path_join
 
+# Make sure that modules placed in the same directory as the jupyterhub config are added to the pythonpath
+configuration_directory = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, configuration_directory)
```
manics: I like it!

One thing I'm not sure about, @consideratio may know better: k8s may symlink config files into place, is there any situation where the real parent directories of this file and z2jh.py would be different?

consideRatio: I don't know, I think though that if this works for the TravisCI test pipeline, than I think it is robust enough as it worked in that k8s setup at least.

consideRatio:
This LGTM! Want to merge @manics ? :)

jupyterhub/files/hub/jupyterhub_config.py
mrow4a: @manics if you mount it like
```
- mountPath: /etc/jupyterhub/z2jh.py
   subPath: z2jh.py
   name: config
```

the file is regular file, not a symlink. Symlink happens only in case of mounting secrets as a directory, which in z2jh chart is not possible to do into `/etc/jupyterhub`.
```
- mountPath: /etc/jupyterhub/config/
  name: config
```

mrow4a: Anyways even if this would be moved to ConfigMap directory, the directory there is still a directory as https://docs.python.org/2/library/os.path.html#os.path.realpath ignores symlinks

```
$ root@hub-755dbdc69c-tt7r4:/srv/jupyterhub# python -c "import os;print(os.path.realpath('/etc/jupyterhub/config/z2jh.py'))"
/etc/jupyterhub/config/..2019_11_13_14_20_12.271060545/z2jh.py
```

manics:
Thanks @mrow4a!

Issue #1479, opened 2019-11-13
Make readiness and liveness configurable


mrow4a:
Currently readiness probe for hub is enabled by default, and liveness probe is disabled. I suggest in this draft commit https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/master...mrow4a:fix_probes to make it configurable in the chart, for few reasons:
- to support with chart 0.9.0 JH <1.0.0
- sometimes it might be desirable to enable/disable certain probes (e.g. liveness)
- issues like https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1357
- these probes as part of separate alerting system

@consideRatio any opinion?

consideRatio:
I think I'd like to see them be made configurable as well!

PR #1480, opened 2019-11-13
[1479 ] Make probes in hub configurable

mrow4a:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1479

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -194,19 +194,26 @@ spec:
           ports:
             - containerPort: 8081
               name: hub
-          ## livenessProbe notes:
-          ## We don't know how long hub database upgrades could take
-          ## so having a liveness probe could be a bit risky unless we put
-          ## a initialDelaySeconds value with long enough margin for that
-          ## to not be an issue. If it is too short, we could end up aborting
-          ## database upgrades midway or ending up in an infinite restart
-          ## loop.
-          # livenessProbe:
-          #   initialDelaySeconds: 30
-          #   httpGet:
-          #     path: {{ .Values.hub.baseUrl }}hub/health
-          #     port: hub
-          readinessProbe:
+          {{- if .Values.hub.livenessProbe.enabled }}
+          # livenessProbe notes:
+          # We don't know how long hub database upgrades could take
+          # so having a liveness probe could be a bit risky unless we put
+          # a initialDelaySeconds value with long enough margin for that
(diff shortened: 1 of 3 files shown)


consideRatio:
I'd like to see both the hub and proxy deployment's configurable with the same options. That way we avoid breaking the expectancy that if we can configure one we should be able to configure the other, something I believe will cause confusion otherwise.

## Sidetrack

Hmmm.. I wonder if we can apply some neat trick to not have to hardcode too much here, but instead merge fields from the passed probes.

Hmmm... One would do something like this.

```
# 1: inspect if the probe is enabled
# 2: load the defaults as a object
# 3: merge the provided values with the default object
```

Nevermind, it requires a lot of added complexity to load the default stuff, I've done similar things like this before and it can be quite a mess, the crux is that anything written in a Helm template file is text and is troublesome to get back in to an object that can be manipulated as an object.

## Back on track

Okay, then lets decide what to support to be configurable among the valid probe fields: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#probe-v1-core

Perhaps we settle with `periodSeconds` and `initialDelaySeconds` for now?

mrow4a:
@consideRatio if we have lots of defaults maybe this makes sense. The goal is to avoid having too much things in values.yaml, is that correct ? Could you give me some example of existing code to see? 

Regarding probe defaults themselves, yes `periodSeconds` and `initialDelaySeconds` seem reasonable as a start, and will be easy to add later.

consideRatio:
@mrow4a sorry to bug you with that sidetrack, I saw Helm 3 was released now, and with Helm 3 things will be easier I think so let's not get stuck in heavy patterns like the one I considered in the sidetrack so to say. (But here are examples: [values.yaml](https://github.com/mattermost/mattermost-helm/blob/a47132a7b907679068df95dd22cb49d1410ab42f/charts/mattermost-team-edition/values.yaml#L169-L175) + [a helm template merging that with other stuff](https://github.com/mattermost/mattermost-helm/blob/a47132a7b907679068df95dd22cb49d1410ab42f/charts/mattermost-team-edition/templates/secret-config.yaml#L11-L40))

> Regarding probe defaults themselves, yes `periodSeconds` and `initialDelaySeconds` seem reasonable as a start, and will be easy to add later.

I agree! Aim for that (without the overly complicated merge stuff i linked to).

mrow4a:
@consideRatio done :)

consideRatio:
Thank you @mrow4a for your work on this!

PR #1482, opened 2019-11-14
Bump proxy's nginx-ingress-controller image

consideRatio:
I think it can make sense to update this for security reasons, but I'd need help reviewing the breaking changes and if they affect us.

This bumps 0.15.0 to 0.26.1: https://github.com/kubernetes/ingress-nginx/releases


consideRatio:
We replaced nginx with traefik before.

PR #1510, opened 2019-12-03
DOC:fixing authentication link in user customization guide

n3o-Bhushan:
The current link directs to the resource which doesn't exist. So changed it to point to the correct link (https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html)

Link to issue #1511

### DIFF
--- a/doc/source/customizing/user-management.md
+++ b/doc/source/customizing/user-management.md
@@ -80,4 +80,4 @@ auth:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](/administrator/authentication).
+[the Authentication guide](../../administrator/authentication).
--- a/doc/source/customizing/user-management.md
+++ b/doc/source/customizing/user-management.md
@@ -80,4 +80,4 @@ auth:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](./authentication).
+[the Authentication guide](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html).
--- a/doc/source/customizing/user-management.md
+++ b/doc/source/customizing/user-management.md
@@ -80,4 +80,4 @@ auth:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html).
+[the Authentication guide](/administrator/authentication).


doc/source/customizing/user-management.md
```diff
@@ -80,4 +80,4 @@ auth:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](./authentication).
+[the Authentication guide](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html).
```
consideRatio: These markdown files will be built into HTML, and we want them to figure it out. I think the way to go is like this, or similar.

```suggestion
[the Authentication guide](./administrator/authentication).
```

consideRatio: Verified to work locally

betatim: Does `../administrator/authentication` not work? You don't want to reference the `.html` explicitly either as sphinx should be generating the links for us.

I think this might be a side effect of moving docs to sub-directories when we switched to the pandas sphinx theme. Do you remember anything/have advice @choldgraf ?

consideRatio: It works like this also, fixed.

betatim: I think you need two `.` at the start to go up one level in the directory hierarchy.

consideRatio: Using `/` made it relative to the root instead of relative to the current location, which felt more robust than navigating backwards then forwards in the directory structure.

choldgraf: will `/` work properly? In other projects I've worked on it tends to obliterate all of the sub-directories in the URL. But maybe Sphinx is smart about this?

choldgraf: (I can't commit this change because I don't have permissions to do so, but I think we should commit it, and see if the docs preview has the links working)

consideRatio: Depends on where you start your webserver I figure, hmmm... this may indeed be a cause of trouble even though it worked locally for me, but then i didn't have /en/latest etc.

doc/source/customizing/user-management.md
```diff
@@ -80,4 +80,4 @@ auth:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](./authentication).
+[the Authentication guide](/administrator/authentication).
```
consideRatio: I got that wrong, this is what we need.

```suggestion
[the Authentication guide](../../administrator/authentication).
```

n3o-Bhushan: No problem.

consideRatio:
Thanks @n3o-Bhushan !

Issue #1508, opened 2019-12-02
Missing page link for our RBAC documentation


raybellwaves:
On
https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/setup-helm.html#initialization
Step 2 says '...See our RBAC documentation for more information'. The page points to https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/security.html#use-role-based-access-control-rbac
which is missing.

doc source: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/setup-jupyterhub/setup-helm.rst#L63

consideRatio:
It should be pointing to https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#use-role-based-access-control-rbac ! Thanks for catching and reporting these issues!

PR #1514, opened 2019-12-04
Missing page link for our RBAC documentation #1508

n3o-Bhushan:
Fixing issue in #1508


consideRatio:
Thanks this solves it I think!

Issue #1300, opened 2019-06-02
User-scheduler doesn't work with rancher/local-path-provisioner


manics:
I've just been trying Z2JH on a single node [K3s](https://github.com/rancher/k3s) cluster with [rancher/local-path-provisioner](https://github.com/rancher/local-path-provisioner) for dynamic local volumes. See https://github.com/manics/k3s-test for my `Vagrantfile`.

`hub-db-dir` is created automatically as expected, but user pods fail to start (version `0.9-9c580a1`)

```
$ kubectl describe pod jupyter-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711
...
    Mounts:
      /home/jovyan from volume-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711 (rw)
Conditions:
  Type           Status
  PodScheduled   False
Volumes:
  volume-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711:
    Type:        PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:   claim-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711
    ReadOnly:    false
QoS Class:       Burstable
Node-Selectors:  <none>
Tolerations:     hub.jupyter.org/dedicated=user:NoSchedule
                 hub.jupyter.org_dedicated=user:NoSchedule
                 node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
  Warning  FailedScheduling  18s (x2 over 21s)  jh-user-scheduler  0/1 nodes are available: 1 node(s) didn't find available persistent volumes to bind.
```
```
$ kubectl describe pvc claim-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711
Name:          claim-c2144b5b-2dbd79-2d4017-2d8c7d-2d5c65e58b0711
Namespace:     default
StorageClass:  local-path
Status:        Pending
Volume:        
Labels:        app=jupyterhub
               chart=jupyterhub-0.9-9c580a1
               component=singleuser-storage
               heritage=jupyterhub
               release=jh
Annotations:   hub.jupyter.org/username: c2144b5b-bd79-4017-8c7d-5c65e58b0711
[...]

PR #1516, opened 2019-12-04
Fixing #1300 User-scheduler doesn't work with rancher/local-path-provisioner

cgiraldo:
Inspecting the user-schedule-xxxx pod log:

error binding volumes: persistentvolumeclaims "claim-test" is forbidden: User "system:serviceaccount:ns:user-scheduler" cannot update resource "persistentvolumeclaims" in API group "" in the namespace "ns"

I fix it by creating a new roleBinding to let user-scheduler update the volume claim:


consideRatio:
@cgiraldo thanks this LGTM!

Issue #1515, opened 2019-12-04
Fix broken links in documentation


consideRatio:
I used a online website service to check our documentation links and if they functioned as they should, and came up with this report: http://website-link-checker.online-domain-tools.com/link/1c23febgtzy0pMUbGj/

@choldgraf and @willingc I've seen you solve link breakage effectively before and I hope to learn from your experience!
- Are you aware of a similar tool like that online based tool above that you would recommend for working to ensure documentation links are maintained?
- Is it possible for us to setup some redirects with RTD so that external links to our documentation can still arrive correctly even after we reorganize it? Where / how is this done? Ah that became a very easy question to answer myself!
  **Answer**: Yes we can! Here is documentation about it: https://docs.readthedocs.io/en/stable/user-defined-redirects.html. Here is this projects redirects: https://readthedocs.org/dashboard/zero-to-jupyterhub/redirects/

## Broken links as of `95fd91f`

<details>
<summary>Click to show broken links and files</summary>

```
writing output... [  2%] administrator/advanced

(line   34) broken    https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-controllers - Anchor 'ingress-controllers' not found
(line  204) broken    https://zero-to-jupyterhub.readthedocs.io/en/latest/reference.html?highlight=pvc#hub-db-type - Anchor 'hub-db-type' not found

writing output... [  7%] administrator/authentication

(line  145) broken    https://zero-to-jupyterhub.readthedocs.io/en/latest/reference.html#hub-extraconfig - Anchor 'hub-extraconfig' not found

writing output... [ 17%] administrator/optimization

(line   71) broken    https://docs.helm.sh/developing_charts/#hooks - Anchor 'hooks' not found
(line  228) broken    https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node - Anchor 'what-types-of-pods-can-prevent-ca-from-removing-a-node' not found

writing output... [ 21%] administrator/troubleshooting

(line   28) broken    https://github.com/jupyterhub/helm-chart#versions-coupled-to-each-chart-release - Anchor 'versions-coupled-to-each-chart-release' not found
writing output... [ 24%] administrator/upgrading

(line   11) broken    https://groups.google.com/forum/#%21forum/jupyter - Anchor '%21forum/jupyter' not found
[...]

PR #1523, opened 2019-12-17
fix link to Helm chart best practices

rpwagner:
Looks like the Helm docs have been consolidated on [new website](https://helm.sh)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -352,4 +352,4 @@ differently.
 We strive to follow the guidelines provided by
 [kubernetes/charts](https://github.com/kubernetes/charts/blob/master/REVIEW_GUIDELINES.md)
 and the [Helm chart best practices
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -352,4 +352,4 @@ differently.
 We strive to follow the guidelines provided by
 [kubernetes/charts](https://github.com/kubernetes/charts/blob/master/REVIEW_GUIDELINES.md)
 and the [Helm chart best practices
-guide](https://github.com/kubernetes/helm/tree/master/docs/chart_best_practices).
```
consideRatio: Thanks this is certainly better than a 404!

We may need to make more updates down the line to not end up confusing that this is made during helm2 era rather than the new helm3 era which the current helm docs have entered without retaining all of their old docs.

rpwagner:
If I'd seen it sooner, I would have referenced #1515. I bet there are more to come because of the Helm docs changes.

PR #1530, opened 2019-12-23
Fix jupyter hub helm chart user scheduler template rbac error

cutiechi:
Fix z2jh 0.9.0 user scheduler update pv error


consideRatio:
Thanks!

PR #1531, opened 2019-12-23
due to a kubernetes change in 1.14 the ssl service must come first

sstarcher:
In kubernetes 1.14 if the type is LoadBalancer and the backend has any SSL configuration it will configure an https health check, but it will use the first port which in this case is http and will fail.  Putting the possible https port first fixes this issue.

https://github.com/kubernetes/kubernetes/issues/83070

### DIFF
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -42,18 +42,6 @@ spec:
     {{- end }}
     release: {{ .Release.Name }}
   ports:
-    - name: http
-      port: 80
-      protocol: TCP
-      {{- if $autoHTTPS }}
-      targetPort: 80
-      {{- else }}
-      targetPort: 8000
-      {{- end }}
-      # allow proxy.service.nodePort for http
-      {{- if .Values.proxy.service.nodePorts.http }}
-      nodePort: {{ .Values.proxy.service.nodePorts.http }}
-      {{- end }}
     {{- if $HTTPS }}
     - name: https
       port: 443
@@ -67,6 +55,18 @@ spec:
       {{- end }}
       {{- if .Values.proxy.service.nodePorts.https }}
       nodePort: {{ .Values.proxy.service.nodePorts.https }}
+      {{- end }}
+    - name: http
+      port: 80
+      protocol: TCP
+      {{- if $autoHTTPS }}
+      targetPort: 80
+      {{- else }}
+      targetPort: 8000
+      {{- end }}
+      # allow proxy.service.nodePort for http
+      {{- if .Values.proxy.service.nodePorts.http }}
+      nodePort: {{ .Values.proxy.service.nodePorts.http }}
       {{- end }}
     {{- end }}
   type: {{ .Values.proxy.service.type }}


consideRatio:
Thanks!

So this is a workaround to solve an upstream problem that should not be there at all. Hopefully no-one else has a hardcoded practice like this that breaks by our re-ordering to workaround this issue now.

sstarcher:
Yep, upstream has not considered it a priority and it's been broken for a few releases.  It's certainly a bug in upstream.  Thanks for the merge.

albertmichaelj:
Interesting, this is literally the exact opposite of the problem that I am having in #1637. However, I've never run into this problem on Kubernetes 1.14 using the previous ordering even when using SSL. I'm curious why that is.

I have just confirmed that one of the two jupyterhub k8 deployments that I manage is kubernetes version 1.14 and the health check port is the http port, ssl is configured, and the health check is not failing. What causes the health check to fail if it uses the http port?

sstarcher:
@albertmichaelj Can you do some debugging on this to find out why you see this issue.

Please look at the ELB and see what port it's trying to hit.  Please attempt to hit that port yourself and see what's happening.

My changes should have no impact when HTTPS is disabled.  If you can tell me your exact configuration I can try and reproduce it, but it would be easier if you were able to debug what exactly is failing and to verify why the ELB is failing it's health check.

albertmichaelj:
I've done a good bit of debugging listed in #1637. If there is something that I should be doing that is not there, then please let me know. The problem is that `https` is enabled by default using the minimal example configuration from the jupyterhub docs, so anyone following the docs for AWS will have a non-working config as far as I can tell.

Additionally, I'm not entirely sure how to re-enable https here because as soon as I enable ssl, it will fail since the certificate will not be valid yet since it takes a little bit of time to get the certificate (plus if you haven't already started the deployment with https disabled in order to get the ip address in order to set up the dns record, the certificate will not be granted until sufficient DNS updating). In the time you wait for an ssl certificate, the health checks seem to fail (at least that was my experience before I started digging down into why this was happening).

Did you observe that the health checks were failing with the previous ordering when ssl was enabled? Again, I'm not observing this behavior on kubernetes 1.14, and I'm not sure why (the health check is checking http port, not the https port, but ssl is working fine).

sstarcher:
@alanjcastonguay If HTTPS is enabled by default it's not by PR that caused that.  It would seem the root question is when was HTTPS enabled by default and what else is needed for that to work.

If HTTPS is disabled my PR should have no impact as those ports should not be open.

I don't remember experiencing your DNS issues.  I would recommend disabling HTTPS or looking for PRs that caused it to be enabled by default.  None of that sounds related to this PR.

As for debugging, I would recommend trying to kubectl port-forward and seeing if that works.  I would also recommend hitting the nodeport directly to see if that works.  It sounds to me that the default Jupyter Hub configuration may be invalid and needs to be fixed.  This does not sound related to the ELB health check or to this PR, but to the default Jupiter hub configuration.

albertmichaelj:
@sstarcher Thanks for taking the time to respond. HTTPS has been enabled by default for over two years on this chart (it was enabled on [this commit](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/427992c3aea53f985e14aa538f70e8dfa9acab3c)). What I'm observing is that things seemed to work fine on 1.14 with ssl enabled and the AWS load balancer health checks checking http (not https) with the ordering of the port specifications the way they were before your pull request. If we re-order the port specifications so that http comes first, we can leave https enabled by default (I'm not sure why it's enabled by default, but it has been for three years now).

My question is, were you experiencing issues with the port specification the way it was before your pull request? I am not, and I'm really trying to figure out why. If it doesn't cause issues to have http specified first, then that would be a minimal change that fixes this issue.

Moreover, as I stated in my previous comment, I don't necessarily think that disabling https by default will fix the issue because the health checks seem to fail until the ssl cert is granted. Given that it takes time to grant the ssl cert, and the chart itself is getting the ssl cert, it's not clear that we want the load balancer checking health using the https port ever, if we don't need it to (and in my experience we don't need to check health using the https port, but I'm assuming your experience differs, so I want to understand why).

Any help in understanding why this pull request was made in the first place would be helpful, because it seems like my problem would be solved if we reverted the ordering. Thanks for your help.

sstarcher:
Yes, please see the linked Kubernetes issue.  It explains everything. 

TLDR With Kubernetes 1.14 if you enabled HTTPS at the load balancer and the first service port is not HTTPS it will fail.

I see you are right https has been enabled forever, but does the default configuration work?

I think the issue is that my PR exposed a bug in the configuration.  HTTPS has been enabled for two years, but it does not function by default.  My reordering has put that invalid https port first which now causes it to fail.  

From my understanding so far my PR did introduce this issue, but the root cause is HTTPS does not work even though it's setup by default.

sstarcher:
Try

```
proxy:
  secretToken: xxx
  https:
    enabled: false
```

sstarcher:
Please see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1641

albertmichaelj:
I have read the kubernetes issue before I commented on this pull request. Please see my previous linked issue which talks about seeing https to false as well. However, as I have stated a few times now, I cannot replicate what you're seeing with the failure of kubernetes 1.14. can you please provide a minimum example that demonstrates what you're seeing? In my experience, I am not affected by the apparent bug in kubernetes that you reference. I am trying to figure out why I'm not affected.

I'm trying to figure this out because I do not think that #1641 is sufficient to fix all possible problems that this reordering seems to introduce. Specifically, as I've stated in my last two comments this does not seem to solve the race condition around the health checks and granting of the let's encrypt certificate.

If this reordering isn't necessary (and unless I can figure out how to reproduce your issue, I'm not sure it is given that I'm not affected by it with kubernetes 1.14 and ssl turned on), the cleanest solution is to revert the ordering.

sstarcher:
The reordering is necessary when you terminate SSL at the ELB.  This only happens when a user annotated the service saying to enable HTTPS at the ELB.  This is needed when you want to use AWS ACM at the ELB.

I verified that my PR I just put in fixes the default configuration and that the default configuration is mistakenly enabling https as it's impossible for it to work.  We should not enable HTTPS in a manner that does not work as it will be confusing for users.

I can't speak to the letsencrypt workflow as I have never used let's encrypt.

sstarcher:
You need to enable HTTPS at the load balancer.  You can offload to HTTP or HTTPS.  The below config passes HTTPS through to the backend.  You will need your own ACM cert to do this.

```
proxy:
  service:
    annotations:
        service.beta.kubernetes.io/aws-load-balancer-backend-protocol: ssl
        service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:us-xxxx-x:123213213:certificate/1232321321
        service.beta.kubernetes.io/aws-load-balancer-ssl-ports: https
```

albertmichaelj:
Ah, I see why I'm not seeing the issue you ran into. I was not asking the load balancer to only use ssl, which doesn't seem to be the most common configuration (or at least expected configuration) based on the Jupyterhub docs. I tried again to set up a very simple jupyterhub set up with ssl, and there didn't seem to be a problem with the health checks even before the ssl cert is granted. I seemed to run into that problem when I was first testing things out, but there may have been other things going on since I hadn't really identified the problem at that point.

Using the below config, ssl seems to work and the health checks don't seem to fail even before the certificate is issued:

```
proxy:
  secretToken: "<REDACTED1>"
  https:
    enabled: true
    hosts:
      - "<REDACTED2>"
    letsencrypt:
      contactEmail: "<REDACTED3>"
```

Given this, I do think the only solution that likely works for both your case and the base case is to disable https by default, as you do in #1641. I think this likely has ramifications for at least one part of the documentation, but I'll mention that on the pull request.

PR #1535, opened 2019-12-27
[WIP] add support for internal_ssl

### LINKED ISSUE #1520: combine proxy and hub into a single pod so TLS gets terminated at the same pod
Currently, the proxy is implemented as a separate deployment.  The proxy supports TLS, but terminates TLS at its pod and forwards unencrypted traffic over to the hub.  This type of behavior is not optimal and for security reasons, it would be preferred to combine the proxy and the hub into the same pod.

Would you accept a PR to combine the proxy and the hub?

sstarcher:
Resolves #1520 this will need to wait until my PR into KubeSpawner is also released.
https://github.com/jupyterhub/kubespawner/pull/386

### DIFF
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -116,6 +116,34 @@ security report generator. Use the following URL structure to test your domain:
 http://ssllabs.com/ssltest/analyze.html?d=<YOUR-DOMAIN>
 ```
 
+### Internal JupyterHub TLS
+
+By default the internal communication of the Proxy, JupyterHub, and user notebooks are not encrypted by TLS.
+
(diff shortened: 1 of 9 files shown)


chancez:
I'm planning on working on an updated version of this. One thing I found is it's storing certs in `/srv/jupyterhub/internal-ssl` which is mounted as a PVC when using sqlite, but not when using an external DB, so it needs some changes to handle persistence of certificates. However, I believe I can also fully automate TLS provisioning using a Kubernetes job at install time (via a helm hook) so that it creates a secret that the pods can mount prior to their startup, also fixing the PVC problem since the certs will be stored in a secret, even for the hub pod.

PR #1539, opened 2020-01-08
Replace kube-lego + nginx ingress with traefik

### LINKED ISSUE #1448: letsencrypt now wants ACMEv2
I just tried to set up another z2jh and the autohttps pod is reporting the following error:  
```"Error while processing certificate requests: 403 urn:acme:error:unauthorized: Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2 / RFC 8555. See https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 for details." context=kubelego```

At https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 they sent out the following PSA yesterday: "The second production brownout for new ACMEv1 registrations has begun."

### PR DESCRIPTION
**Pull request by yuvipanda**
When we
[picked](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/229)
kube-lego to do automatic HTTPS in 2017, it seemed the easiest and
most kubernetes native way to do let's encrypt based HTTPS. It was
fairly simple, and worked pretty well for our use case. Users of
this chart could set up HTTPS without having to know anything about
certificates, challenges or let's encrypt. Life was good.

Life moves on, and so does code. kube-lego was deprecated in
favor of [cert-manager](https://github.com/jetstack/cert-manager),
which was far more general. It wasn't tied to Let's Encrypt,
and could do a lot more things. It was far more kubernetes native,
and slicker. However, this came with a massive increase in
complexity, which is justifyable if you need many certificates in
many situations. However, for the most common case of just one
certificate per hub, this was super massive overkill.

This PR replaces the current autohttps with traefik. traefik's
Let's encrypt implementation requires persistent
storage to put acme.json into, and we would prefer to not. This
was the primary reason I picked kube-lego over other implementations
(like certbot) back in
the day. However, a small wrapper script that sync's the contents
of acme.json to a kubernetes secret when needed is small,
easy to reason about and solves all our problems.

This PR should be fully backwards compatible. traefik should
be able to renew certificates that kube-lego was handling before,
and issue new certificates as well. Since it is just wrapping
upstream certbot, there are fewer chances of a deprecation hitting
us hard.

d22c69f72577809cce71d0499f5045e080234756 has information on why
we used traefik instead of nginx.

Fixes #1448

### CODE REVIEW

**Comment by consideRatio** [comment]
Excellent!!! I like this a lot!

**Comment by consideRatio** [comment]
This is soo valuable work! Lets make a quick 0.10.0 release with this when it lands. I'm looking forward to review this on sunday! I looked it through from mobile now snf it looks good to me so far.

**Comment by yuvipanda** [comment]
I rewrote this again to use traefik instead of nginx. d22c69f72577809cce71d0499f5045e080234756 has information on why.

File: jupyterhub/values.yaml [hunk 1]
```diff
@@ -148,15 +148,22 @@ proxy:
...
-      tag: 0.1.7
+      name: traefik
+      tag: v2.1
+    proxyBodySize: 64m
+    hstsIncludeSubdomains: 'false'
+    resources: {}
+  autocertbot:
+    image:
+      name: jupyterhub/k8s-autocertbot
+      tag: '0.9.0-alpha.1.112.34b7b8e'
```
betatim:
```suggestion
      tag: 'set-by-chartpress'
```
I think

jupyterhub/values.yaml
```diff
@@ -148,15 +148,22 @@ proxy:
...
     hstsIncludeSubdomains: 'false'
     resources: {}
-  lego:
+  traefik:
     image:
-      name: jetstack/kube-lego
-      tag: 0.1.7
+      name: traefik
+      tag: v2.1
+    proxyBodySize: 64m
```
betatim: Do we need this for the certificate issuer? Or does normal traffic flow through this proxy as well?

images/autocertbot/autocert.py
```diff
@@ -0,0 +1,134 @@
...
+certbot expects contents of /etc/letsencrypt to be persistent
+across runs, so it knows when to renew certificates & when to leave
+them alone. This is a problem in Kubernetes, since we try to avoid
+having persistent disks unless we must.
+
+This script saves / restores the contents of /etc/letsencrypt into
+a Kubernetes secret object, thus letting certbot operate unchanged
+without needing any persistent storage.
+
+This script runs as a sidecar to an nginx container that has webroot
```
betatim: I think this comment is outdated now that we use traefik. Can we replace it with the equivalent for the traefik setup or remove it?

jupyterhub/templates/proxy/autohttps/configmap.yaml
```diff
@@ -0,0 +1,69 @@
...
+      address = ":80"
+
+      [entryPoints.https]
+      address = ":443"
+
+    [wss]
+    protocol = "http"
+
+    [certificatesResolvers.le.acme]
+      email = "yuvipanda@gmail.com"
```
betatim: Should we read this from `.Values.proxy.https.letsencrypt.contactEmail` or does this value get overwritten somewhere else?

jupyterhub/values.yaml
yuvipanda: I need to figure out if this is needed. And yes, normal traffic flows through this proxy - it replaces the nginx ingress from earlier.

images/autocertbot/autocert.py
yuvipanda: I agree! I just wanted to push this up before I left. I think this script should be renamed secret-sync or something like that, and redocumented. I'll do that tomorrow.

jupyterhub/templates/proxy/autohttps/configmap.yaml
yuvipanda: Yep, needs to come from that.

images/autocertbot/autocert.py
yuvipanda: Redocumented!

jupyterhub/templates/proxy/autohttps/configmap.yaml
yuvipanda: Fixed!

jupyterhub/values.yaml
yuvipanda: Done!

yuvipanda:
This works really well!

Things left to do: 

- [ ] HSTS
- [ ] verify that saving large notebooks works

consideRatio:
@yuvipanda wieee! I figure we should release 0.9.0 without this and then go for 0.10.0 with this promptly after.

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -6,31 +6,19 @@ kind: Deployment
 metadata:
```
consideRatio: Should this be a deployment or statefulset? A deployment will do a rolling update and have two replicas running at the same time during updates while a statefulset would shut down and then start back up.

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -41,90 +29,66 @@ spec:
...
       nodeSelector: {{ toJson .Values.proxy.nodeSelector }}
       {{- include "jupyterhub.coreAffinity" . | nindent 6 }}
+      volumes:
+        - name: certificates
+          emptyDir: {}
+        - name: traefik-config
+          configMap:
+            name: traefik-proxy-config
+      initContainers:
+      - name: volume-mount-hack-why-god-still
```
consideRatio: I have some ideas to avoid the mount hack! Perhaps read the Woops section first below though.

## 1 - Read [here under "Secret files permissions"](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets)
Perhaps we use volumes `defaultMode` or volumeMounts `mode` configuration? Note that working in YAML, you can use octal specification even though you can't in JSON to specify.

## 2 - Mounting a specific key/value pair
I have two reasons why I figure this may be worth trying.

1. If we mount a specific key/value pair as a file, rather than the whole directory, there may be a significant difference for how the permissions are applied. I figure they may apply on the root of the mount rather than the recursively, leaving me uncertain about what will happen in general.

2. It sais in the official k8s documentation "A container using a Secret as a subPath volume mount will not receive Secret updates.". It probably is a good idea to not have that update mechanism active while having the sync-script do work at the same time.

## Example solution to avoid mount hack

```yaml
volumes:
  - name: certificates
    emptyDir: {}
    defaultMode: 0755
containers:
  volumeMounts:
    - name: certificates
      # Here we specify a file rather than a directory, as we specify a subPath
      # below to select the mounting of a single key/value pair in the secret
      mountPath: /etc/acme/acme.json
      subPath: acme.json
```

## Woops...

I forgot that you do not interact with a k8s secret volume here but a emptyDir volume, you do it all using kubernetes API from the sync-script... Hehehe. Ah well, but perhaps some points were of relevance still so I'll leave the text above unmodified.

jupyterhub/templates/proxy/autohttps/deployment.yaml
yuvipanda: I think deployment is the right call here - there isn't a lot of state here, and deployment will make sure we aren't dropping user requests during deployment.

jupyterhub/templates/proxy/autohttps/deployment.yaml
yuvipanda: Great catch, @consideRatio. I have removed the hack with a defaultMode instead.

yuvipanda:
I removed the proxyBodySize configurable, since traefik does not seem to have a problem with it as far as I can see.

I used the code in https://github.com/jupyterhub/mybinder.org-deploy/issues/210#issuecomment-349353103 to test, and I crashed my browser around the 160M mark before I ran into the entity too large issue

jupyterhub/templates/proxy/autohttps/deployment.yaml
yuvipanda: Correction - emptyDir defaults to the perms we want anyway, so I just removed the hack and defaultDir :)

yuvipanda:
From my POV, this is complete now :)

yuvipanda:
Thanks @consideRatio!

Is there a reason you wanna push out 0.9 without this? HTTPS is broken right now - if you follow our docs for HTTPS, it no longer works. Renewal works for cases where the first certificate was issued a while ago, but that is gonna stop soon though. I have been away from z2jh development for a while, so I will defer to you though :)

1. TLS certificate aquisition is done by traefik with its built in Let's Encrypt support. It stores the credentials (Let's Encrypt account + cert) in /etc/acme.json, which is sync'd with a kubernetes secret by secret-syncer. This replaces the nginx-ingress + kube-lego + internal-ingress object combination that exists now.
2. TLS termination is performed by traefik, in exactly the same way nginx does now. It then transmits decrypted HTTP connections to chp, where things go on as usual.
3. This does not touch on any of the routing for user pods at all. I'd like to keep these conversations separate, since this is fairly self contained and important enough by itself.

Hope this is useful!

consideRatio:
0.9.0 with without this PR makes sense to me to simply avoid breaking assumptions that not much should change between a beta and final release, that i have at least. I worked to long to land 0.9.0 now and i dont want to introduce another big change to verify for me and all users even though it is this important.

I know for example that GitLab has put in a lot of work to verify 0.9.0-beta.2 works for them now to deploy with the click of a button for their users and they wont use this HTTPS solution but instead their own cert-manager solution.

This is a major change after all, it is not hard to release a new version quickly following this so there is little downside doing that I figure. Perhaps we avoid waiting for JH 1.1.0 before cutting 0.9.0 as i currently do to speed it up though, merge this and put out 0.10.0-beta.1. Hmmmm.

scivm:
I think it is a good idea to make a release after the 0.9.0 that contains this and the internal_ssl pull requests in kube_spawner and this repo.  I can help with testing of that new 0.10.0-beta.1 beta.  There is currently no option but to fork if you want internal_ssl and I want to contribute to avoid having to fork.

yuvipanda:
> 0.9.0 without this PR makes sense to me to simply avoid breaking assumptions that not much should change between a beta and final release. I've worked long to land 0.9.0 and i want to avoid introduce a big change for all of us to verify as part of it kow that it is time yo release 0.9.0.

Thanks for the clarification, @consideRatio. IMO, breaking the (internal) assumption is better than putting out a release where our documented way to do HTTPS doesn't work (unless we removed all the automatic https documentation), especially when it is something as important as HTTPS. However, I have been away from the process enough that I trust your judgement :)

I'd love to get this merged and released ASAP regardless

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -41,90 +29,66 @@ spec:
...
-                fieldRef:
-                  fieldPath: metadata.namespace
-          livenessProbe:
-            httpGet:
-              path: /healthz
-              port: 10254
-              scheme: HTTP
-            initialDelaySeconds: 10
-            timeoutSeconds: 1
-          readinessProbe:
```
consideRatio: We could use the /ping endpoint here. Since it is a deployment now that should be able to do rolling updates, we benefit slightly from it. Example:

If we start up the new traefik pod during a rolling update, the old pod will start terminating as soon as the traefik container has started running even though it hasn't become ready. If we have a readiness check k8s wont shut down the previous traefik instance until it has become ready for real and for example able to respond to pings.

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -41,90 +29,66 @@ spec:
...
           {{- end }}
-          resources:
-            {{- .Values.proxy.lego.resources | toYaml | trimSuffix "\n" | nindent 12 }}
+          command:
+            - /usr/local/bin/secret-sync.py
+            - watch-save
+            - --label=app={{ include "jupyterhub.appLabel" . }}
+            - --label=release={{ .Release.Name }}
+            - --label=chart={{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+            - --label=heritage={{ .Release.Service }}
```
consideRatio:
```suggestion
```

jupyterhub/values.yaml
```diff
@@ -168,7 +169,9 @@ proxy:
     type: letsencrypt
     #type: letsencrypt, manual, offload, secret
     letsencrypt:
-      contactEmail: ''
+      contactEmail: 'yuvipanda@gmail.com'
```
```suggestion
      contactEmail: ''
```

jupyterhub/templates/proxy/autohttps/deployment.yaml
consideRatio: This should be a HTTP endpoint I think, because if not, we will never be able to meet an initial http01 challange.

images/secret-sync/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:3.7-alpine
+
+RUN pip install --no-cache kubernetes
+
+COPY secret-sync.py /usr/local/bin/secret-sync.py
```
consideRatio:
```suggestion
COPY secret-sync.py /usr/local/bin/secret-sync.py

```

jupyterhub/templates/proxy/autohttps/deployment.yaml
yuvipanda: Good catch! I have the heritage set to 'secret-sync' now

images/secret-sync/secret-sync.py
```diff
@@ -0,0 +1,162 @@
...
+
+## File system -> Kubernetes secret
+
+traefik might write new contents to the acme.json file over
+time, and we need to sync it to the kubernetes secret object.
+Ideally, we would watch for changes to the file with inotify
+and update the secret object as needed. However, for now we
+just operate in a 30s loop. This is good enough, since
+traefik can always re-generate certs if needed.
+"""
```
consideRatio: I love this excellent documentation about the script!

images/secret-sync/secret-sync.py
```diff
@@ -0,0 +1,162 @@
...
+
+    setup_logging()
+
+    if not args.namespace:
+        try:
+            with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace") as f:
+                args.namespace = f.read().strip()
+        except FileNotFoundError:
+            print("Can not determine a namespace, must be explicitly set with --namespace", file=sys.stderr)
+            sys.exit(1)
```
consideRatio: Oh, I'm learning seeing this. Looked up stuff in the docs found by googling /var/run/secrets/kubernetes.io

https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#service-account-admission-controller

jupyterhub/templates/proxy/autohttps/configmap.yaml
```diff
@@ -0,0 +1,106 @@
+{{- $HTTPS := (and .Values.proxy.https.hosts .Values.proxy.https.enabled) }}
+{{- $autoHTTPS := (and $HTTPS (eq .Values.proxy.https.type "letsencrypt")) }}
+{{- if $autoHTTPS -}}
```
consideRatio: Hmmm, should we only have this configmap if we want autohttps as compared to https with a manually provided secret?

yuvipanda: Right now manually provided secrets are handled at CHP, so we only need this with autoHTTPS

images/secret-sync/secret-sync.py
yuvipanda: Glad you like it!

jupyterhub/templates/proxy/autohttps/configmap.yaml
```diff
@@ -0,0 +1,106 @@
...
+            {{- range $host := .Values.proxy.https.hosts }}
+            [[http.routers.chp.tls.domains]]
+            main = "{{ $host }}"
+            {{- end}}
+
+    [http.services]
+      [http.services.chp.loadBalancer]
+        [[http.services.chp.loadBalancer.servers]]
+          url = "http://proxy-http:8000/"
+{{- end }}
```
consideRatio: This section overall was beyond me. Perhaps you could add some inline comments here, for example regarding [http.routers] and [http.middlewares] or what goes on in general.

jupyterhub/templates/proxy/autohttps/configmap.yaml
consideRatio: Wait what, is CHP doing TLS termination if the secret is manually provided?

yuvipanda: Yep, look at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/proxy/deployment.yaml#L61. I think we should move it to traefik in a different PR

jupyterhub/templates/proxy/autohttps/configmap.yaml
yuvipanda: Makes sense. I will spend some time doing that.

jupyterhub/templates/proxy/autohttps/configmap.yaml
consideRatio: Oh. This sums up my experience of the https matters in z2jh, a failure to get a good overview! I'll try catch up on this as quickly i can

consideRatio: I created #1545 to represent this discussion.

consideRatio: I created #1545 to represent this discussion.

jupyterhub/templates/proxy/autohttps/deployment.yaml
consideRatio: I created a new issue to represent this work to be done in a future PR instead.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1546

consideRatio:
We are currently not testing any of this autohttps logic in our CI system. Could we perhaps use something like xip + lets encrypt staging to try it out?

manics:
I don't think travis will allow incoming external connections which are required for lets-encrypt verification.

There's https://github.com/letsencrypt/boulder which lets you setup your own ACME server, and in theory could be used for testing autohttps. I tried (and failed) to set it up once, but if someone has time it might be worth giving it anpother go, or looking for some other implementation.

consideRatio:
@manics oh thanks for this input, you are right, we don't think we can get incoming traffic routed to our TravisCI worker VM, and if we could, it would be a bit hard to test locally using the same logic etc so it would be best to have a minimalistic something alongside in the k8s cluster I figure.

This has become one of these very fun PRs to mentally process for me :D

betatim:
I think we should try and get this into the current beta so it can be part of the release. I agree with Erik that big changes in the beta is annoying, however I think the risk that we make things worse by including this are small. Right now HTTPS via Let's encrypt/recommended way in the docs is broken, so if we have even a 30% chance that including this will un-break it we should do it.

This is a tradeoff with having to deal with support requests from people who get reminded of Z2JH via the publicity associated with a release. If we have a release with broken HTTPS there will be (my guess) lots of people asking how to get HTTPS support. To which the answer is kind of long and complicated.

---

Two things which I think we should improve (in this PR or a new one):
* I was caught off-guard by the Ingress Controller which is setup here. I didn't expect to find it in this part of the chart. It makes sense that it is here but as you can see from one of my questions above I needed reminding to remember that. Maybe a comment/doc can help with that. This is more of an architecture thing and has always been like this.
* the traefik config toml is mostly gobbledygook to me. I can read it but I can't explain why it is what it is. Erik commented on that as well. So maybe a few more comments in it or the docs will help a future maintainer understand why traefik is configured like it is. The combination of needing traefik knowledge and z2jh knowledge makes it tricky to understand.

I am a huge fan of testing to stop stuff from breaking :) I don't think we can have traffic from the outside reach the VM we use in travis. Can we (easily) configure traefik to issue a self-signed cert and check for that in a test? That way we would find out that traefik is running, issuing certs, routing traffic but we don't need the complexity of getting a (staging) let's encrypt cert.

consideRatio:
> I was caught off-guard by the Ingress Controller which is setup here. I didn't expect to find it in this part of the chart. It makes sense that it is here but as you can see from one of my questions above I needed reminding to remember that. Maybe a comment/doc can help with that. This is more of an architecture thing and has always been like this.

@betatim, this PR removes an NGINX ingress controller (something that inspects k8s ingress resources of a certain ingress-class as spotted by an annotation), and an internal-ingress.yaml resource.

I'm very keen to try outline the architecture better as it is quite complex, by necessity often. Can you try to clarify the point quoted above to make it easier for me or someone to get actionable about, because I didn't grasp it well enough.

yuvipanda:
# Without this PR

## No HTTPS

Incoming traffic hits chp, which then routes it to hub / user pods

## Manual HTTPS

Incoming traffic hits CHP, which does TLS termination, then routes it to hub / user pods

## Let's Encrypt HTTPS (autohttps)

### When it was working
Incoming traffic hit nginx-ingress, which did TLS termination, then routed it to CHP, which then routed to hub / user pods

### Now

If you turn on autohttps, your entire hub is broken, because kube-lego is deprecated and you can not get a new account. So if you want HTTPS, you need external complex options

# With this PR

## No HTTPS

No Change

## Manual HTTPS
s
No change

## Let's Encrypt HTTPS

Incoming traffic hits traefik, which does TLS termination, then routed to CHP, which routes to hub / user pods

So the major difference right now, is that without this PR *Let's Encrypt HTTPS is completely broken*, so nobody can really use the nginx ingress. If they had set up the hub before November 2019 (based on [this](https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430)), then the HTTPS will continue to renew until June 2020. However, if they have to move the hub, or re-install, or install a new hub - no HTTPS.

So IMO, this is a bugfix to unbreak HTTPS. I currently have it deployed from the PR for berkeley's big hubs, and it seems to work out well so far.

I'll try find some time today to write better docs for the traefik config!

Thank you for working on this slow moving, complex but important work <3

jupyterhub/templates/proxy/autohttps/configmap.yaml
yuvipanda: @consideRatio @betatim I've cleaned up and documented this section!

jupyterhub/templates/proxy/autohttps/deployment.yaml
yuvipanda: Thank you!

tools/templates/lint-and-validate-values.yaml
```diff
@@ -84,7 +84,6 @@ proxy:
         cpu: 100m
         memory: 512Mi
   nginx:
-    proxyBodySize: 64m
```
minrk: nginx and lego sections should be gone?

minrk:
Since testing letsencrypt isn't feasible right now, I went through a manual deployment following the guide (with `helm install ./jupyterhub/` instead of `jupyterhub/jupyterhub`) and everything worked great! With the last commit adding nice docs, I think we should go ahead and land this now.

Thanks for fixing this, @yuvipanda!

consideRatio:
@yuvipanda wieeeeeee awesome work! :heart: :tada: this feels sooooo good to have merged!

yuvipanda:
Thank you for the merge and testing, @minrk!

And thank you for reviewing, @consideratio & @betatim!

consideRatio:
0.9.0-beta.3 released, I gained a lot of confidence that this would not disrupt users not using autohttps along the way, thanks for explaining the difference between this TLS termination and CHP termination!

yuvipanda:
Thanks for making the release, @consideRatio!

I also wanted to note two things I discovered:

1. Upgrading from nginx autohttps requires you *delete* the old autohttps deployment. `kubectl -n hub-namespace delete deployment -l component=authottps`. This is necessary because we are changing some immutable fields in the deployment, and kubernetes will not do an in-place change. However, maybe `--force` works - I have not tested it.
2. This gets rid of a lot of our `ClusterRole`s, which will help with using this in higher security contexts

dahjah:
Stupid NOOB question on this (and this may be the wrong place to put this, I'm not sure):
Having just installed jupyterhub, it still uses the stable 0.8.2 release. Would it be safe to run 
`helm upgrade --install jhub jupyterhub/jupyterhub   --namespace jhub    --version=0.9.0-beta.3   --values config.yaml && kubectl --namespace=jhub get pod` to get this working, or would that be a bad idea?

consideRatio:
@dahjah yepp I think that should be safe to do.

dahjah:
Thank you @consideRatio. Works like a charm!
In case anyone comes here with the same problem later, I also had to delete the old autohttps deployment as mentioned above before it would work.

consideRatio:
Ohhh!!! VERY relevant advice thank you for describing this!

Issue #1491, opened 2019-11-25
Deploying on a cluster with strict PodSecurityPolicies


ltupin:
Hello there,

First of all I want to thank you a lot for all the work on this helm charts, its really changing the game and avoid hand-craft we did before to customize my jhub server !!!

### Situation
My cluster is secured with PodSecurityPolicies
```
Name:  privileged

Settings:
  Allow Privileged:                       true
  Allow Privilege Escalation:             0xc000304728
  Default Add Capabilities:               <none>
  Required Drop Capabilities:             <none>
  Allowed Capabilities:                   *
  Allowed Volume Types:                   *
  Allow Host Network:                     true
  Allow Host Ports:                       0-65535
  Allow Host PID:                         true
  Allow Host IPC:                         true
  Read Only Root Filesystem:              false
  SELinux Context Strategy: RunAsAny
    User:                                 <none>
    Role:                                 <none>
    Type:                                 <none>
    Level:                                <none>
  Run As User Strategy: RunAsAny
    Ranges:                               <none>
  FSGroup Strategy: RunAsAny
    Ranges:                               <none>
  Supplemental Groups Strategy: RunAsAny
    Ranges:                               <none>


Name:  restricted

Settings:
  Allow Privileged:                        false
  Allow Privilege Escalation:              0xc0008d0139
  Default Add Capabilities:                <none>
  Required Drop Capabilities:              ALL
  Allowed Capabilities:                    <none>
  Allowed Volume Types:                    configMap,emptyDir,projected,secret,downwardAPI,persistentVolumeClaim
  Allow Host Network:                      false
  Allow Host Ports:                        <none>
  Allow Host PID:                          false
  Allow Host IPC:                          false
  Read Only Root Filesystem:               false
  SELinux Context Strategy: RunAsAny
    User:                                  <none>
    Role:                                  <none>
    Type:                                  <none>
    Level:                                 <none>
  Run As User Strategy: MustRunAsNonRoot
    Ranges:                                <none>
  FSGroup Strategy: MustRunAs
    Ranges:                                1-65535
  Supplemental Groups Strategy: MustRunAs
    Ranges:                                1-65535
```

### Issue
As you can see containers cannot run as root and must run with a UID. This is best practices to deploy k8s like this accordly with documentation and bad issues like [CVE-2019-11245](https://nvd.nist.gov/vuln/detail/CVE-2019-11245).

[...]

PR #1540, opened 2020-01-09
 Deploying on a cluster with strict PodSecurityPolicies

ltupin:
Regarding the issue I've opened here:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1491

I propose those changes to be able to activate a default configuration to run containers with UID if a securityContext who block username exist on the cluster.


jupyterhub/values.yaml
```diff
@@ -94,6 +94,8 @@ rbac:
 
 
 proxy:
+  securityContext:
+    enable: false
```
RAbraham: typo, should it be `enabled`?

ltupin: Yes you right ! thx !

consideRatio:
I need better understanding to review this PR.

I understand, that for some reason, a securityContext is needed, but I've no experience with or overview understanding of it myself. I know that they exist on the pod level and the container level. The container level is a finer scope than the pod level and they don't have exactly the same configuration.

For this or similar PRs to be merged, I'd like this kind of understanding to be written out explicitly. I have not gotten time to read up on this subject yet =/

Action points:
- [ ] Clarify the situation, why does pod specification or their container specifications need a securityContext sometimes, and do we need specific fields?
- [ ] Consider the perspective of maintenance. I currently lean towards allowing the user to provide the full securityContext structure for various pods, as compared to allowing them to provide specific values to specific fields, as that makes the helm chart more complicated and harder to keep up to date with things.

Thank you @ltupin for your work on this and sharing your solution! It is my understanding that these changes solved an issue for you in #1491!

/ ping @metonymic-smokey, regarding being part of the review process as a way to contribute!

metonymic-smokey:
@consideRatio thank you for the opportunity!
Will try to contribute to this review process by learning about PodSecurityPolicies!

jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -82,6 +87,11 @@ spec:
         {{- end }}
         {{- if not .Values.singleuser.cloudMetadata.enabled }}
         - name: image-pull-metadata-block
+          {{- if .Values.hookimagepuller.securityContext.enabled }}
```
yuvipanda: Can we just let the user specify the whole securityContext? That way, they can set other stuff too (like capabilities, gid, etc)

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -58,6 +58,7 @@ spec:
       serviceAccountName: hub
       {{- end }}
       securityContext:
```
yuvipanda: Same here - can we move securityContext to values.yaml and let users override it completely? That would be more flexible, and require fewer overrides in the future.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
yuvipanda: It also means we don't have to use the hub uid everywhere.

jupyterhub/templates/hub/deployment.yaml
mriedem: Somewhat agree, the proposal in #1677 is to be able to specify additional fields in order to integrate [py-spy](https://github.com/benfred/py-spy#how-do-i-run-py-spy-in-kubernetes), in this case `capabilities`. But if we can generically override everything in here via `values.yaml` like @betatim pointed out [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1491#issuecomment-559855666) then that would probably work for us as well.

mriedem: Actually this might not work for our case in #1677 because this is the pod `securityContext` and according to https://github.com/benfred/py-spy#how-do-i-run-py-spy-in-kubernetes we need to set `capabilities` in the `container` spec below.

consideRatio:
@ltupin thank you so much for your work on thoroughly documenting this in #1491 and also pushing towards a resolution with this PR! With your insights as a starting point I've now pushed towards setting a configurable securityContext by default on all our containers except the user pods which already had some parts on them, this was done in #1798.

Thank you for your contribution! :heart: :tada: Note that if you try out a release following #1798, please let me know if you experience any issues or not, if you do I want to work quickly to resolve them and if not it would be a good confirmation that things are alright.

Issue #1532, opened 2019-12-24
mention in docs that the helm section refers to helm2


raybellwaves:
Before https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1509 you could add a line at the top of https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/setup-helm.html to say this section refers to helm2 and is subject to change with the latest version of helm (helm3)

PR #1543, opened 2020-01-12
Helm 3 preview

manics:
Adds Helm 3 to the Travis matrix.

Adds a Helm 3 preview doc.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1532

### DIFF
--- a/ci/common
+++ b/ci/common
@@ -55,12 +55,18 @@ setup_kind () {
 setup_helm () {
     echo "setup helm ${Z2JH_HELM_VERSION}"
     if ! [ -f "bin/helm-${Z2JH_HELM_VERSION}" ]; then
-        curl -sSLo "bin/helm-${Z2JH_HELM_VERSION}.tar.gz" "https://storage.googleapis.com/kubernetes-helm/helm-v${Z2JH_HELM_VERSION}-linux-amd64.tar.gz"
+        curl -sSLo "bin/helm-${Z2JH_HELM_VERSION}.tar.gz" "https://get.helm.sh/helm-v${Z2JH_HELM_VERSION}-linux-amd64.tar.gz"
         tar -xf "bin/helm-${Z2JH_HELM_VERSION}.tar.gz" --directory bin --strip-components 1 linux-amd64/helm
         rm "bin/helm-${Z2JH_HELM_VERSION}.tar.gz"
         mv bin/helm "bin/helm-${Z2JH_HELM_VERSION}"
     fi
     cp bin/helm-${Z2JH_HELM_VERSION} bin/helm
+}
+
+init_helm () {
+    if [ "${Z2JH_HELM_VERSION%%.*}" = "2" ]; then
+        helm init --client-only
+    fi
 }
 
 setup_kubeval () {
--- a/dev
+++ b/dev
@@ -181,17 +181,13 @@ def kind_create(recreate):
 
     if os.getenv('Z2JH_HELM_VERSION', '').startswith('3.'):
         print("Creating K8s namespace for Helm 3 to work in.")
-        k8s_ns = {
-            "apiVersion": "v1",
-            "kind": "Namespace",
-            "metadata": { "name": os.environ["Z2JH_KUBE_NAMESPACE"] },
-        }
         _run(
             cmd=[
-                "sh", "-c", "echo '{}' | kubectl create -f -".format(
-                    json.dumps(k8s_ns))
+                "kubectl", "create", "namespace",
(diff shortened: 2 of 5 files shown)


ci/common
```diff
@@ -55,14 +55,20 @@ setup_kind () {
...
+        curl -sSLo "bin/helm-${Z2JH_HELM_VERSION}.tar.gz" "https://get.helm.sh/helm-v${Z2JH_HELM_VERSION}-linux-amd64.tar.gz"
         tar -xf "bin/helm-${Z2JH_HELM_VERSION}.tar.gz" --directory bin --strip-components 1 linux-amd64/helm
         rm "bin/helm-${Z2JH_HELM_VERSION}.tar.gz"
         mv bin/helm "bin/helm-${Z2JH_HELM_VERSION}"
     fi
     cp bin/helm-${Z2JH_HELM_VERSION} bin/helm
 }
 
+init_helm () {
+    if [ "${Z2JH_HELM_VERSION%%.*}" = "2" ]; then
```
consideRatio:
```suggestion
    if [ "${Z2JH_HELM_VERSION:0:1}" = "2" ]; then
```

I googled `%%` to learn what it meant. It was "remove suffix" matching what follows. And matching everything removed everything except the first character. Aha... But if we want to only show the first character, I figured it made sense to use this logic instead as that may be quicker to understand for someone like me for example that isn't a linux user since birth ;D

dev
```diff
@@ -177,36 +179,51 @@ def kind_create(recreate):
...
+            "kind": "Namespace",
+            "metadata": { "name": os.environ["Z2JH_KUBE_NAMESPACE"] },
+        }
+        _run(
+            cmd=[
+                "sh", "-c", "echo '{}' | kubectl create -f -".format(
+                    json.dumps(k8s_ns))
+            ],
+            print_end="",
+        )
```
consideRatio: Perhaps `kubectl create namespace NAME` is more suitable here? I'm impressed you got the current solution working!

dev
```diff
@@ -56,7 +57,8 @@ def depend_on(binaries=[], envs=[]):
     return decorator_depend_on
 
 
-@depend_on(binaries=["kubectl", "kind", "docker"], envs=["Z2JH_KUBE_VERSION"])
+@depend_on(binaries=["kubectl", "kind", "docker"],
+    envs=["Z2JH_KUBE_VERSION", "Z2JH_HELM_VERSION"])
```
consideRatio: We should probably not specify this here as we don't need it to have a value in the script. It made sense for me to introduce in order to enforce specifications of various variables.

Hmm... Thinking about it, if Z2JH_HELM_VERSION is omitted, we will use the currently available helm CLI, which could be v2 or v3. It would make more sense that the logic below checked for the helm version using the helm cli to determine how to act instead of probing this variable.

ci/common
manics: `%%.*` means remove everything starting from the first `.` that's found, so it'll still work if helm reaches a double digit major version 😀. I'm happy to go with your suggestion though.

dev
manics: I remember now, I wanted to use `kubectl apply` so that it'd work even if the namespace already existed, but must have accidentally used `create` here instead. I'll gop with your suggestion of `exit_on_error=False`

dev
manics: I'm a bit worried about autodetecting config in CI, because if someone makes an incorrect change to any of the CI scripts or Travis config the auto-configuration could skip over it and the tests would appear to pass, when in fact a portion of them might not have been run.

It could be made opt-in (add another flag e.g. `dev --autodetect-env`?) though that's still not very user friendly.

manics:
IIRC Circle CI will build and publish the PR docs, can you remind me how to find them? Can't see anything obvious in the job status https://circleci.com/gh/jupyterhub/zero-to-jupyterhub-k8s/817

manics:
@consideRatio Any more thoughts?

dev
consideRatio: Okay, we don't need to autodetect, but I think we should not require the explicit specification of the helm version using the depend_on decorator that will make the function fail unless it is specified. We are looking at the env var with a default of a blank string later anyhow.

So, not using the `helm version` to get the version of helm, but also not requiring the Z2JH_HELM_VERSION to be specified, makes sense to me short term. I want to revise this dev script again as soon as I find time btw. It is not good enough for local development :(

manics: I understand now, done!

consideRatio:
:heart: thanks for your work on this @manics! :tada:

PR #1553, opened 2020-01-23
Added documentation of secret https mode

RossRKK:
I added documentation explaining how to configure https using a secret resource.

Storing private keys in a secret resource is better practice than stroing secret information like that in a plain text file.

### DIFF
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -54,6 +54,10 @@ changes to your `config.yaml` file:
 
 If you have your own HTTPS certificates & want to use those instead of the automatically provisioned Let's Encrypt ones, that's also possible. Note that this is considered an advanced option, so we recommend not doing it unless you have good reasons.
 
+There are 2 ways to specify your manual certificate, directly in the config.yaml or by creating a `secret`.
+
+#### Specify certificate in config.yaml
+
 1.  Add your domain name & HTTPS certificate info to your `config.yaml`
 
     ```yaml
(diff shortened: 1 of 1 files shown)


doc/source/administrator/security.md
```diff
@@ -54,6 +54,10 @@ changes to your `config.yaml` file:
 
 If you have your own HTTPS certificates & want to use those instead of the automatically provisioned Let's Encrypt ones, that's also possible. Note that this is considered an advanced option, so we recommend not doing it unless you have good reasons.
 
+There are 2 ways to specify your manual certificate, directly in the config.yaml or by creating a `secret`.
```
consideRatio:
```suggestion
There are two ways to specify your manual certificate, directly in the config.yaml or by creating a [Kubernetes `secret`](https://kubernetes.io/docs/concepts/configuration/secret/).
```

doc/source/administrator/security.md
```diff
@@ -73,8 +77,26 @@ If you have your own HTTPS certificates & want to use those instead of the autom
...
 
-2.  Apply the config changes by running helm upgrade ....
-3.  Wait for about a minute, now your hub should be HTTPS enabled!
+2. Apply the config changes by running helm upgrade ....
+3. Wait for about a minute, now your hub should be HTTPS enabled!
+
+
+#### Specify certificate through Secret resource
+
+1. Create a `secret` resource with type `kubernetes.io/tls` containing your certificate. Add your domain and the name of your `secret` to your config.yaml.
```
consideRatio: Make this two points, where the initial one could be a `kubectl create secret` command or similar.

RossRKK: The reason I became aware of this option was because I wanted to load the certificate from an azure key vault. My secret was created by [azure vault to kubernetes](https://github.com/SparebankenVest/azure-key-vault-to-kubernetes). I agree adding the command would be helpful but do you think it's worth including something about cloud service integration (with azure in this case) or do you think that is beyond the scope of this documentation?

RossRKK: I have added an example of how to create a secret resource

consideRatio:
Thank you @RossRKK for your work on this! :heart: :tada:!

PR #1558, opened 2020-01-27
[MRG] Add option to remove named server when culling

betatim:
In a setup like a BinderHub with auth I enable the named server option to allow users to launch the same repository multiple times in an attempt to keep the experience as much like mybinder.org as possible. The reason the bhub has auth is because it has access to private repositories.

This PR adds a flag to the culler to allow it to remove a named server when it culls the server. With the old culler you have to enable culling of users and wait roughly two culling periods before a user who has maxed out their number of allowed named servers can start a new server again. (Or you have to teach users about the control panel and deleting servers there manually.)

I don't really want to cull users as that means you have to click the login button again.

What do people think? What is missing from this PR in terms of helm chart'ing? (it has been a while since I've made my last contribution this deep inside this project).

### DIFF
--- a/jupyterhub/files/hub/cull_idle_servers.py
+++ b/jupyterhub/files/hub/cull_idle_servers.py
@@ -322,7 +322,7 @@ if __name__ == '__main__':
            )
     define('remove_named_servers', default=False,
            help="""Remove named servers in addition to stopping them.
-                This is for use in temporary-user cases such as Binderhub.""",
+                This is useful for a BinderHub that uses authentication and named servers.""",
            )
     define('concurrency', default=10,
            help="""Limit the number of concurrent requests made to the Hub.
--- a/jupyterhub/files/hub/cull_idle_servers.py
+++ b/jupyterhub/files/hub/cull_idle_servers.py
@@ -179,6 +179,10 @@ def cull_idle(
         body = None
         if server_name:
             # culling a named server
+            # A named server can be stopped and kept available to the user
+            # for starting again or stopped and removed. To remove the named
+            # server we have to pass an additional option in the body of our
+            # DELETE request.
             delete_url = url + "/users/%s/servers/%s" % (
                 quote(user['name']),
                 quote(server['name']),
--- a/jupyterhub/files/hub/cull_idle_servers.py
+++ b/jupyterhub/files/hub/cull_idle_servers.py
@@ -80,7 +80,15 @@ def format_td(td):
 
 
 @coroutine
-def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concurrency=10):
+def cull_idle(
+    url,
+    api_token,
+    inactive_limit,
+    cull_users=False,
+    remove_named_servers=False,
+    max_age=0,
+    concurrency=10,
+):
     """Shutdown idle single-user servers
 
     If cull_users, inactive *users* will be deleted as well.
@@ -168,16 +176,24 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
                 log_name, format_td(age), format_td(inactive))
             return False
 
+        body = None
         if server_name:
             # culling a named server
-            delete_url = url + "/users/%s/servers/%s"%(
-                quote(user['name']), quote(server['name'])
+            delete_url = url + "/users/%s/servers/%s" % (
+                quote(user['name']),
+                quote(server['name']),
             )
+            if remove_named_servers:
+                body = json.dumps({"remove": True})
         else:
             delete_url = url + '/users/%s/server' % quote(user['name'])
 
         req = HTTPRequest(
-            url=delete_url, method='DELETE', headers=auth_header,
+            url=delete_url,
+            method='DELETE',
+            headers=auth_header,
+            body=body,
+            allow_nonstandard_methods=True,
         )
         resp = yield fetch(req)
         if resp.code == 202:
@@ -302,7 +318,11 @@ if __name__ == '__main__':
            help="The maximum age (in seconds) of servers that should be culled even if they are active")
     define('cull_users', default=False,
            help="""Cull users in addition to servers.
-                This is for use in temporary-user cases such as tmpnb.""",
+                This is for use in temporary-user cases such as BinderHub.""",
+           )
+    define('remove_named_servers', default=False,
+           help="""Remove named servers in addition to stopping them.
+                This is for use in temporary-user cases such as Binderhub.""",
            )
     define('concurrency', default=10,
            help="""Limit the number of concurrent requests made to the Hub.
@@ -332,6 +352,7 @@ if __name__ == '__main__':
         api_token=api_token,
         inactive_limit=options.timeout,
         cull_users=options.cull_users,
+        remove_named_servers=options.remove_named_servers,
         max_age=options.max_age,
         concurrency=options.concurrency,
     )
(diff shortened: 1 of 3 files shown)


consideRatio:
I think this LGTM, but I don't understand it all yet.

I'd appreciate an inline comment clarifying or referencing the API where we pass a body to a DELETE webrequest of `remove: true`. I found this: https://jupyterhub.readthedocs.io/en/1.1.0/_static/rest-api/index.html#operation--users--name--servers--server_name--delete, but I didn't see anything about `remove: true`.

What is going on with this?

betatim:
I found out about the body argument via https://github.com/jupyterhub/jupyterhub/blob/65de8c49167738fde9d7e88245593301a2a6a42a/jupyterhub/tests/test_named_servers.py#L176. The code for the handler is https://github.com/jupyterhub/jupyterhub/blob/65de8c49167738fde9d7e88245593301a2a6a42a/jupyterhub/apihandlers/users.py#L420

I think what is happening is that you can "stop" a named server and you can "remove" it. Stopping is what the culler normally does. However for a named server you then still have an entry in your named server list that lets you restart or remove the named server. With an unnamed server the "remove" doesn't exist, as soon as it is stopped it is gone.

For my BinderHub setup it doesn't make sense to have the named servers stick around as there is nothing for people to go back to compared to just starting a new one (we don't mount home directories or such).

I am not sure what the best way to make a minimal example to demo this is. Maybe start a local JupyterHub with named servers. Then you get an overview of all your named servers on the user's "home page" (where you can start and stop your server and have the "My Server" button). Once you start a named server it persists in the list, even after getting stopped. To remove it you have to explicitly click "remove".

---

Do you  know where the API doc you linked to is generated from? I was expecting to find a doc string in [the code](https://github.com/jupyterhub/jupyterhub/blob/65de8c49167738fde9d7e88245593301a2a6a42a/jupyterhub/apihandlers/users.py#L4200) but there isn't one :( I think it would be worth adding the body argument to the docs there.

jupyterhub/files/hub/cull_idle_servers.py
```diff
@@ -168,16 +176,24 @@ def handle_server(user, server_name, server):
...
         else:
             delete_url = url + '/users/%s/server' % quote(user['name'])
 
         req = HTTPRequest(
-            url=delete_url, method='DELETE', headers=auth_header,
+            url=delete_url,
+            method='DELETE',
+            headers=auth_header,
+            body=body,
+            allow_nonstandard_methods=True,
```
manics: `method='DELETE'` is unchanged, why is `allow_nonstandard_methods=True` now required?

jupyterhub/files/hub/cull_idle_servers.py
```diff
@@ -302,7 +318,11 @@ def handle_user(user):
            help="The maximum age (in seconds) of servers that should be culled even if they are active")
     define('cull_users', default=False,
            help="""Cull users in addition to servers.
-                This is for use in temporary-user cases such as tmpnb.""",
+                This is for use in temporary-user cases such as BinderHub.""",
+           )
+    define('remove_named_servers', default=False,
+           help="""Remove named servers in addition to stopping them.
+                This is for use in temporary-user cases such as Binderhub.""",
```
manics: These are authenticated not temporary users.

Will this also remove the default server? If not is there a way to enable that?

jupyterhub/files/hub/cull_idle_servers.py
betatim: Because we (sometimes) pass a `body` which isn't allowed by the strict mode.

jupyterhub/files/hub/cull_idle_servers.py
betatim: I think your default server gets removed completely already by the standard culling. As I understand it your named servers as slightly different in that they don't vanish completely if you stop them. You can still see them in the dashboard and start them again.

Updated the help text.

betatim:
I added an inline comment about stop&keep vs stop&remove for named servers.

consideRatio:
I've not been able to spend much time on this PR yet. The key question in my mind regards desired behavior still.

Questions regarding desired behavior:
- Does a z2jh deployment enabling culling of named servers always want the named servers to be removed from the JupyterHub UI after being stopped by culling?
- If not, what do we do? Should we opt for always doing one or the other, or should we opt for making the "stop" or "stop and remove" choice configurable?

Too large for this PR to encompass, but a relevant question:
- What would a minimalistic test be here? We have scripts that does a few things, but we don't have a test of the culler yet. I think we should make a new PR for testing the culler etc, but it would be nice to test both normal culling and named-server culling.

betatim:
I think the default should continue to be to not remove named servers for a JupyterHub setup (default use case of Z2JH). This PR implements this. If you upgrade from an older chart to the one after this is merged, nothing will change. Once this has been merged you can enable the removal of named servers as part of the culling.

Special setups like using a JH for BinderHub might want to enable this (what default to use should be a discussion in the binderhub helm chart repo).

jupyterhub/values.yaml
```diff
@@ -357,6 +357,7 @@ ingress:
 cull:
   enabled: true
   users: false
+  namedServers: false
```
consideRatio: The script has "remove named servers", while the user facing z2jh configuration is simply to "cull" it. I dislike that it isn't apparent that named servers are stopped and removed, when they could also for example simply be stopped.

I got to run atm, if we can improve this a bit, which isn't great to change later, I'd be happy to merge.

We could for example rename this and potentially also add another one that allows for only stopping of named servers.

betatim:
I thought I'd add some screenshots of how things look. I have a BinderHub setup with auth. WIthout named servers this means each user can only launch one repository, unlike with a public BinderHub where you can launch as many different repos as you like simultaneously. The reason for this limit is that each user only has one default server.

This means I enabled the "named servers" feature in JupyterHub. Now a user can launch several servers (they each get a different name). IF you go to the "dashboard" page of the JupyterHub you see something like this

Screenshot 2020-01-31 at 07 35 47

after you launch a repo. If you click the "Stop" button the named server is stopped. This is also what the culler does by default. After the server has stopped the dashboard looks like the following:

Screenshot 2020-01-31 at 07 36 06

This means if you were to start the same repo or another repo now you'd end up with two entries in this list. And so on each time you start a new repository. To remove an entry from this list you have to click the "Remove" button (after first clicking stop). When a user clicks "Remove" here the API request is equivalent to what the culler proposed in this PR does when `cull.namedServers: true`.

There are two motivations for wanting the culler to not only stop but also remove the named server:
* I want to limit how many concurrent named servers a user can have
* I don't want to have to explain to users how they can get to this dashboard and what the procedure for removing servers is (and why they probably don't ever want to click "start").

minrk:
I opened https://github.com/jupyterhub/jupyterhub/pull/2917 because `remove` was *meant* to be in the rest api docs, but it was in the wrong place.

jupyterhub/values.yaml
minrk:
```suggestion
  removeNamedServers: false
```

👍 to adding removeNamedServers, since just having `namedServers` here looks like named servers weren't getting culled at all prior to this PR. Since remove is an additional action to culling (which is already working on named servers), including the additional verb in the config parameter makes sense to me.

betatim: Somehow missed this comment for days. I like the new name. The old one was chosen to try and be consistent with the `users` option, so not a lot of thought went into it.

minrk:
Perfect, thanks!

Issue #1221, opened 2019-04-05
Use pipfile.lock for more reproducible image builds


minrk:
Right now we have pinned some direct dependencies in our images, e.g. [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.8.2/images/hub/requirements.txt). This style is vulnerable to upgraded dependencies, e.g. tornado 6.

We should consider switching to using Pipfile.lock to only use a fully specified environment in our images and explicitly upgrade from a 'loose' Pipfile (or requirements.txt). The main thing this will require is a nice automated freeze step for generating the new Pipfile.lock files.

consideRatio:
Ah I like that, I also like pipenv ;) I've had a good learning experience with that as compared to other virtual environments.

consideRatio:
We now have a system solving this very well. As an example of it in use, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1768.

PR #1560, opened 2020-01-30
simplify hub-requirements

minrk:
remove verbose comments that make it harder to see what's in the hub image and update dependencies

pypi links are always pypi.org/project/$name so duplicating that info in links doesn't make them easier to find

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -1,83 +1,29 @@
 # JupyterHub's version is defined in chartpress.yaml
 
 ## Authenticators
-# ---------------------------------------------------------
-
-# https://github.com/jupyterhub/dummyauthenticator
-# https://pypi.org/project/jupyterhub-dummyauthenticator
 jupyterhub-dummyauthenticator==0.3.*
-
-# https://github.com/jupyterhub/firstuseauthenticator
-# https://pypi.org/project/jupyterhub-firstuseauthenticator
 jupyterhub-firstuseauthenticator==0.12.*
-
-# https://github.com/jupyterhub/tmpauthenticator
-# https://pypi.org/project/jupyterhub-tmpauthenticator
 jupyterhub-tmpauthenticator==0.6.*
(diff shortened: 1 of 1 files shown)


minrk:
I'm sorry, I don't want to make things harder for you. However, I made this PR for exactly the same reason–keeping this file up-to-date is harder *with* these changes. Maybe there is a compromise? What was your workflow like that was helped by these links? Adding a large amount of overhead and links costs a lot of space and makes the file harder to read and understand what is going on, at least for me. I removed the PyPI links because they are necessarily redundant with the package names (unlike repo links, which aren't known for a given package name). I removed the repo links because repos *shouldn't* be relevant when considering modifications for this file. Can I ask where you found a need to link directly to development repos when looking at updating the packages in the Hub image, especially for something like psycopg2?

minrk:
I also think we should be automating keeping these files up-to-date with pip-tools / pipfiles, etc. rather than needing to go searching for dependencies (#1221). That should hopefully have a bigger impact than adding links for manual checks.

consideRatio:
:heart: let's go with pip-tools for example, my colleague @shuyi-li recommended it today as well and I appreciated that generated locked requirements file had comments about why it was listed! I have not tried it yet myself but I think we should indeed opt for using a freeze/lock file of some kind.

My workflow regarding links: I wondered what version had we currently specified, what was available on PyPI, what had the repo currently had unreleased that wasn't part of the latest release, and what the changelog said about upgrading. Then, depending on the situation, I opted to make a release happen in the dependency repo, add notes about something impacting us, just bump it in our requirements file.

minrk:
Thanks for the info! Let's try out pip-tools with some docker-based workflow for automatically keep these things up-to-date and automated so that manual checking isn't as tedious.

I would like to perhaps separate concerns a bit and not consider checking for unreleased changes part of what should be on our minds when checking if the images here are up-to-date.

PR #1562, opened 2020-01-31
freeze environment in hub image

minrk:
separates our 'loose' dependencies that we specify from the pinned environment.

Adds a `dependencies` script to the hub image, which wraps running `pip-compile` and `pip list --outdated` in the image to try to automate the process of keeping these things up-to-date.

How it works:

- builds the hub image, taking build arguments from chartpress config, installing from requirements.txt
- requirements.in contains our human-edited dependencies
- requirements.txt contains a full, frozen environment, autogenerated by pi
- `./dependencies freeze` runs `pip-compile` *in the final image* to generate requirements.txt. This ensures the environment is the same without maintaining a separate image, at the expense of building the image more often than is strictly necessary
- `./dependencies outdated` runs `pip list --outdated` in the final image (after installing from the current requirements.txt) and gives a report of what packages are out-of-date

For example, running `./dependencies outdated` immediately after writing it gave me this output:

> ```
> Checking for outdated dependencies with pip.
> Have jupyterhub-firstuseauthenticator==0.12, latest is jupyterhub-firstuseauthenticator==0.13.0
> There are outdated dependencies!
> To pick up any versions outside the range(s) specified in requirements.in,
> update that file.
> To update the environment, run `./dependencies freeze`
> ```

I updated `requirements.in` to have `jupyterhub-firstuseauthenticator==0.13.*`
and re-ran `./dependencies freeze`, which produced ff9873bf62e48f06449b13f13674e253c2356051

TODO:

- [ ] test/discuss design/API/workflow
- [ ] promote workflow to other images if we like it, either in this PR or after merging and learning from use on this one image for a while
- [x] consider removing most pinning in requirements.in, since for ~everything there we just want 'latest'. The only known pin-down that I think we really want there is kubernetes, which we want to update pretty conservatively and explicitly.
- [x] Maybe allow upgrades of individual packages more explicitly with `pip-compile --upgrade-package X`?
      This only works within the ranges pinned in requirements.in, so suggests that we remove most pins there.


minrk:
Updates with last commit, after learning a bit about pip-compile. Now that `requirements.txt` exists, we must run `pip-compile --upgrade` to upgrade all dependencies or `pip-compile --upgrade-package jupyterhub-kubespawner` to upgrade  one or more specific packages. These options are now passed through by `freeze`, so the workflow would be:

1. `./dependencies outdated` to check for updates, and
1. **if ranges should be changed** edit `requirements.in` 
1. run one (or more) of:
   1. `./dependencies freeze` with no args, should only make changes if requirements.in is still satisfied
   1. `./dependencies freeze --upgrade` to upgrade all dependencies within pinnings of requirements.in
   1. `./dependencies freeze --upgrade-package jupyterhub-kubespawner` to upgrade specific packages within pinnings

images/hub/dependencies
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+"""automatically manage requirements.txt dependencies with pip-tools
+
+./dependencies --help for info
```
consideRatio: I'd appreciate some additional description here to summarize this.

Here are stuff I ended up writing down to try illustrate what I imagine could be useful to write here.

This script can identify and report how frozen representation of `requirements.in` that resides in `requirements.txt` is outdated. Also, when an update has been made to `requirements.in`, this script can update the frozen representation.   

This script is assumed to run in a folder with a `Dockerfile` file, a `requirements.in` file, a `requirements.txt`, and a `chartpress.yaml` file two folders up (`../../chartpress.yaml`) in order to function properly.

images/hub/requirements.txt
```diff
@@ -1,29 +1,78 @@
...
+importlib-metadata==1.5.0  # via jsonschema
+ipython-genutils==0.2.0   # via traitlets
+jinja2==2.11.1            # via jupyterhub, jupyterhub-kubespawner
+jsonschema==3.2.0         # via jupyter-telemetry
+jupyter-telemetry==0.0.4  # via jupyterhub
+jupyterhub-dummyauthenticator==0.3.1
+jupyterhub-firstuseauthenticator==0.13.0
+jupyterhub-hmacauthenticator==0.1
+jupyterhub-kubespawner==0.11.1
+jupyterhub-ldapauthenticator==1.2.2
```
consideRatio: I'd like ldapauthenticator 1.3.0 be part of beta.4 that is now out, so perhaps refreeze -> get this merged and -> release beta.4?

consideRatio:
Builds fails due to sphinx 3.0.0 is out and that have introduced an issue for circleci.

consideRatio:
@minrk I tried this out and I love it, this is brilliant!

PR #1563, opened 2020-01-31
First-class azuread support

minrk:
Adds aliases for `auth.type = azuread`, so

```yaml
auth:
  type: azuread
  azuread:
    clientId: ...
    clientSecret: ...
    tenantId: ...
    usernameClaim: ...
    callbackUrl: "https://<your_jupyterhub_host>/hub/oauth_callback"
```

should work now. OAuthenticator is updated to 0.11, which fixes some azuread issues, including the requirement to set tenant id via env, not just standard config.

/cc @cnf

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -293,6 +293,16 @@ elif auth_type == 'gitlab':
         if cfg_key is None:
             cfg_key = camelCaseify(trait)
         set_config_if_not_none(c.GitLabOAuthenticator, trait, 'auth.gitlab.' + cfg_key)
+elif auth_type == 'azuread':
+    c.JupyterHub.authenticator_class = 'azuread'
+    for trait, cfg_key in common_oauth_traits + (
+        ('tenant_id', None),
+        ('username_claim', None),
+    ):
+        if cfg_key is None:
+            cfg_key = camelCaseify(trait)
+
+        set_config_if_not_none(c.AzureAdOAuthenticator, trait, 'auth.azuread.' + cfg_key)
 elif auth_type == 'mediawiki':
     c.JupyterHub.authenticator_class = 'oauthenticator.mediawiki.MWOAuthenticator'
     for trait, cfg_key in common_oauth_traits + (
--- a/doc/source/administrator/authentication.rst
+++ b/doc/source/administrator/authentication.rst
@@ -172,6 +172,26 @@ tape archive, public cloud, or your own laptop. Start a Globus app
        callbackUrl: "https://<your_jupyterhub_host>/hub/oauth_callback"
        identityProvider: "youruniversity.edu"
 
+
+Azure Active Directory
+^^^^^^^^^^^^^^^^^^^^^^
+
+Azure Active Directory <https://docs.microsoft.com/en-us/azure/active-directory/>`_
+is an identity provider from Microsoft Azure.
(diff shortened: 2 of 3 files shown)


consideRatio:
LGTM, what is your thoughts about cutting the final z2jh release @minrk? Add beta.4 now to test oauthenticator 0.11.0 also a while?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -293,6 +293,16 @@ def camelCaseify(s):
         if cfg_key is None:
             cfg_key = camelCaseify(trait)
         set_config_if_not_none(c.GitLabOAuthenticator, trait, 'auth.gitlab.' + cfg_key)
+elif auth_type == 'azuread':
+    c.JupyterHub.authenticator_class = 'azuread'
```
consideRatio: Looking at other options, I think this will be wrong, and that it should be for example:

```suggestion
    c.JupyterHub.authenticator_class = 'oauthenticator.azuread.AzureAdOAuthenticator'
```

minrk: Actually, no, this is correct. After JupyterHub 1.0, authenticator classes can be specified via entrypoint alias, 'azuread' or 'github' or 'gitlab'. I was adopting the new style for the new option, but maybe it's better to adopt the new style all at once, so I'll make this one match here and do that update separately.

minrk:
Since oauthenticator got a big refactor, let's do beta.4 in case there are regressions we haven't caught in testing (oauth is [notoriously hard to test rigorously](https://github.com/jupyterhub/oauthenticator/issues/329))

cnf:
Running `--version 0.9.0-beta.3.n022.hbc907f7` atm, and AD auth is working with the config mentioned by @minrk at the top.

I do have a question on how to configure this while keeping secrets out of the values.yaml file. maybe read them from k8s secrets?

consideRatio:
@cnf I typically maintain two helm template value files for my helm chart deployments. One without sensitive stuff (values.yaml), and one with sensitive stuff (secret-values.yaml).

```
# values.yaml
auth:
  type: azuread
  azuread:
    clientId: ...
    tenantId: ...
    usernameClaim: ...
    callbackUrl: "https://<your_jupyterhub_host>/hub/oauth_callback"
```

```yaml
# secret-values.yaml
auth:
    clientSecret: ...
```

I sometimes group everything into secret-values.yaml if they relate and I find it troublesome to separate them for readability reasons.

Further, all files named secret, or being in a folder called secret etc, is automatically encrypted on commit by the CLI called `git-crypt`.

I install using @minrk's built binaries of git-crypt: https://github.com/minrk/git-crypt-bin
git-crypt itself is found here: https://github.com/AGWA/git-crypt

## .gitattributes for use with git-crypt

```
# NOTE: The syntax deviates from the .gitignore syntax, see this note for more info:
#       https://github.com/AGWA/git-crypt#gitattributes-file

# Encrypt / Decrypt anything including "id_rsa" or "secret" in the filename
**id_rsa** filter=git-crypt diff=git-crypt
**secret** filter=git-crypt diff=git-crypt
# Encrypt all files _directly under_ a folder with "secret" in its name.
# WARNING:
# - my-secret-folder/a.txt                  would be encrypted
# - my-secret-folder/another-folder/b.txt   would not be encrypted
**/*secret*/* filter=git-crypt diff=git-crypt
*secret*/* filter=git-crypt diff=git-crypt
```

PR #1570, opened 2020-02-12
Fix reference doc link

clkao:


doc/source/administrator/optimization.md
```diff
@@ -221,7 +221,7 @@ where he analyzed its introduction on mybinder.org.
 priority depending on how your cluster autoscaler is configured. This is known
 to work on GKE, but we don't know how it works on other cloud providers or
 kubernetes. See the [configuration
-reference](reference.html#scheduling-podpriority) for more details.
+reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference.html#scheduling-podpriority) for more details.
```
betatim: can't we fix this by putting a `/` at the start of the URL so `/reference.html`? It would be good if we could avoid having to hardcode the hostname into the links (if that is possible).

clkao: yes -there are quite a few other occurrences, we should probably fix them in one go. this one is just fixing a 404 i found

consideRatio: I opened an issue to represent the need to avoid hardcoding internal links in the docs: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1572.

Thanks for your help @clkao !

PR #1571, opened 2020-02-12
Bump kind to 0.7.0 and include k8s 1.17.0

clkao:
Note that `kind create cluster` no longer creates new kubeconfig and provides the `kubeconfig-path` subcommand. Instead a new context is created into the current kubeconfig.

This PR adds `.jh-dev-kubeconfig` as the default KUBECONFIG for CI. However since the code path is in `dev`, this might be an unexpected change. cc/@consideRatio

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,9 @@ env:
   ## ref: https://github.com/kubernetes-sigs/kind/issues/197
   ##
   - Z2JH_KUBE_VERSION=1.13.12
-  - Z2JH_KUBE_VERSION=1.15.6
-  - Z2JH_KUBE_VERSION=1.16.3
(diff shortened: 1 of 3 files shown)


consideRatio:
Thanks for your work on this @clkao and sorry for the slow follow up.

I've felt that the `dev` script and `kind` have been a bit too complicated to work with, greatly because it didn't become a robust setup. So, instead of trying to fix it, I opted to rework it in #1664 that currently is containing those changes.

I'm sorry for the time you was led to invest in this infrastructure that I introduced and now pushing to have us abandon. I hope you understand and it feels okay that I close this PR. Thanks a lot for your work @clkao! :heart:

Issue #1561, opened 2020-01-30
Get ldapauthenticator released and bumped here


consideRatio:
https://github.com/jupyterhub/ldapauthenticator latest release is very outdated and I hoped to get it bumped for 0.9.0 as it apparently contains quite a bit of fixes. It requires additional work from maintainers though, and we are very low on maintainers on that repo.

manics:
What stopping us from releasing the current master now?

consideRatio:
I don't know fully, i lack understanding about ldap in general, so I cannot even answer that ;)

manics:
There are several open issues and PRs, but I'm not aware of any regressions in `master` that prevent a release.

How about we release the current master? Now that the automated deployment work is in place it should be easy to push out smaller releases after.

@yuvipanda @dhirschfeld

consideRatio:
I spent some time to do what I could:
1. Reviewed / Merged some PRs
2. Fixed merge conflicts etc and prepared https://github.com/jupyterhub/ldapauthenticator/pull/106 to be merged - review help wanted!
3. Wrote a changelog - review help wanted for https://github.com/jupyterhub/ldapauthenticator/pull/158!

@manics has already done work to update our CD to PyPI when pushing tags and made a RELEASE.md file, and I've made sure we have a deployment token available in the PYPI_PASSWORD environment variable. So, after review/merge on stuff above, I think we should make the 1.3.0 release of ldapauthenticator.

manics:
Since there several changes and limited knowledge I think this should be tested as part of `beta4` alongside https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1563#issuecomment-581012274

PR #1576, opened 2020-02-14
jupyterhub-ldapauthenticator==1.3

manics:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1561

Significant changes (taken from https://github.com/jupyterhub/ldapauthenticator/blob/1.3.0/CHANGELOG.md):

-    Avoid binding the connection twice
-    Gracefully handle username lookups with list return values
-    Empty DN templates are now ignored, search_filter and allowed_groups are no longer mutually exclusive*.
-    Allow authentication with empty bind_dn_template when using lookup_dn
-    Ignore username returned by resolve_username - use_lookup_dn_username configuration option added
-    Lookup additional LDAP user info - user_info_attributes is now saved in auth_state for a valid user.


consideRatio:
Thanks @manics for your work! :heart:

PR #1583, opened 2020-02-24
Add nativeauthenticator to hub image

consideRatio:
Using this with z2jh configuration in config.yaml will require it to be
treated as a custom authenticator at the moment as we have not written
any specific way to configure it.


consideRatio:
@GeorgianaElena you should make a 0.1.0 release I think :D

It has passed its initial development, and now there could be use of having some resolution to indicate what is a bigger and a smaller change when you bump it.

Issue #1339, opened 2019-07-26
helm init with "--history-max" setting


bitnik:
Helm documentation recommends initializing helm and tiller with `--history-max` setting: https://v2.helm.sh/docs/using_helm/#initialize-helm-and-install-tiller

> TIP: Setting --history-max on helm init is recommended as configmaps and other objects in helm history can grow large in number if not purged by max limit. Without a max history set the history is kept indefinitely, leaving a large number of records for helm and tiller to maintain.

What do you think, should we update the z2jh documentaion?

consideRatio:
100% Kenan! Let's follow their suggestion by going for some default that people can change if they know what they are doing.

PR #1587, opened 2020-02-27
init helm and tiller with history-max settings

bitnik:
Hi, I have just seen that mybinder also had problem with helm history and remembered the issue related to that I opened. So I just want to make the first step and add the  `--history-max 100` setting in the documentation of helm setup. Please feel free to add more, I am not comfortable with my wording and it usually takes more time for me than other people :)

closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1339

### DIFF
--- a/doc/source/setup-jupyterhub/setup-helm.rst
+++ b/doc/source/setup-jupyterhub/setup-helm.rst
@@ -73,7 +73,7 @@ cluster:
 
    .. code-block:: bash
 
-      helm init --service-account tiller --wait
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me. We could add "We limit the history to 100 previous installs as very long histories slow down `helm` commands a lot." to the paragraph after the command to explain why the flag is being set.

bitnik:
thanks! I added it.

yuvipanda:
woop woop, thank you!

PR #1588, opened 2020-02-27
fix: Bump to base-notebook with JH 1.1.0 etc

bitnik:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1549 bumps jupyter/base-notebook:d4e60350af15 which has jhub 1.0.0 installed (https://hub.docker.com/r/jupyter/base-notebook/tags/?page=2), but because in Dockerfile `pip install --upgrade jupyterhub` is executed, at the end image has jhub 1.1.0.

Since https://github.com/jupyter/docker-stacks/issues/992 is fixed, we can use the latest jupyter/base-notebook and dont have to manually upgrade jhub anymore.

### DIFF
--- a/images/singleuser-sample/Dockerfile
+++ b/images/singleuser-sample/Dockerfile
@@ -6,6 +6,10 @@ FROM jupyter/base-notebook:a07573d685a4
 # The jupyter/docker-stacks images contains jupyterhub, jupyterlab and the
 # jupyterlab-hub extension already.
 
+## NOTE: This is a default and be overridden by chartpress using the
+##       chartpress.yaml configuration
+ARG JUPYTERHUB_VERSION=1.1.*
+
 # Example install of git and nbgitpuller.
 # NOTE: git is already available in the jupyter/minimal-notebook image.
 USER root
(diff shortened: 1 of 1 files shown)


consideRatio:
We don't have to manually do it now, but will need to do that whenever there is a minor release or anything on JupyterHub before it has been merged in docker-stacks. I think I prefer to retain the update step to ensure we are always up to date with JupyterHub like we are quick to be in the hub image.

+1 on bumping the base-notebook tag, but could we retain the manual upgrade step of jupyterhub? I find it convenient with a very little drawback of potentially increasing the image size, I'm not sure.

bitnik:
Hmm I understand your point. Then maybe we should upgrade it to `JUPYTERHUB_VERSION`
as we do in [hub's dockerfile](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/hub/Dockerfile). What do you think?

We already have `JUPYTERHUB_VERSION` defined in chartpress.yaml:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/25a303bedb5bf503cdc1f374d4ba494a28f6682f/chartpress.yaml#L20-L23

consideRatio:
Yepp that sounds good to me @bitnik!

Issue #1601, opened 2020-03-17
Broken docs cross-links


manics:
I'm pretty sure these Sphinx cross-references used to work before, was I just imagining it?

- https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/setup-helm.html
  - `preliminary instructions are available for testing.` link to `<setup-helm3>` is broken
  - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/7b4a873b6795ffb5d986ca7f713e4ee635c4600f/doc/source/setup-jupyterhub/setup-helm.rst#L29

- https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/setup-jupyterhub.html
  - `Now that we have a Kubernetes cluster and Helm setup` links to `<create-k8s-cluster>` and `<setup-helm>` are broken
  - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/7b4a873b6795ffb5d986ca7f713e4ee635c4600f/doc/source/setup-jupyterhub/setup-jupyterhub.rst#L6-L7

betatim:
Are these cross-references internal to the docs or to other sphinx documentation?

Maybe they are broken because the pydata-docs-theme doesn't support them? @choldgraf

choldgraf:
I don't think so, more likely the pages just moved. the `doc` role needs a relative path to an actual file, and I don't see a `create-k8s-cluster` file in the `setup-jupyterhub` folder, so I think this will be fixed by hunting down the proper pages and updating the links.

As an aside: another way we could catch this is to make our docs build use the sphinx "strict" mode, which will fail on any warnings. Is that worth doing?

manics:
@choldgraf I made warnings into errors in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1610

This picked up some other warnings, but not the broken references in this issue.

PR #1610, opened 2020-03-28
[MRG] Sphinx: warnings are errors

manics:
As suggested in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1601#issuecomment-602637363 changes Sphinx warnings to errors
1. First commit adds `-W` to convert warnings to errors, circle-ci fails.
2. Second commit fixes the errors detected in (1), circle-ci passes. Note the broken cross references from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1601 **were not detected**
3. Third commit fixes the broken internal links

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1601

Updated:
- Bump sphinx so `--keep-going` flag can be used to show all doc errors

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
           name: Build docs to store
           command: |
             cd doc
-            make html
+            make html SPHINXOPTS='-W'
 
       - store_artifacts:
           path: doc/build/html/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
           name: Build docs to store
           command: |
             cd doc
-            make html SPHINXOPTS='-W'
+            make html SPHINXOPTS='-W --keep-going'
 
       - store_artifacts:
           path: doc/build/html/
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           name: Build docs to store
           command: |
             cd doc
-            make html
+            make html SPHINXOPTS='-W'
```
manics: Currently we require ['sphinx >=1.7'](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/4c5dc3dee1d8ea2fcace39e6967f0c39510a41ad/doc/doc-requirements.txt#L1). If we upgrade to sphinx 1.8 we could add the ` --keep-going` flag to show all errors instead of just the first one.

https://www.sphinx-doc.org/en/1.8/man/sphinx-build.html

choldgraf: +1 to upgrade

choldgraf:
This looks great to me - thanks for the fixes, this will make it easier to avoid the docs going out-of-date (at least in terms of links etc)

PR #1612, opened 2020-03-31
Ensure a default value for hub.existingSecret

sgloutnikov:
Installing the chart without providing hub.existingSecret results in an error. Provide `false` by default. 

Helm 2:
`Error: render error in "jupyterhub/templates/proxy/deployment.yaml": template: jupyterhub/templates/proxy/deployment.yaml:28:32: executing "jupyterhub/templates/proxy/deployment.yaml" at <include (print $.Tem...>: error calling include: template: jupyterhub/templates/hub/secret.yaml:1:18: executing "jupyterhub/templates/hub/secret.yaml" at <.Values.hub.existing...>: can't evaluate field existingSecret in type interface {}`

Helm 3:
`Error: template: jupyterhub/templates/proxy/deployment.yaml:28:32: executing "jupyterhub/templates/proxy/deployment.yaml" at <include (print $.Template.BasePath "/hub/secret.yaml") .>: error calling include: template: jupyterhub/templates/hub/secret.yaml:1:18: executing "jupyterhub/templates/hub/secret.yaml" at <.Values.hub.existingSecret>: nil pointer evaluating interface {}.existingSecret`

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -87,7 +87,7 @@ hub:
     enabled: true
     initialDelaySeconds: 0
     periodSeconds: 10
-  # existingSecret: existing-secret
+  existingSecret: false
 
 rbac:
   enabled: true


consideRatio:
Oh that was new, what version of helm2/3 was this?

I think it may be that it sais hub is actually null, so perhaps an indentation error in your yaml? (null.existingSecret -> error)

sgloutnikov:
Ah, I think you are right, just caught the problem on my end. I had `hub:` definition was followed immediately by a commented out block and then more definitions bellow which I guess tripped up yaml. If I reorder the definitions so that a commented block does not directly follow `hub:` all is fine. 

Closing the PR, Cheers!

PR #1623, opened 2020-04-14
Do not hardcode IPv4 localhost address for IPv6 compatibility

stv0g:
I am currently deploying Z2JH on an IPv6-only Kubernetes cluster.

There are a bunch of changes lined up for being upstreamed.
This is the first an easiest one.


consideRatio:
We don't have a test to cover this I think, but it makes sense to me!

Thank you for helping this project become more future proof for being an active community member @stv0g :heart: :tada:!

PR #1628, opened 2020-04-16
switching to myst markdown in docs

choldgraf:
Over the last few months, we have been launching a few Sphinx projects that support the [new jupyterbook version](beta.jupyterbook.org). One that I am really excited about is a first-class markdown parser for Sphinx. It is a flavor of markdown called **MyST, for "Markedly Structured Text"**.

The goal of MyST is to do **everything you can do in rST, but in pure markdown**. It has its own syntax pieces for roles, directives, etc, which means you shouldn't ever have to write rST if you're writing in MyST markdown.

[check this page for the syntax pieces supported with MyST](https://myst-parser.readthedocs.io/en/latest/using/syntax.html).

This PR replaces our `recommonmark` extension with `myst_parser`, and makes the modifications needed to the markdown files to use myst syntax. Basically, this means replacing all of the `eval_rst` blocks with their MyST counterparts. What do folks think?

Here's an example of a page now built with MyST, it shouldn't look any different from the current docs:

https://930-87849371-gh.circle-artifacts.com/0/html/customizing/user-management.html

### DIFF
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -169,13 +169,11 @@ In your `hub.extraConfig`,
 You need to have a `import z2jh` at the top of your `extraConfig` for
 `z2jh.get_config()` to work.
 
-```eval_rst
-.. versionchanged:: 0.8
-
-  `hub.extraConfigMap` used to be required for specifying additional values
-  to pass, which was more restrictive.
-  `hub.extraConfigMap` is deprecated in favor of the new
-  top-level `custom` field, which allows fully arbitrary yaml.
+```{versionchanged} 0.8
+`hub.extraConfigMap` used to be required for specifying additional values
+to pass, which was more restrictive.
+`hub.extraConfigMap` is deprecated in favor of the new
+top-level `custom` field, which allows fully arbitrary yaml.
 ```
 
 ### `hub.extraEnv`
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -3,4 +3,4 @@ pyyaml
 myst_parser
 sphinx-autobuild
 sphinx-copybutton
-sphinx
+sphinx<3
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -1,6 +1,6 @@
 pydata_sphinx_theme
 pyyaml
-recommonmark
+myst_parser
 sphinx-autobuild
 sphinx-copybutton
 sphinx
--- a/doc/source/administrator/optimization.md
+++ b/doc/source/administrator/optimization.md
@@ -1,7 +1,4 @@
-```eval_rst
-.. _optimization:
-```
-
+(optimization)=
 # Optimizations
 
 This page contains information and guidelines for improving the reliability,
@@ -74,7 +71,7 @@ situations:
     referring to how a [Helm
     hook](https://docs.helm.sh/developing_charts/#hooks) is used to accomplish
     this, a more informative name would have been *pre-upgrade-image-puller*.
-    
+
     **NOTE**: With this enabled your `helm upgrade` will take a long time if you
     introduce a new image as it will wait for the pulling to complete. We
     recommend that you add `--timeout 600` or similar to your `helm upgrade`
@@ -165,10 +162,7 @@ prePuller:
       tag: 2343e33dec46
 ```
 
-```eval_rst
-.. _efficient-cluster-autoscaling:
-```
-
+(efficient-cluster-autoscaling)=
 ## Efficient Cluster Autoscaling
 
 A [*Cluster
@@ -271,7 +265,7 @@ following:
       kubernetes labels, but this label must be a kubernetes label.
 
     - The taint: `hub.jupyter.org/dedicated=user:NoSchedule`
-    
+
       **NOTE**: You may need to replace `/` with `_` due cloud provider
       limitations. Both taints are tolerated by the user pods.
 
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -27,7 +27,7 @@ import yaml
 # ref: http://www.sphinx-doc.org/en/latest/extdev/tutorial.html#the-setup-function
 
 def setup(app):
-    app.add_stylesheet('custom.css')
+    app.add_css_file('custom.css')
 
 
 # -- Project information -----------------------------------------------------
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -20,7 +20,6 @@
 
 from datetime import date
 
-from recommonmark.transform import AutoStructify
 import yaml
 
 
@@ -28,13 +27,7 @@ import yaml
 # ref: http://www.sphinx-doc.org/en/latest/extdev/tutorial.html#the-setup-function
 
 def setup(app):
-    app.add_config_value(
-        'recommonmark_config',
-        { 'enable_eval_rst': True, },
-        True,
-    )
     app.add_stylesheet('custom.css')
-    app.add_transform(AutoStructify)
 
 
 # -- Project information -----------------------------------------------------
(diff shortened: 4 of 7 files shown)


doc/source/administrator/advanced.md
```diff
@@ -169,13 +169,11 @@ In your `hub.extraConfig`,
...
 `z2jh.get_config()` to work.
 
-```eval_rst
-.. versionchanged:: 0.8
-
-  `hub.extraConfigMap` used to be required for specifying additional values
-  to pass, which was more restrictive.
-  `hub.extraConfigMap` is deprecated in favor of the new
-  top-level `custom` field, which allows fully arbitrary yaml.
+```{versionchanged} 0.8
```
choldgraf, line 172: this is an example of "directive" syntax in MyST, the structure is:

````
```{directivename} <directive-args>
<directive content>
```
````

**Comment by choldgraf** [comment]
The documentation build is failing due to a few warnings that are all from missing glossary terms. I don't see how that'd be related to any of these changes, since all of the pages in the warning are `rst` files which shouldn't be touched with myst_parser...is this a known problem right now?

**Comment by manics** [comment]
@rkdarst You've been getting quite deep into markdown/sphinx in https://github.com/jupyterhub/jupyterhub/pull/3021 so you might be interested in this.

**Comment by manics** [comment]
@choldgraf It could be due to the new major release of Sphinx? See also https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1628

**Comment by choldgraf** [comment]
good call - I pinned the sphinx version to 2.x and it works fine now. What do folks think about keeping Sphinx at 2 until, say, 3.1 comes out with some assumed bugfixes?

**Comment by manics** [comment]
Given the other options are broken builds, or removing `-W`, I think pinning an older Sphinx is best.

Do you know whether the warnings are genuine and have arisen due to stricter checks, or if it's due to a Sphinx bug?

**Comment by rkdarst** [comment]
In the case of jupyterhub docs, the problem was the source_parsers was
removed, and recommonmark wasn't being used as an extension the way
that worked anymore.  Once that was fixed + one other thing, it
worked.  But that doesn't seem to be the problem here.

I know a fair amount about rst and sphinx, but don't have any deep
knowledge of markdown, myst, or recommonmark.  So, I don't have any
great ideas here, and other stuff is calling me so I probably won't
have any more time tonight to look.  I can try to take a look later if
it's still needed, but don't know if I can do much that you all can't!

Thanks for letting me know, though!

**Comment by choldgraf** [comment]
My 2 cents is that, in the long run, it'll be easier for us to maintain pure markdown files instead of hybrid markdown-with-rst files (which is what recommonmark does), which is one reason to make this switch given that we're already using the hybrid approach

**Comment by betatim** [comment]
I'll be the grumpy one: all switches in format of the docs takes some amount of effort to work out kinks. All formats that aren't rST come with additional maintenance complexity. Not because of the format or their implementation but because they aren't the default of sphinx.

Which format is more intuitive/less complex/tedious/etc depends a lot on the persons history I think. For me R markdown is completely impenetrable, LaTeX is easy, Markdown is ok but can't do half the stuff I want and rST is a bit tedious when I haven't used it for a while. I am sure for others this list looks totally different. Which is why I think we will never be able to find a answer everyone is happy with.

Which makes me think reducing the complexity of the stack is what we should aim for. The goal would be that we have to spend less time working on the infrastructure and can spend more of our limited time on the content. In the end the reader doesn't care how it is they got excellent docs to read. They just care that they are excellent :)

What do I mean with complexity? recommonmark is higher complexity because even though it is fairly well established and widely used when sphinx released a new version someone had to go and figure out that the config option name had changed. Where as rST just kept working because it is the default choice. This means rST is lower maintenance complexity than recommonmark.

In general the build systems/CI setups in JupyterHub land are on the complex side. Some of them have to be because we do far out there stuff. However where we could choose lower complexity we sometimes don't.

The summary of this opinion: leaving the mainstream comes with additional costs. Do we have spare budget to pay our bills?

**Comment by choldgraf** [comment]
I agree with all of your points (though I think that MyST markdown has less long-term complexity than rST). This PR basically takes a baby-step (converting our hybrid markdown/rst to just markdown) rather than a big-step (which I guess would mean converting all of our markdown docs to rST docs).

Or put another way, I'd argue that:

````
```{note}
My note
```
````
is less-complex and easier to onboard and maintain than:

````
```eval-rst
.. note::
   
   my note
```
````

And that

````
```
(mylabel)=
# My header
```
````

Is easier than

````
```eval-rst
.. _mylabel:
```
My header
=========
````



Not to mention the fact that (at least in my experience) you often run into bugs with recommonmark.

betatim:
The complexity I was thinking of isn't about the complexity of the format of the docs but the complexity of the infrastructure of building the docs. This means a sphinx setup with more plugins/custom processors/themes is more complex than one that only uses builtin things. If this was the only relevant dimension it would be a no brainer to go with rST (less complexity, less time maintaining the build system).

However complexity of writing content is also important and a separate dimension. So we need to find a trade-off between these two dimensions. I wanted to highlight that format complexity isn't the only thing we should consider.

My thinking about this was triggered by the feeling that over the last month or so it feels like we've mostly spent time maintaining our infrastructure/dealing with side-effects of not maintaining it. Where as the fun part really is adding new features/bug fixes and making our products better. This made me think "can we reduce the complexity of our infrastructure?" Some of this we are already doing (issue templates, standardised release procedures, etc).

---

I love our new theme. Though I wonder if I could objectively justify the amount of time we have spent on it vs using that time to improve the content of the docs + using the default RTD theme.

choldgraf:
Ah - on all of those points I agree. I don't think I'd recommend migrating pre-existing documentation from rST to something like MyST, at least not right now, for the reasons that you specify.

manics:
My preference is for whatever gets us closer to standardising on a single format.

If this is merged should new files, or files that are heavily rewritten, be written in MyST instead of RST?

consideRatio:
> The documentation build is failing due to a few warnings that are all from missing glossary terms. I don't see how that'd be related to any of these changes, since all of the pages in the warning are rst files which shouldn't be touched with myst_parser...is this a known problem right now?

@choldgraf I've fixed this in #1634 and ensured that warnings are not crashing circleci any more, which is fine, because they are doing that in travis but in a job that is tolerated to fail - so we can still spot such warnings clearly.

> Which format is more intuitive/less complex/tedious/etc depends a lot on the persons history I think. For me R markdown is completely impenetrable, LaTeX is easy, Markdown is ok but can't do half the stuff I want and rST is a bit tedious when I haven't used it for a while. I am sure for others this list looks totally different. Which is why I think we will never be able to find a answer everyone is happy with.

@betatim I think MyST may be the perfect choice for us. I think markdown is more known than rST because we are a GitHub organization that have relied on Markdown in GitHub rather than rST, and MyST allows us to extend this markdown slightly to get the missed functionality.

> This PR basically takes a baby-step (converting our hybrid markdown/rst to just markdown) rather than a big-step (which I guess would mean converting all of our markdown docs to rST docs).

@choldgraf I think this is an excellent demonstration of MyST, and it may be a good place to test if it feels mature enough to swap away entirely from rST to MyST. I support using Z2JH as a training ground of MyST and a pilot in the JupyterHub ecosystem in general.

@manics wrote:
> If this is merged should new files, or files that are heavily rewritten, be written in MyST instead of RST?

@choldgraf wrote:
>  I don't think I'd recommend migrating pre-existing documentation from rST to something like MyST, at least not right now, for the reasons that you specify.

I vote to get this merged and use it as a pilot to gain some experience, and when we have gained that, we have enough insights to decide if we should make a full switch of this repo from rST + recommonmark to MyST, which would be as a pilot extension.

consideRatio:
@manics @rkdarst @betatim @choldgraf I appreciate being able to collaborate with you :heart: :heart: :heart: :heart:

choldgraf:
> If this is merged should new files, or files that are heavily rewritten, be written in MyST instead of RST?

IMO this is a longer-term question that should be hashed out explicitly for each repository. This PR is more of a taste to see what the syntax is like, and to reduce some of the complexity that comes with recommonmark. Basically the way @consideRatio describes it is what I was aiming for:

> ...use it as a pilot to gain some experience, and when we have gained that, we have enough insights to decide if we should make a full switch of this repo from rST + recommonmark to MyST, which would be as a pilot extension.

consideRatio:
Forgot to get this merged! Let's go for it!

PR #1630, opened 2020-04-19
[resource prefixing] feature(proxy & hub): make hub & proxy configurable in chart

### LINKED ISSUE #1616: make proxy name configurable
Hi, 
in my enterprise case, it would be beneficial if proxy pod (its name to be precise) could be configurable, i.e. I could overwrite it at any point of time - without going into changing template scripts. 

Why: well because there is naming conflict with our apache proxy and i am forced to put jupyter into its own namespace and maintaining an exception to the rule. 
(i know it is best practice, but many services also allow to reconfigure these kinds of names)

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/proxy/deployment.yaml#L6

thanks you for considering it

dmpe:
closes #1616


dmpe:
To be continued with 
`
Error: failed to create resource: Service "jupyterhub-proxy-public" is invalid: spec.ports[1].nodePort: Invalid value: 31212: provided port is already allocated

Error: UPGRADE FAILED: failed to create resource: Service "jupyterhub-proxy-public" is invalid: spec.ports[1].nodePort: Invalid value: 31212: provided port is already allocated

`

dmpe:
@betatim @consideRatio 
Hi, so, I have done adjustments for proxy **and hub**. some comments from #1616:

>but instead does it right once and for all where we prefix the release name to all of our resources.

I tried that here in this PR. If you pass `.Values.{hub/proxy}.fullnameOverride` value, you will use `chart.name-fullnameOverride` construct. If you do not pass anything, a default naming remains the same (is equivalent to the current one) - `hub`/`proxy`. 

I admit that I have done only very simplistic tests - so it may need additional verification in VirtualBox. I hope that i haven't miss any variables either. 

>breaking changes

Well, the chart is still in pre-release 1.0 status.? :) I am ok if I need to adjust something for making chart following best practises. :)

> don't want to end up needing to maintain backward compatibility against it

I agree 1000%. I couln't agree more. 

@consideRatio you proxy graphic was very helpful. Thank you for that. I think it should go into documentation here.

dmpe:
Because of recent master changes, i will need to rebase this. Nonetheless, it can still be reviewed. 😄

manics:
I just ran `helm create xxxx` to see what property names it generates:
```
{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "xxxx.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

It looks like `fullnameOverride` also overrides the chart name, whereas `nameOverride` is appended to the chart name. This PR adds support for the latter so I think `nameOverride` should be used for consistency with other charts.

dmpe:
@manics Hi, thanks for the feedback. Adjusted.

dmpe:
@manics would you find some time to review again please ?

dmpe:
> 
> 
> Sorry for the delay. Could you say a bit about why you chose to prefix the hub and proxy resources, but not the others?

Thank you for the feedback @manics ! The reason was that these 2 are (for me) the most important ones - which I need to override. I could do singleuser/scheduling if they are of any value. Maybe image puller as well....

I will then take a look on that this weekend.

manics:
@consideRatio in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1616#issuecomment-616109272 you suggested prefixing all resources. The current changes only cover the hub and proxy but are straightforward, backwards compatible, and simply replace `hub` or `proxy`. In a future breaking release we could make them the default by setting `*.nameOverride` and removing the conditionals. Any preferences on how to handle the remaining resources?

consideRatio:
Thanks for being an active member of this project @dmpe!

As this is a popular topic that has been discussed and considered in many places without clear resolution for a long time, I've decided to close all existing issues and PRs relating to it and refer to a single issue: #1791. In this issue I've done my best to provide a clear and actionable path towards progress.

PR #1631, opened 2020-04-19
DX: more recognizable port numbers

consideRatio:
Whenever it works to use a port number that indicates what it is, I appreciate it. This does that for the dev-config, not relevant for production use.


consideRatio:

PR #1632, opened 2020-04-19
Dep: Bump deploy/autohttps's traefik to v2.2

consideRatio:
I was working a lot against treafik in a PR to come, and I bumped it to know that I was working against the latest version of documentation. It seems to work fine even though we can't test it in our CI setup - yet!


consideRatio:

PR #1633, opened 2020-04-19
Feat: autohttps readinessProbe for quicker shutdowns

consideRatio:
When the autohttps deployment is updating, it keeps existing connections 
open for a while before it shuts down, but it closes itself for new 
connections.

After it has closed itself for new connections, it is still considered 
ready by the service that delegates traffic, so that makes the pod 
continue to receive traffic that will crash.

With this change, it will stop receiving traffic, which is preferred as 
it isn't actually doing its job to receive it as it should any more at 
this stage.


consideRatio:

consideRatio:
I've tested this locally and it works great, reducing the disruptions of service during upgrades from ~30 seconds of flakyness down to just a few seconds.

PR #1638, opened 2020-04-19
[WIP] CI: Testing infra of autohttps with letsencrypt/pebble

consideRatio:
## Goal of PR
To provide infrastructure to _enable testing_ of this Helm chart's _autohttps functionality_.

### What the autohttps functionality is
1. **It acquires HTTPS certificates**. By communicating with an _ACME server_ acting as a Certificate Authority using an _ACME client_, we acquire HTTPS certificates.
   - ACME? It is [a standardized way](https://tools.ietf.org/html/rfc8555) for a domain owner (jupyter.org etc.) to prove its ownership of a domain to a Certificate Authority.
   - Certificate Authority? Let's Encrypt is a CA, and they give out and sign HTTPS certificates for free for those that can prove they own the domain that the certificate is valid for.
   - ACME Server? Let's Encrypt's has a production and staging ACME server. They will accept requests of ACME clients that wants to be _challenged_, for example using a HTTP-01 ACME challenge.
   - HTTP-01 ACME challenge? The ACME client can prove ownership of a domain to the ACME server if they can respond properly to web requests to the domain. Only those who can that are in control of the domain and deserve to be granted a certificate signed by the ACME server acting as a Certificate Authority.
   - ACME Client? We use Traefik's built in ACME client, which is called LEGO, which runs in the kubernetes deployment called autohttps.
2. **It acts as a TLS termination proxy**. By using the acquired certificates, it listens to incoming HTTPS connections and ensures traffic is decrypted/encrypted when going through it to the actual destination which isn't the proxy itself.

## Why letsencrypt/pebble?
We cannot use Let's Encrypt in a CI environment like TravisCI, because they can't reach us. We can reach them and they can respond, but they can't reach us on their own, and that is part of the HTTP-01 ACME challenge we use to acquire a domain certificate.

I've evaluated the following options, but in the end pebble seemed most suitable because it was lightweight and designed to be used in a testing environment. The other options were more heavy weight and would likely break over time due to their own complexity. Pebble is also [used by cert-manager](https://github.com/jetstack/cert-manager/blob/master/devel/addon/pebble/chart/templates/configmap.yaml), which makes it seem like a quite safe bet.
- [letsencrypt/boulder](https://github.com/letsencrypt/boulder)
- [letsencrypt/pebble](https://github.com/letsencrypt/pebble)
- [smallstep/certificates](https://github.com/smallstep/certificates)
- [smallstep/autocert](https://github.com/smallstep/autocert)
- [smallstep/helm-charts](https://github.com/smallstep/helm-charts)

## Pebble need to resolve jovyan.local to proxy-public
It turns out that pebble couldn't resolve everything for us entirely on its own. Pebble can send out HTTP-01 challenges and such, but if we for example want the pretend domain jupyter.local to point to our ACME client at the kubernetes service called proxy-public, we need to make pebble think it does.

I evaluated many options on how to make pebble think a pretend domain such as jovyan.local could be pointed to the ACME client at proxy-public. They typically had their drawbacks.

### k8s: Pod.spec.hostAlias
This is the Kubernetes way of modifying a pod's `/etc/hosts` file. This allow us to map a domain name to an IP, but not another domain name which `proxy-public` is thanks to it being a Kubernetes Service. But this is perhaps something we could live with.

ref: https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

### k8s: Pod.spec.dnsConfig / pebble's -dnsserver flag
This is the Kubernetes way of modifying a pod's `/etc/resolve.conf` file. This allow us to specify a DNS server to use for domain -> ip lookups of certain domains, but this doesn't help us really unless we have a DNS server that we can modify. Another option to this, is to use the pebble's startup flag called `-dnsserver`.

ref: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config

### k8s: Service.spec.externalName
We could use Kubernetes Service resources named `jovyan` with `spec.externalName` to get a registered DNS entry for `jovyan` pointing where we please, either an IP or another domain name. The limitation is that we are required to name the server without punctuation marks (`.`), so while we can have `jovyan` point to `proxy-public` we cannot have `jovyan.local` point there.

ref: https://kubernetes.io/docs/concepts/services-networking/service/#externalname

### Patching the coredns or kube-dns configuration
coredns is the current default DNS server in Kubernetes now, and previously it was kube-dns. We could perhaps configure them to point make our domain names point to the `proxy-public` service with a CNAME entry?

Maybe. But I don't know how to. And, if we manage, it should preferably not require a standalone command to execute following setup etc. If there were a Kubernetes native resource that would influence the DNS server, that would be something.

ref: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#coredns-configmap-option
ref: https://coredns.io/2017/07/23/corefile-explained/

### Hosting our own DNS server
This was functional in 3fff117 but I removed it as using hostAlias is a simpler solution and because I failed to get the DNS server automatically startup with CNAME entries I could configure from configuration files.

I tried using this, which has a DNS server part of it: https://github.com/letsencrypt/pebble/tree/master/cmd/pebble-challtestsrv

My goal with this solution was to avoid hardcoding the ClusterIP of the proxy-public service, but I failed and ended up hardcoding it anyway.

### Chosen solution
`Pod.spec.hostAlias` is the winner, but we need to hardcode the ClusterIP of the proxy-public service endpoint. If the hardcoding becomes an issue we need to avoid, we need to have a lightweight DNS server that pebble can communicate with, and that can startup with certain CNAME records pointing to proxy-public.

## Local development notes

Inspired by dask/dask-gateway, I'm trying out k3d instead of minikube to see how it works and hope to get a speed increase over minikube as well as avoid needing a virtual machine to back minikube's environment.

### Prerequisites
```
# add the following /etc/hosts entries, where the first is for k3d to work well
# 127.0.0.1 registry.local
# 127.0.0.1 jovyan.local
sudo vim /etc/hosts
```

```
k3d create --wait 300 --enable-registry --publish 8080:30080 --publish 8443:30443
export KUBECONFIG="$(k3d get-kubeconfig --name='k3s-default')"
kubectl config set-context --current --namespace=jh-dev
kubectl create namespace jh-dev
chartpress --skip-build
helm upgrade jh-dev jupyterhub/ --namespace jh-dev --install --cleanup-on-fail --values dev-config-https.yaml --values dev-config.yaml
```

Visit https://jovyan.local:8443 and accept that we don't have the trusted certificates.


yuvipanda:
This is awesome! I also love the detailed explanation you have written - something I'll aspire to :)

I agree that hardcoding the cluster IP sucks, and externalName is the easiest way to do this. Here's a possible solution:

1. For each service, Kubernetes creates the DNS entry for `my-svc.my-namespace.svc.cluster-domain.example`. Each of these usually resolve, thanks to [search domains](https://en.wikipedia.org/wiki/Search_domain) and the [ndots](https://linux.die.net/man/5/resolv.conf) (search for ndots in that page) directive. [this blog post](https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html) has a reasonable overview.
2. If we create a namespace called `local` and an externalName service there called `jupyter`, we can then point it to `proxy-public.default` (assuming we are in the default namespace). This will give us a resolvable DNS entry for `jupyter.local` that points to our proxy-public pod.

I *think* this should work? Wanna try it?

Also, can pebble just hit `proxy-public.default` instead of needing to hit `jupyter.local`? Is the `.local` a pebble restriction?

Also, amazing inspirational detailed work! <3

yuvipanda:
Also is it possible to run the pebble server external to the z2jh chart itself? It feels like autohttps is complex enough, and it might be better to not make it more complicated. This way, we can set a specific nodePort on proxy-public, making the testing easier - if nodePort is 8000, `jupyter.local` points to localhost, then `jupyter.local:8000` will point to proxy-public.

sgibson91:
I feel like the OP of this PR is 90% a blog post :rocket: :wink:

consideRatio:
@yuvipanda just to verify we are on the same page - the hardcoding of the clusterIP is to set it explicitly, not to first read it and then set it which would be really bad.

I want to lead with questions.
1. Do we care if we limit ourselves to testing hostnames without dots?
   I think I do, because if binderhub wants to test its setup later, it probably wants to see the hub as a subdomain as we usually configure it.
2. Do we care if we _set_ a hardcoded ClusterIP of the proxy-public service?
   Maybe we do, but in ephemeral k8s clusters like created by minikube, k3d, kind etc, I think we can avoid issues and at worst need some configuration. This could certainly be a reason for a future issue though.

I lean towards accepting that we set a hardcoded ClusterIP of the proxy-public service and reference that hardcoded ClusterIP later, hoping we will always be allowed to do this.

I think it introduce too much complexity to create N # services for M # hosts. I suggest to either settle with this, or add a dedicated DNS server that we can configure through a configmap to have certain entries, and pointing pebble to it.

---

Regarding running pebble external to the chart - yes, that would be possible, but if BinderHub wants to test HTTPS, then it needs to have the testing stuff setup as well. I don't see a good enough reason to not embed some templates which doesn't render by default, unless `testing.enabled: true` is specified. It will add som bytes to the packaged helm chart.

I tried finding an option where I didn't have to learn about pebble or such, for example by using a precreated pebble helm chart, but there were none. There were a helm chart for smallstep/certificate and smallstep/autocert, and I went all the way of replacing my pebble templates to avoid introducing something we maintain ourselves. But, they were not mature enough and made complexity rise drastically. At that time I felt that it was a bad idea to abandon pebble, and I went back to it.

Hmmm...

We could perhaps make a pebble helm chart standalone, that we conditionally add in requirements.yaml. That could make some sense perhaps. BinderHub would be able to do the same.

Yeah! Should we do that?

consideRatio:
## Open questions
- Extract the pebble deployment / configmap / service and create a standalone helm chart?
- Accept the need to pin a ClusterIP of the proxy-public service, or do something to avoid it such as hosting a DNS server as well which we reference?

yuvipanda:
Thanks for the thoughtful response, @consideRatio. I agree that hardcoding clusterIP isn't a problem for me in testing. But very much yeah, having a separate helm chart for pebble would be much more helpful than embedding that in z2jh. Each conditional in a template produces upto 2x more possible outcomes, so I'd like to avoid ones that won't be set in production if possible.

yuvipanda:
And as you said, we probably aren't the only people looking for that!

consideRatio:
https://github.com/jupyterhub/pebble-helm-chart is now created and published as `pebble` in our jupyterhub helm chart repository, I've also been able to use it successfully in #1664 to test HTTPS.

PR #1639, opened 2020-04-20
Fix broken links in the Reference documentation

bibz:
The Reference documentation page is generated from the JupyterHub
schema. Fix the broken links[1] there and enjoy the propagation to
the documentation.

[1]: Turn relative links into absolute links to the stable
  documentation page on ReadTheDocs. This mimicks how the other
  links are set in the schema.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1172,14 +1172,18 @@ properties:
     properties:
       hook:
         description: |
-          See the [*optimization section*](optimization.html#pulling-images-before-users-arrive) for more details.
+          See the [*optimization
+          section*](https://zero-to-jupyterhub.readthedocs.io/en/stable/administrator/optimization.html#pulling-images-before-users-arrive)
+          for more details.
         type: object
         properties:
           enabled:
             type: bool
       continuous:
         description: |
-          See the [*optimization section*](optimization.html#pulling-images-before-users-arrive) for more details.
+          See the [*optimization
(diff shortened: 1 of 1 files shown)


bibz:
While browsing through the online documentation, I noticed [the](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html#prepuller-hook) [links](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html#prepuller-continuous) [pointing](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html#prepuller-extraimages) back to the "optimization section" were broken. Fixed now.

Note: the link check recipe (`make linkcheck`) is currently ignoring(?) the links and hence do not show they are broken:
```
% make linkcheck
...
writing output... [ 80%] reference/reference
...
(line 1188) -local-   optimization.html#pulling-images-before-users-arrive
(line 1195) -local-   optimization.html#pulling-images-before-users-arrive
(line 1205) -local-   optimization.html#the-images-that-will-be-pulled
...
```

bibz:
Additionally, and to support this change, here are the notable changes:
- Stable at 33 broken links ⇒ the PR does fix silenced broken links (see previous comment).
  ```
  make clean html linkcheck | grep -c broken
  ```
- The difference in the HTML files is limited to the intended generated reference.
  ```
  git checkout origin/master && make clean html && mv build build-master
  git checkout fix-reference-links && make clean html
  diff -r build-master/html build/html
  ```

betatim:
Thanks a lot!

PR #1640, opened 2020-04-20
Fix user storage customization variable

bibz:
The size of the PVC request for each user belongs to the `singleuser`
object, not the root configuration.


consideRatio:
@bibz :heart: :tada: thanks !

Issue #1637, opened 2020-04-19
AWS LoadBalancer health checks failing when using release 0.9.0 but not 0.8.2


albertmichaelj:
### Bug description

I have been successfully running a relatively heavily modified instance of JupyterHub-k8 (modifications primarily around using an NFS server for file storage and per-user customization options) for the past year using JupyterHub-k8 from git commit `d25bcd9e38aca2734f90e51c2bc9a0bf9d90ca3e` (I wasn't using the official `0.8.2` release because I needed hub init containers). I have been running this on AWS, and it has worked flawlessly for teaching multiple classes at the university I work at. I give this background to say that I don't think I'm making a stupid mistake that is causing my error, but I might be.

I have been testing out the `0.9.0` release, and there seems to be some incompatibility with AWS LoadBalancer health checks. I am using `eksctl` to set up the cluster using eks (Kubernetes version `1.15`). I then use a minimal example following the exact steps in the JupyterHub guide (starting from https://zero-to-jupyterhub.readthedocs.io/en/latest/setup-jupyterhub/setup-helm.html). I use the following basic config file:

``` yaml
proxy:
  secretToken: "<REDACTED>"
```

All containers start fine, and the logs for all containers seem normal. However, I cannot access the hub page by going to the specified address from the load balancer. When I check the load balancer, it tells me that the connection to the "instance id" that is associated with the workers associated with the hub containers are "OutOfService" due to "Instance has failed at least the UnhealthyThreshold number of health checks consecutively." Oddly, I have gotten to the hub page successfully a few times for about 30 seconds, and then it fails. I think this is because it takes some time for the health checks to reach a threshold where they fail enough to kick the instance, and during that time, I can successfully reach it.

I then tried the exact above configuration (using the same kubernetes cluster) with `0.8.2` and things worked as expected. So, I do not think it is a misconfiguration of the cluster, but for the life of me, I have no idea where the problem would have been introduced. Given that my existing setup works using commit `d25bcd9e38aca2734f90e51c2bc9a0bf9d90ca3e`, it seems likely the problem was introduced after that commit, but that commit was almost a year ago.

I am very happy to help troubleshoot, but I'm not entirely sure where to start.

#### Expected behaviour

The hub should be reachable at the web address that you get from running `kubectl --namespace=jhub get svc`

#### Actual behaviour
I get an error message saying:

```
This page isn’t workinga
[...]

PR #1641, opened 2020-04-20
Default HTTPS on proxy does not function

sstarcher:
By default the proxy has HTTPS enabled, but no certificates exist.  The user must either setup letsencrypt or supply their own certificates.  We should disable this by default as to not expose something that does not work.  

This resolves - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1637

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -165,7 +165,7 @@ proxy:
     enabled: true
     minAvailable: 1
   https:
-    enabled: true
+    enabled: false
     type: letsencrypt
     #type: letsencrypt, manual, offload, secret
     letsencrypt:


albertmichaelj:
As we've discussed in the comments to #1531, this seems to solve the problem raised in #1637, but the documentation in at least one spot seems to also need to be updated. In the documentation [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/administrator/security.md), the documentation I think needs to have 

``` yaml
proxy:
  https:
    enabled: true
```

for every example in the documentation. This is already the case on the "Off-loading SSL to a Load Balancer" section, but not in "Set up automatic HTTPS" or "Set up manual HTTPS". These sections should be updated as well to keep the documentation updated.

consideRatio:
This kind of solution would cause disruptions for any users of the helm chart that has a configuration that is something like this:

```
proxy:
  https:
    hosts:
      mydomain.com
```

It may make sense to require a user to explicitly enable https itself as it doesn't work without additional configuration anyhow, but i don't want to merge this PR without additional guidance for the user.

I'm still exploring this issue and for now isn't ready to lock in a solution yet.

sstarcher:
Sounds good, I'll leave it up to you.

PR #1624, opened 2020-04-14
Bind proxy on IPv4 and IPv6 for dual stack support

stv0g:
The `configurable-http-proxy` relies on the `http` module which in turn relies on the `server` module:

https://nodejs.org/api/net.html#net_server_listen

We dont make use of the `ipv6Only` option. Therefore we can safely listen on `::` for dual stack support.

PR #1643, opened 2020-04-23
Use DNS names instead of IPv4 addresses to be IPv6 friendly

stv0g:
## PR Summary by @consideRatio

- References to IP addresses through the `*_SERVICE_HOST` environment variables were replaced with the DNS names of the services.
- `hub_connect_ip` and `hub_connect_port` were combined into `hub_connect_url` (how to reach the hub from other pods)
- `hub_ip` were replaced with `hub_bind_url` (what ports the JupyterHub process listen to within the container)
- `ip` and and `port` were not combined to `bind_url` but instead removed, as this is configuration influencing a CHP proxy started up by JupyterHub, which we don't do at all.
- A function (camelCaseify) was refactored to the top of the jupyterhub_config.py file.

## Original PR text

This PR lets JupyterHub use an IPv6 compatible configuration and allows us to run z2jh on an IPv6 only Kubernetes cluster.

I am not really convinced by the use of the old "Docker-style" link environment variables.
From my understanding best practice for service discovery in Kubernetes is by using DNS.

Is there any reason why DNS is not being used by z2jh? 
If there are no objections, I will update my PR to rely on DNS..
This would also have avoided the problems we now encounter with IPv6 compatability.

This PR also eliminates warnings about deprecated JupyterHub configuration settings during startup. Most of the `*_port` and `*_ip` settings have been replaced by an `*_url` variant.

This also closes #1624


jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -53,11 +53,11 @@ spec:
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           command:
             - configurable-http-proxy
-            - --ip=0.0.0.0
-            - --api-ip=0.0.0.0
+            - "--ip=::"
+            - "--api-ip=::"
```
manics, line 57: How will this behave on an ipv4-only cluster?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -12,6 +12,22 @@
 
 from z2jh import get_config, set_config_if_not_none
 
+def serviceUrl(svc, ssl=False):
+    port = int(os.environ['{}_SERVICE_PORT'.fornat(svc.upper())])
```
manics, line 16:
```suggestion
    port = int(os.environ['{}_SERVICE_PORT'.format(svc.upper())])
```
Travis is failing due to this error

jupyterhub/templates/proxy/deployment.yaml
stv0g: It has no effects for IPv4-only clusters.

Binding to `::` actually causes to socket to listen for IPv4 and IPv6 unless the `IPV6_V6ONLY` socket option is added to the socket which is normally not the case by default.

See also: http://man7.org/linux/man-pages/man7/ipv6.7.html

stv0g:
Okay, CI stills fails because Tornados httpserver does special things here:

https://github.com/tornadoweb/tornado/blob/master/tornado/netutil.py#L144

stv0g:
Okay, CI is now green :)

kind has support for IPv6. Is there some interest to extend the CI pipeline with a test case on an IPv6 Kubernetes cluster?

manics:
More tests are always good 🙂 

Perhaps you could combine it with the Helm 3 test to avoid adding another matrix job? https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6f7505fe34afc70f6f8fa8e3b3a827ef6125cdf9/.travis.yml#L64
@consideRatio thoughts on testing this, and also on whether to replace the env-vars with K8s DNS?

yuvipanda:
I choose to use env variables to link rather than DNS because I had not installed kube-dns in the hand-build wikimedia cluster I started this codebase on :D So @manics is right about the age...

consideRatio:
My take on testing
- I like the idea of dropping helm2 in testing in favor of helm3
- I'm not happy about the `dev` script + `kind` I introduced, and is hoping to replace them for a more robust, performant, and transparent experience.

manics:
I'm planning to update the Helm 3 docs for Helm 3.2.0 (includes a `--create-namespace` option which removes one step), so instead of changing the deployment matrix in this PR how about if we update an existing job. Then when I update the Helm 3 docs I'll replace a Helm 2 job with a new Helm 3 one?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -21,7 +29,7 @@
 c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
 
 # Connect to a proxy running in a different pod
-c.ConfigurableHTTPProxy.api_url = 'http://{}:{}'.format(os.environ['PROXY_API_SERVICE_HOST'], int(os.environ['PROXY_API_SERVICE_PORT']))
+c.ConfigurableHTTPProxy.api_url = 'http://proxy-api'
```
consideRatio, line 32: It seems like the proxy-api service is exposed at 8001 rather than 80

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -247,8 +247,7 @@ def camelCaseify(s):
 c.KubeSpawner.volume_mounts.extend(get_config('singleuser.storage.extraVolumeMounts', []))
 
 # Gives spawned containers access to the API of the hub
-c.JupyterHub.hub_connect_ip = os.environ['HUB_SERVICE_HOST']
-c.JupyterHub.hub_connect_port = int(os.environ['HUB_SERVICE_PORT'])
+c.JupyterHub.hub_connect_url = 'http://hub'
```
consideRatio, line 250: Seems like it use port 8081 rather than the default of :80

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -74,11 +73,12 @@ def camelCaseify(s):
         cfg_key = camelCaseify(trait)
     set_config_if_not_none(c.JupyterHub, trait, 'hub.' + cfg_key)
 
-c.JupyterHub.ip = os.environ['PROXY_PUBLIC_SERVICE_HOST']
-c.JupyterHub.port = int(os.environ['PROXY_PUBLIC_SERVICE_PORT'])
+c.JupyterHub.bind_url = 'http://proxy-public'
 
 # the hub should listen on all interfaces, so the proxy can access it
-c.JupyterHub.hub_ip = '0.0.0.0'
+# tornados httpserver only listens on a single address family which deviates from Lux standard behaviour
```
consideRatio, line 79: What is Lux?

stv0g:
@yuvipanda I switched the service discovery to DNS now.

@consideRatio Thanks for the review. I addressed your comments in my last commit

@manics IPv6 in Travis proofs to be difficult as Travis does not support IPv6 in the VM based runners: https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

jupyterhub/files/hub/jupyterhub_config.py
stv0g: fixed

jupyterhub/files/hub/jupyterhub_config.py
stv0g: fixed

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -53,11 +53,11 @@ spec:
...
             - --api-port=8001
-            - --default-target=http://$(HUB_SERVICE_HOST):$(HUB_SERVICE_PORT)
-            - --error-target=http://$(HUB_SERVICE_HOST):$(HUB_SERVICE_PORT)/hub/error
+            - --default-target=http://hub:8081
+            - --error-target=http://hub:8081/hub/error
```
consideRatio, line 60:
```suggestion
            - --default-target=http://hub:$(HUB_SERVICE_PORT)
            - --error-target=http://hub:$(HUB_SERVICE_PORT)/hub/error
```

jupyterhub/templates/image-puller/job.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             - /image-awaiter
             - -ca-path=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
             - -auth-token-path=/var/run/secrets/kubernetes.io/serviceaccount/token
-            - -api-server-address=https://$(KUBERNETES_SERVICE_HOST):$(KUBERNETES_SERVICE_PORT)
+            - -api-server-address=https://kubernetes.default.svc
```
consideRatio, line 39:
```suggestion
            - -api-server-address=https://kubernetes.default.svc:$(KUBERNETES_SERVICE_PORT)
```

As a precaution, lets not assume this service doesn't always use port 443.

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -247,8 +249,7 @@ def camelCaseify(s):
 c.KubeSpawner.volume_mounts.extend(get_config('singleuser.storage.extraVolumeMounts', []))
 
 # Gives spawned containers access to the API of the hub
-c.JupyterHub.hub_connect_ip = os.environ['HUB_SERVICE_HOST']
-c.JupyterHub.hub_connect_port = int(os.environ['HUB_SERVICE_PORT'])
+c.JupyterHub.hub_connect_url = 'http://hub:8081'
```
consideRatio, line 252:
```suggestion
c.JupyterHub.hub_connect_url = f'http://hub:{os.environ['HUB_SERVICE_PORT']}'
```

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -74,11 +73,14 @@ def camelCaseify(s):
...
-c.JupyterHub.port = int(os.environ['PROXY_PUBLIC_SERVICE_PORT'])
+c.JupyterHub.bind_url = 'http://proxy-public'
 
 # the hub should listen on all interfaces, so the proxy can access it
-c.JupyterHub.hub_ip = '0.0.0.0'
+# tornados httpserver only listens on a single address family
+# this deviates from Linux's standard behaviour. Right now JupyterHub
+# does not support dual-stack setups.
+ipv6 = ':' in os.environ['KUBERNETES_SERVICE_HOST']
+c.JupyterHub.hub_bind_url = 'http://[::]:8081' if ipv6 else 'http://0.0.0.0:8081'
```
consideRatio, line 83:
```suggestion
ipv6_enabled_cluster = ':' in os.environ['KUBERNETES_SERVICE_HOST']
hub_container_port = 8081
if ipv6_enabled_cluster:
    c.JupyterHub.hub_bind_url = f'http://[::]:{hub_container_port}'
else:
    c.JupyterHub.hub_bind_url = f'http://0.0.0.0:{hub_container_port}'
```

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -74,11 +73,14 @@ def camelCaseify(s):
         cfg_key = camelCaseify(trait)
     set_config_if_not_none(c.JupyterHub, trait, 'hub.' + cfg_key)
 
-c.JupyterHub.ip = os.environ['PROXY_PUBLIC_SERVICE_HOST']
-c.JupyterHub.port = int(os.environ['PROXY_PUBLIC_SERVICE_PORT'])
+c.JupyterHub.bind_url = 'http://proxy-public'
```
consideRatio, line 76:
```suggestion
c.JupyterHub.bind_url = f'http://proxy-public:{os.environ['PROXY_PUBLIC_SERVICE_PORT']}'
```

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -21,7 +29,7 @@
 c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
 
 # Connect to a proxy running in a different pod
-c.ConfigurableHTTPProxy.api_url = 'http://{}:{}'.format(os.environ['PROXY_API_SERVICE_HOST'], int(os.environ['PROXY_API_SERVICE_PORT']))
+c.ConfigurableHTTPProxy.api_url = 'http://proxy-api:8001'
```
consideRatio, line 32:
```suggestion
c.ConfigurableHTTPProxy.api_url = f'http://proxy-api:{os.environ['PROXY_API_SERVICE_PORT']}'
```

consideRatio (commented):
I added some code suggestions to stop assuming ports used by various k8s services as hardcoded.

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -74,11 +73,26 @@ def camelCaseify(s):
...
+# container should listen to.
+hub_container_port = 8081
+# Tornados httpserver can only listens to either IPv4 or IPv6 traffic, which
+# deviates from Linux's standard behavior. Due to this, JupyterHub currently
+# does not support dual-stacks setup where either IPv4 or IPv6 can be used.
+ipv6_enabled_cluster = ':' in os.environ['KUBERNETES_SERVICE_HOST']
+if ipv6_enabled_cluster
+    c.JupyterHub.hub_bind_url = f'http://[::]:{hub_container_port}'
+else:
+    c.JupyterHub.hub_bind_url = f'http://0.0.0.0:{hub_container_port}'
```
consideRatio, line 91: @minrk @stv0g I was considering this limitation about only being able to listen either on IPv4 or IPv6? I could not find any clear confirmation on this.

Perhaps though JupyterHub ends up limiting us even though Tornado's httpserver does. What JupyterHub ends up doing to actively listen is this call [here](https://github.com/jupyterhub/jupyterhub/blob/6ecda96dd636dfdd0cf34fcf6c1fc44942779c01/jupyterhub/app.py#L2645), but before this call, some processing of the configuration is made.

I wonder if it would work to set this to f`http://:{hub_container_port}` instead, as my understanding is that tornado would support both IPv4 or IPv6 if a blank string or None value was used for the IP.

Hmmm... But looking more in depth, i see that there seem to be three ways to start Tornado, and we seem to follow the "single process" pattern. See: https://www.tornadoweb.org/en/stable/httpserver.html#http-server.

I find that the `listen` function we use is documented [here, in the TCPServer class which HTTPServer is derived from](https://www.tornadoweb.org/en/stable/tcpserver.html#tornado.tcpserver.TCPServer.listen).

The interesting part is defined in the [`bind` function documentation](https://www.tornadoweb.org/en/stable/tcpserver.html#tornado.tcpserver.TCPServer.bind).

> Address may be an empty string or None to listen on all available interfaces. Family may be set to either socket.AF_INET or socket.AF_INET6 to restrict to IPv4 or IPv6 addresses, otherwise both will be used if available.

It sounds to me, that as long as we pass an empty string when calling `listen`, we should be able to listen to both IPv4 and IPv6 - could this be correct?

consideRatio: I tried to use `http://:8081` and that worked well for IPv4, but based on the Tornado documentation I think it will also work for IPv6 as well.

PR #1645, opened 2020-04-24
Refactor: reference ports by name instead of repeating the number

consideRatio:
- Port definitions `protocol: TCP` is no longer explicitly set
- Ports are now referenced by name from services and network policy ingress
- The deploy/hub and deploy/proxy pods got some ports renamed:
  - hub -> http
  - proxy-public -> http
  - proxy-https -> https
- I removed legacy ports allowed by the deploy/proxy pod's network policy for kube-lego and nginx /health. They could not be accessed anyhow as they were not exposed by the pod.

This should not influence anything.

### DIFF
--- a/jupyterhub/templates/proxy/netpol.yaml
+++ b/jupyterhub/templates/proxy/netpol.yaml
@@ -18,30 +18,22 @@ spec:
     - Egress
   ingress:
     - ports:
-      - port: 80
-      - port: 443
-      {{- if not $autoHTTPS }}
-      - port: 8000
-      {{- end }}
+      - port: http
       {{- if or $manualHTTPS $manualHTTPSwithsecret}}
-      - port: 8443
+      - port: https
       {{- end }}
-      # kube-lego /healthz
-      - port: 8080
-      # nginx /healthz
-      - port: 10254
     - from:
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-proxy-http: "true"
       ports:
-        - port: 8000
+        - port: http
     - from:
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-proxy-api: "true"
       ports:
-        - port: 8001
+        - port: api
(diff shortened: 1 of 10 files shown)


jupyterhub/templates/proxy/netpol.yaml
```diff
@@ -18,38 +18,22 @@ spec:
...
-      - protocol: TCP
-        port: 8443
+      - port: https
       {{- end }}
-      # kube-lego /healthz
-      - protocol: TCP
-        port: 8080
-      # nginx /healthz
-      - protocol: TCP
-        port: 10254
```
consideRatio, line 38: This is to allow ingress to the deploy/proxy's pod, but that pod doesn't expose these ports anyhow. This is not relevant any more.

jupyterhub/templates/proxy/service.yaml
```diff
@@ -44,26 +43,18 @@ spec:
...
     {{- end }}
     - name: http
       port: 80
-      protocol: TCP
-      {{- if $autoHTTPS }}
-      targetPort: 80
-      {{- else }}
-      targetPort: 8000
-      {{- end }}
+      targetPort: http
```
consideRatio, line 57: No matter if this goes to the deploy/autohttps or deploy/proxy, both have a port named http even if it is exposed on different port numbers.

jupyterhub/templates/proxy/service.yaml
```diff
@@ -44,26 +43,18 @@ spec:
...
-      protocol: TCP
-      {{- if or $manualHTTPS $manualHTTPSwithsecret }}
-      targetPort: 8443
-      {{- else if $offloadHTTPS }}
-      targetPort: 8000
+      {{- if $offloadHTTPS }}
+      targetPort: http
       {{- else }}
-      targetPort: 443
+      targetPort: https
```
consideRatio, line 49: No matter if this goes to the deploy/autohttps or deploy/proxy, it should always go to the `https` port unless the "offload".

PR #1654, opened 2020-05-05
Update create service principle command.

superyaniv:
https://docs.microsoft.com/en-us/cli/azure/ad?view=azure-cli-latest

should be `az ad sp` create-for-rbac


betatim:
Thanks for helping out make the documentation better!

PR #1659, opened 2020-05-10
autohttps: instant secret-sync shutdown

consideRatio:
During an upgrade, the autohttps pod can restart in a rolling update fashion, which makes us have two autothttps pods, and two secret-sync pods. The new pod may be getting certificates for a new hostname, but suddenly the old pods secret-sync container may override the content. It may not be a big deal, but it will at least slowdown the shutdown of the autohttps pod that the secret-sync container is slow to shutdown and require the full 30 seconds of grace period for no reason.

This fixes that issue by listening to the SIGTERM signal that comes before the SIGKILL as part of kubectl delete or docker stop.

### DIFF
--- a/images/secret-sync/Dockerfile
+++ b/images/secret-sync/Dockerfile
@@ -1,5 +1,10 @@
 FROM python:3.7-alpine
 
+ADD https://github.com/krallin/tini/releases/download/v0.19.0/tini-static /tini
+RUN chmod +x /tini
+
 RUN pip install --no-cache kubernetes
 
 COPY secret-sync.py /usr/local/bin/secret-sync.py
+
+ENTRYPOINT [ "/tini", "--", "/usr/local/bin/secret-sync.py" ]
--- a/images/secret-sync/secret-sync.py
+++ b/images/secret-sync/secret-sync.py
@@ -29,15 +29,16 @@ and update the secret object as needed. However, for now we
 just operate in a 30s loop. This is good enough, since
 traefik can always re-generate certs if needed.
 """
-import sys
+import argparse
+import base64
+import io
+import logging
 import os
 import subprocess
-import argparse
-import time
+import sys
 import tarfile
-import io
-import base64
-import logging
+import time
+
 from kubernetes import client, config
 
 def update_secret(namespace, secret_name, labels, key, value):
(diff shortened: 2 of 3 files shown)


manics:
Do you know why SIGTERM doesn't already stop it?
https://stackoverflow.com/a/41840796 suggests it should terminate by default, though adding the handler allows the shutdown to be graceful

consideRatio:
@manics I don't know about pythons default behavior of SIGTERM in general, but I know that the script in the k8s pod keep running during termination until SIGKILL. If a python process actually reacts correctly to SIGTERM, then perhaps the python process never got the signal.

Okay I've verified that python does indeed react to SIGTERM so there is no need for additional python code. The question is then why it isn't passed to Python and how we ensure it is passed.

I'm debugging this now. I can reproduce it, but I cant fix it at the moment. Here are some relevant things to consider: https://hynek.me/articles/docker-signals/

---

I could relate to us using `chmod +x secret-sync.py` which has a `#!/usr/bin/env python3` in it, and starting it as an executable with `#!/usr/bin/env python3`.

```python
#!/usr/bin/env python3

import time
time.sleep(3600)
```

consideRatio:
Hmmm... well I don't get it actually. It works if I add signal handling, but not without it. So it seems like it actually receives the signal and ignores it.

consideRatio:
I debugged this at length to get it right. Apparently we need to either use tini or to use pythons signal handling, both of those options work. But, one of them is required.

```Dockerfile
# Dockerfile
FROM python:3.7-alpine

ADD https://github.com/krallin/tini/releases/download/v0.19.0/tini-static /tini
RUN chmod +x /tini
ENTRYPOINT [ "/tini", "--" ]

RUN pip install --no-cache kubernetes

COPY secret-sync.py /usr/local/bin/secret-sync.py

# DEBUGGING
COPY sleep.py /usr/local/bin/secret-sync.py

CMD ["python", "/usr/local/bin/secret-sync.py"]
```

```python
# sleep.py
import time
time.sleep(3600)
```

```
docker build -t sigterm-me . && id=$(docker run --detach sigterm-me); sleep 1; docker container ls | grep ${id:0:12}; docker stop $id && docker logs $id
```

consideRatio:
Thanks for the review @manics! I'll self merge this as i actively need this for current development and will thereby also be able to test it properly, and we have no tests for this atm as its part of HTTPS setup.

Issue #1599, opened 2020-03-13
Reference or relocate static IP setup documentation


consideRatio:
It probably make sense to have documentation about static ip setups either referenced or located in this area: https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#set-up-automatic-https

I created this issue before about documenting static IP, and I think we ended up having documentation about it in the configuration reference section automatically generated from schema.yaml: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/797

PR #1663, opened 2020-05-12
Reference static ip docs

GeorgianaElena:
This references the documentation about setting up static IPs in the configuration reference section from the automatic https section.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1599

P.S.  After reading a bit through comments and other issues related to this, it seems that the `loadBalancerIP` should be set on the proxy service and not on the hub (or is it both?).
If it's just the proxy, we should probably remove the hub part from the docs.

### DIFF
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -50,6 +50,23 @@ changes to your `config.yaml` file:
 2. Apply the config changes by running `helm upgrade ...`
 3. Wait for about a minute, now your hub should be HTTPS enabled!
 
+***
+**NOTE:**
+
+If the proxy service is of type `LoadBalancer`, then a specific static IP address can be requested (if available) instead of a dynamically acquired one.  
+Although not essential for HTTPS, using a static IP address is a recommended practice for domain names referencing fixed IPs.
+This ensures the same IP address for multiple deployments.
+The IP can be provided like:
+
+```yaml
+  proxy:
+    service:
+      loadBalancerIP: xxx.xxx.xxx.xxx
+```
+
+More info about this can be found on the [Configuration Reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html) page.
+***
+
 ### Set up manual HTTPS
 
 If you have your own HTTPS certificates & want to use those instead of the automatically provisioned Let's Encrypt ones, that's also possible. Note that this is considered an advanced option, so we recommend not doing it unless you have good reasons.
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -53,7 +53,7 @@ changes to your `config.yaml` file:
 ***
 **NOTE:**
 
-If the proxy service is of type `LoadBalancer`, then a specific static IP address can be requested (if available) instead of a dynamically acquired one.  
+If the proxy service is of type `LoadBalancer`, which it is by default, then a specific static IP address can be requested (if available) instead of a dynamically acquired one.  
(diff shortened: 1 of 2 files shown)


consideRatio:
@GeorgianaElena ah! yes all traffic from outside the k8s cluster to the hub should go through the proxy, so having a loadbalancerIP for the hub service is not relevant at all.

It has been for me, and have for others as well I know, been a confusion if it makes sense to access the hub directly as compared to going through the proxy. I think the summary is that all cluster internal traffic can go directly to the hub but doesn't need to, but all traffic outside should go to the proxy and then be directed to the hub so that we only need to have a single external point of entry for TLS termination etc.

doc/source/administrator/security.md
```diff
@@ -50,6 +50,23 @@ changes to your `config.yaml` file:
...
+This ensures the same IP address for multiple deployments.
+The IP can be provided like:
+
+```yaml
+  proxy:
+    service:
+      loadBalancerIP: xxx.xxx.xxx.xxx
+```
+
+More info about this can be found on the [Configuration Reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html) page.
```
consideRatio, line 67: I'm not a 100%, but I think this could reference the relative path rather than a full internet url. There may be a deviation or two in these docs as there may be differences between markdown in a .md file, or rST in a rST file etc.

Could you have a look for a relative reference?

This LGTM otherwise! Thanks @GeorgianaElena !

GeorgianaElena: Thanks @consideRatio :sparkles: 
I just added a commit that uses a relative path rather than the full url of the docs section.
I also moved the docs from schema.yml about loadBalancerIp from hub to proxy. Is this ok?

jupyterhub/schema.yaml
```diff
@@ -533,7 +511,25 @@ properties:
...
+              This sets the IP address that should be used by the LoadBalancer for exposing the hub service.
+              Set this if you want the hub service to be provided with a fixed external IP address instead of a dynamically acquired one.
+              Useful to ensure a stable IP to access to the hub with, for example if you have reserved an IP address in your network to communicate with the JupyterHub.
+
+              To be provided like:
+              ```
+              hub:
+                service:
+                  loadBalancerIP: xxx.xxx.xxx.xxx
+              ```
```
consideRatio, line 525:
```suggestion
              The public IP address the proxy-public Kubernetes service should
              be exposed on. This entry will end up at the configurable proxy
              server that JupyterHub manages, which will direct traffic to user
              pods at the `/user` path and the hub pod at the `/hub` path.
              
              Set this if you want to use a fixed external IP address instead of
              a dynamically acquired one. This is relevant if you have a domain
              name that you want to point to a specific IP and want to ensure it
              doesn't change.
```

doc/source/administrator/security.md
```diff
@@ -50,6 +50,23 @@ changes to your `config.yaml` file:
 2. Apply the config changes by running `helm upgrade ...`
 3. Wait for about a minute, now your hub should be HTTPS enabled!
 
+***
+**NOTE:**
+
+If the proxy service is of type `LoadBalancer`, then a specific static IP address can be requested (if available) instead of a dynamically acquired one.
```
consideRatio, line 56:
```suggestion
If the proxy service is of type `LoadBalancer`, which it is by default, then a specific static IP address can be requested (if available) instead of a dynamically acquired one.  
```

consideRatio:
@GeorgianaElena thanks for your work! :heart:

PR #1667, opened 2020-05-15
removal: template validation (part of kubectl now)

consideRatio:
Kubectl has introduced a validation by default, so when we use
helm/kubectl to install charts, that rendered configuration is
validated.

It would make sense to keep testing this, but I'd like us to do it using
something like `helm template ... | kubectl apply --dry-run=server -f -`
where we run `helm template` with many various values combinations.

---

## About build failures and docs
I'll fix them in another PR (#1664)


consideRatio:
Closing this in favor of any future PR that removes it along with provides a modern alternative.

PR #1668, opened 2020-05-15
removal: Vagrant for local dev

consideRatio:
I've spent a lot of time thinking about the local development setup, and I'd like to steer away from maintaining any VM related content.

@manics as the author of this setup, what do you think about removing these files?


manics:
I'll take a closer look at this and related PRs later.

The original idea behind the Vagrant setup was to partially simulate a Travis environment for developing the CI setup. From past experience Travis is great once you've figured out your base infrastructure, whether that's setting up a virtualenv, figuring out a Docker config, or getting K8S running, but it's very hard to iron out bugs in your setup, or to modify that base setup. Travis is a bit of a black box and you wait ages to get feedback for each git push.

At some point it morphed into a dev setup 😀. If you're rearchitecturing the CI setup I think it'd be good to keep this around, especially in case there are problems with the new setup.

manics:
Since you're refactoring the CI how about:
- Ignore the Vagrant setup completely whilst you make your changes (i.e. deliberately break it)
- When you're done with your refactoring I'll attempt to update the Vagrant setup following the new docs. In theory this should be easy 🙂 
- This should be a good test of the docs. After that we decide whether to keep or delete it?

yuvipanda:
Death to vagrant for local dev setup, mostly :)

manics:
I'm maintaining my own copy for work on Fedora 32.

PR #1671, opened 2020-05-15
Tweaking readiness/liveness probe: faster startup

consideRatio:
With a readinessProbe that polls every ten seconds, we potentially need
to wait ten seconds after the pod is actually ready. By reducing this
number significantly, we avoid eigght potential seconds of waiting for
nothing.

At the same time, this could clutter logs. But, the proxy will not log
these requests, and I made a PR to JupyterHub to only log requests to
the health endpoint on the debug level: https://github.com/jupyterhub/jupyterhub/pull/3047

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -81,12 +81,12 @@ hub:
   templateVars: {}
   livenessProbe:
     enabled: false
-    initialDelaySeconds: 30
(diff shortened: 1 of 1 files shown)


consideRatio:
The outcomes of a change like this can look something like demonstrated by GIF animations in this PR: https://github.com/rancher/k3s/pull/1762

Issue #1642, opened 2020-04-23
Proposal: New Proxy which mangages Kubernetes Ingress Resources


stv0g:
Hi all,

I am considering the implementation of a new proxy sub-class which would allow us to ged rid of the CHP:

https://github.com/stv0g/nginx-ingress

The idea is simple: `jupyterhub_ingress_proxy` would create / delete Kubernetes `Ingress` resources when instructed.

There are some minor quirks involved:

#### The whole state of the proxy would reside in the Kubernetes APIServer

We can add the route data as a JSON encoded label

#### Ingress resources must match a Service resource. The Proxy API only provides a URL as the target

But this can be resolved:

- Check if URL points to a hostname or IP
- In case of an IP:
  - Create a `Service` resource without pod selector
  - Create an `Endpoint` resource with the IP address which gets bound to the service
  - Create an `Ingress` resource pointing to the new Service
- In case of an hostname:
  - We create an `Service` resource of type `ExternalName`
  - Create an `Ingress` resource pointing to the new Service

This article here describes the technique from above: https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-mapping-external-services

Alternatively, we might extend the Proxy API to include the singleuser pod name or hide it somewhere in the data attribute which gets passed to Proxy.add_route()

#### Are Ingress controllers fast enough to apply the new Ingress rules without a noticeable delay for the users?

Does somebody have experiences here?

stv0g:
This article here describes the technique from above: https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-mapping-external-services

stv0g:
Alternatively, we might extend the Proxy API to include the singleuser pod name or hide it somewhere in the data attribute which gets passed to `Proxy.add_route()`

manics:
Sounds like an interesting idea!

FYI there's also ongoing work to replace CHP with Traefik: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1162

There's also ongoing work to add `internal_ssl`:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1535
- https://github.com/jupyterhub/kubespawner/pull/386 This PR adds a service for each pod, so could be relevant to you?

stv0g:
Hi @manics,

[...]

PR #1673, opened 2020-05-18
Support an alternative proxy implementation: KubeIngressProxy with an external k8s Ingress controller

remche:
The  goal of this PR is to enable the use of native k8s ingress controller as described in #1642.

Changes :
1. Adds a way to skip proxy deployment : 
    - adds `proxy.chp.enabled` value (default to true)
    - templates files to avoid deployment of any proxy ressources 
    - pass value to hub configmap to allow hub configuration in `jupyterhub_config.py`

2. Patch the `Role` manifest when `externalIngressController` is enabled to allow (get, watch, list, create, delete, patch) (endpoints, services, ingresses).


remche:
Here is a minimal config file for testing : 

```
hub:
  image:
    name: remche/k8s-hub
    tag: ingress-extv1b1
  redirectToServer: false
  extraConfig:
    ingress_proxy.py: |
      c.JupyterHub.proxy_class = 'kubespawner.proxy.KubeIngressProxy'
      c.KubeIngressProxy.should_start = False
proxy:
  externalIngressController: true
  chp:
    enabled: false
```

- `remche/k8s-hub:ingress-extv1b1` is an jhub image featuring  jupyterhub/kubespawner#402.
- we use KubeIngressProxy and disable chp.

jupyterhub/templates/hub/rbac.yaml
```diff
@@ -19,6 +19,14 @@ rules:
   - apiGroups: [""]       # "" indicates the core API group
     resources: ["events"]
     verbs: ["get", "watch", "list"]
+  {{- if .Values.proxy.externalIngressController }}
+  - apiGroups: [""]       # "" indicates the core API group
+    resources: ["endpoints", "services"]
+    verbs: ["get", "watch", "list", "create", "delete", "patch"]
```
manics, line 25: To help with review did you figure these out by going through the K8s API calls, or is it a guess?

remche:
> Thanks for opening this PR! I haven't had time to try it yet, but what are your impressions of using it so far? Are you planning to use it in production?

I used it on test deployment with both traefik and nginx ingress controller with no pb so far. We would use it in production if merged.

> `proxy.externalIngressController` modifies the hub RBAC role but based on your configuration you still need to manually set `c.JupyterHub.proxy_class = 'kubespawner.proxy.KubeIngressProxy'`. Can you explain why you made this choice?

That's indeed a good question ;) 
It was my testing setup, but you're right I should've put this in `jupyterhub_config.py` !

> There's some ongoing work to refactor the C2JH CI system #1664
> Do you have any thoughts on adding a test scenario for `externalIngressController` _after_ the CI work is completed?

I need a deeper look but my first feeling is that I can add a test scenario for that case.

jupyterhub/templates/hub/rbac.yaml
remche: Additive hardworking way. I added permission according to hub error logs.

remche:
I simplified logic as @manics suggested, relying only on proxy.externalIngressController value. New minimal config file : 
```yaml
hub:
  image:
    name: remche/k8s-hub
    tag: ingress-extv1b1
proxy:
  externalIngressController: true
```

chancez:
@remche does your image have any changes to jupyterhub besides running kubespawner 0.12?

remche:
@chancez I did not try yet but now that kubespawner 0.12 is released and updated in 29e0d56, I'm pretty sure we dont need custom image anymore.

chancez:
It doesn't seem the underlying `KubeIngressProxy` has enough configuration options to be very useful. I can't configure annotations which means I can't set the ingress class, or custom ingress options. @remche did you have any advice on this?

remche:
In our use case we do not need IngressClass nor annotations but I agree that it should propose more configuration options.. With the GA of Ingress and IngressClass we might work on this soon.

manics:
@remche The CI work was completed a while ago and it seems pretty stable. Are you still able to add a test scenario?

consideRatio:
## Summary of relevant understanding to help discussion

JupyterHub relies on _some proxy_ that it can control and configure dynamically. The control JupyterHub exerts on the proxy is made through a well defined Python API [described here](https://jupyterhub.readthedocs.io/en/stable/reference/proxy.html) that can be implemented by any Python class. JupyterHub will delegate the work to for example add a network route to a user pod to a specified Python class (`c.JupyterHub.proxy_class`). This Python class is assumed to be an implementation of a [Proxy base class](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/proxy.py#L68), because JupyterHub will only use those methods to add and remove network routes.

In this JupyterHub Helm chart, we use [the default `c.JupyterHub.proxy_class`](https://github.com/jupyterhub/jupyterhub/blob/abd3bc13d2d7c9100a366c0862b1429981d5c584/jupyterhub/app.py#L653) configuration. This means that we use the [ConfigurableHttpProxy python class](https://github.com/jupyterhub/jupyterhub/blob/be8f8473093777b8e856c73200df471edf3d3e41/jupyterhub/proxy.py#L242). This Python class is acting based on configuration through `c.ConfigurableHttpProxy.<...>` which we do provide [in the Helm chart's `jupyterhub_config.py`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/8d6b0f6f89116c41d69d670fe621919f0c8464f2/jupyterhub/files/hub/jupyterhub_config.py#L24). This implementation for example needs to know how to speak with the actual Configurable HTTP Proxy server that is then configured using a REST API. The actual proxy that runs [jupyterhub/configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy) is running outside the JupyterHub container and is deployed through the Helm chart. It is running in a k8s Pod (`proxy-123..-.123..`) part of a k8s Deployment (`proxy`) and reached through a k8s Service (`proxy-public` or `proxy-http` depending on automatic HTTPS certificate acquisition is configured).

My understanding of this PR is that it want to accomplish two things. First that it wants to enable opting out of the default proxy implementation, and second that it wants to enable opting in to use the a class defined in [jupyterhub/kubespawner](https://github.com/jupyterhub/kubespawner) far less known than the actual Spawner, which is the [`KubeIngressProxy` class](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/proxy.py#L45).  `KubeIngressProxy` class that works by speaking with k8s API to create k8s `Ingress` and `Service` resources. K8s `Services` resources are entrypoints for network traffic to pods, and K8s `Ingress` resources are entrypoints for network traffic entering the k8s cluster and will point to k8s Services. The Ingress resources require a _ingress controller_ that inspects the Ingress resources and does the actual network traffic routing.

consideRatio:
## My thoughts about this PR

### Opting out of CHP usage
Enabling this Helm chart to let the advanced user opt out of the default proxy implementation seems fine to me. Such Helm chart configuration option should in my mind opt out in in a way agnostic to the replacement alternative. Adding such CHP opt-out feature would make sense to make a standalone PR as its quite a significant feature by its own.

__Notes for an opt-out of CHP PR__
- It would be good to add a conditional `fail` trigger in the `NOTES.txt` Helm template which is part of the Helm chart. This trigger could trigger on having an known invalid configuration such as having asked to opt out of using CHP while at the same time requesting something related to HTTPS as that typically depends on CHP or the k8s Services created along it.
- It would be important to have associated documentation about opting out centered about the consequences of the Helm charts functionality impacted.

### Opting in to use `KubeIngressProxy`
The support of using `KubeIngressProxy` as an alternative to the default CHP implementation is whats most complicated to review in my mind, as I consider it require a more in depth decision for the project as a whole. To motivate a positive decision I would like to feel confident it adds value and that we as a open source community are able to maintain it.

At the moment, I'm curious to know who is using the `KubeIngressProxy` and for what reasons over the CHP, and some history why we in this projected opted to use CHP in favor of `KubeIngressProxy` in the first place. I have not until now properly positioned the KubeIngressProxy class as sometehing that could make sense to use in this Helm chart so I have some thinking to do still.

@minrk and @yuvipanda I'd be interested to hear your thoughts about this and some history about the past thinking regarding CHP / KubeIngressProxy.

yuvipanda:
`KubeIngressProxy` was created to support extremely large deployments - see https://github.com/berkeley-dsep-infra/data8xhub for a config that was stress tested to like, 40k simultaneous users with sharding. It helped with HA, and removed the proxy as a source of scalability problems.

However, it turns out that when people say they expect upto 40k concurrent users, they are wildly optimistic haha. We had 40k total users, but barely 2k concurrent users. The setup I had was not worth it at the scale we *really* had, and I moved it back to CHP. I hadn't touched KubeIngressProxy since then, although it continues to work I think. We never publicized it either...

I think the main reason we didn't make KubeIngressProxy the *default* in z2jh is that it would require we bundle an ingress controller, and I lost a lot of steam at that time due to health issues. I think removing CHP completely is a noble goal that we should work towards, either via traefik-proxy or via default ingress objects. The HA, performance and upstream support make it worth it. 

An issue with ingress objects is that the route is *eventually consistent*, unlike with CHP where it's immediately consistent on return from the API call. I added support for [exponential backoff in user redirects](https://github.com/jupyterhub/jupyterhub/pull/1250) in JupyterHub to work around this, and it worked fairly well. IMO, this was the biggest issue with Ingress controllers - some take more than a second to register new routes...

Anyway, I welcome efforts to remove CHP from here :)

P.S. My first JupyterHub project was to implement a CHP alternative [with nginx](https://github.com/yuvipanda/jupyterhub-nginx-chp), by mimicking the CHP API (before proxy_class was a thing). So this warms my heart haha.

remche:
> @remche The CI work was completed a while ago and it seems pretty stable. Are you still able to add a test scenario?

@manics I worked on the test scenario but faced some weird issue with k3s on travis. I will give another try.
https://travis-ci.com/github/remche/zero-to-jupyterhub-k8s/builds

Besides being more "elegant" and k8s native, using ingress resources enable high availability and better resilience without the need to have a redis server. It would allow to use service mesh, too.

consideRatio:
Thanks for thorough and interesting work @remche!

I'd like to see the Helm chart configuration option be more obvious, as the boolean `externalIngressController` doesn't convey much unless you have a very clear understanding about what it already does in my mind. By configuring `proxy.type: chp` to `proxy.type: kip` it becomes a bit clearer though that we are replacing one approach with another.

I'm not great a reviewing this PR without any experience of using the KubeIngressProxy myself, but I think this can be merged without too much consideration if there is agreement to maintain support for this. I don't have a clear idea of this, and would like to know if @yuvipanda who has worked on KubeIngressProxy thinks this is worthwhile to support in Z2JH. What do you think @yuvipanda?

## My overview of the PR

This PRs goal is to let the JupyterHub Helm chart support using another proxy implementation than the default.

In the default proxy implementation, we run a [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy) within the k8s deployment named proxy, and let it be controlled through the [ConfigurableHTTPProxy class](https://github.com/jupyterhub/jupyterhub/blob/6ecda96dd636dfdd0cf34fcf6c1fc44942779c01/jupyterhub/proxy.py#L411).

To use another proxy implementation, we update both:
- `c.JupyterHub.proxy_class` to use another Proxy class, which is JupyterHub's interface to communicate it needs for certain network routes to be updated for the spawned user pods.
- The k8s proxy Deployment resource and its associated resources that are no longer needed.

Overall, I think what is important for this PR to get merged is that we lock in sensible configuration. I'd also like to see documentation to help future maintainers maintain this as part of Z2JH without for example being active users of this proxy implementation themselves. I think this comment takes us a bit along the way, but it would be great to have a section that introduce this configuration option properly in the z2jh.jupyter.org documentation both for users and maintainers.

## My technical considerations
- Does KubeIngressProxy support adding JupyterHub services and such as well?
- How does usage of KubeIngressProxy influence usage of autohttps? How can KubeIngressProxy handle that through changes to the Ingress resources etc?
- Does this Helm chart need to pass through additional configuration to configure the KubeIngressProxy class?

Id2ndR:
Thank you for this interesting work @remche. I tested it, and here are the remaining problems I found:
- KubeIngressProxy spawn *jupyter--2f-route* ingress that route to the hub just like *jupyterhub* ingress (the last one is created by the z2jh chart when *ingress.enabled* is true). Using the chart ingress however still is necessary to add annotations (like cert-manager / ingress class ones).
- KubeIngressProxy's spawned services and ingresses are not cleaned up when the chart is uninstalled
- configuration of KubeIngressProxy is not possible yet

IMHO, this PR mainly depends of https://github.com/jupyterhub/kubespawner/pull/406 and should be reworked after the latest is merged.

consideRatio:
@remche I'll opt to close this PR at this point as I perceive it as KubeIngressProxy [isn't a well enough maintained project](https://github.com/jupyterhub/kubespawner/pull/568/files) for z2jh to support integration against it with all the complexity that follows of testing, documenting, and adding to the general maintenance burden of the z2jh repo.

Thank you for your work on this nevertheless!

remche:
@consideRatio I agree and I'm really sorry not to have much time to spend on this.

consideRatio:
No worries at all, thank you for pioneering work on this in the open!!

PR #1679, opened 2020-05-22
hub deployment: tcp readiness/liveness probes

consideRatio:
The http probes clogged the jupyterhub log, which wasn't helpful. Using TCP probes that checks if is possible to establish a TCP connection wouldn't do that.

But, at the same time, https://github.com/jupyterhub/jupyterhub/pull/3047 also resolves the same thing to some degree by hiding logs about /health endpoint requests unless a loglevel is set to debug. This isn't yet available though.

I'm not sure, http vs TCP probes, pro's / con's. I assume tcpSocket may be more lightweight, I'm not sure.

Merge or close? It felt a bit simpler with TCP socket readiness/liveness probes as that didn't require us to choose a path etc as well.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -224,17 +224,15 @@ spec:
           # database upgrades midway or ending up in an infinite restart
           # loop.
           livenessProbe:
+            tcpSocket:
+              port: http
             initialDelaySeconds: {{ .Values.hub.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.hub.livenessProbe.periodSeconds }}
-            httpGet:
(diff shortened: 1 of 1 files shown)


manics:
A tcp probe only checks something is listening on the port. It won't check for e.g. 404 or 500 error codes, or some weird misconfiguration where a non-http service is listening.

consideRatio:
Yeah, HTTP probes test more than a TCP probe don't they?

I struggle to think of what makes most sense because I lack experience of this making a difference, and I also lack of experience of what actually is wanted. Do we want a readinessProbe / livenessProbe to fail if HTTP probe fails, or only if it is even worse and the TCP probe fails?

We have a single replica, a readinessProbe that have failed will make us get 404 responses on access to the associated k8s service, there is no fallback anyhow. I'm not sure what I think so I figure I'll close this PR for now!

PR #1681, opened 2020-05-23
[Experiment] helm lint --strict compliance - not fully there

consideRatio:


baboune:
I like the changes!

consideRatio:
I ran into a few issues that would require some work to resolve, hmmm...

1. There were some helm templating logic in the hub's configmap and the hub's secret that was way to complicated to allow for `--strict` flag to approve, but #1682 would resolve it by avoiding that logic. It is a PR that makes sense on its own.

2. There were some issues for the helper functions, but it could be resolved by providing some default values in the values.yaml file, even though they may be pointless for the end user of the Helm chart.

3. Providing even a blank string is required to avoid a common `--strict` error relating to referencing a null value, but providing a blank string will make our configuration in jupyterhub_config.py set it to a blank string down the line, and that in turn gives us errors. Sometimes a blank string should be treated as a blank string though hmm.

4. For integers that needed a value, I could set the value 0 to represent a falsy value. It may be confusing though, hmm...

My conclusion is that this require quite a bit of modifications, and if it is done, I think it should be done with a clear reason - and right now it is mainly based on aesthetics of being compliant with helm's CLI. There were some modifications made that make sense of their own I think, but also some compromises, and I don't think it seems worth it without additional incitements.

---

Oh hey Nicolas btw! I'm Erik in the Data Science team of Sandvik :D

PR #1682, opened 2020-05-23
Complexity reduction - combine passthrough values.yaml data in hub-config (k8s configmap) to hub-secret (k8s secret)

consideRatio:
JupyterHub itself, authenticators, etc, all may want to use the values
we also use to render the Helm charts templates. So, we typically take
them av available in {{ .Values }}, split it apart, and render it to a
k8s configmap as well as a k8s secret.

This separation had little to no benefit and was hard to maintain. In
this commit I gather the {{ .Values }} into the hub-secret file. I also
configure c.JupyterHub.cookie_secret instead of passing that
information by mounting a k8s secrets key's value into a
JPY_COOKIE_SECRET environment variable on the hub pod.

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -62,6 +62,7 @@ for trait, cfg_key in (
     ('active_server_limit', None),
     # base url prefix
     ('base_url', None),
+    ('cookie_secret', None),
     ('allow_named_servers', None),
     ('named_server_limit_per_user', None),
     ('authenticate_prometheus', None),
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -2,6 +2,8 @@ import os
 import re
 import sys
 
+from binascii import a2b_hex
+
 from tornado.httpclient import AsyncHTTPClient
 from kubernetes import client
 from jupyterhub.utils import url_path_join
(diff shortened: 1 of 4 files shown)


jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -62,6 +62,7 @@ def camelCaseify(s):
     ('active_server_limit', None),
     # base url prefix
     ('base_url', None),
+    ('cookie_secret', None),
```
minrk, line 65: I don't think we can treat `cookie_secret` quite as transparently as the rest of the traits; I think it needs special handling because `JupyterHub.cookie_secret` is a `Bytes` trait, which should fail when you try to set it with text from yaml.

Instead, adding the same encoding from yaml text as is [currently done](https://github.com/jupyterhub/jupyterhub/blob/3663d7c8fcba2118dd469a8a4cca540cc63a39cf/jupyterhub/app.py#L1345-L1346) when it's set via env:

```python
from binascii import a2b_hex
cookie_secret_hex = get_config("hub.cookieSecret")
if cookie_secret_hex:
    c.JupyterHub.cookie_secret = a2b_hex(cookie_secret_hex)
```

minrk (commented):
Sorry for not seeing this long ago, I was utterly buried in the Spring, but this is great! I'd love to see this land.

I think cookie_secret needs a little more special handling to come directly from yaml, but other than that, 👍

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Thanks for the review and nice catching this @minrk! I've updated the PR with a commit reflecting this insight!

consideRatio: @minrk, does this look good? https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682/commits/9aff240572627488b7e0501bfbbcea1d25757c88

consideRatio: I'm marking this as resolved as I think I got it right.

PR #1689, opened 2020-06-05
[MRG] Add services API tokens in hub-secret

betatim:
Having an explicit key for the API token of each service is nice because
it allows the service to mount the particular key as an environment
variable.

We used to have this functionality but in #941 we moved the API tokens to the `values.yaml`. This is a bit tricky to use because now your service has to mount that yaml, parse it and get the token out of the yaml (and it gets access to tokens from all services).

### DIFF
--- a/jupyterhub/templates/hub/secret.yaml
+++ b/jupyterhub/templates/hub/secret.yaml
@@ -14,6 +14,11 @@ data:
   {{- if .Values.hub.db.password }}
   hub.db.password: {{ .Values.hub.db.password | b64enc | quote }}
   {{- end }}
+  {{- range $key, $service := .Values.hub.services }}
+  {{- if $service.apiToken }}
+  services.token.{{$key}}: {{ $service.apiToken | b64enc | quote }}
+  {{- end }}
+  {{- end }}
   {{- if .Values.auth.state.enabled }}
   auth.state.crypto-key: {{ (required "Encryption key is required for auth state to be persisted!" .Values.auth.state.cryptoKey) | b64enc | quote }}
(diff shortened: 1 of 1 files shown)


betatim:
Tested this with the following as `test.yaml`:
```yaml
hub:
  services:
    refresher:
      admin: true
      url: http://myservice/
      apiToken: "TIMTIMTIMTIM"

proxy:
  secretToken: "fizzbuzzle"

auth:
  state:
      enabled: true
      cryptoKey: "blahblahblah"
```
and then `helm template jupyterhub -f test.yaml | grep -A20 "^  name: hub-secret"` which outputs:
```
  name: hub-secret
  labels:
    component: hub
    app: jupyterhub
    release: release-name
    chart: jupyterhub-0.0.1-set.by.chartpress
    heritage: Tiller
type: Opaque
data:
  proxy.token: "Zml6emJ1enpsZQ=="
  services.token.refresher: "VElNVElNVElNVElN"
  auth.state.crypto-key: "YmxhaGJsYWhibGFo"
  values.yaml: "YXV0aDoge30KaHViOgogIHNlcnZpY2VzOgogICAgcmVmcmVzaGVyOgogICAgICBhcGlUb2tlbjogVElNVElNVElNVElNCg=="
```

betatim:
I have no super strong opinion on naming of the key. One reason to use this naming scheme is that it is we used before.

BertR:
This would benefit my deployment as well.

betatim:
Could we merge this for now and if we decide to change the name of the key do that in a second PR before the next release?

PR #1691, opened 2020-06-09
Add support for setting hub-wide notebook config

yuvipanda:
Sometimes you want to set some notebook server configurations
for all your users. On mybinder.org, we do this for
tornado settings & idle culler settings
(https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/values.yaml#L5)

This is generally useful, so I've ported that from mybinder.org
here. Once this gets merged, we can migrate the mybinder.org
config to use this.

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -246,6 +246,18 @@ elif storage_type == 'static':
 c.KubeSpawner.volumes.extend(get_config('singleuser.storage.extraVolumes', []))
 c.KubeSpawner.volume_mounts.extend(get_config('singleuser.storage.extraVolumeMounts', []))
 
+if get_config('singleuser.etcJupyter'):
+    c.KubeSpawner.volumes.extend({
+        'name': 'user-etc-jupyter',
+        'configMap': {
+            'name': 'user-etc-jupyter'
+        }
+    })
+
+    c.KubeSpawner.volume_mounts.extend({
+        'name': 'user-etc-jupyter',
+        'mountPath': '/etc/jupyter'
+    })
 # Gives spawned containers access to the API of the hub
 c.JupyterHub.hub_connect_ip = os.environ['HUB_SERVICE_HOST']
 c.JupyterHub.hub_connect_port = int(os.environ['HUB_SERVICE_PORT'])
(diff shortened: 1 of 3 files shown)


consideRatio:
This is specifically for a etc juptyer folder. I think id like to see us include this more generally than settle for etc/jupyter.

I think id like it the most if we could avoid assuming json/yaml/toml etc, while supporting a helm dict passed through values to to render into it, with or without quoting etc, and have an api to mount a collection of individual files with info provided.

I imagine singleuser.mountedFiles where we mount individual files in folders using subMount to avoid overriding an entire directory, and allow input of string/base64 encoded binary/dict objects, that allow us to then render it as string/binary/(any of json/yaml/toml). We would then need to ask for filepath (including name), filepermissions, toJson/yaml/toml transform, squote/quote, 

I think it is a common situation to want to mount more than just to etc/jupyter if one has decided such details are worth considering. For example one can mount not only to etc/jupyter, but also .bashrc, etc/dask, etc.

yuvipanda:
Thanks @consideRatio!

I think in general mounting arbitrary files should be handled by users making configmaps and using `extraVolumes` and `extraVolumeMounts`. IMO, having something specific for this extremely common use case is worth discussing separately from a super generic file mounting option.

yuvipanda:
and re: JSON, it's mostly special cased so you can pass in a dict to helm, and it'll show up as JSON - so you don't have to quote it. This helps make sure that you are in fact passing valid JSON. If you were to pass a .py file instead, or any other extension, it should just put the file there directly.

yuvipanda:
Actually, the more I think about it, I think you're absolutely right in wanting to make this more general purpose :) I'll spend some more time working on that. Will mark this as WIP in the meantime.

Thanks for the awesome idea!

consideRatio:
Wiieee! I'm really excited about a feature like that which really helps with motivation to maintain it!

I would probably often not choose to use such functionality myself unless it was more general as I would anyhow probably want more general functionality, and then it becomes more complicated to introduce another system along the custom but general system.

@yuvipanda we also have the `--set-file` flag available (see [helm options](https://helm.sh/docs/helm/helm_install/#options)), that allow a user of this functionality to avoid embedding for example python, shell, or other formatted scripts inside the helm values file.

consideRatio:
@yuvipanda I'll go ahead and close this for now, but I have captured a lot of the idea and implementation details I think we should aim for in the issue i just mentioned this PR from.

yuvipanda:
AWESOME! Thank you for picking it up, @consideRatio!

PR #1695, opened 2020-06-12
Update Google auth to use a list for hosted_domain

petebachant:
I was getting this error with v0.9.0 and the change in the commit attached fixed it:

```
traitlets.traitlets.TraitError: The 'hosted_domain' trait of a GoogleOAuthenticator instance must be a list, but a value of class 'str'
```


consideRatio:
@petebachant thank you for patching this! :heart:

PR #1699, opened 2020-06-25
Enable the user scheduler to pay attention to CSI volume count

rschroll:
I've been getting the 0.9.0 chart running on an autoscaling cluster on Digital Ocean.  Most things are working, but it would try to schedule more than seven pods with volumes on a given node.  (Digital Ocean has a relatively low limit of 7 on this.)  This only happened when using the userScheduler; if that was disabled, the limits were respected.

Digital Ocean support pointed out that the scheduler did not have the "MaxCSIVolumeCountPred" predicate, which they use to count the attached volumes.  This appears to be a more general predicate, which will replace the provider specific ones we currently have. (See [here](https://github.com/mjaow/kubernetes/blob/master/pkg/scheduler/algorithm/predicates/predicates.go#L61).)

I could only get this to work with an updated version of the kube-scheduler image.  I don't know what was going wrong with the current version, and I also don't know if it needs to be updated this far.  If we're worried about upgrading this far, I can check some of the previous versions and report back.

### DIFF
--- a/jupyterhub/files/userscheduler-defaultpolicy.yaml
+++ b/jupyterhub/files/userscheduler-defaultpolicy.yaml
@@ -11,6 +11,7 @@
     { "name": "MaxEBSVolumeCount" },
     { "name": "MaxGCEPDVolumeCount" },
     { "name": "MaxAzureDiskVolumeCount" },
+    { "name": "MaxCSIVolumeCountPred" },
     { "name": "CheckVolumeBinding" },
     { "name": "NoVolumeZoneConflict" },
     { "name": "MatchInterPodAffinity" }
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -318,7 +318,7 @@ scheduling:
     policy: {}
     image:
       name: gcr.io/google_containers/kube-scheduler-amd64
-      tag: v1.18.4
+      tag: v1.16.11
     nodeSelector: {}
     pdb:
       enabled: true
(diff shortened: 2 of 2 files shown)


rschroll:
It's working on 1.16 and 1.18.  Not sure what 1.17 is unhappy about.

minrk:
Just had to restart 1.17, turns out it was an intermittent error on CI, not a true failure.

Thanks!

rschroll:
Thanks for finding this one -- I had forgotten about it!

consideRatio:
On a k8s 1.15 cluster, I'm running into the following errors in the user-scheduler's logs.

```
I0908 10:05:43.912366       1 reflector.go:158] Listing and watching *v1beta1.CSINode from k8s.io/client-go/informers/factory.go:134
E0908 10:05:43.914298       1 reflector.go:123] k8s.io/client-go/informers/factory.go:134: Failed to list *v1beta1.CSINode: csinodes.storage.k8s.io is forbidden: User "system:serviceaccount:jupyterhub:user-scheduler" cannot list resource "csinodes" in API group "storage.k8s.io" at the cluster scope
```

This is probably because k8s 1.15 clusterrole for the kube-scheduler doesn't provide it, only k8s 1.16's clusterrole for the kube-scheduler. We are augmenting this, so to be compatible with 1.15 we need to augment it with whats available by default in k8s 1.16's kube-scheduler clusterrole.

Trying to add `csinodes` failed though, because k8s 1.15 rejected it as it didn't find it.

```
# * : Invalid value: "The edited file failed validation": the server could not find the requested resource
```

Due to this, the solution as I see it, is to only add this conditionally.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1699/files

PR #1702, opened 2020-06-28
Add annotations to image puller pods

### LINKED ISSUE #1627: Annotate image puller and awaiter
### Proposed change
It is not possible to add annotation to the image puller deployments.

This is useful for example deploying jupyterhub to an istio enabled namespace (where all pods are auto injected unless they are annotated)

Happy to submit a PR on this.

duongnt:
This PR adds the ability to add annotations to the continuous and hook image puller pods.

Fixes #1627

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1163,6 +1163,11 @@ properties:
   prePuller:
     type: object
     properties:
+      annotations:
+        type: object
+        description: |
+          Annotations to apply to the hook and continous image puller pods. One example use case is to
+          disable istio sidecars which could interfere with the image pulling.
       hook:
         description: |
           See the [*optimization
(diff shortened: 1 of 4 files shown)


consideRatio:
Welcome and thanks for your PR @duongnt! I think this makes sense to add, could you do the following?

- [ ] Add a entry for `prePuller.annotations` set to `{}` in values.yaml. Just a practice we have, I find it useful to help myself track what is possible to configure, and I know `helm lint` likes it.
- [ ] Use `with` instead of `if`, use `trimSuffix "\n"`, and pipe into `toYaml` like below. This is just format nitpicking.

```yaml
      {{- with .Values.prePuller.annotations }}
      annotations:
      {{- end }}
```

duongnt:
@consideRatio Thanks for your review! I addressed your comments.

duongnt:
Just realize that this fixes #1627

consideRatio:
@duongnt I think you also need to not use a `"` when you reference `.Values...`. After that this PR is ready for merge I think!

consideRatio:
Thank you @duongnt !!!!! :heart: :tada:

PR #1711, opened 2020-07-05
removing circleci

choldgraf:
This removes CircleCI for documentation previews, because readthedocs now supports its own documentation previewing feature so this reduces some of the complexity of this repo. I've activated it on RTD so let's see if it works!

Currently the tests *should* fail because CircleCI won't be able to find the config. However if folks are +1 then I'll turn off CircleCI building and then we should be good to merge


consideRatio:
Ah nice =)

choldgraf:
yeah we needed to click a box in the RTD config!

PR #1714, opened 2020-07-10
Autohttps reliability fix: bump traefik version

consideRatio:
I [recently pinned](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/66b917d98ba877d9fad70351dba07b9e08af65a2) v2.2.2 of traefik (running in our autohttps pod) from v2.2 as v2.2.2 contain [a fix](https://github.com/containous/traefik/pull/6939) to [an issue](https://github.com/containous/traefik/issues/6871) that will make our autohttps system stop malfunction randomly in our CI system and with users trying to acquire certificates.

Now I noted that [Traefik released v2.2.3 / 2.2.4](https://github.com/containous/traefik/releases/tag/v2.2.4) to fix a regression, so this PR bumps it once again. The docker image [will soon be out](https://hub.docker.com/_/traefik?tab=tags) but isn't yet. When its out, we restart this test and merge on success.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,7 +180,7 @@ __Install Pebble__
 ```shell
 helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
 helm repo update
-helm install pebble jupyterhub/pebble --values dev-config-pebble.yaml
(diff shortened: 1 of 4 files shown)


consideRatio:
The worst bug i've experienced for a long time and certainly this year... Yikes.

See #1716 for a trail of the trouble only to end up reducing the risk of issues...

Issue #1712, opened 2020-07-10
CI: Automatic hub image scanning using trivy


meneal:
### Bug description
Running [trivy image scanning]() against the image for k8s-hub results in a number of vulnerabilities found.

#### Expected behaviour
0 vulnerabilities

#### Actual behaviour
`trivy image --ignore-unfixed jupyterhub/k8s-hub:0.9.0-n033.h8211ad2 > out.txt`

Results in: [`out.txt`](https://github.com/jupyterhub/jupyterhub/files/4900105/out.txt) showing 100 vulnerabilities

### How to reproduce
1. Install [trivy](https://github.com/aquasecurity/trivy)
2. Run `trivy image --ignore-unfixed jupyterhub/k8s-hub:0.9.0-n033.h8211ad2 > out.txt`

### Your personal set up
 - OS: MacOS
 - Version: jupyterhub/k8s-hub:0.9.0-n033.h8211ad2 
 - Configuration: No configuration

I'm reporting on `jupyterhub/k8s-hub:0.9.0-n033.h8211ad2` because the release version `0.9.0` has around 300 vulnerabilities.

consideRatio:
@meneal thanks for introducing these toolkits as part of reporting on the security issues in the image! I'm curious to see the current status if we rebuild it now. I'll do that.

consideRatio:
With it being rebuilt, it ended up like this:

```
hub-dependencies (ubuntu 18.04)
===============================
Total: 8 (UNKNOWN: 0, LOW: 4, MEDIUM: 4, HIGH: 0, CRITICAL: 0)

+----------+------------------+----------+-------------------+-----------------+------------------------------------+
| LIBRARY  | VULNERABILITY ID | SEVERITY | INSTALLED VERSION |  FIXED VERSION  |               TITLE                |
+----------+------------------+----------+-------------------+-----------------+------------------------------------+
| libc-bin | CVE-2018-11236   | MEDIUM   | 2.27-3ubuntu1     | 2.27-3ubuntu1.2 | glibc: Integer overflow in         |
|          |                  |          |                   |                 | stdlib/canonicalize.c on           |
|          |                  |          |                   |                 | 32-bit architectures leading       |
|          |                  |          |                   |                 | to stack-based buffer...           |
+          +------------------+          +                   +                 +------------------------------------+
|          | CVE-2018-11237   |          |                   |                 | glibc: Buffer overflow in          |
[...]

PR #1715, opened 2020-07-10
Upgrade libc to patch vulnerability in hub img

meneal:
Supports #1712

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get update && \
       sqlite3 \
       curl \
       dnsutils \
+      libc-bin=2.27-3ubuntu1.2 \
       $(bash -c 'if [[ $JUPYTERHUB_VERSION == "git"* ]]; then \
         # workaround for https://bugs.launchpad.net/ubuntu/+source/nodejs/+bug/1794589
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks @meneal makes perfect sense! Adding an inline comment to ensure we remember why this is added.

images/hub/Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN apt-get update && \
       sqlite3 \
       curl \
       dnsutils \
+      libc-bin=2.27-3ubuntu1.2 \
```
consideRatio, line 23:
```suggestion
      # libc-bin explicitly added and pinned 2020-06-10 for a CVE found with trivy
      # ref: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1712#issuecomment-656425028
      libc-bin=2.27-3ubuntu1.2 \
```

consideRatio:
@meneal wieeeeeee thanks!!!

@jupyterhub/jupyterhubteam in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1712 @meneal showed how we can use a tool called trivy to inspect our images, perhaps this is something we want to make use of in various repos.

PR #1722, opened 2020-07-17
update kubespawner to 0.12

minrk:
Main changes:

- changes default named-server template to avoid possible collisions
- use slugs to select profiles


minrk:
This resolves security issue GHSA-v7m9-9497-p9gr in KubeSpawner 0.11 (already published chart 0.9.1 with a [smaller patch](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/a80527025ec4d2455322a28ccac5fb6c2f492a36) that results in the same change)

PR #1727, opened 2020-07-21
jupyterhub existing image pull secret configuration load bug fixed

mpolatcan:
When we want to deploy Jupyterhub on Kubernetes, we have had an issue about that singleuser pod spawner which it is also named Kubespawner doesn't use our existing image pull secrets although we had defined on values.yaml. I have found that jupyterhub_config.py configuration doesn't include that config load so anyone can't use existing image pull secrets on their system. Also Kubespawner takes image pull secret name as string not array so that I have changed format of the image pull secret configuration on values.yaml :)

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -48,8 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    # pullSecrets:
-    #   - secretName
+    pullSecrets: secretName
   resources:
     requests:
       cpu: 200m
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -48,7 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    pullSecrets: secretName
+    pullSecrets:
   resources:
     requests:
       cpu: 200m
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,5 @@ pyvenv.cfg
 
 # Emacs
 .#*
+
+.idea
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -111,6 +111,7 @@ set_config_if_not_none(
 for trait, cfg_key in (
     ('start_timeout', None),
     ('image_pull_policy', 'image.pullPolicy'),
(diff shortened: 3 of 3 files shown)


jupyterhub/values.yaml
```diff
@@ -48,8 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    # pullSecrets:
-    #   - secretName
+    pullSecrets: secretName
```
consideRatio, line 51:
```suggestion
    pullSecrets: ""
```

jupyterhub/values.yaml
```diff
@@ -48,8 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    # pullSecrets:
-    #   - secretName
+    pullSecrets: ""
```
consideRatio, line 51:
```suggestion
    pullSecrets:
```

jupyterhub/values.yaml
```diff
@@ -48,8 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    # pullSecrets:
-    #   - secretName
+    pullSecrets:
```
consideRatio, line 51:
```suggestion
    pullSecrets: []
```

consideRatio:
> Also Kubespawner takes image pull secret name as string not array so that I have changed format of the image pull secret configuration on values.yaml :)

Oh

jupyterhub/values.yaml
```diff
@@ -48,8 +48,7 @@ hub:
   image:
     name: jupyterhub/k8s-hub
     tag: 'set-by-chartpress'
-    # pullSecrets:
-    #   - secretName
+    pullSecrets: []
```
consideRatio, line 51:
```suggestion
    pullSecrets:
```

mpolatcan: config c.KubeSpawner.image_pull_secrets = Unicode(None)

The kubernetes secret to use for pulling images from private repository.

Set this to the name of a Kubernetes secret containing the docker configuration required to pull the image.

See the Kubernetes documentation for more information on when and why this might need to be set, and what it should be set to.

We couldn't set pullSecret config to array type because Kubespawner takes that configs as Unicode string so I have set it to string type on my production deployment with existing image pull secrets :)

PR #1735, opened 2020-08-04
Docs/schema.yaml patches

rubdos:
Updating the docs while I am installing z2j-k8s. Our deployment should be up and running by September, so I don't expect to add anything to this PR after then.

Feel free to cherry pick those changes that you want!


minrk:
Excellent, thank you!

PR #1736, opened 2020-08-05
[Documentation] Authenticating with Auth0

asubb:
I've spent some time figuring out how to integrate with yet another OAuth 2.0 / OpenID Connect provider, perhaps one more example would be helpful for others.

### DIFF
--- a/doc/source/administrator/authentication.rst
+++ b/doc/source/administrator/authentication.rst
@@ -345,6 +345,36 @@ This example is equivalent to that given in the
          - 'cn=researcher,ou=groups,dc=wikimedia,dc=org'
          - 'cn=operations,ou=groups,dc=wikimedia,dc=org'
 
+Example Auth0 Configuration
+------------------
+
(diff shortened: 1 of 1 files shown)


manics:
@consideRatio Do you think this is the right place to document auth0? Or would https://github.com/jupyterhub/oauthenticator#installation be better?

consideRatio:
I don't think we need to hold back on merging this as it provides relevant examplification of the not obvious need to configure an environment variable and such alongside and how to specify the use of the specific authenticator class.

If there would only be documented at one place, I think it would be better to have documented in jupyterhub/oauthenticator though. Perhaps a reference to https://github.com/jupyterhub/oauthenticator#installation would be good so if this becomes outdated, the user would know where the actual source of truth is.

I'd say let's merge and put in work down the line if needed, at this point this repo / the forum probably gets more maintenance / help burden of not having such practical example in the documentation so I think this is a win win for users and maintainers at this point.

yuvipanda:
Thanks for the PR, @asubb!

PR #1740, opened 2020-08-09
[resource prefixing] Prefixed release.name to some of the configurations and secrets

jorgecarleitao:
In general, Helm recommends that every chart should be deployable on the same namespace multiple times by prepending the release name to all resources created by it.

This PR prefixes some of our deployables with release.name. It does not solve the issue completely, but IMO it does improve the situation.


consideRatio:
I agree with the goal outcome, but making the upgrade is what I find to be scary. Some resources won't let themselves be renamed without recreating them and such for example.

What we need for this kind of change to land is a clear strategy on the upgrade plan. I'm not sure what I think is sensible.

## Transition ideas
1. To prefix something generated from a helm templates defined in `_helpers.tpl` to all resource names which by default will be a blank string to not introduce trouble for existing deployments.
2. To thoroughly evaluate all the resources we create and consider which can be recreated without issues. I think though that current user pods would at least be forced to be restarted, and I think PVCs created by KubeSpawner would need to be renamed.

I think 1 will be the only viable solution, making this kind of feature something we opt-in to even though it would be preferred to have something you opt out from instead like in most Helm charts.

kristofmartens:
I created an alternative PR for this issue: #1775
@consideRatio I agree with your remarks of this change potentially causing new resources to be recreated. I would also like to make the case to not quickly dismiss option 2 of your transition idea's. Keep in mind that every update can cause pods to restart. The resources where we have to be the most careful are the PVC resources as they contain operational information and I would propose to only use option 1 for these resources only

ashrafgt:
Any progress on this please?

consideRatio:
Thanks for being an active member of this project @jorgecarleitao!

As this is a popular topic that has been discussed and considered in many places without clear resolution for a long time, I've decided to close all existing issues and PRs relating to it and refer to a single issue: #1791. In this issue I've done my best to provide a clear and actionable path towards progress.

jorgecarleitao:
Perfect, I agree with approach. Thanks for taking this on, @consideRatio

PR #1741, opened 2020-08-09
Split Python config in 2 for simpler customization and maintainability

jorgecarleitao:
Currently, customizing the application requires some copy-pasting of the existing Python configuration (for the hub).

This PR splits that file in two parts: one with configuration functions, and another with the execution. This allows users to re-use most of the configuration code without having to copy-paste the existing configuration.

### DIFF
--- a/jupyterhub/files/hub/helpers.py
+++ b/jupyterhub/files/hub/helpers.py
@@ -1,0 +1,503 @@
+import os
+import sys
+
+import kubernetes.client
+
+from z2jh import get_config, set_config_if_not_none, camelCaseify
+
+
+def configure_kube_spawner_volumes(kube_spawner):
+
+    # Configure dynamically provisioning pvc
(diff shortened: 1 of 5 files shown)


minrk:
Thanks for the PR!

> Currently, customizing the application requires some copy-pasting of the existing Python configuration (for the hub).

I'm not sure I quite follow this. Arbitrary configuration can be loaded via `hub.extraConfig`, so no copy-paste from jupyterhub_config.py should be necessary. What is your situation that you've run into a need to copy/paste jupyterhub_config and extraConfig does not work?

I'm also not opposed in principle to organizing things in utility functions, which we do already do in z2jh.py, but I would say as a minor implementation detail to add new helpers to z2jh.py (or make z2jh a package instead of a single module if multiple files is preferred) rather than adding a second helpers.py at the top-level.

Issue #1103, opened 2019-01-14
A secrets friendly version of extraEnv


xrl:
I am looking for design feedback on a more general version of `extraEnv`. I have looked over #440 and it makes valid points but doesn't talk about a design to move forward.

Folks are going to be creating multi-user jupyter hubs where they'll want all environment to share some base secrets. I'm thinking AWS IAM credentials for reading from buckets or athena. We're not talking about "admin for everything" kind of credentials.

While these secrets are low impact I would rather not keep them in my `values.yml` used for the jhub release. What if I wanted to keep my kube config project as an open repo? I should be able to balance transparent kube objects and kube's secret system. Also, I want all my secrets in kubernetes so I can more easily audit and roll them later.

Could we convert `extraEnv` back to a more "raw" format? Instead of the current `envKey: envValue` key-pair, could we do a list of native kubernetes env values? We would absolutely be giving up the convenience in exchange for a more verbose but ultimately far more powerful format. A blind substition is also more the style of helm charts.

So `envFrom` would be something like:

```
  extraEnv:
    - name: AWS_ACCESS_KEY_ID
      value: 1234ABCD
```

and if they wanted to mount a secret instead:

```
  extraEnv:
    - name: AWS_ACCESS_KEY_ID
      valueFrom:
        secretKeyRef:
          name: jupyter-user
          key: AWS_ACCESS_KEY_ID
```

Perhaps we don't want to break folks values.yml, what would be another good `values.yml` key to use?

minrk:
> could we do a list of native kubernetes env values?

I definitely think we should move toward standard kube structures as much as possible. However, we've also learned that list config doesn't work well for hierarchical charts since they don't get merged (setting extraEnv in one chart that includes jupyterhub means deployments must be aware of and duplicate all of the extraEnv in that chart in their deployment if they want to set any env without clobbering). Dicts merge a lot nicer, so I'm somewhat inclined to deviate from the spec only in that the config ought to be a dict somehow. There are two ways I see to do that:

1. a dict of exactly kubernetes [EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#envvar-v1-core) arrays that just get concatenated, or
2. a dict keyed by env name where the body is *either* a simple env value (current), or an EnvVar struct.

[...]

PR #1462, opened 2019-10-29
Added doc example of mounting k8s secrets as ENV variables to user pods without values.yaml

sgloutnikov:
Added example of mounting Kubernetes secrets as environment variables inside user pods without exposing them in values.yaml.

PR #1757, opened 2020-08-26
hub.extraEnv / singleuser.extraEnv in dict format to support k8s EnvVar spec

consideRatio:
# Background
In short, specifying an environment variable for the JupyterHub pod or the user pods could only be done in a limited fashion that excluded specifying an environment variable that got its value from some Kubernetes ConfigMap/Secret or similar. This has only been only possible with intrusive workarounds. This PR closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1103 and an associated KubeSpawner PR has already closed https://github.com/jupyterhub/kubespawner/issues/306#issuecomment-474934945.

## hub.extraEnv advanced syntax support

This PR enables `hub.extraEnv` to function with a format either exactly like Kubernetes [EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core) specification or a shorthand where a `name: ...` is implied. This format is now also documented in the configuration reference ([docs preview](https://zero-to-jupyterhub--1757.org.readthedocs.build/en/1757/reference/reference.html#hub-extraenv)).

## singleuser.extraEnv advanced syntax support

This PR doesn't fully close https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1103 because this only enables the new syntax support for `hub.extraEnv` but _not also for `singleuser.extraEnv`_. https://github.com/jupyterhub/kubespawner/pull/426 is now merged which takes care of the `singleuser.extraEnv` part.

### Technical breakdown regarding singleuser.extraEnv

The Helm chart configuration `singleuser.extraEnv` propegates without issues to `c.KubeSpawner.environment` which is [defined in the Spawner base class](https://github.com/jupyterhub/jupyterhub/blob/29de00ee3c7f1ad60f869e2e3877fb5ab923c60e/jupyterhub/spawner.py#L453), so  it turned out that no changes were needed in this repo. Below are still some of the notes I made along the way to ensure that was the case and to develop the PR (https://github.com/jupyterhub/kubespawner/pull/426) in KubeSpawner that fixed it over there.

1. Either directly configuring `c.KubeSpawner.environment` or doing it through `singleuser.extraEnv` should be fine even if we use the new syntax.
2. KubeSpawner references the `environment` traitlet from the end of the [`jupyterhub.Spawner.get_env` function](https://github.com/jupyterhub/jupyterhub/blob/cc8e7806530466dce8968567d1bbd2b39a7afa26/jupyterhub/spawner.py#L726).
3. The `get_env` function is [then called](https://github.com/jupyterhub/kubespawner/blob/bf1fdca48a4a73278e629bffa6f749136edd5f86/kubespawner/spawner.py#L1435) to provide the `env` argument to the `make_pod` function.
4. The `make_pod` function [use the the passed env vars once](https://github.com/jupyterhub/kubespawner/blob/3d9519ba2c1894a4a8f5e6b211fa92c12fcdb78d/kubespawner/objects.py#L282), that assuming support for new syntax should be able to get a `dummyKey1` with a value of `{name: "MY_ENV", value: "my-value"}` passed to it.
5. The [`V1EnvVar` constructor](https://github.com/kubernetes-client/python/blob/bc79976edfe01b028aef0b93bde54e002b113443/kubernetes/client/models/v1_env_var.py#L47) allows for name/value/value_from, but is only used to pass a name/value combination rather than the possible name/valueFrom combination.
6. The issue found in the point above is fixed in https://github.com/jupyterhub/kubespawner/pull/426
7. Documentation is also added in this PR to use singleuser.extraEnv with this new syntax.

Relevant references for this tech breakdown are the [k8s api specs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core) and the [Python API for Kubernetes](https://github.com/kubernetes-client/python).

## Discussion

- @minrk considered the implementation options in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1103#issuecomment-454336736.
- @manics suggested the implicit syntax where `name: ...` is implied. This is now supported.

## Todo
- [x] Get KubeSpawner 0.13.0 releaased (https://github.com/jupyterhub/kubespawner/pull/429)
- [x] Bump the KubeSpawner version in the hub image we build in this Helm chart (https://travis-ci.org/github/jupyterhub/zero-to-jupyterhub-k8s/builds/724045064)

## Related
- Closing #1462 in favor of this as the added documentation regards the no longer needed workaround.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -308,20 +308,38 @@ properties:
         description: |
           Extra environment variables that should be set for the hub pod.
 
+          Environment variables are usually used to:
+            - Pass parameters to some custom code in `hub.extraConfig`.
+            - Configure code running in the hub pod, such as an authenticator or
+              spawner.
+
           ```yaml
           hub:
             extraEnv:
-              MY_ENV_VARS_NAME: "my env vars value"
+              # short notation, only for hardcoded string values
+              MY_ENV_VARS_NAME1: "my env var value 1"
+
+              # long notation, Kubernetes native syntax
+              dummyKey1:
+                name: HUB_NAMESPACE
+                valueFrom:
+                  fieldRef:
+                    fieldPath: metadata.namespace
+              dummyKey2:
+                name: PREFIXED_HUB_NAMESPACE
+                value: "my-prefix-$(HUB_NAMESPACE)"
+              dummyKey3:
+                name: SECRET_VALUE
+                valueFrom:
+                  secretKeyRef:
+                    name: my-k8s-secret
+                    key: password
           ```
 
-          **NOTE**: We still support this field being a list of
-          [EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core)
-          objects as well.
-
-          These are usually used in two circumstances:
-            - Passing parameters to some custom code specified with `extraConfig`
-            - Passing parameters to an authenticator or spawner that can be directly customized
-              by environment variables (rarer)
+          In the long notation, you can provide any [Kubernetes
+          EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core)
+          specification, and the name of the key (`dummyKey1`) will have no
+          impact.
       extraConfig:
         type: object
         description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -313,7 +313,7 @@ properties:
             - Configure code running in the hub pod, such as an authenticator or
               spawner.
 
-          String literals with $(ENV_VAR_NAME) will be expanded by Kubelet which
+          String literals with `$(ENV_VAR_NAME)` will be expanded by Kubelet which
           is a part of Kubernetes.
 
           ```yaml
@@ -896,10 +896,10 @@ properties:
         description: |
           Extra environment variables that should be set for the user pods.
 
-          String literals with $(ENV_VAR_NAME) will be expanded by Kubelet which
+          String literals with `$(ENV_VAR_NAME)` will be expanded by Kubelet which
           is a part of Kubernetes. Note that the user pods will already have
           access to a set of environment variables that you can use, like
-          JUPYTERHUB_USER and JUPYTERHUB_HOST. For more information about these
+          `JUPYTERHUB_USER` and `JUPYTERHUB_HOST`. For more information about these
           inspect [this source
           code](https://github.com/jupyterhub/jupyterhub/blob/cc8e7806530466dce8968567d1bbd2b39a7afa26/jupyterhub/spawner.py#L763).
 
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -313,33 +313,34 @@ properties:
             - Configure code running in the hub pod, such as an authenticator or
               spawner.
 
+          String literals with $(ENV_VAR_NAME) will be expanded by Kubelet which
+          is a part of Kubernetes.
+
           ```yaml
           hub:
             extraEnv:
-              # short notation, only for hardcoded string values
+              # basic notation (for literal values only)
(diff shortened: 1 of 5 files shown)


jupyterhub/schema.yaml
```diff
@@ -308,20 +308,38 @@ properties:
...
+                    fieldPath: metadata.namespace
+              dummyKey2:
+                name: PREFIXED_HUB_NAMESPACE
+                value: "my-prefix-$(HUB_NAMESPACE)"
+              dummyKey3:
+                name: SECRET_VALUE
+                valueFrom:
+                  secretKeyRef:
+                    name: my-k8s-secret
+                    key: password
```
manics, line 336: I can see why you've done it this way, but my first instinct is it's confusing. Environment variables are already key-value pairs so adding the extra dummy keys seems unnecessary. Is there ever a case where the `name` key would not exist? If not what do you think of:
```yaml
              PREFIXED_HUB_NAMESPACE:
                value: "my-prefix-$(HUB_NAMESPACE)"
              SECRET_VALUE:
                valueFrom:
                  secretKeyRef:
                    name: my-k8s-secret
                    key: password
```
It doesn't match the k8s spec for an env-var, but neither does requiring a dict of env-vars instead of a list.

consideRatio: I considered this and felt that the most important principle to follow was to stick with valid k8s specifications. A compromise could be to accept leaving out the `name: value` field and automatically populating it using the top level key name and let the template rendering fail if a `name: something` field was specified that mismatched the top level key name.

What do you think about that approach? I want to ensure there isn't any confusion about what takes priority etc down the line.

consideRatio: Syntax updated. The `name: something` will take precedence and the top level key name is a fallback now.

jupyterhub/schema.yaml
```diff
@@ -308,20 +308,39 @@ properties:
         description: |
           Extra environment variables that should be set for the hub pod.
 
+          Environment variables are usually used to:
+            - Pass parameters to some custom code in `hub.extraConfig`.
+            - Configure code running in the hub pod, such as an authenticator or
+              spawner.
+
+          String literals with $(ENV_VAR_NAME) will be expanded by Kubelet which
```
manics, line 316:
```suggestion
          String literals with `$(ENV_VAR_NAME)` will be expanded by Kubelet which
```

jupyterhub/schema.yaml
```diff
@@ -872,6 +891,43 @@ properties:
         description: |
           Deprecated and no longer does anything. Use the user-scheduler instead
           in order to accomplish a good packing of the user pods.
+      extraEnv:
+        type: object
+        description: |
+          Extra environment variables that should be set for the user pods.
+
+          String literals with $(ENV_VAR_NAME) will be expanded by Kubelet which
```
manics, line 899:
```suggestion
          String literals with `$(ENV_VAR_NAME)` will be expanded by Kubelet which
```

jupyterhub/schema.yaml
```diff
@@ -872,6 +891,43 @@ properties:
...
+          is a part of Kubernetes. Note that the user pods will already have
+          access to a set of environment variables that you can use, like
+          JUPYTERHUB_USER and JUPYTERHUB_HOST. For more information about these
```
manics, line 902:
```suggestion
          `JUPYTERHUB_USER` and `JUPYTERHUB_HOST`. For more information about these
```

manics (commented):
Thanks, I think this is a good compromise!

consideRatio:
Thanks @manics!

consideRatio:
I consider this PR now ready for merge. @yuvipanda since you just looked at the KubeSpawner equivalent PR, perhaps you could have a look at this as well? Thank you @manics and @yuvipanda for the attention! :heart:

yuvipanda:
Thanks a lot for working on this, @consideRatio!

PR #1758, opened 2020-08-27
https: Only expose port 443 if we really have HTTPS on

### LINKED ISSUE #1637: AWS LoadBalancer health checks failing when using release 0.9.0 but not 0.8.2

yuvipanda:
For HTTPS, we want to expose port 443 on our service, but
only if it is 'turned on'. Unfortunately, in v 0.9, the
default was set to 'enabled' but it didn't act as enabled
unless hostnames were actually set. This exposed port 443
in proxy-public regardless of wether we were actually
set up for HTTPS or not. This caused weird issues with
AWS, since it was health checking all exposed ports. Since
we didn't have automatic HTTPS setup (since that needs
hostnames), this broke all installs on AWS!

This tries to be backwards compatible change, exposing
port 443 only if we *really* have https running. This
should fix issues with AWS, and be more 'correct'.

Manual HTTPS doesn't actually need hosts lists to be set,
so we remove that from our documentation.

We also re-introduce 'https.enabled' in our documentation -
manual action (DNS) is needed to enable HTTPS, so this
being explicit is IMO a good thing.

Fixes #1637

### DIFF
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -1,8 +1,10 @@
 {{- $HTTPS := .Values.proxy.https.enabled -}}
-{{- $autoHTTPS := and $HTTPS (and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
-{{- $offloadHTTPS := and $HTTPS (eq .Values.proxy.https.type "offload") -}}
+# Automatic HTTPS via Let's Encrypt requires list of hostnames to be turned on
+{{- $autoHTTPS := and $HTTPS and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
+# Manual HTTPS with user provided certs does not require list of hostnames
 {{- $manualHTTPS := and $HTTPS (eq .Values.proxy.https.type "manual") -}}
-{{- $manualHTTPSwithsecret := and $HTTPS (eq .Values.proxy.https.type "secret") -}}
+# Offloading HTTPS to external LoadBalancer does not require list of hostnames either
+{{- $offloadHTTPS := and $HTTPS (eq .Values.proxy.https.type "offload") -}}
 apiVersion: v1
 kind: Service
 metadata:
@@ -40,9 +42,12 @@ spec:
     {{- end }}
     release: {{ .Release.Name }}
   ports:
-    {{- if $HTTPS }}
+    {{- if or $autoHTTPS (or $manualHTTPS $offloadHTTPS) }}
     - name: https
       port: 443
+      # When HTTPS is handled outside our helm chart, pass traffic
+      # coming in via port 443 to port 80. Termination happens
+      # outside of kubernetes somehow
       {{- if $offloadHTTPS }}
       targetPort: http
       {{- else }}
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -1,6 +1,6 @@
 {{- $HTTPS := .Values.proxy.https.enabled -}}
 # Automatic HTTPS via Let's Encrypt requires list of hostnames to be turned on
-{{- $autoHTTPS := and $HTTPS and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
+{{- $autoHTTPS := and $HTTPS (and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
 # Manual HTTPS with user provided certs does not require list of hostnames
 {{- $manualHTTPS := and $HTTPS (eq .Values.proxy.https.type "manual") -}}
 # Offloading HTTPS to external LoadBalancer does not require list of hostnames either
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -182,7 +182,7 @@ proxy:
     enabled: true
     minAvailable: 1
   https:
-    enabled: true
+    enabled: false
     type: letsencrypt
     #type: letsencrypt, manual, offload, secret
     letsencrypt:
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -82,8 +82,6 @@ There are two ways to specify your manual certificate, directly in the config.ya
     proxy:
       https:
         enabled: true
-        hosts:
-          - <your-domain-name>
         type: manual
         manual:
           key: |
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -41,6 +41,7 @@ changes to your `config.yaml` file:
    ```yaml
    proxy:
      https:
+       enabled: true
        hosts:
          - <your-domain-name>
        letsencrypt:
@@ -80,6 +81,7 @@ There are two ways to specify your manual certificate, directly in the config.ya
     ```yaml
     proxy:
       https:
+        enabled: true
         hosts:
           - <your-domain-name>
         type: manual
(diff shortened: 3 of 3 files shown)


jupyterhub/values.yaml
```diff
@@ -182,7 +182,7 @@ proxy:
     enabled: true
     minAvailable: 1
   https:
-    enabled: true
+    enabled: false
```
consideRatio, line 185: I think this makes sense as a default, but is something that probably breaks peoples configurations which left it out given our past documentation did so.

Hmmm, should we perhaps mitigate this? We could for example make template rendering fail using Helm's `fail` function given a certain condition, or we could provide a warning in NOTES.txt. For example, that hosts is specified and `https.enabled` is false, that would be a typical indication of an issue. Another one would be to not set a default value for this at all in values.yaml, and if it isn't rendering to be true or false but instead to be a blank string we assume it to be true...

Hmm... I'm not sure. What do you think about this potential need to help users transition?

jupyterhub/templates/proxy/service.yaml
```diff
@@ -1,8 +1,10 @@
 {{- $HTTPS := .Values.proxy.https.enabled -}}
-{{- $autoHTTPS := and $HTTPS (and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
-{{- $offloadHTTPS := and $HTTPS (eq .Values.proxy.https.type "offload") -}}
+# Automatic HTTPS via Let's Encrypt requires list of hostnames to be turned on
+{{- $autoHTTPS := and $HTTPS and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
```
consideRatio, line 3: Syntax error due to a mismatch of parenthesis.

```suggestion
{{- $autoHTTPS := and $HTTPS (and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
```

consideRatio:
I created a PR to merge into this PR, see https://github.com/yuvipanda/zero-to-jupyterhub-k8s/pull/1.

jupyterhub/values.yaml
consideRatio: I think it would be fine to make it a breaking change, where we notify users of in the release notes, saying they need to avtively set it to enabled if they had it implicitly set enabled before.

I think it would also be fine to let one z2jh release look for a dependency on the default truthiness and actively fail template rendering if we think it would crash some configurations.

minrk:
Thanks folks!

yuvipanda:
Turns out this wasn't actually backwards compatible somehow. If I don't set `https.enabled` to `true` explicitly, it strips HTTPS away from my installs!

consideRatio:
@yuvipanda yes =/ I considered that in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1758#discussion_r481009825

Issue #1760, opened 2020-08-31
insufficient pods and the interaction with the hook-image-puller daemonset


consideRatio:
### Background

On GKE for example, one is limited to 110 pods per node no matter what configuration option is used. Other k8s deployments will have different limits but still have limits.

We have some challenges with regards to this limitation and the system to automatically download new images ahead of time on certain nodes. We do it by creating a deamonset that creates one pod per node. The hook puller daemonset is created temporary with a Helm hook as part of chart upgrades to ensure the new images is available before the actual chart upgrade. The continuous puller daemonset is created as part of the chart to ensure newly added nodes get the images.

### Goal outcomes
1. To avoid the [upgrade failures](https://github.com/neurohackademy/nh2020-jupyterhub/issues/86) following having a hook daemonset trying to create a pod on a node capped by a pod limit.
2. To avoid using unnecessary pods.

### Solution brainstorming
I'm not sure, but we could consider:
1. changing the pod priority which influences scheduling order of pending pods and could potentially get pods evicted if they have a negative priority if i recall correctly.
2. making the hook puller not want to schedule on nodes that have full capacity somehow, i think this may be hard though.
3. modifying the hook-image-awaiter that inspects the hook pre-puller daemonset to see if its pods are ready and avoid waiting for pods in a pending state that hasn't been scheduled to avoid this kind of reliability issue.

I think I like option 3 the most, its quite simple logic update in the image awaiter pod. It will need to introduce some delay if pods are pending though as they will always be pending the initial second they are scheduled or so I figure. This tackles the both goal 1 and 2 for the hook-image-puller daemonset, but would fail to address the second goal for the continuous image puller daemonset. Perhaps that daemonset should get a placeholder kind of priority as well?

### Suggested changes
1. We update the logic of the hook-image-awaiter go binary we have developed to not just wait for the number of pods created to all be ready, but to instead wait mindful that some pods may get stuck in pending and accept that.
2. We update the continuous image puller daemonsets pod priority to be like the user placeholders if scheduling.podPriority is enabled, which allows pre-pulling to run only on nodes which isn't full of users or other pods.

Ping @yuvipanda as I know you are already mindful of these issues, what do you think about these suggested changes to address them?

yuvipanda:
Thanks for writing this up, @consideRatio.

If my understanding is correct, what's happening here is:

1. Daemonset *schedules* pods on all nodes
[...]

PR #1762, opened 2020-09-01
Make continuous image puller pods evictable

consideRatio:
This was discussed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1760. The gist of this PR which was only part of the discussion, is to allow the continuous image puller pods to be evictable in order to prefer letting a user pod schedule than a puller pod if the node becomes full. Sometimes a node can become full because of the sheer number of pods as k8s limits this as there is a limit of IPs for each node.﻿

### DIFF
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -33,13 +33,19 @@ spec:
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


yuvipanda:
I wanted to check if daemonset pods will be treated differently wrt eviction, and found they [will not be](https://github.com/kubernetes/kubernetes/issues/76930) so this looks great to me!

consideRatio:
Thanks @yuvipanda!

PR #1763, opened 2020-09-01
hook-image-puller: -pod-scheduling-wait-duration flag added for reliability during helm upgrades

### LINKED ISSUE #1760: insufficient pods and the interaction with the hook-image-puller daemonset

consideRatio:
## PR Summary

This PR address a risk of `helm upgrade` failing for a bad reason - that the hook-image-puller pods fail to schedule because the nodes cannot accept more pods on them.

### How `helm upgrade` currently can fail for a bad reason

When `helm upgrade` is run, certain k8s resources are rendered and deployed before others using [Chart hooks](https://helm.sh/docs/topics/charts_hooks/). We deploy a `hook-image-awaiter` Pod and a `hook-image-puller` daemonset. The `hook-image-awaiter` pod awaits the pods of the `hook-image-puller` daemonset to become ready, as they do when they have been scheduled and pulled all images in dummy init containers.one node

The failure we can avoid is when the `hook-image-awaiter` pod get stuck waiting for one of the image puller pods that fails to schedule on a node, because a node is out of capacity to schedule more pods - probably mostly because a maximum pod per node constraint.

GKE has a max 110 pods per node for example, and there is no workaround for this limit as far as I know.

### How this PR address the issue

This PR updates the logic of the `hook-image-awaiter` pod's binary which we have written ourselves in Go. Before this PR it is waiting for all image puller pods to scheduled, start, and get the main container running/ready. But in this PR, it is made to await only the pods that actually have scheduled, but only to do that after an initial duration which is configurable with a `-pod-scheduling-wait-duration`.

We can accomplish this by using the status set on the hook-image-puller DaemonSet resource that describes `CurrentNumberScheduled`, `DesiredNumberScheduled`, and `NumberReady`.

### Configuration added

This duration is configurable with `prePuller.hook.podSchedulingWaitDuration` and defaults to `-1`. It can be configured to be an infinite duration with `-1`.

### DIFF
--- a/images/image-awaiter/main.go
+++ b/images/image-awaiter/main.go
@@ -85,8 +85,8 @@ func main() {
 	flag.StringVar(&daemonSet, "daemonset", "hook-image-puller", "The name DaemonSet that will perform image pulling")
 	var debug bool
 	flag.BoolVar(&debug, "debug", false, "Communicate through a 'kubectl proxy --port 8080' setup instead.")
-	var strictCheckDuration int
-	flag.IntVar(&strictCheckDuration, "strict-check-duration", 10, "Duration of seconds to wait for the desired number of scheduled pods to be ready until transitioning to wait for the currently scheduled pods to be ready instead. Set to -1 for an infinite duration.")
+	var podSchedulingWaitDuration int
+	flag.IntVar(&podSchedulingWaitDuration, "pod-scheduling-wait-duration", 10, "Duration of seconds to await the desired number of scheduled pods to become ready until transitioning to awaiting the currently scheduled pods to become ready instead. Set to -1 for an infinite duration.")
 
 	flag.Parse()
 
@@ -110,8 +110,8 @@ func main() {
 			log.Fatal(err)
 		}
 
-		strictCheck := strictCheckDuration == -1 || i < strictCheckDuration
-		if areDaemonSetPodsReady(ds, strictCheck) {
+		waitForPodsToSchedule := podSchedulingWaitDuration == -1 || i < podSchedulingWaitDuration
+		if areDaemonSetPodsReady(ds, waitForPodsToSchedule) {
 			log.Printf("Image download on nodes awaited successfully: shutting down!")
 			break
 		}
--- a/images/image-awaiter/main.go
+++ b/images/image-awaiter/main.go
@@ -4,9 +4,8 @@
 // started when this job starts. When all images are pulled, this job exits.
 
 /*
-K8s API options - currently using 1.9
-- K8s 1.8 API: curl http://localhost:8080/apis/apps/v1beta2/namespaces/<ns>/demonsets/<ds>
-- K8s 1.9 API: curl http://localhost:8080/apis/apps/v1/namespaces/<ns>/demonsets/<ds>
+K8s API access of relevance
+- curl http://localhost:8080/apis/apps/v1/namespaces/<ns>/demonsets/<ds>
 */
 
 package main
@@ -86,6 +85,8 @@ func main() {
 	flag.StringVar(&daemonSet, "daemonset", "hook-image-puller", "The name DaemonSet that will perform image pulling")
 	var debug bool
 	flag.BoolVar(&debug, "debug", false, "Communicate through a 'kubectl proxy --port 8080' setup instead.")
+	var strictCheckDuration int
+	flag.IntVar(&strictCheckDuration, "strict-check-duration", 10, "Duration of seconds to wait for the desired number of scheduled pods to be ready until transitioning to wait for the currently scheduled pods to be ready instead. Set to -1 for an infinite duration.")
 
 	flag.Parse()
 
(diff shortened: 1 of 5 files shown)


images/image-awaiter/main.go
```diff
@@ -86,6 +85,8 @@ func main() {
 	flag.StringVar(&daemonSet, "daemonset", "hook-image-puller", "The name DaemonSet that will perform image pulling")
 	var debug bool
 	flag.BoolVar(&debug, "debug", false, "Communicate through a 'kubectl proxy --port 8080' setup instead.")
+	var strictCheckDuration int
```
yuvipanda, line 88: What do you think of a slightly more descriptive name, like 'schedule-wait-time'? with just `strict-check-duration` it's unclear what the check is for.

jupyterhub/values.yaml
```diff
@@ -355,6 +355,7 @@ prePuller:
     image:
       name: jupyterhub/k8s-image-awaiter
       tag: 'set-by-chartpress'
+    strictCheckDuration: 10
```
yuvipanda, line 358: My suggestion is to keep this at -1, which is the current default. Then we can run this with a 10s timeout in production for a couple weeks, and then change the default if that works. How does that sound?

yuvipanda (changes requested):
Minor suggestions, but otherwise LGTM. Thanks for working on this, @consideRatio

GeorgianaElena (approved):
Everything looks super nice @consideRatio :+1: :tada: :heart: 

Is it possible for other pods to get scheduled after the `strict-check-duration` runs out and the `hook-image-awaiter` gets "unblocked"?

jupyterhub/values.yaml
consideRatio: Hmmm yes we can do that, but i consider it a bug to fail as it currently is where a full node can make a deployment crash - due to that i believe it could make sense to go for the fix by default.

Although, this PR could introduce some other bug, so then its ansafeguard for not making it worse. Sure! Going for that

consideRatio:
@yuvipanda @GeorgianaElena thank you for your reviews :heart: :tada:

@GeorgianaElena yes a pod that was pending for say 30 seconds can still get scheduled. The scheduling of these pods are done by a kubernetes pod scheduler and isn't influenced by the image-awaiter pod which only monitors the status of the daemonset object, which in turn contains information about: the number of scheduled pods (_scheduled_), the number of pods that should be scheduled (_desired_), the number of ready pods (_ready_).

So assuming a strict-check-duration of 10, what could happen is...

1. After 5 seconds --- _desired_: 5, _scheduled_: 4, _ready_: 4 --- keep waiting
2. After 9 seconds --- _desired_: 5, _scheduled_: 5, _ready_: 4 --- keep waiting
3. After 10 seconds --- image-awaiter will now accept scheduled==ready rather than desired==ready
4. After 15 seconds --- _desired_: 6, _scheduled_: 5, _ready_: 4 --- keep waiting
5. After 30 seconds --- _desired_: 6, _scheduled_: 5, _ready_: 5 --- completed!

images/image-awaiter/main.go
consideRatio: I went with `pod-scheduling-wait-duration`

PR #1764, opened 2020-09-01
Minimal and explicit resource requests for image-puller pods

### LINKED ISSUE #1761: Possibility to have resource requests/limits set on all our containers
To specify resource quotas on all our containers or provide an ability to do so. There are multiple reasons for this.

1. If we don't do it, we may end up with 100m CPU for a pause pod for example, which is a waste.
2. If we don't allow configuration of the resources, someone using a resourcequota may fail to use the Helm chart involving these pods because its may be required to have them explicitly set, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1251.

consideRatio:
If a k8s cluster has a LimitRange resource, it may end up for example
allocating 100m CPU for containers which is a waste of resources. This
would be a waste as the running container only does `sh -c "Done!"`
anyhow. While there is image pulling happening, that is done before the
actual container starts so its worth noting we don't need any resources
specified for that.

Another reason for specifying resources, and making them configurable,
is that sometimes there is a ResourceQuota resource that requires the
resources to be explicitly specified on pods.

Related:
- Default CPU requests - https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/
- LimitRange https://kubernetes.io/docs/concepts/policy/limit-range/
- ResourceQuota - https://kubernetes.io/docs/concepts/policy/resource-quotas/

Closes #1761

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -350,6 +350,13 @@ scheduling:
 
 prePuller:
   annotations: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
+    limits:
+      cpu: 10m
+      memory: 10Mi
   hook:
     enabled: true
     image:
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1181,6 +1181,16 @@ properties:
         description: |
           Annotations to apply to the hook and continous image puller pods. One example use case is to
           disable istio sidecars which could interfere with the image pulling.
+      resources:
+        type: object
+        description: |
+          These are standard Kubernetes resources with requests and limits for
+          cpu and memory. They will be used on the containers in the pods
+          pulling images. These should be set extremely low as the containers
+          shut down directly or is a pause container that just idles.
+
+          They were made configurable as usage of ResourceQuota may require
+          containers in the namespace to have explicit resources set.
       hook:
         description: |
           See the [*optimization
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -72,6 +72,8 @@ spec:
             - /bin/sh
             - -c
             - echo "Pulling complete"
+          resources:
+            {{- .Values.prePuller.resources | toYaml | trimSuffix "\n" | nindent 12 }}
         {{- range $k, $container := .Values.singleuser.profileList }}
         {{- if $container.kubespawner_override }}
         {{- if $container.kubespawner_override.image }}
@@ -81,6 +83,8 @@ spec:
             - /bin/sh
             - -c
             - echo "Pulling complete"
+          resources:
+            {{- .Values.prePuller.resources | toYaml | trimSuffix "\n" | nindent 12 }}
         {{- end }}
         {{- end }}
         {{- end }}
@@ -94,6 +98,8 @@ spec:
             - /bin/sh
             - -c
             - echo "Pulling complete"
+          resources:
+            {{- .Values.prePuller.resources | toYaml | trimSuffix "\n" | nindent 12 }}
         {{- end }}
         {{- range $k, $v := .Values.prePuller.extraImages }}
         - name: image-pull-{{ $k }}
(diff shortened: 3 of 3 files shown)


jupyterhub/values.yaml
```diff
@@ -350,6 +350,13 @@ scheduling:
 
 prePuller:
   annotations: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
+    limits:
+      cpu: 10m
```
yuvipanda, line 358:
```suggestion
      cpu: 1m
```

jupyterhub/values.yaml
```diff
@@ -350,6 +350,13 @@ scheduling:
...
+      memory: 10Mi
```
yuvipanda, line 359:
```suggestion
      memory: 1Mi
```

yuvipanda (commented):
Yeah, the limitrange has bit me more than once.

Does setting the request to '0' work well? I remember seeing that it would count it as unset and set request == limit, but not sure.

Small suggestion to reduce the limit even more!

jupyterhub/values.yaml
consideRatio: ## About 0 requests

These were the resources reported back from `kubectl get pod -o yaml` after trying to set the resources like described. I suspect that KubeSpawner may have a logic checking for a truthy value of the cpu request and leaving it out if not set or similar could be an issue though.

```yaml
    resources:
      limits:
        cpu: 10m
        memory: 10Mi
      requests:
        cpu: "0"
        memory: "0"
```

Relevant to this is that I've now learned that `LimitRange` resources can specify a `maxLimitRequestRatio` which would enforce the resource type influenced to not have any field set to zero. Do we want to avoid this by having a request of 1?

## Reducing limits further

We can do this, but are you sure that the containers will stay below 1Mi of memory usage? My main concern is that I'll have concern about this when debugging some other issue and spend time thinking about if this narrow limit could have caused problems.

I decided to try using a extremely narrow limit of 1m / 10Mi and found that in my k8s cluster with Calico etc the pod with 1m / 1Mi limits got stuck in `ContainerCreating` while the 10m / 10Mi limited pod got started successfully.

image

The issues were the following on the 1m / 10 Mi pod.

```
Events:
  Type     Reason                  Age                   From                                                Message
  ----     ------                  ----                  ----                                                -------
  Normal   Scheduled               3m18s                 default-scheduler                                   Successfully assigned default/tmp-6f699cc94f-62sm6 to gke-nh-2020-core-standard-2-b80a0890-kfvt
  Warning  FailedCreatePodSandBox  104s (x3 over 2m47s)  kubelet, gke-nh-2020-core-standard-2-b80a0890-kfvt  Failed create pod sandbox: rpc error: code = Unknown desc = failed to start sandbox container for pod "tmp-6f699cc94f-62sm6": Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:319: getting the final child's pid from pipe caused \"read init-p: connection reset by peer\"": unknown
  Warning  FailedCreatePodSandBox  33s (x2 over 68s)     kubelet, gke-nh-2020-core-standard-2-b80a0890-kfvt  Failed create pod sandbox: rpc error: code = Unknown desc = failed to start sandbox container for pod "tmp-6f699cc94f-62sm6": Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"read init-p: connection reset by peer\"": unknown
  Normal   SandboxChanged          32s (x5 over 2m46s)   kubelet, gke-nh-2020-core-standard-2-b80a0890-kfvt  Pod sandbox changed, it will be killed and re-created.
```

I think the newly introduced [Pod Overhead](https://kubernetes.io/docs/concepts/configuration/pod-overhead/) could mitigate this and allow us to have extremely low limits, but since that is a lot of extra infrastructure, I suggest we stick with the current 10 m / 10 Mi to be safe.

consideRatio: @yuvipanda what do you think, merge as it is?

yuvipanda:
<3 for checking through this, @consideRatio. I always appreciate how thorough you are.

jupyterhub/values.yaml
yuvipanda: I think we should actually remove the limits. On my cluster (AKS), setting this to 10Mi causes the pods to constantly crash as well. Why don't we just remove the limits?

consideRatio: @yuvipanda ah yes deal. Are we unlimited by leaving them out?

Issue #1410, opened 2019-09-13
When&where is HUB_SERVICE_HOST set?


qzchenwl:
I saw those environment variables in jupyterhub_config.py but cannot find out how is this set.

Similar environment variables:
```
HUB_PORT
HUB_PORT_8081_TCP
HUB_PORT_8081_TCP_ADDR
HUB_PORT_8081_TCP_PORT
HUB_PORT_8081_TCP_PROTO
HUB_SERVICE_HOST
HUB_SERVICE_PORT
PROXY_API_PORT
PROXY_API_PORT_8001_TCP
PROXY_API_PORT_8001_TCP_ADDR
PROXY_API_PORT_8001_TCP_PORT
PROXY_API_PORT_8001_TCP_PROTO
PROXY_API_SERVICE_HOST
PROXY_API_SERVICE_PORT
PROXY_PUBLIC_PORT
PROXY_PUBLIC_PORT_80_TCP
PROXY_PUBLIC_PORT_80_TCP_ADDR
PROXY_PUBLIC_PORT_80_TCP_PORT
PROXY_PUBLIC_PORT_80_TCP_PROTO
PROXY_PUBLIC_SERVICE_HOST
PROXY_PUBLIC_SERVICE_PORT
PROXY_PUBLIC_SERVICE_PORT_HTTP
```

consideRatio:
I've stumbled on this as well and our code base could certainly benefit from some inline comment about this. It is done by kubernetes and described [here](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables).

PR #1766, opened 2020-09-01
Leave a comment about where HUB_SERVICE_* values come from

mriedem:
While wondering where HUB_SERVICE_HOST and HUB_SERVICE_PORT values
come from I stumbled across this comment [1] so I'm adding an inline
comment to the code.

[1] https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1410#issuecomment-531130122

Related #1410


consideRatio (approved):
Excellent thanks!

Issue #1732, opened 2020-07-24
Default settings of the Hub pod's readinessProbe are problematic


consideRatio:
We have a k8s readinessProbe defined, which means that we declare the hub only to be "Ready" to receive traffic when it has passed the test of responding to a HTTP request. This is only used as an indicator of successfully starting up currently. For example, the CI tests should not try send traffic to the hub pod before the hub pod has started.

> **NOTE:** that there is also a livenessProbe which if that fails, it would trigger the container in the pod to restart.

To make a pod not-ready at a later time doesn't make much sense - we don't have a pod to fall back to, and we cannot have multiple instances running. Due to this, the setting of the readinessProbe should be such that the hub pod becomes ready when it should be, but that it doesn't ever fail because it was slow for a while.

Instead of disabling the readinessProbe entirely, I think we should try to make it not fail, this can be done using failureThreshold for example: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#probe-v1-core

## 503 (Service unavailable) errors

These happen when traffic is routed to a k8s Service that direct traffic to pods without any pod in a Ready status, which is influenced by the readinessProbe.

[![image](https://user-images.githubusercontent.com/3837114/88401651-4c92d300-cdca-11ea-9ce9-9457b8dc1713.png)](https://snapshot.raintank.io/dashboard/snapshot/oA6Kyl3zZZZ2Oh5ZifMc7KEn6WIZcbEg?orgId=2)

## Current default configuration and its use
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0c88d30d32f00a27c491a50986169bdf77e86f4d/jupyterhub/values.yaml#L90-L97

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/d080b1d24bc1eb4d8faac0726f5d5ea87605459b/jupyterhub/templates/hub/deployment.yaml#L224-L246

yuvipanda:
lololol, I just made #1800 and came here to file this exact issue :D 



> To make a pod not-ready at a later time doesn't make much sense - we don't have a pod to fall back to, and we cannot have multiple instances running. Due to this, the setting of the readinessProbe should be such that the hub pod becomes ready when it should be, but that it doesn't ever fail because it was slow for a while.

[...]

PR #1767, opened 2020-09-03
Add additional liveness and readiness probe properties

rmoe:
This allows one to set failureThreshold and timeoutSeconds for the hub probes.

Related: #1732


consideRatio:
The provided default values in values.yaml are also the k8s default values, so this should not make any difference other than getting a bit more explicit output which now can be configured.

PR #1772, opened 2020-09-04
hub: Switch base image to latest LTS

yuvipanda:
18.04 is a while ago :)

This also brings a newer version of python, I believe

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 ## NOTE: This is a default and be overridden by chartpress using the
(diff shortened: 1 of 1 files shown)


yuvipanda:
Running dependency freezer didn't seem to make any difference.

consideRatio:
@yuvipanda, @manics initiated this kind of work for the JupyterHub project's Docker image as well. See https://github.com/jupyterhub/jupyterhub/pull/3156.

minrk:
I think we need to remove the pinning of libc-bin, etc. that should only be relevant for the old release

yuvipanda:
Thanks for the fix, @minrk! This LGTM

consideRatio:
Nice! :D

PR #1776, opened 2020-09-10
Add terraform resources to the community resources section

### LINKED ISSUE #1745: Links to Terraform Guides for Z2JH-K8s on AWS
### Background
[Terraform](https://www.terraform.io/) is a very useful tool for deploying cloud infrastructure. I've used it a fair bit on the Pangeo project for hackweeks support and infrastructure documentation (see the [ICESat-2 Hackweek 2020](https://github.com/ICESAT-2HackWeek/terraform-deploy/tree/master) and [OceanHackWeek 2020](https://github.com/oceanhackweek/ohw-terraform-deploy/tree/main) infrastructure repos). I think it could be a good way for new users to see and manipulate infrastructure in their quest to deploy JupyterHubs. I've had some conversation about this over on [this Discourse post](https://discourse.jupyter.org/t/contributing-a-terraform-guide-for-zero-to-jupyterhub-k8s/4830). In particular, @manics suggested:

> Having a set of Terraform instructions on a separate repository is very valuable, and since it’s in a separate repo you can use your judgement as to how complicated or configurable you want to make it.
> I think it’s best in a separate repo because unlike the existing guides it’s not just a set of commands to run, it involves Terraform files. Having the docs and example of files in their own repo makes it easier to test and update.

### Proposed change

My proposed change is to link two Terraform deployments in the [Z2JH Community Resources](https://zero-to-jupyterhub.readthedocs.io/en/latest/community/index.html). 

One of these [is already up](https://github.com/pangeo-data/terraform-deploy/tree/blog-post) and is the subject of a blog post I wrote. It is the configuration I've used for the VPC and EKS cluster that backs the hackweeks mentioned above. It also contains configuration for the `cluster-autoscaler`, autoscaling to zero, spot instancing, and an EFS drive with automatic home directory provisioning. It's quite opinionated and deploys a lot of different pieces, which can be confusing for new users. I recognize that it would be a bad template for a Z2JH option, but is still valuable as a real-world example.

The other Terraform deployment will be much more minimal, featuring the bare minimum that Terraform needs to configure the VPC and EKS cluster with two modules that I have found indispensable in making the previous deployment. I'm in the process of making this deployment. It will probably live in the same repo as the above deployment, but on a different branch.

### Alternative options

An alternative option would be to integrate that second deployment into the Z2JH-K8s docs. Mentioned above, we are currently looking to just link to the repo, but this is an option if there are strong opinions about it.

These repos could serve as a starting point for new users who want to use a cloud-agnostic, industry tool for deploying infrastructure. I recognize that the example is exclusively AWS at this time. I hope to branch out to other cloud providers in time or inspire others to make contributions to reach this goal at the `terraform-deploy` repo.

I have worked with rST / Sphinx a bit in developing documentation for `hubploy`. I am willing to do all the legwork in documenting these links and some descriptions in the Community Resources section once the second deployment is ready (maybe next week?).

salvis2:
I also put a commit in to fix formatting in the same file. Tested locally and it looked all right, I'd love for there to be more whitespace between some the first sublist and the list I just put in, but I'm not sure how to enforce that in rST.

Closes #1745 .

### DIFF
--- a/doc/source/community/additional-resources.rst
+++ b/doc/source/community/additional-resources.rst
@@ -15,7 +15,9 @@ Links to community project resources
 ------------------------------------
 
 * **Automatically deploy a JupyterHub on AWS** from the Space Telescope Software Institute
+
   * `github repository <https://github.com/spacetelescope/z2jh-aws-ansible>`_
+
   * `blog post <https://github.com/spacetelescope/z2jh-aws-ansible>`_
 
 * **Setup Kubernetes with Terraform**
--- a/doc/source/community/additional-resources.rst
+++ b/doc/source/community/additional-resources.rst
@@ -17,3 +17,17 @@ Links to community project resources
 * **Automatically deploy a JupyterHub on AWS** from the Space Telescope Software Institute
   * `github repository <https://github.com/spacetelescope/z2jh-aws-ansible>`_
   * `blog post <https://github.com/spacetelescope/z2jh-aws-ansible>`_
+
+* **Setup Kubernetes with Terraform**
+
+  * `terraform-deploy repository <https://github.com/pangeo-data/terraform-deploy>`_
+	This repo houses several example Kubernetes cluster deployments on AWS.
+
+  * `Terraform tutorial example <https://github.com/pangeo-data/terraform-deploy/tree/master/aws-examples/minimal-deployment-tutorial>`_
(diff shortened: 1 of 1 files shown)


salvis2:
For context, it renders like this:

z2jh-pr-render

yuvipanda:
This is great! Thank you :)

consideRatio:
Thank you @salvis2 !

PR #1778, opened 2020-09-12
Updates to user-scheduler's coupling to the kube-scheduler binary

### LINKED ISSUE #1773: kube-scheduler binary bump to 1.16 make user-scheduler in 1.15 k8s fail
#1699 introduced an breaking bug using this Helm chart on Kubernetes clusters versioned 1.15 and lower. The `user-scheduler` pods responsible for scheduling the k8s pods representing the users in a smart manner is now failing to schedule them to the k8s nodes with the following error message.

```
I0908 10:05:43.912366       1 reflector.go:158] Listing and watching *v1beta1.CSINode from k8s.io/client-go/informers/factory.go:134
E0908 10:05:43.914298       1 reflector.go:123] k8s.io/client-go/informers/factory.go:134: Failed to list *v1beta1.CSINode: csinodes.storage.k8s.io is forbidden: User "system:serviceaccount:jupyterhub:user-scheduler" cannot list resource "csinodes" in API group "storage.k8s.io" at the cluster scope
```

The error describes that the user-scheduler pod isn't allowed to ask the k8s api-server about `storage.k8s.io/v1beta/csinodes` resources.

To understand why this is, we need to understand what grants the user-scheduler pod permission to potentially do something like this in the first place.

1. The user-scheduler pod is configured to use a k8s ServiceAccount named `user-scheduler`.
2. The k8s ServiceAccount `user-scheduler` is coupled to have the Cluster permissions of...
   - the ClusterRole `system:kube-scheduler` through the ClusterRoleBinding `jupyterhub-user-scheduler-base`
   - the ClusterRole `jupyterhub-user-scheduler-complementary` through the ClusterRoleBinding `jupyterhub-user-scheduler-complementary`.
3. The permissions defined in ClusterRole `system:kube-scheduler` depends on the k8s version, its not part of this Helm chart. And, k8s 1.15 doesn't include the permission to inspect CSINode resources, while k8s 1.16's version of the `system:kube-scheduler` ClusterRole does.

Due to this, the solution as I see it, is to add the permission to inspect CSINode resources in the ClusterRole `jupyterhub-user-scheduler-complementary`. This is a one-liner fix right in this section, by adding `csinodes` next to `storageclasses`.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/cd1eff7a5f093453de4d5868d1c4148044c0db23/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml#L39-L42

---

/cc: @deepitpro and @ArijitBhattacharyya11 who have also seen the consequences of this issue - perhaps you could create a PR to fix this also? =) Following its merge, a new developer release version will be published herehttps://jupyterhub.github.io/helm-chart/ which we can upgrade to.

### Related

Closes #1773.

consideRatio:
Closes #1773 which also contains the findings that made me do what I summarize having done below.

## Important review
- [ ] We should review this on a modern k8s cluster (1.17+) and verify that the user-scheduler does indeed schedule pods on the most busy node even with the new configuration style. I suggest this is done after merge though.

## Summary of changes
- Fixes a bug introduced by bumping to kube-scheduler binary of version 1.16 relating to not having the permission to look for CSINodes resources.
- Bumping kube-spawner binary to the latest k8s version (1.19.1), but is able to fallback to 1.16.15) if the CSINode resource is not registered as a resource type in the k8s cluster.
- Configures the modern kube-scheduler binary in its new configuration format.
- Provides some notes as to when we can remove the fallback (when 1.17+ is assumed)

## More detailed summary of changes
- I made us mount the configuration as a file, which is what we need to do in the modern configuration format and it also made us no longer need to provide RBAC rules for inspecting configmaps.
- I made the new `plugins` configurable _for the advanced user_ wanting to try another scheduling policy.
- I opted to fully specify the RBAC permissions instead of complement what could end up missing by augmenting the `system:kube-scheduler` and `system:volume-scheduler` ClusterRole's as doing this felt easier to maintain.
- I removed a helm template function to get the current k8s version as a fallback for the kube-scheduler image tag which would be unsuitable to use and has not been used.


consideRatio:
I have the time to test this both on k8s 1.15 and on k8s 1.17 today if I merge this now, so I'll go ahead and do it.

consideRatio:
I've now reviewed this PR in a k8s 1.17 cluster, where it works but not yet good enough. It tends to schedule on one node over another, but I need to disable one more plugin that skews it back from only scheduling on the most resource intensive node.

Here is the scoring as indicated by the logs from the `user-scheduler` pod when scheduling a pod and one node had more pods scheduled on it than the other. As we can see, the `SelectorSpread` plugin scores against the scoring by `NodeResourcesMostAllocated` which we don't want. I'll disable it in another PR.

```
SelectorSpread             scores on pod1 => [{node1 27     } {node2 0      }]
NodeResourcesMostAllocated scores on pod1 => [{node1 18     } {node2 65     }]
ImageLocality              scores on pod1 => [{node1 0      } {node2 0      }]
InterPodAffinity           scores on pod1 => [{node1 0      } {node2 0      }]
NodeAffinity               scores on pod1 => [{node1 0      } {node2 0      }]
NodePreferAvoidPods        scores on pod1 => [{node1 1000000} {node2 1000000}]
PodTopologySpread          scores on pod1 => [{node1 200    } {node2 200    }]
TaintToleration            scores on pod1 => [{node1 100    } {node2 100    }]
```

PR #1780, opened 2020-09-16
Remove memory / cpu limits for pre-puller

yuvipanda:
This memory limit seems to include pod overhead, which
can be variable based on cluster config. In my cluster (AKS)
with Calico enabled, the memory limit was too low and kept
crashing these pods with very weird reasons - the usual
reasons aren't reported since it's not *our* process that
is causing OOM kills, but the cluster's overhead.

By removing the default resource limits, we prevent thi
from being a problem. Theoretically this would let the pods
have unbounded memory - but since we control the contents,
it wouldn't be the case.

We leave the explicit default requests be, for reasons
outlined in
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1764.
The capacity for admins to specify their own limits is
still in place, in case you want that for your cluster.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -358,9 +358,6 @@ prePuller:
     requests:
       cpu: 0
       memory: 0
-    limits:
-      cpu: 10m
-      memory: 10Mi
   hook:
     enabled: true
     image:


consideRatio:
Thank you @yuvipanda, this LGTM!

A pod may still crash due to these resource requests because the share of CPU allocated every 100ms relates is set proportional to the resource request, so I think this would be CPU starved if something else clogs all the CPU available. I still lean towards thinking it makes the most sense to keep the requests at 0 as thats the share of CPU we expect them to take in the long run.

Hmm... at the same time. We allow the continuous image puller pods to be evicted now, which makes it fine to request a bit to ensure some CPU time. Well. this PR is an improvement in the right direction!

PR #1781, opened 2020-09-16
Update to newer version of 'pause' container

yuvipanda:
Why not?

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -371,7 +371,7 @@ prePuller:
   extraImages: {}
   pause:
     image:
-      name: gcr.io/google_containers/pause
+      name: k8s.gcr.io/pause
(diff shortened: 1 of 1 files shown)


consideRatio:
I think k8s.gcr.io should be used now (`k8s.gcr.io/pause:3.2`), see an indication of this [in the 1.19 release notes](https://kubernetes.io/docs/setup/release/notes/#production-images-moved-to-community-control).

I fail to find a place to list the latest tags for k8s.gcr.io though, and your reference is the one I've used before.

jupyterhub/values.yaml
```diff
@@ -372,7 +372,8 @@ prePuller:
   pause:
     image:
       name: gcr.io/google_containers/pause
```
consideRatio, line 374:
```suggestion
      name: k8s.gcr.io/pause
```

PR #1782, opened 2020-09-21
user-scheduler: tweak modern configuration

consideRatio:
This is a followup on #1778 which didn't ensure that pods are scheduled on the most used node correctly because there were some remnant competing logic. This PR disables the competition that makes sense to give lower priority to than the `NodeResourcesMostAllocated` critiera.

See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1778#issuecomment-696079664 for a discussion on this PR.

As this PR is a complementary fix to #1778 recently merged, I'll go ahead and merge this directly if tests succeeds.﻿

### DIFF
--- a/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
@@ -22,8 +22,10 @@ data:
           {{- else }}
           score:
             disabled:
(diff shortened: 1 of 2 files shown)


consideRatio:
Tested to have worked out well on k8s 1.17 woop woop! :tada:

PR #1788, opened 2020-09-26
Fix legacy version in DigitalOcean Kubernetes setup doc

subwaymatch:
In the [Kubernetes on Digital Ocean](https://zero-to-jupyterhub.readthedocs.io/en/latest/digital-ocean/step-zero-digital-ocean.html) doc page, there are two minor issues.

image

The `doctl k8s cluster create` command is using an outdated Kubernetes version. Using `1.12.1-do.2` throws a 422 validation error.<br />
image
image

When creating a Kubernetes cluster through Digital Ocean's dashboard, only versions 1.16.14 and up seem to be supported.
image

I've replaced the legacy version with latest `1.18.8-do.0`.
<br />
The `doctl k8s cluster create` command is also missing a double quote at the end. I've added the missing double quote.

Confirmed by testing in Ubuntu 20.04:
image


manics (approved):
Thanks for helping us keep the docs up to date 😀

PR #1794, opened 2020-10-04
Support chart wide and pod specific config of imagePullSecrets

### LINKED ISSUE #981: Improve the imagePullSecrets documentation
We are missing an example of working configuration, and could perhaps now avoid duplicating the documentation within the hub's imagePullSecret and the singleuser's imagePullSecret.

**Suggested action:**
- visit schema.yaml and add an example of a working configuration
- update the hub.imagePullSecret to reference the singleuser.imagePullSecret

**The current generated documentation, based on the schema.yaml file:**
https://zero-to-jupyterhub-with-kubernetes.readthedocs.io/en/latest/reference.html#singleuser-imagepullsecret

### LINKED ISSUE #1465: Can't pull private image for SingleUser
I am using helm 0.9.0-alpha.1.  In pod/continuous-image-puller-2hlk8, it shows
```

 Normal  Scheduled  6m2s   default-scheduler                           Successfully assigned jupyterhub/continuous-image-puller-2hlk8 to aks-nodepool1-37655849-vmssXXXX3x
  Normal  Pulling    6m1s   kubelet, aks-nodepool1-37655849-vmssXXXX3x  Pulling image "<private_repo>.azurecr.io/jupyter-minimal-notebook:latest"
  Normal  Pulled     6m1s   kubelet, aks-nodepool1-37655849-vmssXXXX3x  Successfully pulled image "<private_repo>.azurecr.io/jupyter-minimal-notebook:latest"
  Normal  Created    6m     kubelet, aks-nodepool1-37655849-vmssXXXX3x  Created container image-pull-singleuser
  Normal  Started    6m     kubelet, aks-nodepool1-37655849-vmssXXXX3x  Started container image-pull-singleuser
  Normal
```

but when i use jupyther hub to spin up a user with similar private image <private_repo>.azurecr.io/jupyter-minimal-notebook:latest , i get 

```
  Normal   Pulling                 8s    kubelet, aks-nodepool1-37655849-vmssXXXX3x  Pulling image "<private_repo>.azurecr.io/jupyter-minimal-notebook:latest"
  Warning  Failed                  7s    kubelet, aks-nodepool1-37655849-vmssXXXX3x  Failed to pull image "<private_repo>.azurecr.io/jupyter-minimal-notebook:latest": rpc error: code = Unknown desc = Error response from daemon: Get https://<private_repo>.azurecr.io/v2/jupyter-minimal-notebook/manifests/latest: unauthorized: authentication required
  Warning  Failed                  7s    kubelet, aks-nodepool1-37655849-vmssXXXX3x  Error: ErrImagePull
  Normal   BackOff                 7s    kubelet, aks-nodepool1-37655849-vmssXXXX3x  Back-off pulling image "<private_repo>.azurecr.io/jupyter-minimal-notebook:latest"
  Warning  Failed                  7s    kubelet, aks-nodepool1-37655849-vmssXXXX3x  Error: ImagePullBackOff
```

In config.yml file, i have
```
singleuser:
  imagePullSecret: 
      name : my-existing-secret
```

### LINKED ISSUE #1467: Adding optional imagePullSecrets for proxy Image
Currently proxy image does not support imagePullSecrets, in case someone wants to use a custom image.

I have created #1391 which solves this and it passes the tests

### LINKED ISSUE #1504: Fix fragile configuration of pullPolicy for singleuser, and complement for the proxy pod.
#1426 added the ability to use an existing imagePullSecret, but didn't propagate it to the singleuser pods but only to the pre-puller pods. If someone would use a pulling policy of Always this would fail or if the imagepullers didnt pull the images to the nodes first it would also fail.

So, to avoid the fragility of that, we should also add the existing image pull secrets to the singleuser pods directly. 

Also, we should have this for the proxy pod as that would close #1467

consideRatio:
## Motivation
Before a k8s pod can starts its containers, _Kubelet_, a k8s component will download the images from the image registry. If this image registry requires credentials to access it, the Pod can specify `spec.imagePullSecrets`. If that is done, Kubelet will look in the referenced Kubernetes Secret resources and use the credentials within them to download the images.

This Helm chart both supports the creation of such k8s Secret with credentials, and the configuration of the hub pods and user pods `spec.imagePullSecrets` that reference the created Secret. What is missing though, is the same kind of configuration for all other pods.

This PR aims to resolve this once and for all by providing a chart wide configuration that augments all pods `spec.imagePullSecrets`, either with a manually created k8s Secret, or by a k8s Secret created by the Helm chart given configured credentials.

### Added

- `imagePullSecrets` - A place to specify manually created k8s Secret references for all pods.
- `imagePullSecret.create|registry|username|password` - A place to specify credentials for the creation of a k8s Secret that will be added to the list of all pods `spec.imagePullSecrets`.

### Updated

- `...image.pullPolicy` configuration is now available for all pods
- `...image.pullSecrets` configuration is now available for all pods

### Deprecated

- `hub.imagePullSecret` - Deprecated in favor of chart wide `imagePullSecret`
- `singleuser.imagePullSecret` - Deprecated in favor of chart wide `imagePullSecret`

## Implementation steps

- [x] Add config `imagePullSecret.create` etc, to create a k8s Secret
  resource with image registry credentials.
- [x] Define a helper function (`"jupyterhub.imagePullSecrets"`) that returns a
  JSON formatted list of Kubernetes Secret names or a blank string representing:
  the Secret created by `imagePullSecret`, the `imagePullSecrets`
  list, and the pod's specific config in `...image.pullSecrets`.
- [x] Enable all pods to use these credentials to pull images with the helper
  function.
- [x] Add a hard deprecation of `hub.imagePullSecret` and
  `singleuser.imagePullSecret`. A message will show asking the user to use
  `imagePullSecret` instead from now on.
- [x] Add config `imagePullSecrets` to allow users to specify manually
  created Kubernetes Secret's for all pods to make use of.
- [x] Add config `...image.pullSecrets` and `...image.pullPolicy` where it was
  missing for certain pods.
- [x] Update schema.yaml to reflect changes.
- [x] Update `jupyterhub_config.py` logic of the user pod's `imagePullPolicy`
  and `imagePullSecrets`
  - [x] Update KubeSpawner to support configuration where mutliple imagePullSecrets are specified, like in the k8s native specification. https://github.com/jupyterhub/kubespawner/pull/442
  - [x] Release KubeSpawner 0.14.0
  - [x] Bump KubeSpawner to 0.14.0 in this repo

Closes #981, Closes #1504, Closes #1467, Closes #1465


manics:
It's quite a big change to review without knowledge of exactly how this all works. Is it possible someone wants to pull their singleuser image from a different registry to the hub and proxy?

Do you have any idea how many users rely on a private registry? If it's a significant number maybe we should convert one of the existing CI jobs to test it?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -163,8 +163,17 @@ def camelCaseify(s):
...
+# (list), and singleuser.image.pullSecrets (list).
+image_pull_secrets = []
+if get_config('global.imagePullSecret.create') or get_config('global.imagePullSecret.enabled'):
+    image_pull_secrets.append('image-registry-credentials')
+if get_config('global.imagePullSecrets'):
+    image_pull_secrets.extend(get_config('global.imagePullSecrets'))
+if get_config('singleuser.image.pullSecrets'):
+    image_pull_secrets.extend(get_config('singleuser.image.pullSecrets'))
+if image_pull_secrets:
+    c.KubeSpawner.image_pull_secrets = image_pull_secrets
```
consideRatio, line 176: This is the logic that configures KubeSpawner in the same way that other pods are configured through a Helm chart helper function defined in `_helpers.tpl`, but KubeSpawner must be configured here with different syntax.

jupyterhub/schema.yaml
```diff
@@ -1,55 +1,21 @@
 title: Config
 type: object
 properties:
-  hub:
+  global:
```
consideRatio, line 4: `schema.yaml` in its entirety is mostly just documentation, but its also some adjustment to avoid repeating the same stuff over and over by using YAML anchors (`$`) and references (`*`) to the previously defined anchors.

This section could be reviewed in the PR review by RTD. See https://zero-to-jupyterhub--1794.org.readthedocs.build/en/1794/reference/reference.html#global

jupyterhub/templates/NOTES.txt
```diff
@@ -48,4 +48,16 @@ resources.
...
+{{- if .Values.hub.imagePullSecret.enabled }}
+{{ "HARD DEPRECATION: hub.imagePullSecret has renamed to global.imagePullSecret" | fail }}
+{{- end }}
+{{- end }}
+
+{{- if hasKey .Values.singleuser "imagePullSecret" }}
+{{- if .Values.singleuser.imagePullSecret.enabled }}
+{{ "HARD DEPRECATION: singleuser.imagePullSecret has renamed to global.imagePullSecret" | fail }}
+{{- end }}
+{{- end }}
```
consideRatio, line 63: This is the breaking changes that the user will be aware of if needed during Helm template rendering. This is because when this template renders, the `| fail` part will emit the message piped to it.

I've considered a sot deprecation, but the Helm templates logic to allow for that is messy, and we don't want to silently let users get stuck failing to pull images suddenly because their configuration isn't working any more, then its better to fail hard before any changes are made to k8s.

jupyterhub/templates/_helpers.tpl
```diff
@@ -172,12 +172,12 @@ component: {{ include "jupyterhub.componentLabel" . }}
 
     - https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
 */}}
-{{- define "jupyterhub.dockersingleuserconfigjson" -}}
-{{ include "jupyterhub.dockersingleuserconfigjson.yaml" . | b64enc }}
+{{- define "jupyterhub.dockerconfigjson" -}}
+{{ include "jupyterhub.dockerconfigjson.yaml" . | b64enc }}
 {{- end }}
```
consideRatio, line 177: The chart previously allowed for the creation of separate secrets through the Helm chart, one for the hub pod, and one for the user pods, but now we only let them create a global or reference those they have manually created. This is what relates to the hard deprecation and `fail` command in the NOTES.txt file.

jupyterhub/templates/_helpers.tpl
```diff
@@ -194,33 +194,33 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
-    {{ .registry | default "https://index.docker.io/v1/" | quote }}: {
-      "username": {{ .username | quote }},
-      "password": {{ .password | quote }},
-      {{- if .email }}
-      "email": {{ .email | quote }},
-      {{- end }}
-      "auth": {{ (print .username ":" .password) | b64enc | quote }}
-    }
-  }
-}
```
consideRatio, line 223: Here is a section removed which was one of two almost identical sections, one for hub, and one for singleuser. The other section is made into the global version.

jupyterhub/templates/_helpers.tpl
```diff
@@ -194,33 +194,33 @@ component: {{ include "jupyterhub.componentLabel" . }}
...
+{{- if eq (typeOf .) "string" }}
+{{- $__ := set $_ "res" (append $_.res (dict "name" .)) }}
+{{- else }}
+{{- $__ := set $_ "res" (append $_.res .) }}
+{{- end }}
+{{- end }}
+
+{{- /* Write the results */}}
+{{- $_.res | toJson }}
+
```
consideRatio, line 223: To review this is really not possible, this requires testing locally pretty much. I used the tools/templates/lint-and-validate.py script with the associated values.

```console
# this was used a lot to get things right
helm template ./jupyterhub --values tools/templates/lint-and-validate-values.yaml --show-only templates/image-registry-credentials.yaml
helm template ./jupyterhub --vlues tools/templates/lint-and-validate-values.yaml --show-only templates/hub/deployment.yaml
```

To use Helm templates for somewhat more complicated stuff isn't fun, I hope as few people as possible ever need to learn about this.

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -59,15 +59,8 @@ spec:
...
-        {{- if .Values.hub.imagePullSecret.enabled }}
-        - name: hub-image-credentials
-        {{ else }}
-        {{- range .Values.hub.image.pullSecrets }}
-        - name: {{ . }}
-        {{- end }}
-        {{- end }}
+      {{- with include "jupyterhub.imagePullSecrets" (dict "root" . "image" .Values.hub.image) }}
+      imagePullSecrets: {{ . }}
       {{- end }}
```
consideRatio, line 64: I embedded the logic to render a JSON formatted output such as `["secret-name-a", "secret-name-b"]` or a falsy blank string in a template helper function to avoid adding this otherwise fluffy logic in each pod.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -110,9 +97,6 @@ spec:
         {{- range $k, $container := .Values.singleuser.extraContainers }}
         - name: image-pull-singleuser-extra-container-{{ $k }}
           image: {{ $container.image }}
-          {{- with $container.imagePullPolicy }}
-          imagePullPolicy: {{ . }}
-          {{- end }}
```
consideRatio, line 115: I figured that it doesn't make sense for the prePuller to have a imagePullPolicy configured like the singleuser has it configured, so I removed these from here. I'm not 100% on this, but it also reduced the complexity and I could not see it turning out to be especially bad.

consideRatio:
> It's quite a big change to review without knowledge of exactly how this all works. 

Yeah =/ Thank you for trying your best! I added some review comments which can potentially make it a bit easier to understand what various sections are about.

> Is it possible someone wants to pull their singleuser image from a different registry to the hub and proxy?

They can reference `singleuser.image.pullSecrets` to add references to k8s secrets in addition to anything declared on the `global.imagePullSecret(s)` level influencing all pods. This is because the logic will augment them in a list.

> Do you have any idea how many users rely on a private registry? If it's a significant number maybe we should convert one of the existing CI jobs to test it?

Not sure, but it's quite common. I've tested it quite well manually by rendering templates and ensuring the output makes sense for various pods. We could add this as a test quite easily without much complexity added by adding a dummy image to a fake-private registry with credentials that are public, and then we try it as an extra `initContainer` to some pod which we configure to use these credentials for.

I'm not sure how to go about getting such registry wich public credentials that we could make use of right now though, if I had that I could go ahead and add that to this PR as well right now.

yuvipanda (changes requested):
This is great work! I'm happy to dogfood this in one of the clusters I maintain that require imagePullSecrets.

A few comments!

1. Can we just use `imagePullSecrets` instead of `global.imagePullSecrets`? This is what we do for other global things (like rbac), and it would be nice to keep that consistent, instead of adding another layer.
2. Does this mean you can't have a separate secret for hub & user pods? I don't fully understand that, can you explain why?
3. Pre-puller modes might need more than just single-user secrets, since they might also be pulling in additional images that are from different image sources - something specified with `singleuser.extraContainers`, for example. I *think* this is also configurable here, but wanted to make sure.

In general, <3 the globalized imagePullSecrets along with per-pod overrides!

consideRatio:
> 1. Can we just use imagePullSecrets instead of global.imagePullSecrets? This is what we do for other global things (like rbac), and it would be nice to keep that consistent, instead of adding another layer.

Yeah, not sure about this, I considered both, but since I'm not sure, I can go ahead and change to not using global.

---

> 2. Does this mean you can't have a separate secret for hub & user pods? I don't fully understand that, can you explain why?

No you can, I have this part in the docs currently, but I'll make it clearer and reference that it will augment with hub/singleuser/etc.image.pullSecrets entries.

> This will not override or get overridden by pod specific configuration, but instead augment the pod specific configuration.

---

> 3. Pre-puller modes might need more than just single-user secrets, since they might also be pulling in additional images that are from different image sources - something specified with singleuser.extraContainers, for example. I think this is also configurable here, but wanted to make sure.
In general, <3 the globalized imagePullSecrets along with per-pod overrides!

You can only use the Helm chart to _create_ one k8s Secret that will be automatically added to appended to all pods imagePullSecrets lists. But, you can use the Helm chart to specify any additional entries to the k8s Secret names in individual or all pods imagePullSecrets specification. Individual pods within their respective image.pullSecrets config, and chart wide config in `global.imagePullSecrets`.

consideRatio:
@yuvipanda, similarly to a comment you made in the rbac.enabled -> rbac.create PR, you can perhaps help me decide on something.

We have `imagePullSecret.create`, but it also implies an automatic appending of the created secret to all pods spec.imagePullSecrets list. Should we add a separate flag (true by default), to separate the `create` functionality from _reference by all pods spec.imagePullSecrets` functionality? If so, what name should we use? Perhaps `create` and `referenceByAllPods`?

consideRatio:
@yuvipanda I've now addressed your comments in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1794#issuecomment-705548350 and feel happy about this PR in general.

If you think this looks good, perhaps we can merge @yuvipanda?

yuvipanda:
Thanks for thoroughly working on this, @consideRatio!

PR #1797, opened 2020-10-04
values: rename rbac.enabled to rbac.create

### LINKED ISSUE #616: Allow to provide a service account for user-scheduler, and potentially other pods
Hello,

On our baremetal cluster, our ops teams restrict the creation of service account (more precisely we do not have rights to use `clusterrolebinding`, so we have to request it and it usually ends with the creation by them of a "admin"-like service account, restricted to our namespace.

I have patched the jupyterhub chart so that I provide the RBAC service account instead of letting the chart create it. I would like to propose it to upstream, will you agree with this feature?

Thanks,
Gaetan

### LINKED ISSUE #921: Use rbac.create convention ?

consideRatio:
This is a breaking change for anyone having rbac.enabled set to false. The reason to rename this is only to align with other Helm charts' conventions.

It is very possible to add logic to make this a non-breaking change, my experience with the GitLab Helm chart makes me love these clear "fix this" messages during upgrades before anything has actually changes though, so I like the idea of instead doing that and not cluttering the Helm template logic.

This PR is not at all important, but it feels good to align with other Helm charts.

Closes #921, and closes #616

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -54,7 +54,7 @@ spec:
           persistentVolumeClaim:
             claimName: hub-db-dir
         {{- end }}
-      {{- if .Values.rbac.enabled }}
+      {{- if .Values.rbac.create }}
       serviceAccountName: hub
       {{- end }}
       securityContext:
--- a/jupyterhub/templates/hub/rbac.yaml
+++ b/jupyterhub/templates/hub/rbac.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -37,6 +37,10 @@ enable pod priority or stop using the user placeholders to avoid wasting cloud
 resources.
 {{- end }}
 
+{{- if eq (.Values.rbac.enabled | default true) false }}
+{{ fail "DEPRECATION: rbac.enabled has been renamed to rbac.create" }}
+{{- end }}
+
 {{- if hasKey .Values.hub "uid" }}
 
 DEPRECATION: hub.uid is deprecated in jupyterhub chart 0.9. Set the hub.containerSecurityContext.runAsUser value
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -12,7 +12,7 @@
 
   When you ask a helper to render its content, one often forward the current
   scope to the helper in order to allow it to access .Release.Name,
-  .Values.rbac.enabled and similar values.
+  .Values.rbac.create and similar values.
 
   #### Example - Passing the current scope
   {{ include "jupyterhub.commonLabels" . }}
--- a/jupyterhub/templates/image-puller/job.yaml
+++ b/jupyterhub/templates/image-puller/job.yaml
@@ -29,7 +29,7 @@ spec:
       {{- end }}
     spec:
       restartPolicy: Never
-      {{- if .Values.rbac.enabled }}
+      {{- if .Values.rbac.create }}
       serviceAccountName: hook-image-awaiter
       {{- end }}
       containers:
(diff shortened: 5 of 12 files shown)


yuvipanda:
In my head, `create` and `enabled` do slightly different things. `create` implies creation of `Role`, `RoleBinding`, `ServiceAccount`, etc - but doesn't say anything about actually *using* them. So if your chart needs appropriate service accounts with roles & role bindings, and you set `create` to `False`, now it is your responsibility to create those RBAC things out-of-band, and *explicitly* provide service account names to the components that need them.

`enabled` is a bit more catch-all, and does 'all' the things. I'm guessing if you set enable to false, you can't just provide your own service accounts.

I think aligning with the other charts is a good idea, but if we call it `create` it should then probably require you specify service accounts for hub and autohttps. How does that sound?

consideRatio:
@kuahyeow (omg from New Zeeland, across the world for me in Sweden - Hi!!) suggested this kind of change a long time ago, and also outlined some notes on how to go about the service accounts to mimic best practices at that time.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/921#issuecomment-419692264

## Action points
- [ ] Verify these best practices are still best practices
- [ ] Consider if we want to do something special for the user-scheduler RBAC stuff, which is currently the only pod that requires ClusterRole and ClusterRoleBinding (probably not)

kuahyeow:
(hello from NZ to Sweden, @consideRatio !)

Yes, the best practices still remain - https://helm.sh/docs/chart_best_practices/rbac/#rbac-resources-should-be-created-by-default

But I think I left off `service_account.create`.

```
rbac:
  # Specifies whether RBAC resources should be created
  create: true

serviceAccount:
  # Specifies whether a ServiceAccount should be created
  create: true
  # The name of the ServiceAccount to use.
  # If not set and create is true, a name is generated using the fullname template
  name:
```

Examples of each:

- https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-role.yaml#L1
- https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-serviceaccount.yaml

jupyterhub/templates/hub/rbac.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.create -}}
```
kuahyeow, line 1: Ideally the name of the file describes the type of resource, so we will have a `hub/serviceaccount.yaml` which is fenced by `.Values. serviceAccount.create`, and a `hub/role.yaml` and a `hub/rolebinding.yaml` which are fenced by `.Values.rbac.create`

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -54,7 +54,7 @@ spec:
           persistentVolumeClaim:
             claimName: hub-db-dir
         {{- end }}
-      {{- if .Values.rbac.enabled }}
+      {{- if .Values.rbac.create }}
       serviceAccountName: hub
```
kuahyeow, line 58: Maybe this can be user specified instead, like in https://github.com/helm/charts/blob/5cd842068b06fcb41dd70085af5b070c07b4c671/stable/nginx-ingress/templates/controller-deployment.yaml#L240

This might be an extra breaking change, because I think the current of setting `.Values.rbac.enabled` to `false` is to basically use the default service account

consideRatio:
I'm closing this as I don't consider it worth merging any more. My thinking is:

1. It's nice to follow typical helm chart practices -> we should merge
2. It's not fun to introduce a `fail` in template rendering as it will cause people to need to do something.
3. It's not fun to introduce a dual support, as it increase complexity.
4. `rbac.enabled` or `rbac.create` will probably not be used in general anyhow, so its no big deal if those that do need to change it need to learn about the choice of having this named `enabled` rather than `create`.

PR #1798, opened 2020-10-04
Set container securityContext by default

consideRatio:
This can help us avoid usage of root by default which can cause clusters with [PodSecurityPolicy resources](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) to stop the Helm chart from functioning.

In practice, what I do in this PR is to create Helm chart configuration to set all Pod's containers' `securityContext`, and let this configuration default to have a securityContext like this be injected to all of our managed containers within the pods.

```yaml
# some of our k8s Deployment resources
spec:
  template:
    spec:
      containers:
        - image: ...
          # ...
          securityContext:
            runAsUser: 65534
            runAsGroup: 65534
            allowPrivilegeEscalation: false
```

A securityContext of a pod with a `runAsUser` directive will force the container to startup as that user instead of the default: `root`. This can add another layer of security to by letting the containers we run not end up escalating privileges and ending up abusing some k8s vulnerability to do harmful stuff.

Fixes #1386, but not #1408.

## singleuser.containerSecurityContext ?

This PR does not add `singleuser.containerSecurityContext`, yet, because it is quite a bit more complicated at the moment. KubeSpawner expose some configuration that in the end goes into either the pod's securityContext or the container's securityContext. Here is an overview.

KubeSpawner.* | [PodSecurityContex](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podsecuritycontext-v1-core) | [SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#securitycontext-v1-core)
-|-|-
`fs_gid` | `fsGroup` | -
`supplemental_gids` | `supplementalGroups` | -
`run_as_uid` | - | `runAsUser`
`run_as_gid` | - | `runAsGroup`
`privileged` | - | `privileged`

## Breaking change
- `proxy.containerSecurityContex` has been renamed to `proxy.chp.containerSecurityContex`

---

I deem that this can be said to close #937 good enough, but only properly by setting `singleuser.cloudMetadata.blockWithIptables: false`.

Closes #1491, Closes #1677

PR #1799, opened 2020-10-05
Added a security context to run as non-root user.

NerdSec:
Hello Everyone,

As discussed on the forum with @manics, raising a PR for fixing the security context of the `user-placeholder` pods.

This is my first of many (hopefully) PR on an OSS project. Do let me know in case I missed something.

Regards,
Nachiket

### DIFF
--- a/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
+++ b/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
@@ -47,4 +47,6 @@ spec:
           image: {{ .Values.prePuller.pause.image.name }}:{{ .Values.prePuller.pause.image.tag }}
           resources:
             {{- include "jupyterhub.resources" . | nindent 12 }}
+      securityContext:
+        runAsUser: 1000
 {{- end }}


meeseeksmachine:

https://discourse.jupyter.org/t/restricted-psp-z2jh-fails-in-local-gitlab-oauth-workflow/6202/5

consideRatio:
Welcome and thanks for being an active community member @NerdSec ! :heart: :tada:

Could you help me review #1798 perhaps? I recently opened a PR that this duplicates in order to resolve this kind of issues for all the Helm chart's pods!

jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
```diff
@@ -47,4 +47,6 @@ spec:
           image: {{ .Values.prePuller.pause.image.name }}:{{ .Values.prePuller.pause.image.tag }}
           resources:
             {{- include "jupyterhub.resources" . | nindent 12 }}
+      securityContext:
+        runAsUser: 1000
```
consideRatio, line 51: This is a podSecurityContext, which can influence sidecar containers for example from `istio` which I've seen can lead to issues before, so I think we should set it on the container level only influencing the containers we actually manage.

NerdSec: Yes, this does make sense. Adding it as a `podSecurityContext` seems better than for the entire deployment. Maybe we can add the `runAsGroup` parameter as well to ensure that any changes to the image does not affect us in the future?

The indentation in the code above seems to miss a tab, I think it should be as follows:

```
  image: {{ .Values.prePuller.pause.image.name }}:{{ .Values.prePuller.pause.image.tag }}
  resources:
    {{- include "jupyterhub.resources" . | nindent 12 }}
  securityContext:
    runAsGroup: 1000
    runAsUser: 1000
```

consideRatio: Thanks for thinking about this @NerdSec ! I've made it a container securityContext in #1798, and also added runAsGroup there, what do you think?

PR #1800, opened 2020-10-05
Match initialDelaySeconds for all hub probes

yuvipanda:
Right now, k8s starts checking if the hub is up as soon
as it starts. If the hub isn't fully up in about 10s,
it *restarts the hub*, which then has top start initializing
all over again! If you're lucky, the hub will actually be
able to finish initializing before CrashLoopBackoff, but
very easy to be stuck in CrashLoopBackoff with even a moderately
large hub.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -94,7 +94,7 @@ hub:
     timeoutSeconds: 1
   readinessProbe:
     enabled: true
-    initialDelaySeconds: 0
+    initialDelaySeconds: 60
     periodSeconds: 2
     failureThreshold: 3
     timeoutSeconds: 1


jupyterhub/values.yaml
```diff
@@ -94,7 +94,7 @@ hub:
     timeoutSeconds: 1
   readinessProbe:
     enabled: true
-    initialDelaySeconds: 0
+    initialDelaySeconds: 60
```
consideRatio, line 97: @yuvipanda this only influence the Ready state of the hub pod, which only influence if the `hub` k8s Service resources will respond with `Service not available` on traffic, so increasing this is not wanted in my mind. The `livenessProbe` above is already set to 60 seconds, but perhaps we want something longer than 60 seconds?

consideRatio (changes requested):
@yuvipanda this only influence the Ready state of the hub pod, which only influence if the hub k8s Service resources will respond with Service not available on traffic, so increasing this is not wanted in my mind. The livenessProbe above is already set to 60 seconds, but perhaps we want something longer than 60 seconds?

Perhaps setting the livenessProbe's initial delay to 3 minutes make sense?

PR #1805, opened 2020-10-05
Soft deprecate singleuser.cloudMetadata.enabled in favor of blockWithIptables

### LINKED ISSUE #1408: Blocking access to cloudmetadata endpoint requires "unsecure" PodSecurityPolicy
Fix of issue [#407 Block access to cloud metadata endpoint by default](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/407) unfortunately introduced the requirement to grant `NET_ADMIN` capability and `privileged: true` to the user Pod to successfully start up as the `initContainer` manipulates `iptables`. 

A better approach would be to use Egress rules in a `NetworkPolicy` as usually cluster admins that use PSPs also have NetPols in place.

So for example if JupyterHub is running in `Namespace` jhub a possible Network Policy could look like this:

```
kind: NetworkPolicy
metadata:
  name: block-cloudmetadata
  namespace: jhub
spec:
  podSelector:
    matchLabels:
      app: jupyterhub
      component: singleuser-server
  policyTypes:
  - Egress
  egress:
  - to:
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 169.254.169.254/32
```

Couldn't check yet but will do asap.

### LINKED ISSUE #1803: To disable the initContainer running as root to block an insecure IP to expose by default?
We currently have `singleuser.cloudMetadata.enabled=false`, which mean we actively disable access to a metadata REST API endpoint on k8s nodes on the cloud that can expose some potentially sensitive stuff. This is an endpoint that we already disable access to by default in our k8s NetworkPolicy for the singleuser pod's.

The drawback of disabling this in the way we do it, is that we require to startup an extra initContainer for the user pods, and it requires a lot of privileges compared to normal pods, which is the cause of #1408 for example.

I think it could make sense to disable this fix by default now, but a drawback is that the NetworkPolicy resources while created by default, isn't enforced without a NetworkPolicy controller available in the k8s cluster, and that isn't installed by default in all k8s clusters.

Overall though, I lean towards thinking we should stop having a default that adds this privileged initContainer to the user pods in favor of relying on the NetworkPolicy we create.

consideRatio:
Closes #1803 and closes #1408.

### DIFF
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -248,7 +248,7 @@ only need one.
 ### Block metadata with a NetworkPolicy enforced by a NetworkPolicy controller
 
 If you have _NetworkPolicy controller_ such as Calico in the Kubernetes cluster,
-it will enforce the NetworkPolicy resource created by this chart that blocks
+it will enforce the NetworkPolicy resource created by this chart (`singleuser.networkPolicy.*`) that blocks
 user access to the metadata server. We recommend relying on this approach if you
 you had a NetworkPolicy controller, and then you can disable the other option.
 
--- a/doc/source/administrator/security.md
+++ b/doc/source/administrator/security.md
@@ -248,30 +248,24 @@ only need one.
 ### Block metadata with a NetworkPolicy enforced by a NetworkPolicy controller
 
 If you have _NetworkPolicy controller_ such as Calico in the Kubernetes cluster,
-it will enforce the NetworkPolicy resource created by this chart (`singleuser.networkPolicy.*`) that blocks
-user access to the metadata server. We recommend relying on this approach if you
-you had a NetworkPolicy controller, and then you can disable the other option.
-
-```yaml
-# default configuration
-singleuser:
-  netpol:
-    enabled: true
-```
+it will enforce the NetworkPolicy resource created by this chart
+(`singleuser.networkPolicy.*`) that blocks user access to the metadata server.
+We recommend relying on this approach if you you had a NetworkPolicy controller,
+and then you can disable the other option.
 
 ### Block metadata with a privileged initContainer running `iptables`
 
 If you can't rely on the NetworkPolicy approach to block access to the metadata
-server, we suggest relying on this option. When `singleuser.cloudMetadata.block`
-is true as it is by default, an `initContainer` is added to the user pods. It
-will run with elevated privileges and use the `iptables` command line tool to
-block access to the metadata server.
+server, we suggest relying on this option. When
+`singleuser.cloudMetadata.blockWithIptables` is true as it is by default, an
+`initContainer` is added to the user pods. It will run with elevated privileges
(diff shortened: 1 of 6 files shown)


doc/source/administrator/security.md
```diff
@@ -244,6 +230,51 @@ traditional computing environment), so be very careful.
...
+only need one.
+
+### Block metadata with a NetworkPolicy enforced by a NetworkPolicy controller
+
+If you have _NetworkPolicy controller_ such as Calico in the Kubernetes cluster,
+it will enforce the NetworkPolicy resource created by this chart that blocks
+user access to the metadata server. We recommend relying on this approach if you
+you had a NetworkPolicy controller, and then you can disable the other option.
+
+```yaml
```
manics, line 255: I don't think you need to include this snippet since it's the default and isn't needed, having the full config here implies it needs to be copied. I've added a suggestion to just mention the config property in-line.

doc/source/administrator/security.md
```diff
@@ -244,6 +230,51 @@ traditional computing environment), so be very careful.
...
+server, we suggest relying on this option. When `singleuser.cloudMetadata.block`
+is true as it is by default, an `initContainer` is added to the user pods. It
+will run with elevated privileges and use the `iptables` command line tool to
+block access to the metadata server.
+
+```yaml
+# default configuration
+singleuser:
+  cloudMetadata:
+    block: true
```
manics, line 274: This is potentially confusing since if it's false cloud-metadata is still blocked if network policies are active (which is fine, I don't think it needs to be configurable since you can override the whole network policy). What do you think of changing the name of the property to e.g. `blockWithIptables` or something else?

doc/source/administrator/security.md
```diff
@@ -244,6 +230,51 @@ traditional computing environment), so be very careful.
...
+38_](https://schd.ws/hosted_files/kccncna17/d8/Hacking%20and%20Hardening%20Kubernetes%20By%20Example%20v2.pdf)
+provides more information on the dangers presented by this attack.
+
+This Helm chart blocks access to this metadata in two ways by default, but you
+only need one.
+
+### Block metadata with a NetworkPolicy enforced by a NetworkPolicy controller
+
+If you have _NetworkPolicy controller_ such as Calico in the Kubernetes cluster,
+it will enforce the NetworkPolicy resource created by this chart that blocks
```
manics, line 251:
```suggestion
it will enforce the NetworkPolicy resource created by this chart (`singleuser.networkPolicy.*`) that blocks
```

doc/source/administrator/security.md
consideRatio: Nice point, I fully agree, let's be more explicit!

doc/source/administrator/security.md
consideRatio: Deal!

PR #1807, opened 2020-10-05
deprecation: warn when proxy.https is modified and proxy.https.enabled=true

### LINKED ISSUE #1806: Disallow https configurations that likely are faulty
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1758 introduced a breaking change by letting `proxy.https.enabled` default to `false` in order to solve another issue, but the Helm chart users have previously not required specifying that in order to use HTTPS.

I think we should stick with forcing users to explicitly set it, but we could do something to help users migrate. We can spot where it seems like they have actively configured HTTPS but not actively enabled it through the `proxy.https.enabled` flag, and then trigger a Helm template rendering `fail` if so.

To identify these situations, we can look for the following pattern, which indicates that they have tried to use HTTPS by adding configuration for it, but not explicitly enabled it.

```yaml
proxy:
  https:
    enabled: false # if this is falsy and...
    type: letsencrypt # a) this isn't letsencrypt, or
    letsencrypt:
      contactEmail: '' # b) this isn't falsy
```

@yuvipanda what do you think?

consideRatio:
## Update

I'm making this PR to emit a warning instead of forcing the template rendering to `fail` as part of `helm upgrade` or `helm install`.

## Original PR

This checks for attempts to configure proxy.https without explicitly setting proxy.https.enable to true, and if so, let the template rendering fail. This is meant to help users that configured https retain the expected functionality by forcing them to use the configuration that is required to maintain such functionality.

The default value was changed in #1758.

This PR implements the idea in #1806 and thereby closes #1806.

---

In this PR I included a commit fixing a warning in our CI system.

### DIFF
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -25,6 +25,16 @@ DEPRECATION: singleuser.cloudMetadata.enabled is deprecated, instead of singleus
 
 {{- end }}
 
+
+{{- /* Warn about an attempt to configure HTTPS but not having enabled it. */ }}
+{{- if eq .Values.proxy.https.enabled false }}
+{{- if or (not (eq .Values.proxy.https.type "letsencrypt")) (not (eq .Values.proxy.https.letsencrypt.contactEmail "")) }}
+
+WARNING: Configuring proxy.https without setting proxy.https.enabled to true is no longer allowed.
+{{- end }}
+{{- end }}
+
+
 {{- if .Values.hub.extraConfigMap }}
 
 DEPRECATION: hub.extraConfigMap is deprecated in jupyterhub chart 0.8.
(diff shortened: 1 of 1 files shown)


yuvipanda (changes requested):
I often use this - when you're deploying a hub, you want all your config set (hostnames, etc) but https turned off (while DNS is cooking). I think that's sort of the point of `enabled` flags - so you can leave everything else configured, but turn the feature off.

consideRatio:
@yuvipanda okay! So the path forward is then:
- Make it clear in the changelog that users that have configured to use proxy.https and relied on proxy.https.enable to be true now need to set it to true explicitly
- Decide if we care to provide a warning in the NOTES.txt or not. What do you think?

yuvipanda:
@consideRatio yeah, I think that's a good way to do that. Unfortunately, because HTTPS requires manual steps, we can't really have `enabled` on by default, so reverting that is gonna break something...

consideRatio:
> Decide if we care to provide a warning in the NOTES.txt or not. What do you think?

consideRatio:
Self merging this to fix some conflicts in other PRs now rather than later, this PR is a harmless deprecation warning in NOTES.txt.

PR #1813, opened 2020-10-07
Added config option: proxy.chp.extraCommandLineFlags

### LINKED ISSUE #1302: Ability to pass extra command line options to configurable-http-proxy
## Description
The z2jh-k8s deployment disables the managed configurable-http-proxy (CHP) in favor of running it in its own pod. This is perfectly fine, but we lose the ability to configure the proxy's command line options since using `c.ConfigurableHTTPProxy.command` in `hub.extraConfig` no longer has any effect. Digging deeper, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/proxy/deployment.yaml builds a set of command line options based on Helm chart values, but it doesn't expose the breadth of options available in https://github.com/jupyterhub/configurable-http-proxy#command-line-options.

## Proposal
I propose adding a `proxy.extraConfig` field in the Helm chart that appends to the CHP command line options that are derived. This would mirror `hub.extraConfig` conceptually, but could be specified as a list of command line options.

## Use Cases:
* Disabling TLS 1.0/1.1 - command line option exists in `configurable-http-proxy` but no way to set it when using the Helm chart.
* Restricting allowed ciphers - command line option exists in `configurable-http-proxy` but no way to set it when using the Helm chart.
* Adding [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) response header - command line option to add custom headers implemented in PR https://github.com/jupyterhub/configurable-http-proxy/pull/206 but even when/if it is merged there won't be a way to set it when using the Helm chart.

consideRatio:
Closes #1302, thanks for documenting this request well and exploring what happens if you provide a flag twice @ivan-gomes, and as usual @manics - thank you for your presence! I love it :heart:!

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -473,6 +473,26 @@ properties:
           Configure the configurable-http-proxy (chp) pod managed by jupyterhub to route traffic
           both to itself and to user pods.
         properties:
+          extraCommandLineFlags:
+            type: list
+            description: |
+              A list of strings to be added as command line options when
+              starting
+              [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy#command-line-options)
+              that will be expanded with Helm's template function `tpl` which
+              can render Helm template logic inside curly braces (`{{ ... }}`).
+
+              ```yaml
+              proxy:
+                chp:
+                  extraCommandLineFlags:
+                    - "--auto-rewrite"
+                    - "--custom-header {{ .Values.myCustomStuff }}"
+              ```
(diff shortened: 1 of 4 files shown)


ivan-gomes:
This is fantastic, thank you so much for working on this!

yuvipanda:
Thanks a lot, @consideRatio & @ivan-gomes!

@consideRatio TIL about `tpl`! We should use it more!

PR #1816, opened 2020-10-07
CI: fix RTD builds on push to master

consideRatio:
It appears that when we build the docs in the PR preview, it works. But,
when we push to master, it doesn't any more. It appears that using
pydata_sphinx_theme to build PDFs and/or ePub files makes it crash if we
using sphinx3 rather than sphinx2. Since a doc dependency that
(executablebooks/myst-parser) previously was limited to sphinx 2 but not
longer is, the PDF / ePub build made our builds when pushing to master
fail.

I'd like to avoid maintaining these things, and would like to stop
building PDFs and ePub files. I don't know if they are used, but I'd
like to focus attention to https://z2jh.jupyter.org anyhow I think.

---

I'll self merge this now as to trial that this actually resolved the issue. If we want to have htmlzip / PDF / ePub builds again, I think we can pin sphinx<3 in doc-requirements.txt and configure `formats: all` in `.readthedocs.yml` instead of `formats: []`.

Here is an related issue about the build failure, which stems from the interaction of pydata_sphinx_theme and RTD builds like `htmlzip` it seems: https://github.com/pandas-dev/pydata-sphinx-theme/issues/177

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,7 +10,7 @@ sphinx:
   configuration: doc/source/conf.py
 
 # Optionally build your docs in additional formats such as PDF and ePub
-formats: all
+formats: []
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -1,5 +1,5 @@
 # Order matters:
-# - myst-parser depends on "sphinx>=2,<3"
+# - myst-parser depends on "sphinx>=2,<4"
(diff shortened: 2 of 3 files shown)


consideRatio:
@jupyterhub/jupyterhubteam, do we wish to keep putting in work to ensure htmlzip, pdf, and ePub builds to function with our documentation, or only do standard html builds?

manics:
I didn't know we had zip/pdf/epub builds, nor can I think of a situation where I'd want to download those formats.

Issue #1130, opened 2019-01-30
Add tolerations to everything!


AXington:
Our organization uses taints on literally all our nodes to separate our node pools. It would be rather nice to have tolerations and affinities added to all of these resources so that we don't have to maintain a fork or this repo in order to deploy to the right node pools. 

Further problems are caused when your node pools have multiple availability zones on EKS and using PVCs. PVCs by default can be in any AZ the cluster has nodes provisioned in, however the volume has to be deployed to the same AZ as the node it's being mounted to, and therefore there are issues when trying to use the default sqlite pvc database for the hub.

guohenghaha:
same problem

zzh8829:
My solution to EKS EBS PVC problem is create a separate single AZ autoscaling group dedicated to jupyter. It works great for user pods, but since we currently can't set dedicated tolerance on "hub", hub pods has to share "dedicated" node with other pods.

stevenstetzler:
I ran in to this issue as well, where I wanted to have tolerations for every schedulable object in the chart. I tweaked the Chart to allow for this. I haven't had time to run extensive tests, but I've been able to deploy a JupyterHub and everything works as expected. I've forked the repo and made a new branch with my changes, here is the commit: 456b78a25a237e7cf83bd2241d75f45a09f272c4 I think it would be a helpful addition to the Helm chart. Should I open a PR for this?

betatim:
sounds like a good idea/PR to make. One thing that @consideRatio commented on in another PR is that things should be wrapped in if statements so that they only get added to (say) an objects template when they are set to "not their empty value". Another thing to double check is if kubespawner adds the right tolerations to the pods it starts. Potentially with a short bit of documentation for how to make it all work together.

consideRatio:
This is probably still relevant, and we should support this.

AXington:
we're not even using jupyterhub anymore, but a better solution that we've been using for cross az storage using EBS is to use rook+ceph with EBS backed OSDs. 

But yeah, would open a PR for this myself but getting my employer's legal team to approve of contributions to FOSS projects is... so far not allowed.

mausch:
I'm using EFS for cross-az storage... Really easy to set up, however tolerations might still be necessary for other things (or if you don't want to use EFS)!

stevenstetzler:
I've opened a PR to merge the small change I made to add tolerations to the Helm chart.

[...]

Issue #1486, opened 2019-11-21
Core pods (hub/proxy/autohttps/user-scheduler) doesn't get the hub.jupyter.org/dedicated=core:NoSchedule toleration


jaskiratr:
I'm unable to leverage taints on an instance group dedicated for core pods. 
Following is the configuration for `jhub-core` instance group created for core pods.
```yaml
apiVersion: kops/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: null
  name: jhub-core
spec:
  image: kope.io/k8s-1.12-debian-stretch-amd64-hvm-ebs-2019-09-26
  machineType: t2.medium
  maxSize: 1
  minSize: 1
  nodeLabels:
    hub.jupyter.org/node-purpose: core
    kops.k8s.io/instancegroup: jhub-core
  role: Node
  subnets:
  - us-west-2a
  taints:
  - hub.jupyter.org/dedicated=core:NoSchedule
```
The core pods don't seem to have the relevant tolerances applied when deployed via helm. Following is the `config.yaml` used by helm
```yaml
proxy:
  secretToken: "-------------"
  pdb:
    enabled: false
singleuser:
  defaultUrl: "/lab"
  hub:
    pdb:
      enabled: false
    extraConfig:
      jupyterlab: |
        c.Spawner.cmd = ['jupyter-labhub']
  image:
    name: jupyter/all-spark-notebook
    tag: latest

scheduling:
  corePods:
    nodeAffinity:
      matchNodePurpose: require
  userPods:
    nodeAffinity:
      matchNodePurpose: require
  userScheduler:
    pdb:
      enabled: false
```
**Jhub Version** = `0.9-445a953`
`helm upgrade -i jhub jupyterhub/jupyterhub --namespace jhub --version=0.9-445a953 -f .\aws-manifests\jhub\helm\config.yaml
`
For example the `hub` pod appears to have tolerations
```json
"tolerations": [
      {
        "key": "node.kubernetes.io/not-ready",
        "operator": "Exists",
        "effect": "NoExecute",
        "tolerationSeconds": 300
      },
      {
        "key": "node.kubernetes.io/unreachable",
        "operator": "Exists",
        "effect": "NoExecute",
        "tolerationSeconds": 300
      }
    ],
```
How can I add `hub.jupyter.org/dedicated=core:NoSchedule` to all the core pods in order to prevent any other pods getting scheduled on this jhub-core instance group?

Thank you!

[...]

PR #1827, opened 2020-10-09
Add nodeSelector and tolerations config for all pods of Helm chart

stevenstetzler:
This PR is meant to fix https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1130

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -463,6 +463,41 @@ properties:
 
           NOTE: if you choose to manage the secret yourself, you are in charge of ensuring the
           secret having the proper contents.
+      nodeSelector: &nodeSelector-spec
+        type:
+          - object
+        description: |
+          An object with key value pairs representing labels. K8s Nodes are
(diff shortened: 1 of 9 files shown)


consideRatio:
This PR is related to, but does not fix https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1486, which should be fixed in a separate PR.

consideRatio:
Thank you @stevenstetzler for getting this started :heart: :tada:! I made some adjustments to: a) align configuration with other chart configuration, b) test this with dummy values in our linting script, and c) document this configuration in schema.yaml.

It would be reasonable to have a chart wide nodeSelector/nodeAffinity for core pods to some predefined label, and we already intend to have that, but we need to implement it correctly first - this is issue #1486.

It would also be reasonable to have a similar chart wide toleration of some taint for core pods, but I don't want to push for that unless its clearly wanted by multiple people as it would add quite a bit of complexity as usual - mostly because augmenting values in Helm templates is a mess, and deprecations etc will make it become a terrible blob of code to maintain.

PR #1837, opened 2020-10-13
user-scheduler: let image locality etc matter again

### LINKED ISSUE #1836: PR Discussion - block user-placeholder pod eviction temporarily based on node's single user image locality
The user-placeholder Statefulset and the continuous-image-puller Daemonset relies completely on Kubernetes pod eviction algorithm (based on the given pod priority) to choose which user-placeholder pod will be evicted to schedule a new user pod.

Considering a Jupyterhub setup like the following:

- Google Kubernetes Engine 
- User-placeholder pod resources limits/guarantees equal to VM’s resources
- A heavy single user docker image bigger than 5GB
- Considerable amount of users launching and stopping servers 
- Every user gets its own VM, the base VM image does not contain the docker single user image, we rely on continuous-image-puller to do its job.
- Three or more user-placeholder replicas

Based on my experience, Kubernetes always evicts the youngest user-placeholder pod, and in my case the youngest pod is always a recently scaled node/VM instance. This is a problem because there is a time window of approximately 3-5 min where the new node is downloading the docker image. When a user launches a server inside that time window, Kubernetes will assign the new user pod to a node that does not have that heavy docker image pulled, yet ignores the older nodes that already have the image.

As a result, users are experiencing a longer boot time than expected even when the user-placeholder is enabled.

### Proposed change

The only solution that I have found to tackle the issue described above is combining the usage of user-placeholder and continuous-image-puller with an application that adds and removes temporal `node taints`. [Nidhogg](https://github.com/uswitch/nidhogg) seems to do the job quite well.

```
Nidhogg is a controller that taints nodes based on whether a Pod from a specific Daemonset is running on them.
Sometimes you have a Daemonset that is so important that you don't want other pods to run on your node until that Daemonset is up and running on the node. 
Nidhogg solves this problem by tainting the node until your Daemonset pod is ready, preventing pods that don't tolerate the taint from scheduling there.
```
These nodes’ taints are added and removed based on the continuous-image-puller Daemonset status, which is only `ready` when the single user docker image has been pulled. 

The usage of Nidhogg + a small modification on the helm chart that adds an exclusive Nidhogg toleration in the continuous-image-puller Daemonset solves the issue described above. 

Would you be interested in adding this modification and perhaps a Nidhogg dependency to the upstream Jupyterhub Helm Chart?

Anyone that has a considerable amount of users, and is experiencing user-placeholder pod eviction order issues.

consideRatio:
I consider this to be fix #1836 by considering image locality when scheduling users.﻿

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ script:
   - await_autohttps_tls_cert_acquisition
 
   # Run tests
-  - pytest --verbose --exitfirst ./tests
+  - pytest --verbose --exitfirst ./tests || (full_namespace_report && exit 1)
 
 env:
   # Global environment variables will act as defaults, jobs will override them
@@ -124,7 +124,7 @@ jobs:
         - await_autohttps_tls_cert_acquisition
 
         # Run tests
-        - pytest --verbose --exitfirst ./tests
+        - pytest --verbose --exitfirst ./tests || (full_namespace_report && exit 1)
 
     - stage: test
       name: chart:k8s-1.19
--- a/ci/common
+++ b/ci/common
@@ -156,3 +156,35 @@ setup_kubeval () {
     curl -sfL https://github.com/instrumenta/kubeval/releases/download/0.15.0/kubeval-linux-amd64.tar.gz | tar xz kubeval
     sudo mv kubeval /usr/local/bin/
 }
+
+full_namespace_report () {
+    # list config (secret,configmap)
+    kubectl get secret,cm
+    # list networking (service,ingress)
+    kubectl get svc,ing
+    # list workloads (deployment,statefulset,daemonset,pod)
+    kubectl get deploy,sts,ds,pod
+
+    # if any pod has any non-ready -> show its containers' logs
+    kubectl get pods -o json \
+    | jq '
+        .items[]
+        | select(
+            any(.status.containerStatuses[]?; .ready == false)
+        )
+        | .metadata.name' \
+    | xargs --max-args 1 --no-run-if-empty \
+    sh -c 'printf "\nPod with non-ready container detected\n - Logs of $0:\n"; kubectl logs --all-containers $0'
+
+    # if any pods that should be scheduled by the user-scheduler are pending ->
+    # show user-scheduler's logs
+    (
+        kubectl get pods -l "component in (user-placeholder,singleuser-server)" -o json \
+        | jq -r '
+            .items[]
+            | select(.status.phase == "Pending")
+            | .metadata.name
+        '
+    ) | xargs --max-args 1 --no-run-if-empty --max-lines \
+    sh -c 'printf "\nPending user pod detected ($0)\n - Logs of deploy/user-scheduler:\n"; kubectl logs --all-containers deploy/user-scheduler'
+}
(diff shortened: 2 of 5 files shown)


consideRatio:
@donotpush I added a script to provides logs of failing containers in the CI system, the user-scheduler failed with:

```
couldn't create scheduler using provider "DefaultProvider": initializing profiles: creating profile for scheduler name jupyterhub-user-scheduler: plugin "NodePreferAvoidPods" already registered as "ScorePlugin"
```

My guess is that it reacts to the fact that I added `NodePreferAvoidPods` to `scheduling.userScheduler.plugins.enabled`, and that it was probably already enabled by default so under some wrapper plugin that contained all of them already.

## Remaining steps
- [ ] Resolve the current CI error.
- [ ] Preferably also manually verify it does what it should by inspecting the scheduler logs when `loglevel: 10` and you can see the scoring be evaluated whenever it schedules a pod.

donotpush:
@consideRatio I am going to reproduce this on my environment, and get back to you with my findings.

consideRatio:
Thanks @donotpush! Btw, perhaps the error `plugin "NodePreferAvoidPods" already registered as "ScorePlugin"` will go away by also adding it to the disabled list, my guess is that kube-schedulers implementation to read this is to first disable then enable.

donotpush:
@consideRatio I cannot seem to influence the pod eviction order even when using a custom scheduler with image locality as a unique/highest priority. The eviction is skipping the whole scoring process.

consideRatio:
@donotpush okay, I'm not sure how the scheduler will influence the eviction choice in order to get a certain pod to be scheduled.

I consider it out of scope for this PR to also make the scheduler, if possible, influence the eviction of a user-scheduler pod, rather than just making sure that image locality is considered given two separate nodes where there is room to schedule without eviction.

---

Did you get the user-scheduler to run at all properly by some additional change?

donotpush:
@consideRatio I agree with you about the scope.  Considering the scheduler stages (QueueSort, PreFilter, PreScore, Score, ...) I do not seem to reach the Score stage when trying to influence eviction due to the predicate `PodFitsResources` - which is probably causing the eviction. I am still investigating - I will report back my findings.

Have you noticed that --policy-config-file=/etc/user-scheduler/policy.cfg is always using the default config instead of the one provided via `values.yaml`?

consideRatio:
@donotpush have you rendered the Helm templates on k8s 1.16+? If not, you are working with the old user-scheduler configuration still.

```
            {{- if .Capabilities.APIVersions.Has "storage.k8s.io/v1/CSINode" }}
            - --config=/etc/user-scheduler/config.yaml
            {{- else }}
            - --scheduler-name={{ .Release.Name }}-user-scheduler
            - --policy-config-file=/etc/user-scheduler/policy.cfg
            - --lock-object-name=user-scheduler-lock
            - --lock-object-namespace={{ .Release.Namespace }}
            {{- end }}
```

consideRatio:
I want to try out if disable + enable of the scoring parts helps resolve the bug so I've pushed one more commit now.

consideRatio:
Looks like that did the trick - @donotpush, I'd like to go ahead and merge this as it is, but leave the issue it was meant to close open just a bit longer to summarize the state.

Issue #1580, opened 2020-02-20
Dedicated .py file instead of snippets in extraConfig


albertmichaelj:
I'm opening this issue primarily for advice, but I also think that there may have been an unintentional side effect of a recent change to helm chart. #1407 moved the config files for the hub image to the helm chart `files/` directory. Overall, this seems to be a good thing, but there seems to be a problem with trying to use a custom hub image with a custom configuration with this helm chart as a sub-chart. Maybe I'm just misunderstanding something.

Before this change, if I wanted to change `jupyterhub_config.py`, all I needed to do was build a custom hub image with my new `jupyterhub_config.py` file and then point the hub image in the helm chart to my custom image. Now though, even with a custom docker image for the hub, the helm chart version of `jupyterhub_config.py` gets mounted and run. Moreover, if I'm using this helm chart as a sub-chart in a larger helm app, I can't figure out a way I could overwrite the `jupyterhub_config.py` file without just forking this chart entirely and modifying the `jupyterhub_config.py` file directly.

Am I missing something? Did the hub customization become harder when using this chart as a sub-chart?

I'm very open to ideas, and just to be clear, using the `extra_config` is not enough because I want to sub-class KubeSpawner.

consideRatio:
> I'm very open to ideas, and just to be clear, using the extra_config is not enough because I want to sub-class KubeSpawner.

I think you can actually sub class kubespawner from the extraConfig. Consider it as a python file that will run. You can define the sub class and then assign it.

---

If you really want to replace everything, then hmm... Perhaps override the command to start jupyterhub to use the specified configuration that you build with the image could accomplish it?

albertmichaelj:
Your right that I certainly could sub class KubeSpawner in `extra_config`, but I think I would basically have to redo any part of the config file that sets values for the `c.KubeSpawner...` stuff, right? I would certainly have to do that as well if I had my own `jupyterhub_config.py` file, but it would be easier to do a find and replace instead of maintaining most of the `jupyterhub_config.py` as stuff in `extra_config`. Or maybe I'm misunderstanding how the `c.KubeSpawner...` stuff works.

[...]

PR #1839, opened 2020-10-15
Provide a way to disable default hub configmap

cbanek:
The jupyterhub chart is ultra-configurable and good for deploying
my other jupyterhub environments.  But it provides no way to not
use the default jupyterhub_config.py via the configmap.  I want
to be able to use the extraConfigMaps to mount my own configmap
that contains the jupyterhub_config.py and not include the original
jupyterhub_config.py.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -303,6 +303,10 @@ properties:
                 image: busybox:1.28
                 command: ['sh', '-c', 'command2']
           ```
+      enableConfigMap:
+        type: boolean
+        description: |
+          Whether the configmaps are created and used for the hub deployment.
       extraEnv:
         type: object
         description: |
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.hub.enableConfigMap }}
 kind: ConfigMap
 apiVersion: v1
 metadata:
@@ -8,3 +9,4 @@ data:
   {{- /* Glob files to allow them to be mounted by the hub pod */ -}}
   {{- /* key=filename: value=content */ -}}
   {{- (.Files.Glob "files/hub/*").AsConfig | nindent 2 }}
+{{- end }}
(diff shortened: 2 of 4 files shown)


cbanek:
This one is ready and should be pretty simple!  Let me know if you think there's a better way to do this.

consideRatio:
You want to opt ouf of mounting jupyterhub_config.py, to allow you to create your own and mount it at that specific location jupyterhub loads it from?

I'm actually strongly opinionated against it, because i feel it is a problematic solution by itself, and because i think it will be problematic for the open source projects maintainability - considering for example that issues may end up stemming from this which cause a lot of extra time lost by others to umderstand issues stemming from replacing the config entirely.

See #1580 for a related discussion if i recall myself correctly. A change i think could be acceptable, is to let users of the helm chart configure the commandflags passed to jupyterhub, like for proxy.extraCommandLineFlags, but for hub. See #1813.

I'll close this for the time being at least, let me know if you think the hub.extraCommandLineFlaga solution would work, i think it could make sense to accept a PR for that.

cbanek:
> You want to opt ouf of mounting jupyterhub_config.py, to allow you to create your own and mount it at that specific location jupyterhub loads it from?

Yes that's exactly right.  I'm making a production jupyterhub docker image that already has the jupyterhub_config.py built in, and I'd rather not have the chart changes making code changes to that - in essence, I'm exactly trying to replace the helm config for the hub, but keep all the good deployment parts and the rest of it.

Your idea of doing command line flags was another possibility I was thinking (since then I could just point to a different config file) and that makes sense.  I didn't know that providing the same flag twice will use the latest one, although that does make the implementation simpler.  I'll give that method a shot and come back with another PR.

consideRatio:
@cbanek thank you for suggesting a solution to a problem you have btw 🎉 ❤️ I hate closing PRs as it feels like a bummer thing to do, i hope you don't feel discouraged. Thank you for investing your time in the Jupyter ecosystem and welcome!

cbanek:
:heart: No worries at all. Opened up https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1840 instead, and I'm glad I tested it first!

PR #1840, opened 2020-10-15
hub.command and hub.args configuration added

cbanek:
This allows the chart to override the JupyterHub commandline.
While I started doing an extraCommandLineArguments idea, that
did not work as specifying the config option twice will not
allow the hub to start.  So this makes the command line in
general overrideable.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -4,6 +4,29 @@ properties:
   hub:
     type: object
     properties:
+      command:
+        type: string
+        description: |
+          The command to start jupyterhub, used in the deployment to set the
+          commandline, along with args and other flags set by the chart.
+      args:
+        type: list
+        description: |
+          A list of strings to be used as the command line arguments for JupyterHub
+          that will be expanded with Helm's template function `tpl` which
+          can render Helm template logic inside curly braces (`{{ ... }}`).
+
+          ```yaml
+          hub:
+            args:
+              - "--config"
+              - "/etc/jupyterhub/jupyterhub_config.py"
+          ```
+
+          Note that these replace the command line arguments provided by default,
+          but not those added by using other helm options.
+
+          The default args in values.yaml provides the typical behavior.
       cookieSecret:
         type:
           - string
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -5,28 +5,62 @@ properties:
     type: object
     properties:
       command:
-        type: string
+        type: list
         description: |
-          The command to start jupyterhub, used in the deployment to set the
-          commandline, along with args and other flags set by the chart.
+          A list of strings to be used to replace the JupyterHub image's
+          `ENTRYPOINT` entry. Note that in k8s lingo, the Dockerfile's
+          `ENTRYPOINT` is called `command`. The list of strings will be expanded
+          with Helm's template function `tpl` which can render Helm template
+          logic inside curly braces (`{{... }}`).
+
+          This could be useful to wrap the invocation of JupyterHub itself in
+          some custom way.
+
+          For more details, see the [Kubernetes
+          documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).
       args:
         type: list
         description: |
-          A list of strings to be used as the command line arguments for JupyterHub
-          that will be expanded with Helm's template function `tpl` which
-          can render Helm template logic inside curly braces (`{{ ... }}`).
+          A list of strings to be used to replace the JupyterHub image's `CMD`
+          entry as well as the Helm chart's default way to start JupyterHub.
+          Note that in k8s lingo, the Dockerfile's `CMD` is called `args`. The
+          list of strings will be expanded with Helm's template function `tpl`
+          which can render Helm template logic inside curly braces (`{{... }}`).
+
+          ```{warning}
+          By replacing the entire configuration file, which is mounted to
+          `/etc/jupyterhub/jupyterhub_config.py` by the Helm chart, instead of
+          appending to it with `hub.extraConfig`, you expose your deployment for
+          issues stemming from getting out of sync with the Helm chart's config
+          file.
+
+          These kind of issues will be significantly harder to debug and
+          diagnose, and can due to this could cause a lot of time expenditure
+          for both the community maintaining the Helm chart as well as yourself,
+          even if this wasn't the reason for the issue.
+          
+          Due to this, we ask that you do your _absolute best to avoid replacing
+          the default provided `jupyterhub_config.py` file. It can often be
+          possible. For example, if your goal is to have a dedicated .py file
+          for more extensive additions that you can syntax highlight and such
+          and feel limited by passing code in `hub.extraConfig` which is part of
+          a YAML file, you can use [this
+          trick](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1580#issuecomment-707776237)
+          instead.
+          ```
 
           ```yaml
           hub:
             args:
+              - "jupyterhub"
               - "--config"
               - "/etc/jupyterhub/jupyterhub_config.py"
+              - "--debug"
+              - "--upgrade-db"
           ```
 
-          Note that these replace the command line arguments provided by default,
-          but not those added by using other helm options.
-
-          The default args in values.yaml provides the typical behavior.
+          For more details, see the [Kubernetes
+          documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).
       cookieSecret:
         type:
           - string
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -47,6 +47,10 @@ hub:
     password:
   labels: {}
   annotations: {}
+  command: jupyterhub
+  args:
+    - --config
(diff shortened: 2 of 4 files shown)


cbanek:
Sad about the config option twice not working, although I did try it and it gave me this error:

[C 2020-10-15 23:15:08.790 JupyterHub application:88] Bad config encountered during initialization: Error loading argument JupyterHub.config_file=['/etc/jupyterhub/jupyterhub_config.py', '/app/jupyterhub_config.py'], config_file only accepts one value, got 2: ['/etc/jupyterhub/jupyterhub_config.py', '/app/jupyterhub_config.py']

consideRatio:
@cbanek I think you ran into an issue with multiple `--config` flags because you used an older version of JupyterHub. It seem to work for me using 1.2.0b1 which is currently used by the Helm chart as of some hours ago. Let's aim for hub.extraCommandLineFlags again, and I guess you will be forced to update the JupyterHub version you've been using as well to use this functionality - does this sound okay?

image

cbanek:
That's interesting, I'm using the jupyterhub 1.2 image, but not the beta one.  I can give that a shot, although it seems fairly equivalent?

cbanek:
On second thought, it seems even better to not explicitly rely on doubling up on command line args, especially if depending on that just changed?  It also seems a bit cleaner to not have to worry about passing something twice, which honestly I think adds a bit of confusion since the resulting commandline looks wrong (and did used to throw an error)?  And if a flag or something else is used that can't be easily overridden by providing it twice, that sounds like trouble?

consideRatio:
The JupyterHub image is not made for the Helm chart, but the jupyterhub/k8s-hub is - in general. I've also concluded that specifying --config multiple times is fine for many jupyterhub versions, but it depends on ipytraitlets version rather than jupyterhub itself I think, or at least something that jupyterhub depends to parse its command line options, on rather than jupyterhub itself.

Note how the fact that you use a custom image cause some issues, and this is my experience in general - anyone using a custom image, or custom jupyterhub_config.py, will end up with issues that are hard to spot caused by that fact, and it makes the community effort to keep this repo maintained properly becomes harder. This is a key reason wanting to avoid this kind of practice.

I think adding a extraCommandLineFlags option is okay, but I really hope to see users of the Helm chart avoid it to override the config, and will want to add a warning about it before we merge such feature.

I wrote this in a haste, must join a meeting now, thank you for your effort with this PR @cbanek!

cbanek:
Can you let me know what pushed docker image you are using to test this?  I've tried a few, and I can't reproduce your results on the multiple config flag working, even when using JupyterHub 1.2.0b1 on the image pushed on Friday:

chb@AMBITION:~/work/nublado2$ docker run -it jupyterhub/k8s-hub:0.9.0-n358.h11c2aeb8 /bin/bash
jovyan@917ace6a327b:/srv/jupyterhub$ jupyterhub
[I 2020-10-19 17:01:54.015 JupyterHub app:2331] Running JupyterHub version 1.2.0b1
...
jovyan@917ace6a327b:/srv/jupyterhub$ jupyterhub --config jupyterhub_config.py  --config jupyterhub_config2.py
[C 2020-10-19 17:02:23.079 JupyterHub application:89] Bad config encountered during initialization: Error loading argument JupyterHub.config_file=['jupyterhub_config.py', 'jupyterhub_config2.py'], config_file only accepts one value, got 2: ['jupyterhub_config.py', 'jupyterhub_config2.py']

Then I tried the same on the jupyterhub/k8s-hub:0.9.1, and that seemed to work, but that was running JupyterHub 1.1.0 (and this image is quite old).

In any case, I think it's better to be able to override the whole command-line, rather than just a piece of it, especially when this behavior seems so ill-defined?  I don't understand the need for a warning as this doesn't change any behavior (the default is how it is now), it just allows people (power users) to override it if they need to.  I understand the worries about maintainability, but this also seems to be the only helm chart for jupyterhub?  Is there another less specific jupyterhub helm chart that I should be using instead that I've missed?

consideRatio:
Try the latest here for example: https://hub.docker.com/r/jupyterhub/k8s-hub/tags

0.9.1 is older than 0.9.0-nxxx due to a bug in naming.

cbanek:
Yes, that's the same image as the one I used above (you can see it in the commandline).  I'm using the image that was updated 3 days ago and it doesn't seem to do what yours does.

consideRatio:
Oh.. ill debug this further when im home, currently in transit

consideRatio:
@cbanek yikes okay so I drew the wrong conclusion when trying different JupyterHub versions before. The command line parsing was made by an outdated version of `traitlets` for me, and its version is what matters for the command line parsing, not the JupyterHub version. When I upgraded it to traitlets==5.* from ==4.* I could no longer send of two --config flags.

I'm leaning towards not accepting this PR. I think it hints users toward a usage pattern I think would be bad for the user and the repo's ability to remain well maintained. I don't want to just close it without further deliberation though.

I'll raise this PR as a discussion point with other JupyterHub team members during the upcoming JupyterHub team meeting which everyone is welcome to join: https://github.com/jupyterhub/team-compass/issues/340

yuvipanda:
ooh, this would definitely be useful!

We do *something* like this with `singleuser.cmd`, which specifies the command to execute in the spawned user pod. Can we support `hub.cmd` as well, set default to what's currently hardcoded in our `deployment.yaml`? This allows admins to override it without it, in line with how other helm charts seem to do it.

manics:
@consideRatio Would it help allay your concerns if we had a higher-level team discussion about the overall direction and aims of Z2JH? Maybe scheduled separately from the monthly team-meeting?

consideRatio:
@manics yes... Really sad to have overslept todays meeting and missed the change to discuss this :(

consideRatio:
Lets go with a hub.cmd override of everything since the extraCommandLineFlags approach wont work then.

---

I've been hesitant about this PR because i believe that replacing the jupyterhub_config.py is a big loose/loose practice for the user doing it and for the community maintenance workload of the project, and the driving force of this PR is to be able to do that.

jupyterhub/values.yaml
```diff
@@ -47,6 +47,10 @@ hub:
     password:
   labels: {}
   annotations: {}
+  command: jupyterhub
```
consideRatio, line 50: I recall there is a recommendation to always use an array rather than a string, I don't remember why but let's go with that unless someone say the opposite.

```suggestion
  command:
    - "jupyterhub"
```

jupyterhub/schema.yaml
```diff
@@ -4,6 +4,29 @@ properties:
...
+
+          ```yaml
+          hub:
+            args:
+              - "--config"
+              - "/etc/jupyterhub/jupyterhub_config.py"
+          ```
+
+          Note that these replace the command line arguments provided by default,
+          but not those added by using other helm options.
```
consideRatio, line 27: Currently, they will replace the config specifically, but not --debug or --upgrade-db.

consideRatio:
Since I have some opinions on how to do this, I'll push some changes and we'll request a review following that.

consideRatio:
@cbanek @yuvipanda @manics I've pushed a commit implementing this in a way I find acceptable, let me know what you think.

Relevant to note:
- used `hub.command` and `hub.args` (Kubernetes naming convention) as compared to hub.entrypoint and hub.cmd or similar (Dockerfile naming convention)
- hub.command and hub.args default to an empty array
- hub.command can be used to wrap the invocation of jupyterhub with something else, such as `tini` or like first running a script and then starting jupyterhub.
- hub.args can be used to replace `--config` entirely, but doing so comes with a clear warning about the downsides I hope to avoid within the configuration reference, which you can see [in preview here for this PR](https://zero-to-jupyterhub--1840.org.readthedocs.build/en/1840/reference/reference.html#hub-command).
- hub.args if specified, will override how the Helm chart otherwise append `--debug` and `--upgrade-db` based on configuration. I didn't find an easy way to make them merge without introducing complicated Helm template logic that also is a lot of work.
- hub.args includes `jupyterhub` by default, because if I don't, it makes it tricky as the `args` part of the k8s manifest may or may not end up with content, and then I need to have a lot of complicate Helm template logic to render it conditionally.

cbanek:
That looks great to me, thanks so much for polishing that up!

consideRatio:
@cbanek thank you for the feedback, your work on this, and your patience with me :sunflower:

yuvipanda:
LGTM! Happy to merge once tests pass :)

consideRatio:
I had forgot to add the `--config` and `/etc/jupyterhub/jupyterhub_config.py` arguments by default.

## Off topic about travis being slow
TravisCI is extremely slow as of lately, they seem to have lowered their amount of available workers, and end up getting a large queue that have now not gone to 0 in over 24 hours.

image

More info at https://www.traviscistatus.com/#week

manics:
@consideRatio also off-topic: I converted your K3s/helm setup script to a GitHub action a few months ago, so we could consider switching to a GitHub workflow:
https://github.com/manics/action-k3s-helm/

consideRatio:
@manics nice!

I appreciate github actions:
- feature to not trigger job builds based on what files have changed
- being able to compose parts together, which can be useful for z2jh/binderhub/kubespawner for example within the jupyterhub org

consideRatio:
I'll merge now since tests succeeded except one that ran into an intermittency error: https://github.com/jupyterhub/jupyterhub/issues/3222

PR #1841, opened 2020-10-16
try out jupyterhub 1.2.0b1

minrk:
and bumps dependencies in hub image

I had to tweak things a bit since our freezing code didn't play nice with the fact that the jupyterhub version is specified both inside requirements.txt and as a separate arg, which results in a conflict when the two versions don't match. I'm not sure what exactly is the best way to deal with that. Perhaps put jupyterhub version in requirements.in instead?

several were removed due to https://github.com/googleapis/google-auth-library-python/pull/619 making aiohttp an optional dependency of google-auth. I *think* this is fine and shouldn't affect us, since this is a transitive dependency and we don't directly use google-auth + aiohttp. Presumably tests will reveal if we should add `google-auth[aiohttp]` as a dependency...

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -12,7 +12,7 @@ charts:
         valuesPath: hub.image
         buildArgs:
           # NOTE: Also bump the Chart.yaml's appVersion if this is bumped
-          JUPYTERHUB_VERSION: 1.1.0
+          JUPYTERHUB_VERSION: 1.2.0b1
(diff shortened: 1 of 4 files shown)


consideRatio:
Thanks @minrk for making a note about the duplicate version specification, I'm not sure either, but I really love your tooling with requirements.in being built in a dockerfile to requirements.txt! :heart:

Issue #534, opened 2018-02-23
Support for NetworkPolicy ingress/egress


manics:
I'm interested in using a singleuser pod [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to limit egress. The use-case I have in mind is providing a public but very locked-down jupyterhub deployment to give people a taster with minimal barriers, and to provide full access behind a second authenticated deployment.

I found a related discussion on https://github.com/jupyterhub/mybinder.org-deploy/issues/146 and can see pros and cons to including it in this chart:
- Configuring everything in zero-to-jupyterhub-k8s is very convenient
- A NetworkPolicy is very specific to a deployment, and can be applied outside the chart since the pod labels are known

If you think it's worthwhile here's a proposal:
```
singleuser:
  networkPolicy:
    ingress:
      - List of ingress rules from https://kubernetes.io/docs/concepts/services-networking/network-policies/
    egress:
      - List of egress rules from https://kubernetes.io/docs/concepts/services-networking/network-policies/
```
applied to
```
  podSelector:
    matchLabels:
      app: jupyterhub
      component: singleuser-server
```

minrk:
I 100% think we should apply network policies here.

The one thing I was bitten by in jupyterhub/mybinder.org-deploy#146 is that kubernetes doesn't check at all for using unsupported features, so you can happily load an egress network policy that is completely ignored without any warnings or errors. I haven't checked in a bit, but I assume gke-1.9 will have sufficiently recent calico (>= 2.6.1) to enable egress policies.

We should start by sketching out a network policy of who needs to talk to whom, and what the default should be. To get started:

ingress:

1. single-user servers need only accept ingress directly from the hub and proxy
2. hub needs only accept ingress from proxy, services, and single-user server
3. proxy should accept ingress from the world

egress:

- proxy only needs egress to hub, services, single-user servers
- hub needs egress to proxy-api, services, single-user servers, and whatever may be needed by the Spawner (kube itself?)
- single-user servers need egress to the hub and the outside world, but no other in-cluster addresses

Figuring out the right labeling scheme / default policies for that would be great!

manics:
[...]

PR #1842, opened 2020-10-16
netpol: allowedIngressPorts and interNamespaceAccessLabels config added with defaults retaining 0.9.1 current behavior

### LINKED ISSUE #1863: 502 Bad Gateway in Hub after upgrading chart (0.9.0-alpha.1 -> 0.9.0-n4xx)
### Bug description

Our chart has not been upgraded in quite some time and we need the features that were added for handling `imagePullSecrets` recently, so I tried upgrading and ran into an odd situation. We were all the way back on `0.9.0-alpha.1.060.6698eb9` and upgraded to `0.9.0-n409.hce116620`. 

#### Expected behaviour
When the chart upgrade was completed I expected that I would be able to visit the URL of jupyterhub and login through GitHub and get a newly spawned pod.

#### Actual behaviour
When visiting the jupyterhub URL I get an error from nginx saying "502 Bad Gateway". 

#### Diagnostic

When looking through pod logs, I don't see anything awry except for in the `user-scheduler` pod where I've noticed the following failure:

```
E1022 13:36:29.100644       1 reflector.go:127] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:toolbox-datalake:user-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
```

This might be completely unrelated, but I'm not finding much else.

### How to reproduce

1. Create a `values.yaml` file 
2. Run `helm upgrade ${HELM_RELEASE_NAME} https://jupyterhub.github.io/helm-chart/jupyterhub-0.9.0-n409.hce116620.tgz -f values-jupyterhub.yaml --install --set proxy.secretToken=${PROXY_TOKEN} --namespace ${K8S_NAMESPACE}`
3. Visit hub URL
4. See 502 gateway error

### Your personal set up

 - OS: 
    - Kubernetes on IBM Cloud version `1.18.8_1527`,
    - helm client:`version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}`
 
 - Version: helm chart`0.9.0-n409.hce116620`
 
 - Configuration: I'm happy to share any sections of the `values.yaml` that may be of assistance here as well as any of the image versions.

Please let me know if this is inappropriate as a bug. Thanks!

consideRatio:
# PR changes summarized

### Breaking change
- `proxy.networkPolicy` is deprecated, and `proxy.chp.networkPolicy` and `proxy.traefik.networkPolicy` is introduced.

### Restores access to `proxy` pod's http/https ports by default (bugfix)
Since #1670 the proxy pod would not accept incoming traffic from pods in other namespaces, due to a bugfix that caused a breaking change. This can be observed in #1863 for example I think.

```diff
     - ports:
         - port: http
         {{- if or $manualHTTPS $manualHTTPSwithsecret }}
         - port: https
         {{- end }}
-    - from:
+      from:
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-proxy-http: "true"
```

The bug fixed was that `hub.jupyter.org/network-access-proxy-http: "true"` was entirely ignored and had no meaning, but now that it got meaning, we ended up restricting access too much. As an example, the proxy would reject incoming traffic from a nginx-ingress-controller pod that terminated TLS in another namespace.

In this PR I introduce the networkPolicy configuration `allowedIngressPorts` as suggested by @manics, and configures it for `proxy.chp.networkPolicy` and `proxy.traefik.networkPolicy` to be set to `[http, https]` by default, while the networkPolicy configuration for hub/singleuser will not allow any of this by default.

This provides us with a configuration concept applicable to all networkPolicy resources we deploy, while retaining ability to work with even finer grained control with network access labels.

### Conditional use of access labels across namespaces (enhancement)
This PR let's the respective `networkPolicy` configuration configure `interNamespaceAccessLabels`. If this configuration option is set to `accept`, the access labels valid for the respective networkPolicy configuration, such as `hub.jupyter.org/network-access-proxy-api: "true"`, will be accepted even on pods in other namespaces, and on entire namespaces, to allow these pods as sources for incoming network traffic.

```diff
+        # source 1 - labelled pods
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-proxy-api: "true"
+        {{- if eq .Values.proxy.chp.networkPolicy.interNamespaceAccessLabels "accept" }}
+          namespaceSelector:
+            matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
+        # source 2 - pods in labeled namespaces
+        - namespaceSelector:
+            matchLabels:
+              hub.jupyter.org/network-access-proxy-api: "true"
+        {{- end }}
```

# Motivation
Network policies are enabled by default as of #1271, and we need ability to configure access in a way that won't force users and maintainers to debug k8s native network policy configuration which is notoriously hard to understand correctly. Closes #1863.

### allowedIngressPorts config
We needed to come up with a solution after #1670, which as described below ended up restricting ingress, for example from a nginx-ingress-controller pod in another namespace, towards the proxy pod. At the same time, we should have a flexible and cohesive configuration of networkPolicies.

### interNamespaceAccessLabels config
`interNamespaceAccessLabels` is ensuring that users are not forced to add fragile workarounds if they need to access something from another namespace in the future, and by avoiding fragile workarounds, we avoid complicated issue reports following the fragile configurations future failures.

# Related
### NetworkPolicy documentation
[This section of the k8s documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors) is very relevant details on the rules that we can set in a NetworkPolicy from the k8s docs, distinguishing between: a) podSelector, b) namespaceSelector, c) podSelector AND namespaceSelector, and d) ipblock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Here you can find upgrade changes in between releases and upgrade instructions.
 
 ## [0.10]
 
-### [0.10.0]
+### [0.10.0-beta.1] (UNRELEASED)
 
 #### Breaking changes:
 
@@ -16,16 +16,16 @@ Here you can find upgrade changes in between releases and upgrade instructions.
   known to be needed, with the exception of the user pods outgoing
   communication, where all outgoing communication is allowed by default.
   
-  Note that these network policies only influence network communication in a
+  Note that these NetworkPolicies only influence network communication in a
   Kubernetes cluster if a NetworkPolicy controller enforce them, such as Calico.
-
-  With network policies enabled, you may require additional configuration,
-  especially for deployments that include additional components that access
-  JupyterHub pods directly (i.e. not through the `proxy-public` service).
 
   See the [security
   documentation](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#kubernetes-network-policies)
   for more details on this.
+
+- The Helm chart configuration `proxy.networkPolicy` has been deprecated, and
+  `proxy.chp.networkPolicy` (proxy pod) and `proxy.traefik.networkPolicy`
+  (autohttps pod) is recommended to use instead.
 
 ## [0.9]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,9 @@ Here you can find upgrade changes in between releases and upgrade instructions.
   documentation](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#kubernetes-network-policies)
   for more details on this.
 
-- The Helm chart configuration `proxy.networkPolicy` has been deprecated, and
+- The Helm chart configuration `proxy.networkPolicy` has been removed,
   `proxy.chp.networkPolicy` (proxy pod) and `proxy.traefik.networkPolicy`
-  (autohttps pod) is recommended to use instead.
+  (autohttps pod) must be used instead.
 
 ## [0.9]
 
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -240,7 +240,7 @@ properties:
               `proxy.chp` and `proxy.traefik`'s networkPolicy configuration is
               `[http, https]`, while it is `[]` for other networkPolicies.
               
-              Note that these port names or numbers targets a Pod's port name or
+              Note that these port names or numbers target a Pod's port name or
               number, not a k8s Service's port name or number.
       db:
         type: object
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -204,11 +204,15 @@ properties:
           ingress:
             type: list
             description: |
-              Additional ingress rules to add.
+              Additional ingress rules to add except those that is known to be needed by the respective pods in the Helm chart.
           egress:
             type: list
             description: |
-              Additional egress rules to add.
+              Additional egress rules to add except those that is known to be needed by  the respective pods in the Helm chart.
+
+              The default value of this egress is to allow all traffic, except for the `singleuser.networkPolicy.egress`, which is also limiting access to a metadata server that can be exploited.
+
+              If you want to restrict egress, you can override this permissive default to be an empty list.
           interNamespaceAccessLabels:
             type: string
             enum:
@@ -216,7 +220,7 @@ properties:
               - ignore
             description: |
               This configuration option determines if both namespaces and pods
-              in other namespaces that have specific access labels, should be
+              in other namespaces, that have specific access labels, should be
               accepted to allow ingress (set to `accept`), or, if the labels are
               to be ignored when applied outside the local namespace (set to
               `ignore`).
@@ -235,6 +239,9 @@ properties:
               what the origin of the request is. The default setting for
               `proxy.chp` and `proxy.traefik`'s networkPolicy configuration is
               `[http, https]`, while it is `[]` for other networkPolicies.
+              
+              Note that these port names or numbers targets a Pod's port name or
+              number, not a k8s Service's port name or number.
       db:
         type: object
         properties:
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -191,6 +191,50 @@ properties:
               pullSecrets:
                 - my-k8s-secret-with-image-registry-credentials
               ```
+      networkPolicy: &networkPolicy-spec
+        type: object
+        description: |
+          This configuration regards the creation and configuration of a k8s
+          _NetworkPolicy resource_.
+        properties:
+          enabled:
+            type: bool
+            description: |
+              Toggle the creation of the NetworkPolicy resource for this pod.
+          ingress:
+            type: list
+            description: |
+              Additional ingress rules to add.
+          egress:
+            type: list
+            description: |
+              Additional egress rules to add.
+          interNamespaceAccessLabels:
+            type: string
+            enum:
+              - accept
+              - ignore
+            description: |
+              This configuration option determines if both namespaces and pods
+              in other namespaces that have specific access labels, should be
+              accepted to allow ingress (set to `accept`), or, if the labels are
+              to be ignored when applied outside the local namespace (set to
+              `ignore`).
+
+              The available access labels for respective NetworkPolicy resources
+              are:
+
+              - `hub.jupyter.org/network-access-hub: "true"` (hub)
+              - `hub.jupyter.org/network-access-proxy-http: "true"` (proxy.chp, proxy.traefik)
+              - `hub.jupyter.org/network-access-proxy-api: "true"` (proxy.chp)
+              - `hub.jupyter.org/network-access-singleuser: "true"` (singleuser)
+          allowedIngressPorts:
+            type: list
+            description: |
+              A rule to allow ingress on these ports will be added no matter
+              what the origin of the request is. The default setting for
+              `proxy.chp` and `proxy.traefik`'s networkPolicy configuration is
+              `[http, https]`, while it is `[]` for other networkPolicies.
       db:
         type: object
         properties:
@@ -535,6 +579,7 @@ properties:
           Configure the configurable-http-proxy (chp) pod managed by jupyterhub to route traffic
           both to itself and to user pods.
         properties:
+          networkPolicy: *networkPolicy-spec
           extraCommandLineFlags:
             type: list
             description: |
@@ -806,6 +851,7 @@ properties:
 
               See the [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)
               to learn more about labels.
+          networkPolicy: *networkPolicy-spec
           extraEnv:
             type: object
             description: |
@@ -880,6 +926,7 @@ properties:
     description: |
       Options for customizing the environment that is provided to the users after they log in.
     properties:
+      networkPolicy: *networkPolicy-spec
       podNameTemplate:
         type: string
         description: |
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -29,6 +29,14 @@ spec:
   # Read more at: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors
   #
   ingress:
+    {{- with .Values.hub.networkPolicy.allowedIngressPorts }}
+    # allow incoming traffic to these ports independent of source
+    - ports:
+      {{- range $port := . }}
+      - port: {{ $port }}
+      {{- end }}
+    {{- end }}
+
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -33,14 +33,19 @@ spec:
     - ports:
         - port: http
       from:
+        # source 1 - labelled pods
+        - podSelector:
+            matchLabels:
+              hub.jupyter.org/network-access-hub: "true"
+          namespaceSelector:
+            matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
+        # source 2 - pods in labelled namespaces
         - namespaceSelector:
-            matchLabels: {}   # allow the label below to be set on a pod in any namespace
-          podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
 
     {{- with .Values.hub.networkPolicy.ingress }}
-    # default: nothing --> hub
+    # depends, but default is nothing --> hub
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end }}
 
@@ -62,7 +67,7 @@ spec:
               {{- $_ := merge (dict "componentLabel" "singleuser-server") . }}
               {{- include "jupyterhub.matchLabels" $_ | nindent 14 }}
 
-    # hub -> Kubernetes internal DNS
+    # hub --> Kubernetes internal DNS
     - ports:
         - protocol: UDP
           port: 53
@@ -70,7 +75,7 @@ spec:
           port: 53
 
     {{- with .Values.hub.networkPolicy.egress }}
-    # hub --> default: everything
+    # hub --> depends, but the default is everything
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end }}
 {{- end }}
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -33,16 +33,27 @@ spec:
     - ports:
         - port: http
       from:
+        {{- with .Values.hub.networkPolicy.interNamespaceIngressForHTTPPort }}
         # source 1 - labelled pods
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
+          {{- if eq . "requireEitherPodOrNamespaceLabel" }}
           namespaceSelector:
             matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
+          {{- else if eq . "requireBothPodAndNamespaceLabel" }}
+          namespaceSelector:
+            matchLabels:
+              hub.jupyter.org/network-access-hub: "true"
+          {{- end }}
+
+        {{- if or (eq . "requireEitherPodOrNamespaceLabel") (eq . "requireNamespaceLabel") }}
         # source 2 - pods in labelled namespaces
         - namespaceSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
+        {{- end }}
+        {{- end }}
 
     {{- with .Values.hub.networkPolicy.ingress }}
     # depends, but default is nothing --> hub
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -13,6 +13,21 @@ spec:
     - Ingress
     - Egress
 
+  # IMPORTANT:
+  # NetworkPolicy's ingress "from" and egress "to" rule specifications require
+  # great attention to detail. A quick summary is:
+  #
+  # 1. You can provide "from"/"to" rules that provide access either ports or a
+  #    subset of ports.
+  # 2. You can for each "form"/"to" rule provide any number of
+  #    "sources"/"destinations" of four different kinds.
+  #    - podSelector                        - targets pods with a certain label in the NetworkPolicy's namespace
+  #    - namespaceSelector                  - targets all pods within namespaces with a certain label
+  #    - namespaceSelector and podSelector  - targets pods with a certain label within namespaces with a certain label
+  #    - ipBlock                            - targets network traffic from/to a set of IP address ranges
+  #
+  # Read more at: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors
+  #
   ingress:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -33,26 +33,17 @@ spec:
     - ports:
         - port: http
       from:
-        {{- with .Values.hub.networkPolicy.interNamespaceIngressForHTTPPort }}
-        # source 1 - labelled pods
+        # source 1 - labeled pods
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
-          {{- if eq . "requireEitherPodOrNamespaceLabel" }}
+        {{- if eq .Values.hub.networkPolicy.interNamespaceLabeledIngress "allow" }}
           namespaceSelector:
             matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
-          {{- else if eq . "requireBothPodAndNamespaceLabel" }}
-          namespaceSelector:
-            matchLabels:
-              hub.jupyter.org/network-access-hub: "true"
-          {{- end }}
-
-        {{- if or (eq . "requireEitherPodOrNamespaceLabel") (eq . "requireNamespaceLabel") }}
-        # source 2 - pods in labelled namespaces
+        # source 2 - pods in labeled namespaces
         - namespaceSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
-        {{- end }}
         {{- end }}
 
     {{- with .Values.hub.networkPolicy.ingress }}
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -23,7 +23,7 @@ spec:
   #    "sources"/"destinations" of four different kinds.
   #    - podSelector                        - targets pods with a certain label in the same namespace as the NetworkPolicy
   #    - namespaceSelector                  - targets all pods running in namespaces with a certain label
-  #    - namespaceSelector and podSelector  - targets pods with a certain label within namespaces with a certain label
+  #    - namespaceSelector and podSelector  - targets pods with a certain label running in namespaces with a certain label
   #    - ipBlock                            - targets network traffic from/to a set of IP address ranges
   #
   # Read more at: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -45,7 +45,7 @@ spec:
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
-        {{- if eq .Values.hub.networkPolicy.interNamespaceLabeledIngress "allow" }}
+        {{- if eq .Values.hub.networkPolicy.interNamespaceAccessLabels "accept" }}
           namespaceSelector:
             matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
         # source 2 - pods in labeled namespaces
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -18,7 +18,9 @@ spec:
     - ports:
         - port: http
       from:
-        - podSelector:
+        - namespaceSelector:
+            matchLabels: {}   # allow the label below to be set on a pod in any namespace
+          podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
 
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -19,10 +19,10 @@ spec:
   #
   # 1. You can provide "from"/"to" rules that provide access either ports or a
   #    subset of ports.
-  # 2. You can for each "form"/"to" rule provide any number of
+  # 2. You can for each "from"/"to" rule provide any number of
   #    "sources"/"destinations" of four different kinds.
-  #    - podSelector                        - targets pods with a certain label in the NetworkPolicy's namespace
-  #    - namespaceSelector                  - targets all pods within namespaces with a certain label
+  #    - podSelector                        - targets pods with a certain label in the same namespace as the NetworkPolicy
+  #    - namespaceSelector                  - targets all pods running in namespaces with a certain label
   #    - namespaceSelector and podSelector  - targets pods with a certain label within namespaces with a certain label
   #    - ipBlock                            - targets network traffic from/to a set of IP address ranges
   #
--- a/jupyterhub/templates/proxy/autohttps/netpol.yaml
+++ b/jupyterhub/templates/proxy/autohttps/netpol.yaml
@@ -48,7 +48,7 @@ spec:
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-proxy-http: "true"
-        {{- if eq .Values.proxy.traefik.networkPolicy.interNamespaceLabeledIngress "allow" }}
+        {{- if eq .Values.proxy.traefik.networkPolicy.interNamespaceAccessLabels "accept" }}
           namespaceSelector:
             matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
         # source 2 - pods in labeled namespaces
--- a/jupyterhub/templates/proxy/autohttps/netpol.yaml
+++ b/jupyterhub/templates/proxy/autohttps/netpol.yaml
@@ -72,7 +72,7 @@ spec:
               {{- $_ := merge (dict "componentLabel" "proxy") . }}
               {{- include "jupyterhub.matchLabels" $_ | nindent 14 }}
 
-    # proxy --> Kubernetes internal DNS
+    # autohttps --> Kubernetes internal DNS
     - ports:
       - protocol: UDP
         port: 53
@@ -80,7 +80,7 @@ spec:
         port: 53
 
     {{- with .Values.proxy.traefik.networkPolicy.egress }}
-    # proxy --> depends, but the default is everything
+    # autohttps --> depends, but the default is everything
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end }}
 {{- end }}
--- a/jupyterhub/templates/proxy/autohttps/netpol.yaml
+++ b/jupyterhub/templates/proxy/autohttps/netpol.yaml
@@ -31,12 +31,14 @@ spec:
   # Read more at: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors
   #
   ingress:
-    {{- if .Values.proxy.traefik.networkPolicy.allowHttpIngressUnconditionally }}
-    # allow any incoming traffic to these ports
+    {{- with .Values.proxy.traefik.networkPolicy.allowedIngressPorts }}
+    # allow incoming traffic to these ports independent of source
     - ports:
-      - port: http
-      - port: https
-    {{- else }}
+      {{- range $port := . }}
+      - port: {{ $port }}
+      {{- end }}
+    {{- end }}
+
     # allowed pods (hub.jupyter.org/network-access-proxy-http) --> proxy (http/https port)
     - ports:
         - port: http
@@ -54,7 +56,6 @@ spec:
             matchLabels:
               hub.jupyter.org/network-access-proxy-http: "true"
         {{- end }}
-    {{- end }}
 
     {{- with .Values.proxy.traefik.networkPolicy.ingress}}
     # depends, but default is nothing --> proxy
--- a/jupyterhub/templates/proxy/autohttps/netpol.yaml
+++ b/jupyterhub/templates/proxy/autohttps/netpol.yaml
@@ -1,12 +1,10 @@
 {{- $HTTPS := .Values.proxy.https.enabled -}}
 {{- $autoHTTPS := and $HTTPS (and (eq .Values.proxy.https.type "letsencrypt") .Values.proxy.https.hosts) -}}
-{{- $manualHTTPS := and $HTTPS (eq .Values.proxy.https.type "manual") -}}
-{{- $manualHTTPSwithsecret := and $HTTPS (eq .Values.proxy.https.type "secret") -}}
-{{- if .Values.proxy.chp.networkPolicy.enabled -}}
+{{- if and $autoHTTPS .Values.proxy.traefik.networkPolicy.enabled -}}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: proxy
+  name: autohttps
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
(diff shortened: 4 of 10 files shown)


manics:
If you're running multiple JupyterHubs segregated by namespace you wouldn't want cross-namespace traffic.
Let me think about this a bit more....

consideRatio:
> If you're running multiple JupyterHubs segregated by namespace you wouldn't want cross-namespace traffic.
Let me think about this a bit more....

I guess yes, hmmm, maybe? I mean, you don't want it for pods without clear permissions to do so at least - right? But, is it acceptable for a pod with a dedicated label?

manics:
The labels were originally added to make the definitions of policies between JupyterHub components easier: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/534#issuecomment-368903903

I don't think it's as useful to enforce those labels outside of the JupyterHub namespace. In general an application owner should have control over who can access a service, but this effectively hands over control to the owner of the other namespace. This is a potential problem when namespaces are run by people with different privileges, or where it's used to reduce the impact of an exploit in a service.

How about if we say the labels are only intended for use within the same namespace, and for cross namespace traffic labels are irrelevant since it doesn't add much security. Then if you want to allow ingress from a namespace `other-ns` it should be enough to add `other-ns` to the allowed list and not require labels to be added to the other app.

If the other app owner wants to limit a pod to only connect to JupyterHub they can do that by defining their own egress network policy.

manics:
Do you have any examples of how other Helm charts configure network policies? For instance if all charts follow the Z2JH approach then if you had say 50 applications that need to interact with each other each of those 50 applications would need 49 annotations like `<other-application-X>/network-access: "true"`. This sounds a bit wrong to me, but if that's what other charts do then we could go along with it.

What ingress did you have problems with? I've always used nginx-ingress in its own namespace without a specific exemption in the Z2JH network policy- I just assumed ingress had some sort of special access. Maybe something's changed recently?

consideRatio:
> Do you have any examples of how other Helm charts configure network policies? For instance if all charts follow the Z2JH approach then if you had say 50 applications that need to interact with each other each of those 50 applications would need 49 annotations like <other-application-X>/network-access: "true". This sounds a bit wrong to me, but if that's what other charts do then we could go along with it.

Hmmm I  mean right now, all traffic is banned from all other namespaces, and this is a mechanism to open up communication, so unless the policy is to allow communication from all other namespaces, it doesn't become much easier other than creating a namespace label that will allow traffic for all pods within the namespace. That would make sense to add as well, because why not.

If we don't have this PR merged, then I'd do this to resolve it is...

## Option 1

```yaml
proxy:
  networkPolicy:
    ingress:
      from:
        - namespaceSelector:
            matchLabels: {}
          podSelector:
            matchLabels:
              hub.jupyter.org/network-access-proxy-http: "true"
```

Then update the nginx-ingress helm chart, like I'd do otherwise as well.

```yaml
controller:
  podLabels:
    hub.jupyter.org/network-access-proxy-http: "true"
```

## Option 2

```yaml
proxy:
  networkPolicy:
    ingress:
      from:
        - namespaceSelector:
            matchLabels:
              hub.jupyter.org/network-access-proxy-http: "true"
```

```
kubectl label ns hub.jupyter.org/network-access-proxy-http="true"
```

# Summary of current situation
We need to have a suggestion on how users should configure access to various jupyterhub pods that otherwise would block access from any pod outside the jupyterhub namespace. The concrete case that made me realize this, is that I had a nginx-ingress-controller in another k8s namespace that directed traffic to the proxy-public service / the proxy pod, but that traffic was denied by the network policy rules. It was a surprise to me that a ingress rule without a namespaceSelector rule explicitly part of the traffic to the local namespace.

The proposal of this PR is to simply allow the various JupyterHub pods to allow traffic incoming traffic from pods in all namespaces with a dedicated label, such as `hub.jupyter.org/network-access-proxy-http: "true"`.

Another option to this, would be to ask users to add a custom ingress rule they manage themselves, like below. This rule would require both a namespace being labeled correctly, and have a pod within it labelled correctly, to send traffic to a specific port, on on the proxy pod.

```yaml
proxy:
  networkPolicy:
    ingress:
      from:
        - namespaceSelector:
            matchLabels:
              my-specific-namespace-label: my-specific-namespace-label-value
          podSelector:
            matchLabels:
              my-specific-pod-label: my-specific-pod-label-value
          ports:
            - port: http  # or https depending on other configuration
```

To learn how to specify networkPolicies is quite tricky, I'd say it is the most complicated resource to get right of all the resources we manage in this Helm chart.

---

> I've always used nginx-ingress in its own namespace without a specific exemption in the Z2JH network policy- I just assumed ingress had some sort of special access. Maybe something's changed recently?

I'm not sure, to me it sounds weird that a in-cluster-pod in another namespace should be allowed access on the merit it can be considered functioning as a loadbalancer, because how would it proove it was acting as a loadbalancer to the networkpolicycontroller that enforce the rules?

Perhaps I locked down the proxy policy in a recent update to the networkpolicy resources without thinking about it? Hmm... yes...

image

In the past, we had a rule saying that as long as anyone wants to speak specifically with the proxy pod's HTTP or HTTPS port, that was fine, because specifying only port doesn't block ingress from other namespaces I assume.

I didn't realize that I changed the behaviour when I made the change in #1670 - a testiment to networkPolicy resources are complicated to manage in my mind.

manics:
My suggestion for allowing ingress from other namespaces is for the Z2JH config to list those other namespaces it's allowing ingress from (which may include all) and not bother checking labels for pods outside the current namespace. i.e. the resulting network ingress policy is the union of:
- current ingress policy (specific label for pod in the same namespace)
- ingress from any pods in a list of allowed namespaces regardless of labels

I think this avoids reducing security when namespaces are used for segregation of users with different privileges, and avoids users having to add labels to their own pods to access jupyterhub which can be tricky as each helm chart may do it differently.

consideRatio:
[This section of the k8s documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors) is very relevant details on the rules that we can set in a NetworkPolicy from the k8s docs, distinguishing between: a) podSelector, b) namespaceSelector, c) podSelector AND namespaceSelector, and d) ipblock.

yuvipanda:
I currently have the following lines to enable network policy and also allow ingress from a set of namespaces:

https://github.com/2i2c-org/pilot-hubs/blob/master/hub/values.yaml#L71
https://github.com/2i2c-org/pilot-hubs/blob/master/support/values.yaml#L6

I wrote https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1380 a while ago to make this possible.

I think Option 1 you described is the way to go, @consideRatio.

consideRatio:
@yuvipanda thank you for your input! I just updated https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1842#issuecomment-710641494 with two concrete examples of: a) making it possible to out of Option 1, and b) adding a new way to allow all pods in namespaces with a certain label.

Do you think we should write such configuration, or just go with them by default without exposing configuration to explain and maintain?

---

@yuvipanda as a separate note, in your pilot-hubs configuration, you have a from rule like this added with two separate sources.

```yaml
      ingress:
      - from:
        - namespaceSelector:
            matchLabels:
              name: support
        - podSelector:
            matchLabels:
              hub.jupyter.org/network-access-proxy-http: "true"
        ports:
        - port: http
```

With the two separate sources of traffic that you allow in this rule, either any pod from the support namespace, or from a pod in the same namespace labelled with the access label. The second source is already allowed by default currently. I suspect that you wanted to use a "podSelector AND namespaceSelector" source rather than two separate sources.

consideRatio:
@manics @yuvipanda I updated to also allow labelled namespaces and not only labelled pods in other namespaces.

In my mind, merge or not now depends on: do we see it as a problem that we expose hub/singleuser-server/proxy for communication _from_ other pods in other namespaces assuming they have the correct label?

I would argue that this is not an escalation compared to in 0.9.1 where all http/https traffic is allowed to the proxy pod from other namespaces until #1670 was merged, and the proxy pod expose the singleuser servers and the hub server.

I'm open to the idea that we make the inter-namespace ingress for hub/proxy http(s)/proxy api/singleuser-server configurable by a default, which we set to false by default for all pods except proxy pods http(s) ingress. The downside of this though, is it becomes quite a bit more complicated to correctly communicate this to users who may assume adding a label like `hub.jupyter.org/network-access-hub: "true"` would be enough.

jupyterhub/templates/hub/netpol.yaml
```diff
@@ -13,17 +13,39 @@ spec:
...
+        # source 1 - labelled pods
         - podSelector:
             matchLabels:
               hub.jupyter.org/network-access-hub: "true"
+          namespaceSelector:
+            matchLabels: {}   # without this, the podSelector would only consider pods in the local namespace
+        # source 2 - pods in labelled namespaces
+        - namespaceSelector:
+            matchLabels:
+              hub.jupyter.org/network-access-hub: "true"
```
consideRatio, line 45: This is the key part to review, which is repeated four times across the PR.

manics: What do you think about moving source 2 to values.yaml? Currently `networkPolicy.ingress: []`, if source 2 is removed from here and `networkPolicy.ingress` is changed to:
```yaml
hub:
  networkPolicy:
    ingress:
      - namespaceSelector:
            matchLabels:
```
This makes it easy to override, either to allow all traffic without labels, or to block this traffic restoring the original behaviour?

manics: On second thoughts maybe I meant:
```
      from:
        # source 1 - labelled pods
        - podSelector:
            matchLabels:
```
_Without_ the namespaceSelector:

then in Values.yaml
```yaml
hub:
  networkPolicy:
    ingress:
      - namespaceSelector:
            matchLabels:
            namespaceSelector:
              matchLabels: {}
      - namespaceSelector:
          matchLabels:
```

In other words `netpol.yaml` contains the rules that are strictly necessary for JupyterHub to function, then Values.yaml contains the rules that will work for most people but with enough flexibility for them to be relaxed or restricted.

consideRatio:
```yaml
hub:
  networkPolicy:
    ingress:
      - ports:
        - port: http
        from:
        - podSelector:
            matchLabels:
            namespaceSelector:
              matchLabels: {}
        - namespaceSelector:
            matchLabels:
```

This would be the equivalent thing I think. Hmm... I think this is just massively complicated, and I think the maintenance complexity and such for anyone making anything custom will be too much for what we can manage.

I'd say we have it as I suggest, and if strictly required, we have a toggle to influence that default to not accept inter-namespace labeling.

consideRatio: Imagine for example if a user wants to add a rule, without changing the defaults - since hub.networkPolicy.ingress is a list, it will be replaced by the user which in my mind likely will be a not-intentional action.

I think it may be only you and me who will be able to help when users run into issues stemming from such configuration changes, because I doubt anyone else will be able to do that without significant code inspection - which you and me have right now from working on this. I think we are stretched thin as it is to maintain z2jh, kubespawner, binderhub, and the k8s related jupyterhub repos.

manics: Maybe if we work through one or more examples that'll help to see what the impact is on users?

E.g. On clusters where namespaces are used to segregate users:
- I don't have admin on other namespaces
- I don't want ingress from other namespaces unless I say so
- I'd like to enable ingress from the namespace `nginx-ingress` where there's a cluster level ingress controller

What would my Z2JH config be?

Feel free to add the other examples you have in mind.

manics: Expanding further on my point, my understanding of this PR is that if you have cluster admin rights there should be no need to modify the default ingress policy since the recommendation is to add labels to your pods? Can you give some examples of ingress rules that you can see being required?

manics: Expanding further on my point, my understanding of this PR is that if you have cluster admin rights there should be no need to modify the default ingress policy since the recommendation is to add labels to your pods? Can you give some examples of ingress rules that you can see being required?

consideRatio: @manics deal!

## Overview of what we configure ingress for

We have networkPolicy resources declaring allowed ingress from four locations, for three pods.
- proxy's API port
- proxy's HTTP(S) port(s)
- hub's HTTP port
- singleuser-server's HTTP port

There is already some predefined escalation of access by going through the proxy's HTTP(s) port(s) though, which typically are publicly exposed on the internet, as you can it proxies to the singleuser-server's HTTP port through it (/user/myusername) and the hub's port through it (/hub). The proxy's API port should probably not be publicly exposed, or perhaps ever used except by JupyterHub.

## Ability to configure zero trust
Assuming we have no trust for other namespaces, and think they maliciously want to sent crap to our pods in the JupyterHub namespace, then we can make it configurable to disallow that.

The "risky" configuration discussed is that we allow a) under "source 1" pods with the right labels in other namespaces with to access specific pods in our namespace, and b) pods in namespaces the right labels.

### Source 1 lockdown
We can lock down source 1 in two steps compared to this PR. Source 1 is now is a `podSelector and namespaceSelector`, and we can make it just become a `podSelector`, that is the most tight lock down. The second tightest lockdown is to require that the namespace is also correctly labelled.

Note that we cannot specify individual namespaces, only select them based on labels.

### Source 2 lockdown
We can lock down source 2 only with a single step. Source 2 is now a `namespaceSelector`, and we can make it not be added with a boolean flag per networkPolicy resource

### Source 1 and 2 configuration states
We get 3x2 permutations with some nonsensical combinations.

Permutation # | Source 1 configuration | Source 2 configuration | Note
-|-|-|-
1 | Local only | Disabled | __No inter-namespace ingress allowed.__
2 | Namespace labelled | Disabled | __Inter-namespace ingress allowed if both namespace and pod is labelled.__
3 | All namespaces | Disabled | __Practically nonsensical__. To only allow Pod labels, but not namespace labels, is probably to allow a lower set of permissions I think?
4 | Local only | Enabled | __Practically nonsensical__ - This just limit the configuration resolution in comparison to other options.
5 | Namespace labelled | Enabled | __Theoretically nonsensical__ - Source 1 will never be needed.
6 | All namespaces | Enabled | __Inter-namespace ingress is allowed if either namespace or pod is labelled.__

There are in other words three reasonable states:
1. No inter-namespace ingress allowed.
2. Inter-namespace ingress allowed if both namespace and pod is labelled.
3. Inter-namespace ingress is allowed if either namespace or pod is labelled.

### Conclusion
We would provide a configuration that has three states for our four different ingress sources and associated labels.

```yaml
hub|singleuser|proxy:
  networkPolicy:
    interNamespaceIngressForHTTPPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel
    # for proxy.networkPolicy, we would also have...
    interNamespaceIngressForAPIPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel
```

consideRatio: I guess permutation 4 is potentially relevant, because it may be deemed safer to allow a namespace label than a pod label, and one may not want care about ending up allowing all pods in that namespace.

```yaml
hub|singleuser|proxy:
  networkPolicy:
    interNamespaceIngressForHTTPPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel | requireNamespaceLabel
    # for proxy.networkPolicy, we would also have...
    interNamespaceIngressForAPIPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel | requireNamespaceLabel
```

manics: Thanks for explaining, I now understand what your thinking is!
Is the latest proposal effectively three ingress network policies, two controlled by booleans and one `<anything>`?
```yaml
hub|singleuser|proxy:
  networkPolicy:
    interNamespaceIngressForHTTPPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel | requireNamespaceLabel
    interNamespaceIngressForAPIPort: deny | requireBothPodAndNamespaceLabel | requireEitherPodOrNamespaceLabel | requireNamespaceLabel
    ingress: []
```
If so it's not my preferred option but I can live with it :smile: 

An alternative to `interNamespaceIngressForAPIPort` is to add hub-api to the list of components: `hub|hubapi|singleuser|proxy` so you just have `interNamespaceIngress`, dunno if that's better or worse though.

My other concern was whether we're being "good citizens" in the helm/k8s ecosystem by imposing the use of labels this way? It's fine to say Z2JH is aimed at people with beginner to intermediate experience of K8S, and that we recommend people use labels because we think this is the easiest solution (and if you're an advanced user these are the instructions to do your own thing....), but for me it'd be good to be explicit about that. Getting off topic, but stating this will also help evaluate feature requests- do we add an easy to configure option, or do we add an override and say "override this here, but at your own risk"

consideRatio: > Is the latest proposal effectively three ingress network policies, two controlled by booleans and one <anything>?

Hmmmm, I don't understand... But I meant....

```yaml
# ingress / egress are additional rules that can permit
# other kinds of traffic other than what the default configured rules will provide

# Allowed options for interNameSpaceIngressForXXXXPort are:
# - deny
# - requireBothPodAndNamespaceLabel
# - requireEitherPodOrNamespaceLabel
# - requireNamespaceLabel

# Default values suggested below
hub:
  networkPolicy:
    interNamespaceIngressForHTTPPort: deny 
    ingress: []
    egress:
      - to:
          - ipBlock:
              cidr: 0.0.0.0/0

proxy:
  networkPolicy:
    interNamespaceIngressForHTTPPort: requireEitherPodOrNamespaceLabel
    interNamespaceIngressForAPIPort: deny
    ingress: []
    egress:
      - to:
          - ipBlock:
              cidr: 0.0.0.0/0

singleuser:
  networkPolicy:
    interNamespaceIngressForHTTPPort: deny
    ingress: []
    egress:
      - to:
          - ipBlock:
              cidr: 0.0.0.0/0
              except:
                - 169.254.169.254/32
```

> If so it's not my preferred option but I can live with it smile

Hehe, pweh! I would not be able to live with the idea of users copy-pasting complicated configuration which end up becoming outdated minimal requirements in the future etc, or just the fact that it becomes extremely hard to understand how something has gone wrong if that becomes part of the bug reports - then everyone need to be able to perfectly spot syntax errors in that configuration to rule out its an issue, and even those who can will loose time on ruling it out as the cause of the issue reported.

> My other concern was whether we're being "good citizens" in the helm/k8s ecosystem by imposing the use of labels this way?

We are imposing them as part of secure defaults, which can be disabled. If you use network policy resources, that actually has some meaning by locking thigns down, then in my mind, we are not imposing anything other than what is stemming from secure practices part of k8s - they have chosen the NetworkPolicy resource API where labels of pods/namespaces have impact.

An option is to allow ingress from all namespaces without any label to the proxy pod specifically by default, this would align with the 0.9.0/0.9.1 configuration, which I consider a bug as it made the label we had pointless.

Ah, okay, so we have a final option in practice: `allowAllIngress: false`.

> do we add an easy to configure option, or do we add an override and say "override this here, but at your own risk"

I feel that there has been a bit of requests like this, and even though people get to use things at their own risk, it typically end up being problematic for maintainers in the end anyhow. Example, someone overrides the hub image, then they report an issue, and suddenly the issue's cause can stem from another source that it couldn't if the user had not had the option to override the hub image. I have spent several hours debugging these kinds of issues to end up concluding it was because this. Due to that experience, I don't think it helps that much to add a "use at your own risk" note.

consideRatio: >An alternative to interNamespaceIngressForAPIPort is to add hub-api to the list of components: hub|hubapi|singleuser|proxy so you just have interNamespaceIngress, dunno if that's better or worse though.

@manics The hub doesn't expose its API on a dedicated port, but the proxy server does. It has a REST API to get configured by JupyterHub in how it should route traffic on its port that proxies traffic to hub/users/reigstered services

consideRatio:
@manics okay so I implemented `interNamespaceIngressForAPIPort` options for the networkPolicy resources.

But... Note that we have an autohttps pod which is not covered by any policy, and will route traffic to the proxy pod, and onwards to the hub pod and user pods.

Hmm... In general, if you have no policy targetting a pod, it can ingress/egress like it wants. But if you have default network policy targetting all pods regarding ingress specifically, and no rules allowing traffic, then it implies they are denied all ingress unless another policy allows it.

Hmmmm hmmm hmmm... I think we should really cover the autohttps pod with these network policies as well at least... Perhaps not create a default deny rule to influence pods in the namespace we don't know about though.

manics:
Cheers!

After sleeping on it I wonder if we could simplify the options to just one per component? Your intention was to have the defaults "just work" for the majority of people, my wish was to have an easy way to lock-down the setup with additional ingress done through the `<component>.netpol.ingress: [rules]`. To me this suggests the fine grained configuration offered by multiple options isn't needed. What do you think of:

```yaml
hub:
  networkPolicy:
    interNamespaceLabeledIngress: allow|deny (or enabled|disabled, true|false)
    ingress: []

proxy:
  networkPolicy:
    # Covers both HTTP and API, since 
    interNamespaceLabeledIngress: allow|deny (or enabled|disabled, true|false)
    ingress: []

singleuser:
  networkPolicy:
    interNamespaceLabeledIngress: allow|deny (or enabled|disabled, true|false)
```
where `interNamespaceLabeledIngress` controls both sources 1 and 2, and the default is whatever you prefer.

consideRatio:
@manics thanks for iterating with me on this!

I'll go with your suggestion of having a `interNamespaceLabeledIngress` which toggle the ability to add a label to any pod in any namespace, or a label to an entire namespace. In other words, a toggle between `requireEitherPodOrNamespaceLabel` and  `deny`.

## Currently aiming for...
- [ ] _Create a NetworkPolicy resource for autohttps (coupled with `proxy.traefik.networkPolicy`)_
- [ ] _Deprecate `proxy.networkPolicy` in favor of `proxy.chp.networkPolicy`_
- [ ] _Add `interNamespaceLabeledIngress` options with `allow|block`_
- [ ] _Update `schema.yaml`_

manics:
With https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1852 it might be better to rename `allowHttpIngressUnconditionally` to something like `allowAllIngressPorts`?

consideRatio:
I'm unhappy about exposing the proxy API port by default, I think that should be locked down.

manics:
Good point. Perhaps `allowAllowedIngressPorts`, default [80, 443] (or 8443?)? And if you've add ports using https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1852 you also have to add them to `allowAllowedIngressPorts`? Or do you think that should be automatic?

consideRatio:
> Good point. Perhaps allowedIngressPorts, default [80, 443] (or 8443 or whatever?)? And if you've added ports using #1852 you also have to add them yourself to allowedIngressPorts? Or do you think that should be automatic?

I think this is an excellent idea!

consideRatio:
@manics I've updated the original post to reflect the most recent changes: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1842#issue-504786225

jupyterhub/templates/proxy/autohttps/netpol.yaml
```diff
@@ -0,0 +1,86 @@
...
+    # autohttps --> proxy (http port)
+    - ports:
+        - port: 8000
+      to:
+        - podSelector:
+            matchLabels:
+              {{- $_ := merge (dict "componentLabel" "proxy") . }}
+              {{- include "jupyterhub.matchLabels" $_ | nindent 14 }}
+
+    # proxy --> Kubernetes internal DNS
```
consideRatio, line 75:
```suggestion
    # autohttps --> Kubernetes internal DNS
```

jupyterhub/templates/proxy/autohttps/netpol.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - ports:
+      - protocol: UDP
+        port: 53
+      - protocol: TCP
+        port: 53
+
+    {{- with .Values.proxy.traefik.networkPolicy.egress }}
+    # proxy --> depends, but the default is everything
```
consideRatio, line 83:
```suggestion
    # autohttps --> depends, but the default is everything
```

jupyterhub/schema.yaml
```diff
@@ -191,6 +191,50 @@ properties:
...
+            description: |
+              Toggle the creation of the NetworkPolicy resource for this pod.
+          ingress:
+            type: list
+            description: |
+              Additional ingress rules to add.
+          egress:
+            type: list
+            description: |
+              Additional egress rules to add.
```
consideRatio, line 211:
```suggestion
              Additional egress rules to add except those that is known to be needed by the Helm chart.

              The default value of this egress is to allow all traffic, except for the `singleuser.networkPolicy.egress`, which is also limiting access to a metadata server that can be exploited.

              If you want to restrict egress, you can override this permissive default to be an empty list.
```

jupyterhub/schema.yaml
```diff
@@ -191,6 +191,50 @@ properties:
...
+          _NetworkPolicy resource_.
+        properties:
+          enabled:
+            type: bool
+            description: |
+              Toggle the creation of the NetworkPolicy resource for this pod.
+          ingress:
+            type: list
+            description: |
+              Additional ingress rules to add.
```
consideRatio, line 207:
```suggestion
              Additional ingress rules to add except those that is known to be needed by the respective pods in the Helm chart.
```

jupyterhub/schema.yaml
```diff
@@ -191,6 +191,50 @@ properties:
...
+            type: string
+            enum:
+              - accept
+              - ignore
+            description: |
+              This configuration option determines if both namespaces and pods
+              in other namespaces that have specific access labels, should be
+              accepted to allow ingress (set to `accept`), or, if the labels are
+              to be ignored when applied outside the local namespace (set to
+              `ignore`).
```
consideRatio, line 222:
```suggestion
              This configuration option determines if both namespaces and pods
              in other namespaces, that have specific access labels, should be
              accepted to allow ingress (set to `accept`), or, if the labels are
              to be ignored when applied outside the local namespace (set to
              `ignore`).
```

jupyterhub/schema.yaml
```diff
@@ -191,6 +191,50 @@ properties:
...
+              - `hub.jupyter.org/network-access-proxy-http: "true"` (proxy.chp, proxy.traefik)
+              - `hub.jupyter.org/network-access-proxy-api: "true"` (proxy.chp)
+              - `hub.jupyter.org/network-access-singleuser: "true"` (singleuser)
+          allowedIngressPorts:
+            type: list
+            description: |
+              A rule to allow ingress on these ports will be added no matter
+              what the origin of the request is. The default setting for
+              `proxy.chp` and `proxy.traefik`'s networkPolicy configuration is
+              `[http, https]`, while it is `[]` for other networkPolicies.
```
consideRatio, line 237:
```suggestion
              `[http, https]`, while it is `[]` for other networkPolicies.
              
              Note that these port names or numbers targets a Pod's port name or
              number, not a k8s Service's port name or number.
```

consideRatio:
The build failure is unrelated - I don't want to restart it to make it go green because it contains valuable logs for debugging it.

jupyterhub/templates/proxy/autohttps/netpol.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - Ingress
+    - Egress
+
+  # IMPORTANT:
+  # NetworkPolicy's ingress "from" and egress "to" rule specifications require
+  # great attention to detail. A quick summary is:
+  #
+  # 1. You can provide "from"/"to" rules that provide access either ports or a
+  #    subset of ports.
+  # 2. You can for each "form"/"to" rule provide any number of
```
consideRatio, line 24:
```suggestion
  # 2. You can for each "from"/"to" rule provide any number of
```

jupyterhub/templates/hub/netpol.yaml
```diff
@@ -13,17 +13,49 @@ spec:
...
 
+  # IMPORTANT:
+  # NetworkPolicy's ingress "from" and egress "to" rule specifications require
+  # great attention to detail. A quick summary is:
+  #
+  # 1. You can provide "from"/"to" rules that provide access either ports or a
+  #    subset of ports.
+  # 2. You can for each "form"/"to" rule provide any number of
+  #    "sources"/"destinations" of four different kinds.
+  #    - podSelector                        - targets pods with a certain label in the NetworkPolicy's namespace
```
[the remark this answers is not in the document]
consideRatio:
```suggestion
  #    - podSelector                        - targets pods with a certain label in the same namespace as the NetworkPolicy
```

jupyterhub/templates/proxy/autohttps/netpol.yaml
```diff
@@ -0,0 +1,86 @@
...
+    {{- include "jupyterhub.labels" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "jupyterhub.matchLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
+
+  # IMPORTANT:
```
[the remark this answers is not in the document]
consideRatio: Duplicating them breaks the DRY principle, but it also ensures this information is made available to everyone working with the resource which I felt was the most important part.

NetworkPolicy resources are sooooooo hard to get right, and our tests doesn't test all our rules etc.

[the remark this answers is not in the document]
consideRatio: I updated all comments to reflect your suggestions in the hub/netpol.yaml file

minrk:
The other use case that I can see coming up is folks deploying *users* in their own namespaces. These deployments definitely need cross-namespace access from singleuser pods -> hub API.

I wonder if our lack of using the release name in our resources is coming up here again. Would it be simpler or more confusing to have two labels, such as:

1. `hub.jupyter.org/network-access-hub` - within namespace only (current behavior)
2. `hub.jupyter.org/network-access-hub-{{.Release.Name}}` - across namespaces

or more a descriptive `hub.jupyter.org/cross-namespace-network-access-hub`

The idea being that rather than having network policies with our own customized configuration options, we have two simpler/more static ingress policies. Is that a good idea, or no? The main downside I can see of this being unconditional is that it provides a way for pods in another namespace who know a release name to grant themselves access. Maybe that's not ideal, and what we have here is the best option.

CHANGELOG.md
```diff
@@ -16,17 +16,17 @@ Here you can find upgrade changes in between releases and upgrade instructions.
...
 
-  With network policies enabled, you may require additional configuration,
-  especially for deployments that include additional components that access
-  JupyterHub pods directly (i.e. not through the `proxy-public` service).
-
   See the [security
   documentation](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#kubernetes-network-policies)
   for more details on this.
 
+- The Helm chart configuration `proxy.networkPolicy` has been removed,
```
minrk, line 26: This change is my biggest concern here. I don't love it because it seems to be baking too much of the implementation into the configuration. When we finally manage to land the traefik proxy implementation (#1162), this configuration structure will change again when nothing relevant to *users* should really need. What do *users* need to know about the separation of autohttps and chp that we couldn't represent without separating them?

consideRatio: We will still have a dedicated proxy for TLS termination i think, given that traefik doesnt support HA with lets encrypt.

This helm chart has a separate pod for TLS termination atm, and combining the config of two pods into one has caused it to be confusion in general. I'm think im for doing this even if it was a standalone refactoring without any other purpose to reduce that confusion.

If you feel strongly about it, i can revert it back, locating chp netpol config under proxy.netpol again.

consideRatio:
Thank you for reviewing this @minrk!

Hmmmmm, if we get away with simpler logic in all cases, i'm for simpler logic. My guiding compass feel very strongly about avoiding users needing to add their custom netpol rule injections at this point, because it is just so hard to understand for everybody involved and will become fragile with changes.

If we can get away with simpler logic partially relates to if inter-namespace traffic can accept always going through the proxy pod, or if they need direct access without tls termination etc to the hub pod etc.

> The main downside I can see of this being unconditional is that it provides a way for pods in another namespace who know a release name to grant themselves access.

I had a very long discussion with @manics related to the security implications in a now resolved thread. I think that access to the outer proxy (autohttps, or proxy chp) seems reasonable to allow by active choice for someone powerful to label pods in some namespace.

Now we lock down inter namespace communication by default except for the autohttps and proxy pods http/https ports, but leave the proxy api port blocked by default.

---

It is a complicated API to design, and i can make changes as wished but i feel strongly about the following:
- that we ensure to avoid making users need to specify k8s native ingress rules
- that we get a solution merged without too much further delay, in other words clear decisions is what i want at this point and is willing to compromise on the perfect choice, so we fix a known issue and unblock 0.9.0-beta.1 for release.

manics:
Other than the last discussion about whether to separate CHP and the http pod or to have a single proxy config I think this PR is a good solution. It provides:
1. a clear way to enable inter-namespace ingress to the pods/ports we think are reasonable by adding labels to pods or namespaces (`interNamespaceAccessLabels: accept`)
2. a way to disable default inter-namespace ingress (`interNamespaceAccessLabels: ignore`)
3. a way to add your own NetworkPolicies _independent_ of the previous two points- so you can set the above to allow no ingress and add you own rules, or you can have the default rules and add to them without needing to redefine them

(1) and (3) cover @consideRatio's wishes, (2) and (3) cover mine. My preference is to default to more secure `interNamespaceAccessLabels: ignore` but I'm not opposed to changing that to `accept` if the docs are clear.

minrk:
Thanks for your comments, and for thinking so carefully about this! What I was trying to get at is that I think defining network policies separately for traefik and chp means that users who need to touch this configuration need to know more about how our https handling *works* than they should have to.

1. I do not think that cross-namespace access to the CHP API is something that's important to enable except for *extremely* advanced users, so separating those for traefik and CHP doesn't seem useful to me
2. access to the public entrypoint is the main thing that folks are likely to want to control, if they want to control anything, and which pod that targets is *different* depending on https values (manual or no ssl goes directly to chp while autohttps goes to traefik). I don't think the user's network policy configuration should change based on whether SSL is terminated at CHP, traefik, or outside.

That's why I think we should have only a *single* configuration to set the accessibility of the "external entrypoint", and it should be up to us to make sure that's applied to the right target, not the user.

Other than that detail, I think this PR is a good solution, and I would just like to consolidate the two network policy configurations into one (or assign them more practical and less implementation-detail names if there's a reason they need to be separate that I haven't understood such as `public` and `api` to match the service names)

CHANGELOG.md
minrk: To me, this change seems like it would *increase* confusion to have two network policies applying to what users should see as one resource (and indeed will be one resource if we manage to land the traefik proxy). I am thinking of the example of

> Q: I set proxy.chp.networkPolicy, why is it not having any effect?
> A: that's because you have enabled autohttps, which means that proxy.traefik.networkPolicy governs proxy-public, not proxy.chp.networkPolicy. If you change how you configure https, your network policy will do what you expect, or you need to move the configuration to proxy.traefik.networkPolicy

Or is that not an accurate description of the behavior here?

consideRatio:
> I would just like to consolidate the two network policy configurations into one (or assign them more practical and less implementation-detail names if there's a reason they need to be separate that I haven't understood such as public and api to match the service names)

There is a lot of things to say, and when it comes down to it everything is possible, but I find consolidating them to be too complicated to document for users and maintainers to motivate a change. Here is a list of thoughts.

- For a tight restriction, we need a network policy for the autohttps pod, because if not it will expose the proxy pod through the autohttps traffic.
- For a tight restriction within the namespace, we need a network policy for the proxy pod as well so we can say that only the hub can access the proxy pod.
- Since we need to have two network policies, configuring them at a single location will make a very hard to understand configuration API both for users and maintainers in my mind.
- Thankfully, the current configuration defaults, which are like 0.9.1, is that you don't need to do anything because access to both autohttps/proxy is open to normal HTTP and HTTPS traffic by default
- The advanced used that wants to limit the 0.9.1 config further, can tweak `allowedIngressPorts: [http, https]` from `proxy.chp|traefik.networkPolicy`, which still will allow the hub pod for example to access the proxy pod's API and the autohttps pod to send traffic to the CHP proxy pods HTTP port.

@minrk perhaps we can have a chat about this to arrive at a resolution to go with? I'd like to hash out a concrete action plan and validate that you prefer that to the current PR suggestion.

consideRatio:
@minrk thank you for the video chat! I wanted to summarize a relevant point that we can pursue in the future.

## Idea to followup this PR
1. We want to target all individual pods that accepts network traffic with network policies to tighten the accepted communication. We do that in networkpolicy resources associated with the pods rather than the k8s services. This is how this PR is done.
2. We also want to target _JupyterHub as an installed application_ with a network policy, and this PR accomodates this by still being pod centered but letting the autohttps pod and proxy pod by default allow ingress on their http and https ports by default no matter what labels are used, but allows this to be tightened.
3. We can followup this PR by adding a network policy dedicated to represent the proxy-public k8s service! After we have restricted access to a pod, additional network policies will just open up access by adding more ways to allow access. We can therefore make a new one dedicated to the proxy-public service, targetting the autohttps or proxy pod dynamically. With such networkpolicy in place, we can set `proxy.chp.netorkPolicy.allowedIngressPorts=[]` and `proxy.traefik.netorkPolicy.allowedIngressPorts=[]` from their current defaults of `[http, https]` and instead rely on this proxy-public representative networkpolicy open up the http/https ports on the outermost pod.

consideRatio:
I'll merge this now given agreement that it can be good enough for now with @minrk in a video chat and that @manics seem positive in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1842#issuecomment-715242987 about the current state.

With it merged, I plan to update the changelog and cut 0.10.0-beta.1 finally!

minrk:
👍 thanks for talking me through it!

PR #1844, opened 2020-10-17
Add dependabot config to update dependencies automatically

### LINKED ISSUE #1753: Automate updating image dependencies
As mentioned in #1752 by @jgwerner, it could be cool/convenient to automate PRs bumping the requirements in the hub image, either with a GitHub action or [dependabot config](https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates).

We have started using pip-tools (with some wrapping to run it inside our image to ensure equivalence) for the hub image, but we could probably drop the wrapper if we ran pip-compile in CI or via dependabot. Dependabot [does support pip-compile](https://dependabot.com/blog/dependabot-now-supports-pip-compile/) so this might be pretty simple!

jgwerner:
- Update pip dependencies with `lock-file` versioning strategy 
- Update docker dependencies (EDIT: remove for now)
- Set scheduled dependency checks to weekly (EDIT: daily)

EDIT: May end up closing #1753 if we get this right - keeping it open a while

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -13,6 +13,3 @@ updates:
     - dependencies
   commit-message:
     prefix: "hub image: "
-- package_manager: "docker"
-  directory: "/"
-  update_schedule: "weekly"
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/images"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  versioning-strategy: lockfile-only
+- package_manager: "docker"
+  directory: "/"
+  update_schedule: "weekly"
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,11 +1,18 @@
+# dependabot.yml reference: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates
 version: 2
 updates:
+# Automatically update python packages for the jupyterhub/k8s-hub image
 - package-ecosystem: pip
-  directory: "/images"
+  directory: "/images/hub"
   schedule:
-    interval: weekly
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/images"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  versioning-strategy: lockfile-only
```
consideRatio, line 8: I think the directory need to be /images/hub, and just lowered open-pull-request-limit to a month worth of PRs not that I think this will matter.

```suggestion
# Automatically update python packages for the jupyterhub/k8s-hub image
- package-ecosystem: pip
  directory: "/images/hub"
  schedule:
    interval: weekly
  open-pull-requests-limit: 4
  versioning-strategy: lockfile-only
```

.github/dependabot.yaml
```diff
@@ -0,0 +1,11 @@
+version: 2
```
consideRatio, line 1:
```suggestion
# dependabot.yml reference: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates
version: 2
```

.github/dependabot.yaml
```diff
@@ -0,0 +1,11 @@
...
+updates:
+- package-ecosystem: pip
+  directory: "/images"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  versioning-strategy: lockfile-only
+- package_manager: "docker"
+  directory: "/"
+  update_schedule: "weekly"
```
consideRatio, line 11:
```suggestion
```

consideRatio: Let's experiment with this later, but it was not documented what this meant, so I wanted to leave it out for now. I assume it means to inspect Dockerfile and bump FROM images etc though, but could not find dependabot.yml documentation about this.

PR #1852, opened 2020-10-19
Add proxy.service.extraPorts to add ports to the k8s Service proxy-public

yuvipanda:
Now that we have proxy.traefik.extra(Static|Dynamic)Config,
we can configure traefik to do more proxying. Specifically,
I want it to proxy ssh traffic to an instance of
[jupyterhub-ssh](https://github.com/yuvipanda/jupyterhub-ssh).

This works fine, but we need to have proxy-public expose
port 22 to the world. Currently this is impossible -
this PR allows adding additional ports here.

----

## PR summary by consideRatio
- proxy.service.extraPorts configuration added, influencing the proxy-public service, which goes either to autohttps (traefik) or the proxy pod (chp).
- NetworkPolicy rules for ingress may need to be configured separately to this for the receiving pod to not get this kind of new traffic be blocked.

### DIFF
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -63,7 +63,7 @@ spec:
       {{- with .Values.proxy.service.nodePorts.http }}
       nodePort: {{ . }}
       {{- end }}
-    {{- with .Values.proxy.extraPublicPorts }}
+    {{- with .Values.proxy.service.extraPublicPorts }}
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end}}
   type: {{ .Values.proxy.service.type }}
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -63,7 +63,7 @@ spec:
       {{- with .Values.proxy.service.nodePorts.http }}
       nodePort: {{ . }}
       {{- end }}
-    {{- with .Values.proxy.service.extraPublicPorts }}
+    {{- with .Values.proxy.service.extraPorts }}
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end }}
   type: {{ .Values.proxy.service.type }}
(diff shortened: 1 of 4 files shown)


jupyterhub/templates/proxy/service.yaml
```diff
@@ -63,6 +63,9 @@ spec:
       {{- with .Values.proxy.service.nodePorts.http }}
       nodePort: {{ . }}
       {{- end }}
+    {{- with .Values.proxy.extraPublicPorts }}
+    {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
+    {{- end}}
```
consideRatio, line 68: I think we should nest this under `proxy.service`, which in values.yaml also should get an empty array as a default value, along with an entry in lint-and-validate-values.yaml where we put some mocked port value.

```suggestion
    {{- with .Values.proxy.service.extraPublicPorts }}
    {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
    {{- end }}
```

consideRatio (commented):
This is called extraPublicPorts, but with networkPolicies enabled, do you want to see an integration into this? I figure we allow ingress to these ports unconditionally in our netpol's.

Hmmm... The target of proxy-public is also dynamic, based on having autohttps or not, is this assumed to only be used with traefik as autohttps currently then? Hmmm

Note that the networkPolicyResources will soon be available for both autohttps pod and proxy pod.

yuvipanda:
Done, thanks @consideRatio!

I don't think we need to / can do any automatic netpol integration - we should let admins do what is necessary, since we're just doing a passthrough here. 

This isn't necessarily just for autohttps - if you have a sidecar to chp, it works there too. Use cases are endless, since we're really just allowing customization by admins.

Thanks!

jupyterhub/templates/proxy/service.yaml
```diff
@@ -63,6 +63,9 @@ spec:
       {{- with .Values.proxy.service.nodePorts.http }}
       nodePort: {{ . }}
       {{- end }}
+    {{- with .Values.proxy.service.extraPublicPorts }}
+    {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
+    {{- end}}
```
consideRatio, line 68:
```suggestion
    {{- end }}
```

PR #1862, opened 2020-10-21
Allow adding extra labels to the traefik pod

yuvipanda:
Allows admins to configure the traefik pod, particularly
to support network policies of *other* charts deployed
in the same namespace


consideRatio:
LGTM!

PR #1866, opened 2020-10-23
Add app.kubernetes.io/instance label

droctothorpe:
This PR adds an `app.kubernetes.io/instance: {{ .Release.Name }}` label to JupyterHub objects via the `jupyter.labels` field in `_helpers.tpl`.

This label is a Helm idiom per [their documentation](https://helm.sh/docs/chart_best_practices/labels/#standard-labels) and is [uniformly adopted](https://github.com/helm/charts/search?q=app.kubernetes.io%2Finstance).

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -147,6 +147,7 @@ heritage: {{ .heritageLabel | default .Release.Service }}
     Provides labels: component, app, release, (chart and heritage).
 */}}
 {{- define "jupyterhub.labels" -}}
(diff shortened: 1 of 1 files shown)


consideRatio:
This PR is currently a one liner, but since this kind of change is a breaking chance, it require significant work to be viable for a merge. I try to provide an of the challenge in #1867. I'd be very thankful to see this get done, but it requires quite extensive work to ensure we don't crash something badly for those with charts installed already for this change.

consideRatio:
I'll close this PR for now @droctothorpe so it doesn't just remain open in a stale state.

Issue #1865, opened 2020-10-23
Manage redirects following docs restructures with sphinxext-rediraffe instead of RTD


consideRatio:
As introduced in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1825#issuecomment-713759673, and suggested in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1825#issuecomment-714587547, we can use a tool called [sphinxext-rediraffe](https://github.com/wpilibsuite/sphinxext-rediraffe) to redirect.

Recently I added redirects [in RTD's admin section](https://readthedocs.org/dashboard/zero-to-jupyterhub/redirects/) which also work, but...
- only on the published site has redirects when using RTD's redirect system
- only RTD admins can view and configure the redirects
- the redirects won't be version controlled

So, having rediraffe do it for us as part of the built website comes with quite a bit of benefits. Anyone wanting to help us solve this would be appreciated. Currently, we have the following redirects on RTD that can be migrated.

RTD redirects this path... | to this path...
-|-
/customizing/user-management.html | /jupyterhub/customizing/user-management.html
/customizing/user-storage.html | /jupyterhub/customizing/user-storage.html
/customizing/user-resources.html | /jupyterhub/customizing/user-resources.html
/customizing/user-environment.html | /jupyterhub/customizing/user-environment.html
/customizing/extending-jupyterhub.html | /jupyterhub/customizing/extending-jupyterhub.html
/reference/glossary.html | /resources/glossary.html
/reference/tools.html | /resources/tools.html
/reference/reference-docs.html | /resources/reference-docs.html
/reference/reference.html | /resources/reference.html
/community/additional-resources.html | /resources/community.html
/community/users-list.html | /resources/community.html
/community/tips.html | /resources/community.html
/setup-jupyterhub/turn-off.html | /jupyterhub/uninstall.html
/setup-jupyterhub/setup-jupyterhub.html | /jupyterhub/installation.html
/setup-jupyterhub/setup-helm.html | /kubernetes/setup-helm.html
/ovh/step-zero-ovh.html | /kubernetes/ovh/step-zero-ovh.html
/digital-ocean/step-zero-digital-ocean.html | /kubernetes/digital-ocean/step-zero-digital-ocean.html
/ibm/step-zero-ibm.html | /kubernetes/ibm/step-zero-ibm.html
[...]

PR #1876, opened 2020-10-26
Added rediraffe redirecgtion

NerdSec:
As discussed in #1865, added the ability to perform redirects outside of RTD by leveraging rediraffe.


doc/source/conf.py
```diff
@@ -77,6 +78,63 @@ def setup(app):
...
+    "index-setup-jupyterhub.html": "jupyterhub/index.html",
+    "tools.html": "reference/tools.html",
+    "reference-docs.html": "reference/reference-docs.html",
+#    "index-reference.html": "reference.html",
+    "glossary.html": "reference/glossary.html",
+    "user-storage.html": "customizing/user-storage.html",
+    "user-resources.html": "customizing/user-resources.html",
+    "user-management.html": "customizing/user-management.html",
+    "user-environment.html": "customizing/user-environment.html",
+#    "index-customization-guide.html": "customization.html",
```
consideRatio: Why are these added as comments?

consideRatio (approved):
I tried this locally, and it works nice! I pushed a commit to make `make devenv` still work again after the refactoring.

I only wonder about the commented out redirects, why are they made? Ah they were an older redirect to a file that was moved in this recent refactoring. I updated them to reference directly to the new location.

choldgraf:
wohoo! can we all also appreciate how great the rediraffe logo is?

consideRatio:
Haha yes :D :1st_place_medal:

PR #1877, opened 2020-10-27
Helm template aesthetics fixes

consideRatio:
This PR are some remnant ideas I had from inspecting our Helm chart with the then stricted `helm lint --strict` policy in #1681. `helm lint --strict` is in v3.3.4 less strict than it was before and we now pass with flying colors!

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -42,11 +42,7 @@ spec:
             name: hub-config
         - name: secret
           secret:
-            {{- if .Values.hub.existingSecret }}
-            secretName: {{ .Values.hub.existingSecret }}
-            {{- else }}
-            secretName: hub-secret
-            {{- end }}
+            name: {{ .Values.hub.existingSecret | default "hub-secret" }}
         {{- if .Values.hub.extraVolumes }}
         {{- .Values.hub.extraVolumes | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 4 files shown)


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -42,11 +42,7 @@ spec:
             name: hub-config
         - name: secret
           secret:
-            {{- if .Values.hub.existingSecret }}
-            secretName: {{ .Values.hub.existingSecret }}
-            {{- else }}
-            secretName: hub-secret
-            {{- end }}
+            name: {{ .Values.hub.existingSecret | default "hub-secret" }}
```
consideRatio, line 45: Woops... Should had been secretName...

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -163,43 +159,27 @@ spec:
             - name: CONFIGPROXY_AUTH_TOKEN
               valueFrom:
                 secretKeyRef:
-                  {{- if .Values.hub.existingSecret }}
-                  name: {{ .Values.hub.existingSecret }}
-                  {{- else }}
-                  name: hub-secret
-                  {{- end }}
+                  secretName: {{ .Values.hub.existingSecret | default "hub-secret" }}
```
consideRatio, line 162: Should had been `name`...

consideRatio:
```
WARN - jupyterhub/templates/hub/deployment.yaml contains an invalid Deployment (hub) - secretName: Additional property secretName is not allowed
WARN - jupyterhub/templates/hub/deployment.yaml contains an invalid Deployment (hub) - name: Additional property name is not allowed
```

I should had run tools/templates/lint-and-validate.py locally instead of just assuming i got it right. TravisCI delays made me merge without waiting for tests. Will fix.

PR #1881, opened 2020-10-27
Remove globus_sdk and update various Docker images

consideRatio:
This PR bumps some core images to avoid known vulnerabilities and just keep up to date in general.

`globus_sdk[jwt]` as a dependency could be removed because it was no longer needed for the GlobusOAuthenticator any more, but the azure authenticator still require PyJWT so we still install that.

I choose to keep pinning the docker-stacks/base-notebook version to avoid risking issues introduced without us understanding why.


consideRatio:
As it takes forever for travis to run tests, I'll merge this and inspect the builds tomorrow morning.

Issue #1851, opened 2020-10-19
user-scheduler ranking is bypassing full nodes - can we avoid it?


consideRatio:
We have a _user-scheduler_ that is responsible for assigning real user pods and user-placeholder pods to run the nodes where they are to run. This process is done in multiple steps. This is an example of what could happen if all nodes are full during the scheduling process of a real user pod.

1. User-scheduler performs node filtering
2. User-scheduler concludes there are no available nodes
3. This triggers [Pod eviction by Kubelet](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#interactions-of-pod-priority-and-qos)
4. User-scheduler performs node filtering again
5. Now there is _one_ node available as one pod was evicted, and it will score the highest no matter what
6. The pod is assigned

The user-scheduler first looks at nodes where the pod would fit, which excludes nodes partially full of user-placeholder pods, and then scores the nodes where it would fit given a set of criteria. But, if it really wants to fit on the most used node, or for example where an image is available, it would be good if it could consider nodes full with user-placeholder pods as well.

consideRatio:
I still don't see a clear strategy to manage this that I find to be good enough to motivate the added complexity and maintenance burden that comes with it.

consideRatio:
This issue is a duplicate of #1414 in many ways I think.

minrk:
I really wish the cluster-autoscaler had a built-in "headroom" parameter of minimum allocatable resources to trigger scaling, rather than waiting for a failed scheduling event. That's what we are trying to mimic with our placeholders, and as we can see there are downsides to the fact that it's not *quite* what we are able to do. Is it appropriate to try to formulate a feature request for upstream?

I realize @yuvipanda [already did this in 2017](https://github.com/kubernetes/autoscaler/issues/148), but the resolution [here](https://github.com/kubernetes/autoscaler/issues/148#issuecomment-397209457) doesn't appear to be accurate: that our approach of placeholder pods works, when we have this exact case where it doesn't a lot of the time.

As discussed on gitter, would fewer, larger placeholders tend to behave better? Or are we always going to have this problem of nodes that are 'full' due to placeholders, to which placeholder-triggered headroom nodes are preferred, unless we pick mutually-exclusive resource reservations such that 1 placeholder pod + 1 user pod can never fit on the same node?

[...]

PR #1883, opened 2020-10-28
prePuller.extraTolerations added for the image-puller daemonsets

jerkern:
Adding exclusive extra tolerations to the prePuller daemonset, to make it easier to integrate into a deployment that uses eg. nidhogg for dynamic tainting.
Related to issue #1836 and #1851 . This doesn't fully solve the problem, it just makes it easier to do a custom solution.

Example use-case
- nidhogg taints all nodes while the prePuller is running, prohibiting anything except the prePuller from being assigned to the node
- prePuller completes and nidhogg removes the taint
- now placeholder and actual user pods can be deployed to the node

This PR accomplishes being able to add a taint that only is tolerated by the prePuller and nothing else.

### DIFF
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -48,7 +48,7 @@ spec:
       {{- end }}
       tolerations:
         {{- include "jupyterhub.userTolerations" . | nindent 8 }}
-        {{- when .Values.prePuller.extraTolerations  }}
+        {{- with .Values.prePuller.extraTolerations  }}
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -48,6 +48,9 @@ spec:
       {{- end }}
       tolerations:
         {{- include "jupyterhub.userTolerations" . | nindent 8 }}
+        {{- if .Values.prePuller.extraTolerations  }}
+        {{- .Values.prePuller.extraTolerations  | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- end }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
       {{- if include "jupyterhub.userNodeAffinityRequired" . }}
       affinity:
--- a/jupyterhub/templates/image-puller/_daemonset-helper.yaml
+++ b/jupyterhub/templates/image-puller/_daemonset-helper.yaml
@@ -48,8 +48,8 @@ spec:
       {{- end }}
       tolerations:
         {{- include "jupyterhub.userTolerations" . | nindent 8 }}
-        {{- if .Values.prePuller.extraTolerations  }}
-        {{- .Values.prePuller.extraTolerations  | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- when .Values.prePuller.extraTolerations  }}
+        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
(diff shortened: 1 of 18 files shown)


jupyterhub/values.yaml
```diff
@@ -480,7 +481,6 @@ prePuller:
       allowPrivilegeEscalation: false
     podSchedulingWaitDuration: 10
     nodeSelector: {}
-    tolerations: []
```
donotpush, line 483: Replaced prePuller.hook.tolerations for prePuller.extraTolerations because both daemonsets/jobs should have similar/same tolerations

consideRatio: It is not obvious to the location of configuration etc, but prePuller.hook.tolerations is currently just for the hook-image-awaiter, which is a single pod that monitors the readiness state of the hook-image-puller daemonset's pods.

I think both deamonsets (hook-image-puller, continuous-image-puller) should have the same tolerations, but the hook-image-awaiter Job should have its own still.

jupyterhub/templates/image-puller/_daemonset-helper.yaml
```diff
@@ -48,6 +48,9 @@ spec:
       {{- end }}
       tolerations:
         {{- include "jupyterhub.userTolerations" . | nindent 8 }}
+        {{- if .Values.prePuller.extraTolerations  }}
+        {{- .Values.prePuller.extraTolerations  | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- end }}
```
consideRatio, line 53:
```suggestion
        {{- with .Values.prePuller.extraTolerations  }}
        {{- .  | toYaml | trimSuffix "\n" | nindent 8 }}
        {{- end }}
```

Helm chart template practices has transitioned to using {{ with "something" }} which only renders if "something" is truthy, such as a non-empty array or string.

consideRatio (commented):
Thanks for your work on this @jerkern! This looks sensible and thank you for providing documentation for everyone else about the motivation of this PR even though we discussed it verbally before.

After these parts are addressed I think we can merge.
- [ ] Let the configuration for job.yaml (image-awaiter) in prePuller.hook.tolerations remain
- [ ] Add a valid entry in lint-and-validate-config.yaml for the new tolerations, then run `tools/templates/lint-and-validate.py` to see that we render into valid k8s resources still.

jupyterhub/values.yaml
donotpush: @consideRatio  Are you proposing to keep prePuller.hook.tolerations, and add prePuller.extraTolerations tolerations for deamonsets (hook-image-puller, continuous-image-puller) and also the single pod hook-image-awaiter?

consideRatio: > Are you proposing to keep prePuller.hook.tolerations, and add prePuller.extraTolerations tolerations for deamonsets (hook-image-puller, continuous-image-puller) and also the single pod hook-image-awaiter?

Yes, but I don't think it makes sense to add the tolerations also for the image-awaiter Job's pod. I find the image-awaiter to be something that make sense to run on nodes like the hub pod etc rather than end up on a node that is busy pulling images and potentially blocked by image pulling in the worst case.

donotpush: what about only having then prePuller.continuous.tolerations?

consideRatio: I'm open to relocating image-awaiter job's configuration someplace so its less confusing, because it sure is! For example as part of prePuller.hook.imageAwaiter.tolerations. This kind of change would need to be accompanied with a deprecation plan as well though, and we don't have very clear procedures for that at the moment, but typically by adding a fail statement in NOTES.txt or the more complicated option of adding a warning in NOTES.txt followed by logic to migrate the old configuration to work with the new config.

But without a strong case for having dedicated tolerations for only the continuous/hook image-puller daemonsets' pods, I don't want to set tolerations on the daemonset-specific level.

If one goes for DS-specific config, we end up maybe wanting to have 3 config options instead of one. DS-unspecific, and DS-specific x2, and that is a lot more work to ensure they merge with eachother in a sensible way etc and document and communicate that - it is simply more complicated without clear benefits I think.

donotpush:
@consideRatio we updated the PR based on your comments

consideRatio:
Thank you @jerkern and @donotpush! :tada: :heart:!

Issue #1872, opened 2020-10-24
Replace Travis with GitHub workflows


manics:
### Proposed change
Replace Travis CI with GitHub workflows

### Alternative options
Do nothing

### Who would use this feature?
People who are bored of waiting for Travis to run :smile: See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1840#issuecomment-715262697
Anecdotally Travis seems to be getting slower and slower.

### (Optional): Suggest a solution
I've previously converted the [K3s and helm setup script from this repository into an action](https://github.com/manics/action-k3s-helm/) (was planning to use it for a different project but never got round to it) and I'm happy to convert this repo to use it.

There are other potential benefits mentioned in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1840#issuecomment-715286827 but as a first step I suggest just reproducing the existing tests.

consideRatio:
> I've previously converted the K3s and helm setup script from this repository into an action (was planning to use it for a different project but never got round to it) and I'm happy to convert this repo to use it.

This was a good viability demonstration for me! I have not seen any announcement or similar why TravisCI limited their workers as they have done the last days, which has caused the spike of jobs being queued up for too long durations.

![image](https://user-images.githubusercontent.com/3837114/97081220-29620480-1601-11eb-94ab-0dc35bfef614.png)

## Other significant reasons to use GitHub Actions over TravisCI
1. While it becomes a bit more complicated to introduce a composition of scripts to run, it helps us avoid duplicating logic which is common across for example z2jh, binderhub, and kubespawner and other outside org repositories like DaskHub.
2. GitHub Actions have the nice functionality of conditionally being triggered only when certain files change, which means that we don't have to helm chart tests of charts when we update the documentation only.

Overall, I'm in favor of switching to GitHub actions in this repo at least.

manics:
I think your reason 1 is one of the best arguments for it. If there's a K8S/Helm problem in future it'll be a lot easier to debug the self-contained action repo without the rest of Z2JH getting in the way. It also means we have the option of adding tests to verify the base deployment of K8s (e.g. that network policies are functioning) without adding noise in this repo.

consideRatio:
The TravisCI issues were resolved by a .org -> .com migration, but this is still relevant in my mind (https://github.com/jupyterhub/team-compass/issues/348)

minrk:
[...]

PR #1896, opened 2020-11-01
Replace Travis with GitHub workflow

manics:
publish workflow is untested
~The tests are hanging on K3S 1.17.3 and 1.16.15, but passing for 1.18.10 and 1.19.3~ Seems to be an ipv6 problem with older K3s versions.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1872


consideRatio:
Wieeee thank you for your work on this @manics! Really looking forward to be able to exclude triggering builds when there is no point to them by selecting paths to watch for changes.

```yaml
# Only run this when the master branch changes
on:
  push:
    branches:
    - master
    # If your git repository has the Jupyter Book within some-subfolder next to
    # unrelated files, you can make this run only if a file within that specific
    # folder has been modified.
    #
    # paths:
    # - some-subfolder/**
```

I'm putting in a lot of work on jupyterhub/binderhub and hope that work can us help us transition to using this tooling in the long run.

manics:
I think this is ready! I had to add a delay in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1896/commits/bcff6bf1d6b231481ddaa6f0a591eb3b6feb8f9c. A run was failing around 10% of the time, since there are 5 runs this means a failure ~50% of the time. Adding a delay seems to give all K8s resources time to stabilise.

This uses [manics/action-k3s-helm@v0.2.1](https://github.com/manics/action-k3s-helm/), should I transfer it to the jupyterhub GitHub org or leave it under my account?

.github/workflows/test.yml
```diff
@@ -0,0 +1,122 @@
...
+  pull_request:
+  push:
+
+jobs:
+
+  lint_and_validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: manics/action-k3s-helm@v0.2.1
```
consideRatio, line 14: Will this mean we start an entire k3s cluster just to install Helm 3? If so I favor retaining setup_helm in ci/common and let run that here instead.

.github/workflows/test.yml
```diff
@@ -0,0 +1,122 @@
...
+          metrics-enabled: false
+          traefik-enabled: false
+          docker-enabled: true
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - name: Install and run chartpress
+        run: |
+          . ./ci/common
+          install_and_build
```
consideRatio, line 81: Perhaps we can name this `install_pebble_run_chartpress` or somehow communicate we install pebble.

.github/workflows/test.yml
```diff
@@ -0,0 +1,122 @@
...
+          . ./ci/common
+          helm upgrade --install jupyterhub ./jupyterhub --values dev-config.yaml
+          await_jupyterhub
+          await_autohttps_tls_cert_acquisition
+      - name: Wait 1 minute for everything to be ready
+        # TODO: find out exactly which resources are slow and wait for thoses
+        # So far errors have included:
+        # - some pods not being ready
+        # - network policies not being ready
+        run: sleep 1m
```
consideRatio, line 100: Awww =/ I struggled so much to get the CI system run without these kinds of tiresome delays... I guess we can try debug it over time together though.

ci/common
```diff
@@ -1,100 +1,6 @@
...
-    #
-    # ref: https://rancher.com/docs/k3s/latest/en/installation/network-options/
-    #
-    curl -sfL https://docs.projectcalico.org/v3.14/manifests/calico.yaml \
-  | sed '/"type": "calico"/a\
-          "container_settings": {\
-            "allow_ip_forwarding": true\
-          },' \
-  | kubectl apply -f -
-}
```
consideRatio, line 71: setup_calico is referenced in the CONTRIBUTING.md file. I suggest to copy paste the content of it there or retain it here.

await_... functions are also referenced in CONTRIBUTING.md in comments, which can simply be cut away.

Can you give the CONTRIBUTING.md file an update with regards to this?

.github/workflows/test.yml
consideRatio: I'd prefer it if we let it fail hard for a while, I really want this gone ^^. Perhaps also stop running tests when pushing to master to avoid too much errors triggering notifications etc?

.github/workflows/test.yml
```diff
@@ -0,0 +1,122 @@
+---
+name: Test
+
+on:
+  pull_request:
+  push:
```
consideRatio, line 6:
```suggestion
```

Perhaps skip running tests on merges to master?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+          python-version: '3.7'
+      - name: Install dependencies and run chartpress
+        run: pip install --no-cache-dir -r dev-requirements.txt
+      - name: Publish chart
+        run: |
+          . ./ci/common
+          ./ci/publish
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
consideRatio, line 28: There are environment variables used to decrypt an encrypted SSH key registered with the jupyterhub/helm-chart repo so we can push to it. We have to either replace that SSH key with a new one we have encrypted or create a new deploy key for jupyterhub/helm-chart, encrypt it, and use an environment variable as key to decrypt it.

Let's create a new SSH public/private key, encrypt the private in this repo using some secret, and register the public part for push access in the jupyterhub/helm-chart repo here: https://github.com/jupyterhub/helm-chart/settings/keys.

Or, perhaps there is a GitHub action way to help us get a SSH key encrypted into the repo? Ah yes! Let's try out this action perhaps? https://github.com/marketplace/actions/install-ssh-key

consideRatio (commented):
Thank you @manics for your work on this! I know from experience this kind of work can take quite a bit of time! :heart: :tada:

I made some change requests, let me know if you want help to get this PR all the way!
- [x] Update CONTRIBUTING.md following the changes
- [x] Give the publishing script a pass by creating a new private SSH deployment key to inject in this repo in the publishing job
- [ ] Configure DOCKERHUB_USERNAME and DOCKERHUB_PASSWORD secrets to be available

.github/workflows/test.yml
manics: Are there any disadvantages to running CI tests on master? The publish workflow runs independently.

Advantages of master CI tests are:
- Don't have to worry about PRs being out of sync with master. Since CI tests are run when the PR is updated if running the tests on master is an easy way to handle this possiblity.
- Easy to find record of all CI runs. When debugging a CI failure it is often useful to have a known good reference for comparison. If they're not run against master you'll have to go through the PR runs and possibly cross reference them with a commit. Similarly if an external change causes CI to start failing or a test to become flakier it's useful to hav the record of master CI runs to see when the change occured

.github/workflows/test.yml
manics: Will do....

.github/workflows/test.yml
manics: I hate it too 😢 

At the moment the tests take around 8 mins, compared to 7 without this. However as already mentioned one of the runs fails roughly 10% of the time, so we can expect a CI failure around 50% of the time. This means on average CI will take around 10-11 mins. In addition it takes maintainers 1-2 minutes to check the CI logs to see if it's a usual flakey failure or a genuine failure. Overall I think removing the delay increases the maintenance burden.

ci/common
manics: Will do!

.github/workflows/test.yml
consideRatio: Deal let's retain it!

.github/workflows/test.yml
consideRatio: Let's leave it there, I'm glad you are also motivated to get rid of it as well!

I believe I'm afraid of being alone in working to optimize this away if we don't let it make some noise, but it will indeed be a burden if it add noise as well.

.github/workflows/publish.yml
consideRatio: I added a commit and registered a deployment key etc, but credentials for dockerhub remain still. I have asked @minrk for help about this.

> @minrk I'd like to provide GitHub actions the password for the dockerhub jupyterhubbot account in order to let pushes for z2jh work. This will probably also be done for binderhub in the end, could you provide me with the password or add DOCKER_USERNAME and DOCKER_PASSWORD here:
>
>https://github.com/jupyterhub/zero-to-jupyterhub-k8s/settings/secrets/actions
>https://github.com/jupyterhub/binderhub/settings/secrets/actions

betatim:
@manics how do the tests that run outside the k8s cluster figure out what the IP is on which they can reach the JupyterHub API? It looks like `proxy-public` is deployed as a "NodePort" type service with a fixed port. If this was minikube I'd expect something like `minikube ip` in order to find the IP to go with the port. However I can't find the equivalent for k3s :-/ I did find that coreDNS was being patched so that `local.jovyan.org` always resolves to the porxy service, but I think that magic trick only works if you are inside the cluster.

consideRatio:
@betatim local.jovyan.org resolves to 127.0.0.1 unless hacked inside the cluster to point to proxy-public svc.

Minikube ip isnt needed with k3s as there is no virtual machine with separate network i think.

betatim:
Ah yes. I'm still too much in my OSX world where you can't run kubernets "on the host", instead you are always inside a VM (k3d, minikube, minishift, etc). The answer is that `http://localhost:<NodePortPort>/hub/api` is where you can reach a service inside the k3s cluster.

manics:
I'm a bit short on time at the moment, I can see you've already pushed to this branch, feel free to continue doing so :smile:

ci/common
consideRatio: I pushed a commit to resolve this!

consideRatio:
We still haven't DOCKER_USERNAME and DOCKER_PASSWORD configured, so the publish job will fail on merge, but let's go with this for now until we get help from Min with this.

consideRatio:
@manics thank you so much for your work on this! The extraction of the k8s cluster setup step into a dedicated part is very appreciated apart from the big help with transitioning to GitHub actions from Travis CI.

:tada: :heart:

consideRatio:
@manics everything seem to work and I managed to get rid of the 1m delay in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1896! :tada:

PR #1897, opened 2020-11-03
DOCS: Adding note on limit to guarantee ratio

choldgraf:
This adds a short guide to explain the "limit to guarantee ratio" for memory and CPU. It's a tricky subject that we've found a lot of hub admnistrators run into, but it's also a bit of an art-form without a single strict answer. So this tries to add some explanation of the problem and its nuance in the hopes that administrators will be able to more effectively provision their resources.

WDYT @yuvipanda , who was the one to originally explain this to me.


yuvipanda:
This is a great starting point, thanks :)

PR #1901, opened 2020-11-07
Allow exposing extra ports in traefik deployment

yuvipanda:
Follow-up to #1852. Useful to explicitly expose ports
in deployment too - otherwise you can't use port names like
`ssh` in service definitions. To allow ingress on a port named
`ssh` in traefik right now, you have to know it is listening
on port 8022, and set the following:

```yaml
proxy:
  autohttps:
    networkPolicy:
      allowedIngress: [http, https, 8022]
```

This requires knowledge that `8022` is the ssh port. Instead,
it would be nice to specify `ssh` instead - and that requires
defining that in the deployment. This is true for other places
too - like the service definition.

----

#

### DIFF
--- a/tools/templates/lint-and-validate-values.yaml
+++ b/tools/templates/lint-and-validate-values.yaml
@@ -146,7 +146,7 @@ proxy:
   traefik:
     extraPorts:
     - name: ssh
-      containerPort: 22
+      containerPort: 8022
     labels:
       hub.jupyter.org/test-label: mock
     resources:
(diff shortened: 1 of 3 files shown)


tools/templates/lint-and-validate-values.yaml
```diff
@@ -144,6 +144,9 @@ proxy:
       interNamespaceAccessLabels: accept
       allowedIngressPorts: [http, https]
   traefik:
+    extraPorts:
+    - name: ssh
+      containerPort: 22
```
consideRatio, line 149: Lets examplify with a port number over 1000 to be compliant with strict pod security policies.

```suggestion
      containerPort: 8022
```

consideRatio:
Nice! There is no publishing CI atm though as travis is down for us.

PR #1902, opened 2020-11-07
Feature/code server integration

JoeJasinski:
The goal of this PR is to add documentation on how to run [Code Server](https://github.com/cdr/code-server) Notebooks in JupyterHub instead of Jupyter Notebook/Lab instances. JupyterHub is a perfect spawner for launching Code Server notebooks. 
 This enables full Visual Studio instances to be launched via the JupyterHub framework into private, user-specific pods in a Kubernetes cluster. User can have their own private cloud-based VS Code development environment that is managed behind the JupyterHub proxy.

We've been able to get working on our private Helm charts. As are Helm charts are quite different than this one here, the goal of this MR is to document how Code Server can be integrated with this zero-to-jupyterhub-k8s Helm chart.

### DIFF
--- a/doc/source/jupyterhub/customization.rst
+++ b/doc/source/jupyterhub/customization.rst
@@ -19,3 +19,4 @@ used configurable helm chart fields.
    customizing/user-resources
    customizing/user-storage
    customizing/user-management
+   customizing/running-code-server-from-jupyterhub
--- a/doc/source/jupyterhub/customizing/running-code-server-from-jupyterhub.md
+++ b/doc/source/jupyterhub/customizing/running-code-server-from-jupyterhub.md
@@ -75,7 +75,7 @@ RUN yarn build:vscode
 RUN yarn release && yarn release:standalone && yarn package && mv release-packages /tmp/ && rm -r ./* && mv /tmp/release-packages/ .
 
 # Build the Code Server Notebook image
-FROM ubuntu:20.04
+FROM python:3.8
 
 RUN apt-get update && apt-get install -y \
     curl
@@ -97,6 +97,8 @@ RUN ARCH="$(dpkg --print-architecture)" && \
     chmod 4755 /usr/local/bin/fixuid && \
     mkdir -p /etc/fixuid && \
     printf "user: coder\ngroup: coder\n" > /etc/fixuid/config.yml
+
+RUN pip install jupyterhub==1.2.1 notebook==6.1.5
 
 # Run code-server on port 8888
 CMD ["code-server", \
@@ -221,6 +223,9 @@ Notebooks.
     proxy:
       secretToken: "$RAND_KEY"
 
+    debug:
+      enabled: true
+
     ingress:
       hosts:
       - minikube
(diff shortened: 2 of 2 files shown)


manics:
This looks pretty cool, I'm looking forward to giving it a go!

I'm worried about the maintenance burden though, as it's quite detailed and isn't going to be regularly tested. In particular I don't think embedding a Dockerfile in the docs is great, it'd be better to just reference an existing image on a public registry.

Alternatively could you host this guide on your own website, and simply add a paragraph and a link to it here? I think that would make things easier for you too, since if code-server changes you can easily update your instructions. It also means you can recommend a version of JupyterHub/Z2JH that you've most recently tested. Obviously we don't expect things to break on the Z2JH side but there are always bugs 😀.

What do you think?

consideRatio:
Thank you for your work to help provide relevant and useful documentation for others @JoeJasinski :tada: :heart:!

I share the concerns @manics voices about maintainability, which I think also relate to the scope of the z2jh.jupyter.org guide that. I consider this PR to bee too ambitious with regards to what we can be able to sustain maintenance for in this guide. I think it contains parts that can be reworked to be more general and be in scope of the guide though.

I consider the following contributions to be in scope of the guide to maintain that relates to this PR.
- About the ability to configure and do to start something else than jupyter servers like `jupyter notebook` and `jupyter lab`, such as rstudio or code-server exposed at `/rstudio` and `/code` or similar.
- How to extend the hub image where JupyterHub / KubeSpawner / the Authenticator run and the constraints and things to consider.
- About the constraints of a custom singleuser that image that is to function with a z2jh deployment.

jordanbettisnielsen:
Hi everyone, hope I can provide a bit more context behind this PR:

About a year ago, we (Nielsen) developed an internal project we've been calling "codehub." The product is a multi-user installation of vscode in kubernetes using Jupyterhub, Kubespawner and code-server.

This product sits alongside our kubernetes-based jupyter notebooks and allows our data scientists to more easily productionalize their work by switching to traditional IDE and SDLC as their models get more sophisticated from a software perspective. 

While our data scientists still love their notebooks and would never give them up, we've been noticing a rapid adoption of codehub, and we think that it would be a useful product for the broader open source community, since productionalization is the future of data science.

This PR is the first step in our efforts to open source codehub. We also have additional PRs on the way for both projects, but codehub is 90% configuration so we don't see value in forking existing projects and we would prefer to offer this as a feature of jupyterhub/code-server if possible.

betatim:
It is cool to see Nielsen putting in the extra effort to open-source their work. Especially if a lot of it is "just configuration" (some say that is the least sexy new "product"). 

I think the question/trade off that needs figuring out is: Z2JH contains a lot of things which require attention to keep them working. The people who actively maintain the guide often don't have access to resources needed to check if things still work (like say cloud vendor X based instructions) or they don't use all the different ways you can configure things. As well as having only a limited amount of time ;). This means that in the past we have asked people if they'd rather add a link to their material to the guide and host the content themselves. That way it is clear to readers who to ask when things don't work.

Some ideas for possible compromises (items are numbered for easier reference, not to indicate preference):
1. make instructions more "general" (similar to Erik's suggestion) in order to make "issues with code-server itself" out of scope
1. is it possible to add a test to check that the guide still works (provides a "known good" solution to refer to)
1. rephrase things so describe the general case with code-server being used as a "concretely this might look like this" example

manics:
Here's a possible crazy idea 😄. This will require discussion amongst the maintainers so don't assume it'll be done.

We've moved to using GitHub Workflows for CI testing this chart, and this includes splitting off the Kubernetes cluster setup into it's own reusable GitHub Action. If we did some work to make our CI infrastructure easier to use in other projects that build on top of Z2JH, e.g. a template workflow or evena. GitHub action to install JupyterHub that accepts a config file, would you consider using it in your own code-server guide repo?

betatim:
While I love the idea (especially as a user of z2jh as part of mybinder.org), I am not sure where it goes in the whole "let's not create huge amounts of extra maintenance effort" ranking :-/

manics:
Maybe we wouldn't need to do anything and it's just an example workflow? There are GitHub actions for deploying Helm charts. We've got an action to deploy a K3S cluster. In _theory_ this should therefore be straightforward- pass the kubeconfig and the Z2JH config file to the Helm action. The difficult bit is testing whether the singleuser server i.e. code-server is working properly, but this would be required whichever test strategy we used.

rokroskar:
@JoeJasinski this is a really interesting idea! I'm wondering though why you want to have code-server-specific images/sessions instead of adding code-server to an existing jupyter notebook image? You are installing the jupyter stack in the end anyway, if I'm reading things correctly? If you install the code-server into a jupyter image and expose it through the little [jupyter server extension](https://github.com/betatim/vscode-binder) written by @betatim you basically get the same functionality in the end. You can configure the sessions to skip JupyterLab entirely if you want and go straight to the `/vscode` path. But maybe I'm missing the specific use-case needs?

JoeJasinski:
Thank you all for the great feedback! There's still some more for me to do to get this work fully with this repo. I like the idea of breaking out the Dockerfile and some of the VSCode-specific logic. As @jordanbettisnielsen mentions, we have some other contributions related to this for different repos to get auth integrated and such. 

We like that JupyterHub has the ability to spawn different types of notebooks, and that's partly what inspired us to look into this, along with providing our users the ability to have a full-featured cloud-based VSCode editor. I think it would be great to have some documentation around setting up a framework for spawning different notebooks, like what @consideRatio and @betatim mention, if I understand correctly. I can look into how to document that as more general steps (perhaps using this VSCode as an example use-case?)

@rokroskar, that looks like a cool Binderhub integration for code-server. Our implementation uses a pure code-server notebook (not as an extension), so it runs directly as a webservice in the independent of Jupyter Lab. That being said, I'll take a look at that extension, as it might take care of the authentication automatically (which would be really nice).

rokroskar:
Hi @JoeJasinski, it does indeed take care of authentication - we use it with a JupyterHub with OAuth. You can use the Dockerfile [from a related issue](https://github.com/betatim/vscode-binder/issues/25#issuecomment-727970423) as a starting point.

consideRatio:
To summarize, this PR currently is considered out of scope due to how specific and detailed it is which makes it hard to maintain. It would be fine to link from the [Community resources](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/community.html) section though. PRs to address something more general, such as ideas in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1902#issuecomment-723653781, are very welcome!

Thank you @JoeJasinski for your thorough work to write this up :heart:, I'm sorry for rejecting such a proper contribution to be merged. I'm closing this PR now looking to minimize the amount of stale PRs and such. I suggest we take further discussion into a new issue.

PR #1903, opened 2020-11-07
Add use_lookup_dn_username parameter for LDAP

JarnoRFB:
Enable the parameter to be set in values.yaml. Also add a refernce to
the code that does the mapping, so it is easier for user to figure out
correspondence between ldapauthenticator params and params in
values.yaml.


consideRatio:
Thanks! We will need to update all authentication configuration to be easier to understand and less complex to maintain in the future, for now this is a good improvement! :tada:

Issue #1791, opened 2020-09-30
Helm release unique resource names


consideRatio:
It is a common practice to use the Helm _release name_ as a prefix to the k8s resource names. A Helm _release name_ is the name of a specific installation of a Helm chart, which can of course be installed multiple times in a Kubernetes cluster. Without a Helm release name prefixed to our resources, we can get name collisions that makes installations fail or misbehave after installation.

## Cluster resources - solved
The JupyterHub helm chart has following its launch learned to ensure we can install two releases in the same k8s cluster but in different namespaces, that only required us to prefix a release name to the cluster wide resources like ClusterRole and ClusterRoleBinding.

## Namespaced resources - an open issue
The JupyterHub helm chart still doesn't prefix namespaced resources, and this makes us unable to install the JupyterHub Helm chart twice in the same k8s namespace.

## Goal discussion
I think it will be too complicated to support a transition where a Helm chart that is already installed gain a prefix during an upgrade. To understand this complexity, consider for example how we are to ensure Kubespawner's dynamically created PVCs are to be migrated without creating new PVs or even deleting PVs associated with the old PVCs.

What I do think is a plausible goal, is to allow a fresh installation to be configured to use a prefix. And, I have a clear suggestion on how to get this done.

## Implementation specification
- We define a helper template in `_helpers.tpl` which we call to get a string back with a prefixed name depending on the configuration. We are mindful to mimic typical practices in Helm charts for the name of this template and its associated configuration.
- We systematically use the helper template to prefix our k8s resource names to ensure they are unique for each Helm release.
- We ensure the helper template doesn't add a prefix unless we explicitly configure it to do so, in order to ensure we don't break things for existing deployments.
- We document explicitly that changing the prefix related configuration after the initial install can have severe consequences, and strongly suggest to not do so.
- We update KubeSpawner properly.
  - When KubeSpawner monitors pods, it should only consider those with a label matching the specific Helm release.
  - When KubeSpawner creates pods and PVCs, it should use another template to avoid name collisions for users with the same username but associated with different Helm releases of the JupyterHub Helm chart
- We update jupyterhub_config.py properly.
  - There are references to resource names such as `proxy-public` and `hub` hardcoded here that should now also adjust given the resource prefix.

### Development advice
While developing this, I suggest using `watch`, `helm template`, and `grep` to identify the resources that we want to update, and while making changes to the templates observing that we successfully render and change the resource names we want to change.

[...]

PR #1905, opened 2020-11-10
Adds serviceAccount/rbac support to all available services

DerekHeldtWerle:
This PR closes both #921 and #616 

This takes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1797 and extends upon it to adhere to helm best practices (as pointed out [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1797#issuecomment-704696195)) to allow users to create and manage individual service accounts for every service in this chart. 

Some notes:

* I went ahead and followed the vast majority of charts in setting up both a `jupyterhub.name` as well as a `jupyterhub.fullname` to better integrate with the `_helpers.tpl` of setting up the service account names.
* Currently the `proxy` doesn't have any associated rules for its `Role` as it doesn't explicitly need them. Future looking, this chart could additionally bundle `podSecurityPolicies` for all associated services, which would then require the inclusion of the ability to `use` the psp.

Similar to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1797,  this does introduce breaking changes with the shift from `rbac.enabled` to `rbac.create`. I do like what @consideRatio did [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1797/files#diff-ee9f5ad3ebff22b16eef6fda8dd8651909036ad5a9abd8f92c4543296072d82dR40-R43) but didn't want to pluck that code into this PR without explicit permission.

Signed-off-by: Derek Heldt-Werle <derek.heldt-werle@viasat.com>


consideRatio:
Hi @DerekHeldtWerle, thank you for working towards making improvements to this repo!

This PR is a too large in scope at the moment, and I'd like to verify the changes for the various parts are suitable to make separately.

- [ ] Using `.fullname` helper functions to name the pods. This is a very big PR that can cause a lot of trouble for users during upgrades. See #1791 more information related to this.
- [ ] Renaming .enabled to .create - there is a PR open and I considered resvoling it in #1797 but I think it isn't worth doing just to align with other charts convention.
- [ ] Ability set a service account for the hub pod and user-scheduler pod make sense in my mind, because they create Role/ClusterRole resources that are tied to a dedicated service account for the pod.
- [ ] I don't understand the reason to add Roles/RoleBinding/ServiceAccounts for pods not needing certain permissions with the k8s api-server. We have two pods that need permissions to interact with the k8s api-server I think, it is the hub pod that creates a pod with KubeSpawner for example, and it is the user-scheduler pod that for example list pods across the entire cluster.
- [ ] Create a PodSecurityPolicy - I think it can be nice to run our Helm chart in our CI systems etc with a somewhat strict PSP to help us tighten down pods of the Helm chart, I'm not sure about making it part of the chart as opt-in or opt-out though as I have too little experience with them.

@DerekHeldtWerle could you open an issue suggesting a specific change you would like to contribute a PR for? My heart cracks a bit every time I need to reject a PR someone has worked on. Thank you for being an active member in the Jupyter ecosystem! :heart:

DerekHeldtWerle:
Hey @consideRatio no worries in the least bit!

> Using .fullname helper functions to name the pods. This is a very big PR that can cause a lot of trouble for users during upgrades. See #1791 more information related to this.

I really only set this up for ease of templating the [default](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1905/files#diff-26310cade9b106b0ccc2f3bdc9ce5b084adfb66d233037e18e88e0d53ee21914R341) service account name when `.create` is set to true, but no explicit name is provided.  I'd be happy to setup a new variable set to `{{ .Release.Name }}` and supply that to all the service accounts created [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1905/files#diff-26310cade9b106b0ccc2f3bdc9ce5b084adfb66d233037e18e88e0d53ee21914R339-R400)

> Renaming .enabled to .create - there is a PR open and I considered resvoling it in #1797 but I think it isn't worth doing just to align with other charts convention.

Makes sense

>Ability set a service account for the hub pod and user-scheduler pod make sense in my mind, because they create Role/ClusterRole resources that are tied to a dedicated service account for the pod.
 I don't understand the reason to add Roles/RoleBinding/ServiceAccounts for pods not needing certain permissions with the k8s api-server. We have two pods that need permissions to interact with the k8s api-server I think, it is the hub pod that creates a pod with KubeSpawner for example, and it is the user-scheduler pod that for example list pods across the entire cluster.

In clusters that enable the pod security policy admission controller there is a hard requirement that _every_ pod must have an associated `serviceAccount` that has a `RoleBinding` that maps to a `Role` that defines its ability to [use](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#via-rbac) a pod security policy that exists in the cluster. The exception to this rule, is when a pod is spun up by another pod that has a valid mapping to a psp (e.g. the hub spinning up the `singleuser` container). In the current state, there are no `serviceAccount`'s for the `proxy`, the `image-puller`, or the `user-scheduler` statefulset. Without the ability to set or create these service accounts, the pods will default to using the `default` service account in the namespace. Best practices for managing RBAC within a cluster urge users to avoid using the `default` service account for pods as any additional permissions granted to it by `Role`'s will then allow any pod in that namespace that doesn't set a `serviceAccountName` the same level of access as specified in the newly associated `Role`.  The best way to avoid granting any unwanted additional access is to explicitly set a `serviceAccountName` for every pod and map an `Role` + `RoleBinding` that grants access to the minimum required resources to properly run. 

In our cluster, for example, attempts to install this chart into the cluster is met with the following:
```
          Error creating: pods "hook-image-awaiter-" is forbidden: unable to validate against any pod security policy: []
```
If we had the ability to set the `serviceAccountName` we could either:

a. Utilize an existing service account in the namespace that is already properly mapped to an existing psp
b. Create a new service account where out of band of this chart we have setup the proper `Role` and `RoleBinding` to map it to an existing psp.

At this time, the only route we could take with the current chart is to grant the `default` user the ability to `use` a psp through a role that would then expose that service account to unintended permissions. With that, the setting up the `Roles`, `RoleBinding`'s, and `ServiceAccounts` is entirely about keeping security best practices in mind, and has nothing to do with the pods ability to reach out to the k8s api server.

> Create a PodSecurityPolicy - I think it can be nice to run our Helm chart in our CI systems etc with a somewhat strict PSP to help us tighten down pods of the Helm chart, I'm not sure about making it part of the chart as opt-in or opt-out though as I have too little experience with them.

Totally reasonable. My main goal in this PR is to provide the minimum amount of changes (creating/setting service accounts for all services) necessary to avoid having to grant the `default` service account in the namespace this gets deployed any additional permissions. 

Let me know if any of the above isn't clear or if you have further questions about the implementation 🙂

consideRatio:
> Hey @consideRatio no worries in the least bit!

I'm glad! Thanks for the feedback!

> In clusters that enable the pod security policy admission controller there is a hard requirement that every pod must have an associated serviceAccount that has a RoleBinding that maps to a Role that defines its ability to use a pod security policy that exists in the cluster.

Ah, I understand this to be close to 100% correct from https://kubernetes.io/docs/concepts/policy/pod-security-policy/#authorizing-policies, it seems that the user or controller that creates the pod needs the `use` permission of a PSP, or, the pod's own service account needs the `use` permission of a PSP.

For you situation, would it be a problem to specify permissions to use a certain PSP of the default service account in the namespace?

DerekHeldtWerle:
> I'm confused about the need to avoid letting the default service account get permissions to use a PSP in general. Can you help me understand PSPs a bit better by clarifying what benefit can come from not granting all pods the ability to use the specific PSP you want to validate against?

So the main benefit would be tailoring each PSP to each service to ensure they have the bare minimum amount of permissions to run. If we allow the `default` service account the ability to `use` any given psp, any pod deployed to that namespace with either `default` set as the `serviceAccountName`, or if no service account was provided at all, would then have the same permissions as whatever service the PSP was originally defined for. 

If we look at the `hub` for example, for the PSP's `spec.volumes` it would require [configmap](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/deployment.yaml#L41), [secret](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/deployment.yaml#L43), [persistentVolumeClaim](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/deployment.yaml#L51), and [downwardAPI](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/deployment.yaml#L158) with the default installation to start up. If, for whatever reason, a nefarious actor (or an unknowing developer) gets access to this namespace and simply has the ability to create pods, they would be able to mount _every_ configmap and secret into their pod as well as provision persistent volumes of any size at will. If instead we could set and create service accounts for each service, in a scenario where a user can only create pods, they wouldn't even be able to create a _single_ pod since no role in the namespace maps to a rolebinding that gives the `default` service account the ability to `use` a psp. 

While the `hub` deployment does set a `serviceAccount`, the idea extends to all potential services in the cluster. By providing _any_ permissions to the `default` service account, the potential for unintended issues related to security or otherwise is significantly increased.

consideRatio:
Thank you @DerekHeldtWerle for your patience with helping me understand more about this feature proposal!

I've tried to triage all ServiceAccount / RBAC related issues/prs and the result is #2680. My current understanding is that PSPs are deprecated and will be removed with k8s 1.25, and as part of this will make the need to create service accounts for all pods something that may no longer be relevant.

I'm sorry we couldn't more quickly come up with a path forward and this PR became stale. At this point I'll close it with a reference to #2680 where its mentioned.

PR #1907, opened 2020-11-12
Support node labels with _ instead of / for EKS

consideRatio:
@yuvipanda it is my understanding that it isn't supported to use `/` in EKS clusters that want autoscaling based on your comment in https://github.com/2i2c-org/pangeo-hubs/commit/955bed4b74cacd09099f82c98915e2801825d078. Is this correct? If so, does this PR make some sense perhaps?

### References:
- [Cluster autoscaler docs for use on AWS](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md)
- [K8s docs on labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) - support one `/` separating a "prefix" and "name", requires name to be max 63 chars.
- [AWS tags restrictions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions) - Say they support `/`, max 128 chars

### DIFF
--- a/doc/source/administrator/optimization.md
+++ b/doc/source/administrator/optimization.md
@@ -259,18 +259,19 @@ following:
 
 1. Setup a node pool (with autoscaling), a certain label, and a certain taint.
 
-    If you need help on how to do this, please refer to your cloud providers
-    documentation. A node pool may be called a node group.
+    If you need help on how to label and taint your nodes in a node pool (also
+    known as node group), please refer to your cloud providers documentation.
 
-    - The label: `hub.jupyter.org/node-purpose=user`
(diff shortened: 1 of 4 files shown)


yuvipanda:
@consideRatio I'm not entirely sure if the issue in the PR is 'real' - it's possible that autoscaling worked for unrelated reasons, as I was changing a lot of things there (ugh on my part). Have you discovered this to be the case in your testing? I'd suggest waiting for your experience before merging this. 

Sorry for the possibly misleading message!

consideRatio:
@yuvipanda ah! I tried with `/` and everything worked! Closing this!

yuvipanda:
Thanks for validating this, @consideRatio!

PR #1911, opened 2020-11-14
Require helm v2.16.0 explicitly and minor CI updates

consideRatio:
I made use of a helm template function (list `concat`) made available in helm v2.16 but not helm v2.15, due to that we need to explicitly require helm v2.16+

Besides that, I also add some small changes to the CI, such as strict linting of helm and passing a commit message when chartpress pushes a new helm chart package commit to our git based helm-chart repo.


michzimny:
@consideRatio, may I ask why the required `kubeVersion` from `1.11` to `1.14`? Is it really needed for z2jh 0.10.x or it's a mistake?

Even for z2jh 0.10.6 [the docs](https://zero-to-jupyterhub.readthedocs.io/en/0.10.6/) say the required k8s version is 1.11.

PR #1913, opened 2020-11-15
Add changelog for 0.10.3 release

consideRatio:
I think this may contain a relevant bugfix by bumping kubernetes/python-client used by kubespawner from 12.0.0 to 12.0.1, it also contains a few smaller enhancements.

I have set my eyes on the 1.0.0 release of z2jh btw, I think it would be helpful to be able to use major/minor/patch to signal changes impact better.

I set the release date to monday tomorrow as a hopeful guess this will be reviewed and merged by then ;)
- @yuvipanda this contains your extra ports release
- @matthew-brett this may contain a fix to the hub pod restarts, I don't know though, see #1912. It will also remove the harmless warning you have observed in the user-scheduler pod, see #1895.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,50 @@ Here you can find upgrade changes in between releases and upgrade instructions.
 
 ## [0.10]
 
+### [0.10.3] - 2020-11-16
+
+This release contain minor enhancements and bugfix in a dependency that could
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for the effort in making quicker releases, @consideRatio!

PR #1923, opened 2020-11-25
Support fullnameOverride / nameOverride and reference resources by named templates

consideRatio:
# Summary

Most helm charts provide a `fullnameOverride` configuration and a `nameOverride` configuration, something that also ships with the `helm create` boilerplate. This PR introduce such configuration options, but the default values are set in a way that retain the current naming convention.

With the new configuration options, there are five kinds of naming schemes that can be used depending on if `fullnameOverride` and `nameOverride` are null, blank, or non-blank strings.

Name format                 | Resource types | fullnameOverride | nameOverride  | Note
-|-|-|-|-
component                   | namespaced     | `""`             | *             | Default
release-component           | cluster wide   | `""`             | *             | Default
fullname-component          | *              | str              | *             | -
release-component           | *              | null             | `""`          | -
release-(name-)component    | *              | null             | str           | omitted if contained in release
release-(chart-)component   | *              | null             | null          | omitted if contained in release

## Changes
- [x] All references to resource names are now replaced with statements like `{{ include "jupyterhub.hub.fullname" . }}`.
- [x] Named templates are made usable with the `.` scope from a chart with JupyterHub as a dependency chart.
  This will be useful for binderhub for example.
- [x] The named templates rendered values are also exposed in the hub configmap.
  - They are directly available as keys such as `hub`, `proxy-public`, and `user-scheduler-ref` which correspond to their named template name. This is useful if you want to mount one name to an environment variable for example.
  - They are also available as a name-templates.yaml blob which is can be easier other times.
- [x] `fullnameOverride` and `nameOverride` config options added
  - [x] With documentation in schema.yaml

## Periphery changes
- [x] A script to help diff changes locally is added
- [x] The configmaps `traefik-proxy-config` and `hub-config` are renamed to `autohttps` and `hub`.
  By doing so we become more systematic in how we name things and avoid introducing two additional named templates to reference specifically for these configmaps.
- [x] We no longer try to load `/etc/jupyterhub/config/config.yaml` which isn't mounted anyhow
  It's not possible for someone to put a file there either because we mount the configmap in a way that would replace that folder.

## Some benefits ?

- A large step towards supporting multiple installations in the same namespace
- Aligns with a common practice of providing fullnameOverride and nameOverride chart config.
- Not backward incompatible (due to our default values of fullnameOverride and nameOverride)
- No more duplication of logic for names ("hub-secret" or .Values.hub.existingSecret etc.)
- Provides a future proof way to reference resources for parent charts

## Some downsides ?

- The chart complexity rose a bit, but I think and we have lines of code now, but I think the readability improved overall.

## Review notes

In #1978 I added the ability to see a diff on the rendered Helm templates between the latest dev release of the chart and the PR. It isn't helpful for changes to the hub configmap or the autohttps configmap as those were renamed. For those specifically, it's better to not use this diff but the unrendered diff with GitHub's normal UI.

Here is [an example helm diff](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1923/checks?check_run_id=1672478064) for 1329fa9 which currently is the latest change.

The documentation is updated in the configuration reference, see [the PR docs preview](https://zero-to-jupyterhub--1923.org.readthedocs.build/en/1923/resources/reference.html#fullnameoverride).

### New default naming convention?
Do we want to transition to a new default naming convention, namely where fullnameOverride and nameOverride is `null` and our resources are named with a `release-(jupyterhub-)component` pattern.

- [x] Decision
  Perhaps, but not in this PR.

### Disallow changes to fullnameOverride / nameOverride?
Should we allow a change fullnameOverride / nameOverride where it probably leads to issues, or should we make template rendering fail if attempting to do so?

- [x] Decision
  Perhaps, but not in this PR.
  
### Should this PR solve #1791?

It currently doesn't. I think the following pieces are missing.

- KubeSpawner's created resources must not have conflicts, so pod_name_template, pvc_name_template and more needs to be configured to include the helm release name or similar.
- KubeSpawner's resource inspections must filter out only the pod it manages.
  Currently, only one label is used as a selector, the configurable component_label which is `component: singleuser-server` by default. While this could be modified to include more information, KubeSpawner should instead support the ability to include more selector labels.

- [x] Decision
  No, let's work towards that in future PRs. This PR is already a sizable chunk that is quite cohesive on its own.


consideRatio:
# Original PR text - now outdated

## Summary

This PR is just refactoring. As part of it, it will rename two resources, `hub-config` will be named `hub`, and `traefik-proxy-config` will be named `autohttps`. This should be fine, as I don't know any Helm chart depending on this Helm chart that mounts these or similar. The main change is to stop use hardcoded string literals to reference resources, but instead reference resources by a named template allowing the actual resource name to change without freely.

This kind of change will help dependent charts templates needing to reference resources in this chart be able to rely on a reliable reference even though the actual resource name would change.

## Actual changes

When the Helm chart is rendered, these are the actual changes.

image

---

## Self reflection

### Practical situations considered

__Chart dependency: reference a resource a Helm template__

If a Helm chart depends on this chart either directly on through another chart that depends on this chart, then they can use the
`"jupyterhub.<proxy|hub|autohttps|proxy-public|etc.>.fullname"` defined templates within their own templates.

[mybinder.org-deploy](https://github.com/jupyterhub/mybinder.org-deploy/) depends on this chart through a dependency of BinderHub. [Here](https://github.com/jupyterhub/mybinder.org-deploy/blob/563c2f2983d05366202a9572fc9fe3b05a38042a/mybinder/templates/proxy-patches/deployment.yaml#L60-L64) the mybinder-org-deploy chart mounts this Helm chart's `hub-secret` resource using a hardcoded name _from a Helm template_. This can easily be fixed after this PR by instead referencing `{{ include "jupyterhub.hub-secret.fullname" .
}}`.

__Chart dependency: reference a resource from values__

[DaskHub](https://github.com/dask/helm-chart/tree/master/daskhub) depends on
this chart. [Here](https://github.com/dask/helm-chart/blob/b62d05e9bc8e41439fa97ae7e990f9d9fd4b8c7d/daskhub/values.yaml#L12-L30) the daskhub chart configures the JupyterHub Helm chart _from its own chart values_. I've explored this in depth, and there just isn't any way to make use
of the named templates before submitting the values for the dependency chart.

DaskHub can pin the name of JupyterHub using `fullnameOverride` when it has become available for the JupyterHub Helm chart, but would then itself lock itself into the same issue that JupyterHub did where its 

No clear recommendation is available for this situation.

__No chart dependency: reference a resource from a Helm template__

[Here](https://github.com/dask/dask-gateway/blob/09526b1de9308cb6bc61351d1562db8ca14e78f8/resources/helm/dask-gateway/templates/gateway/configmap.yaml#L92-L104) dask-gateway assumes the service name of a JupyterHub chart deployed in the same namespace. Perhaps this kind of access should be made configurable as a solution for the dask-gateway Helm chart's coupling.

### Resource naming and fullnameOverride

__What is fullnameOverride?__

The fullnameOverride chart config option is part of new Helm charts created with `helm create` and commonly seen in published Helm charts. When set, the names given all resources will be what this config option is set to.

__fullnameOverride's typical impact__

If fullnameOverride is falsy as it is by default in new charts, the default impact it has is to name the chart's resources by combining the release name and chart name. The combination logic is to only use the release name
(`<release_name>`) if it literally contains the chart name, but if it doesn't, name it to `<release_name>-<chart_name>`. It all comes down to how the defined template in `_helpers.tpl` is configured.

__Handling larger Helm charts__

Most larger Helm chart, like ours, needs more names then just a single name, what if for example you have two separate Deployment resources? What typically is done, is that you append something to the chart's fullname template, such as `-web-server` and `-database-server`.

__What we do in our chart currently__

- _Hardcoded literals_: We currently don't make use of named templates such as `"jupyterhub.fullname"` or `"jupyterhub.proxy.fullname"` as reference to resource names. We use hardcoded string literals to the names across the chart.
- _Mixed naming convention_: When we name Cluster wide resources that could conflict with chart installations in other namespaces, we have prefixed `{{.Release.Name}}-` to the resource name. Examples of cluster wide resources
  are the ClusterRole and ClusterRoleBinding for the user-scheduler, the PriorityClass for normal pods and the user-placeholder pods.

__My plan for the future__

We let our named templates make our resources named `<release_name>-<resource_name>`, which for example will result in `jhub-proxy`. This will align with our current practice for cluster wide resource names.

Also, if we would include `.Chart.Name` we would allow a chart depending on this chart to use a named template in its context where .Chart.Name is different and still get the same name as the `.Release.Name` will be the same while `.Chart.Name` would differ between the contexts.

### Changes in future PRs

__Namespace scoped resources renamed__

`hub-secret` is actually a resource used by other Helm charts, such as mybinder.org-deploy, so changing it will break some behavior for those. We also provide the `hub.existingSecret` configuration which makes it convenient to have a dedicated named template for it. Due to this, I opt to not introduce a name change here even though its weird to have it suffixed with `-secret` as it is a k8s Secret anyhow.

- hub-secret -> ...-hub (secret)

__Cluster scoped resources renamed__

These resources should be fine to rename without trouble I think.

- {{ .Release.Name }}-user-scheduler -> ...-user-scheduler
- {{ .Release.Name }}-default-priority -> ...-default
- {{ .Release.Name }}-user-placeholder -> ...-user-placeholder

__service URL names and env var names changes__

If we use a differently named service, it will change its name, and that will make references to for example http://hub:$HUB_SERVICE_PORT fail.

__Renaming the hub-db-dir__

We can use the helm 3 based lookup function to inspect if there is a PVC named hub-db-dir, and if it is, we can use it, but otherwise use the name it would have been given if we just installed the Helm chart without no history. This way, we solve the backward compatibility issue of renaming a PVC which would lead to loss of data.

jupyterhub/files/hub/z2jh.py
```diff
@@ -9,25 +9,42 @@
 
 import yaml
 
-
 # memoize so we only load config once
 @lru_cache()
 def _load_config():
-    """Load configuration from disk
+    """Load Helm chart configuration passed through a mounted ConfigMap."""
```
minrk, line 15: Isn't this a secret now?

minrk (commented):
> hub-secret is actually a resource used by other Helm charts

I found only [one reference](https://github.com/jupyterhub/mybinder.org-deploy/blob/369e3ff8afb0b8ddfcf132faffadcc3dc1ee6fcd/mybinder/templates/proxy-patches/deployment.yaml#L63) to this in mybinder.org-deploy and none elsewhere, so I don't think that's enough of a reason to leave it out of this consistency update. We will need to be aware of this and make that one line update to mybinder.org-deploy when we next bump this chart, but it is an okay requirement, I think. Plus, it would probably be more prudent for consumers of that info to use `.Values.jupyterhub.proxy.secretToken` in its own secret rather than accessing the secret itself. I'd call that an 'unprotected private api access' like accessing a `_private` attribute in Python. Not really a problem to do, but it's okay to expect more churn if you do use them. The error will also be pretty informative ("failed to mount nonexistent hub-secret volume: no such secret" or similar).

minrk:
I'm just running some `helm diff upgrade` tests with this, but the principle makes sense!

jupyterhub/files/hub/z2jh.py
consideRatio: @minrk we have a configmap for passing files, such as z2jh.py and jupyterhub_config.py, which no longer passes any values.yaml content.

consideRatio:
> I'm just running some `helm diff upgrade` tests with this, but the principle makes sense!

Thank you for helping out with review @minrk!!!

This PR made it essential to do `helm diff`, so I put in some effort in #1978 to provide a `helm diff upgrade` as part of a new upgrade job in the CI system, that upgrades from the latest dev version to the current version.

image

consideRatio:
@minrk I rebased to solve a merge conflict, the commits since last reviews are starting with "Rename: k8s Secret hub-secret to be named hub"

consideRatio:
The `helm diff` was quite clean when I temporarily (in f2ca3c0) made this PR not rename hub-secret to hub, hub-config to hub, and traefik-proxy-config to autohttps. Renaming those resources made large sections go red and green in `helm diff` making it harder to use it for review of this PR specifically.

```diff
NOTE: For the purpose of the diff, we have replaced the new
      chart version with the old to minimize clutter.

default, hub, Deployment (apps) has changed:
...
          hub.jupyter.org/network-access-singleuser: "true"
        annotations:
          # This lets us autorestart when the secret changes!
-         checksum/config-map: 24a8d0a9d5b0ba352fe12229eca1a7fc8727f06d4c0f42d1f0736e2a3090698a
-         checksum/secret: fb96f2d961b0284da7957de984be6d29d1c99f9c51825f1faa8a106f81f58130
+         checksum/config-map: 7d58d2cad5ca6a66816ddf8e1136a13bcbab8f1115f0ffbc1e8ab16696daef7b
+         checksum/secret: 51e6cdc5c51bea0ed2125e062d48000a9ff925c2b637c4acc58340ebf0aef209
      spec:
        nodeSelector: {}
        tolerations: []
...
default, hub-config, ConfigMap (v1) has changed:
...
      configuration_directory = os.path.dirname(os.path.realpath(__file__))
      sys.path.insert(0, configuration_directory)
    
-     from z2jh import get_config, set_config_if_not_none
+     from z2jh import get_config, set_config_if_not_none, get_name, get_name_env
    
    
      def camelCaseify(s):
...
      # Connect to a proxy running in a different pod. Note that *_SERVICE_*
      # environment variables are set by Kubernetes for Services
      c.ConfigurableHTTPProxy.api_url = (
-         f"http://proxy-api:{os.environ['PROXY_API_SERVICE_PORT']}"
+         f'http://{get_name("proxy-api")}:{get_name_env("proxy-api", "_SERVICE_PORT")}'
      )
      c.ConfigurableHTTPProxy.should_start = False
    
...
      # hub_connect_url is the URL for connecting to the hub for use by external
      # JupyterHub services such as the proxy. Note that *_SERVICE_* environment
      # variables are set by Kubernetes for Services.
-     c.JupyterHub.hub_connect_url = f"http://hub:{os.environ['HUB_SERVICE_PORT']}"
+     c.JupyterHub.hub_connect_url = (
+         f'http://{get_name("hub")}:{get_name_env("hub", "_SERVICE_PORT")}'
+     )
    
      # implement common labels
      # this duplicates the jupyterhub.commonLabels helper
...
      if get_config("imagePullSecret.automaticReferenceInjection") and (
          get_config("imagePullSecret.create") or get_config("imagePullSecret.enabled")
      ):
-         image_pull_secrets.append("image-pull-secret")
+         image_pull_secrets.append(get_name("image-pull-secret"))
      if get_config("imagePullSecrets"):
          image_pull_secrets.extend(get_config("imagePullSecrets"))
      if get_config("singleuser.image.pullSecrets"):
...
    
      # scheduling:
      if get_config("scheduling.userScheduler.enabled"):
-         c.KubeSpawner.scheduler_name = os.environ["HELM_RELEASE_NAME"] + "-user-scheduler"
+         c.KubeSpawner.scheduler_name = get_name("user-scheduler")
      if get_config("scheduling.podPriority.enabled"):
-         c.KubeSpawner.priority_class_name = (
-             os.environ["HELM_RELEASE_NAME"] + "-default-priority"
-         )
+         c.KubeSpawner.priority_class_name = get_name("priority")
    
      # add node-purpose affinity
      match_node_purpose = get_config("scheduling.userPods.nodeAffinity.matchNodePurpose")
...
      import os
    
      import yaml
-   
    
      # memoize so we only load config once
      @lru_cache()
      def _load_config():
-         """Load configuration from disk
+         """Load the Helm chart configuration used to render the Helm templates of
+         the chart from a mounted k8s Secret."""
+   
+         path = f"/etc/jupyterhub/secret/values.yaml"
+         if os.path.exists(path):
+             print(f"Loading {path}")
+             with open(path) as f:
+                 return yaml.safe_load(f)
+         else:
+             raise Exception(f"{path} not found!")
+   
+   
+     @lru_cache()
+     def _load_name_templates():
+         """Load dynamically determined k8s resource names so we can reference them
+         from within the container."""
+   
+         path = f"/etc/jupyterhub/config/name-templates.yaml"
+         if os.path.exists(path):
+             print(f"Loading {path}")
+             with open(path) as f:
+                 return yaml.safe_load(f)
+         else:
+             raise Exception(f"{path} not found!")
+   
+   
+     def get_name(name):
+         return _load_name_templates()[name]
+   
    
-         Memoized to only load once
-         """
-         cfg = {}
-         for source in ("config", "secret"):
-             path = f"/etc/jupyterhub/{source}/values.yaml"
-             if os.path.exists(path):
-                 print(f"Loading {path}")
-                 with open(path) as f:
-                     values = yaml.safe_load(f)
-                 cfg = _merge_dictionaries(cfg, values)
-             else:
-                 print(f"No config at {path}")
-         return cfg
+     def get_name_env(name, suffix=""):
+         env_key = _load_name_templates()[name] + suffix
+         env_key = env_key.upper().replace("-", "_")
+         return os.environ[env_key]
    
    
      def _merge_dictionaries(a, b):
...
          data = get_config(key)
          if data is not None:
              setattr(cparent, name, data)
+   name-templates.yaml: |
+     fullname: ""
+     hub: "hub"
+     hub-secret: "hub-secret"
+     hub-pvc: "hub-db-dir"
+     proxy: "proxy"
+     proxy-api: "proxy-api"
+     proxy-http: "proxy-http"
+     proxy-public: "proxy-public"
+     proxy-public-tls: "proxy-public-tls-acme"
+     proxy-public-manual-tls: "proxy-public-manual-tls"
+     autohttps: "autohttps"
+     user-scheduler-deploy: "user-scheduler"
+     user-scheduler-lock: "user-scheduler-lock"
+     user-placeholder: "user-placeholder"
+     hook-image-awaiter: "hook-image-awaiter"
+     hook-image-puller: "hook-image-puller"
+     continuous-image-puller: "continuous-image-puller"
+     singleuser: "singleuser"
+     image-pull-secret: "image-pull-secret"
+     ingress: "jupyterhub"
+     priority: "jupyterhub-default-priority"
+     user-placeholder-priority: "jupyterhub-user-placeholder-priority"
+     user-scheduler: "jupyterhub-user-scheduler"
+   fullname: ""
+   hub: "hub"
+   hub-secret: "hub-secret"
+   hub-pvc: "hub-db-dir"
+   proxy: "proxy"
+   proxy-api: "proxy-api"
+   proxy-http: "proxy-http"
+   proxy-public: "proxy-public"
+   proxy-public-tls: "proxy-public-tls-acme"
+   proxy-public-manual-tls: "proxy-public-manual-tls"
+   autohttps: "autohttps"
+   user-scheduler-deploy: "user-scheduler"
+   user-scheduler-lock: "user-scheduler-lock"
+   user-placeholder: "user-placeholder"
+   hook-image-awaiter: "hook-image-awaiter"
+   hook-image-puller: "hook-image-puller"
+   continuous-image-puller: "continuous-image-puller"
+   singleuser: "singleuser"
+   image-pull-secret: "image-pull-secret"
+   ingress: "jupyterhub"
+   priority: "jupyterhub-default-priority"
+   user-placeholder-priority: "jupyterhub-user-placeholder-priority"
+   user-scheduler: "jupyterhub-user-scheduler"
default, hub-secret, Secret (v1) has changed:
...
  data:
    proxy.token: 'REDACTED # (64 bytes)'
    services.token.test: 'REDACTED # (64 bytes)'
-   values.yaml: '-------- # (11810 bytes)'
+   values.yaml: '++++++++ # (11831 bytes)'
  type: Opaque
  
default, proxy, Deployment (apps) has changed:
...
          hub.jupyter.org/network-access-singleuser: "true"
        annotations:
          # This lets us autorestart when the secret changes!
-         checksum/hub-secret: 1c26ef587f00f0468bcb8e11d44cef773a5000ac08b8a756d9d23181948ad1a7
+         checksum/hub-secret: bb228719db3e9673289a743f5a0d7d51ab80ae4c8b0ce868ebe0ddef8c949f1f
          checksum/proxy-secret: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
      spec:
        terminationGracePeriodSeconds: 60
...
default, user-placeholder, StatefulSet (apps) has changed:
...
        component: user-placeholder
        app: jupyterhub
        release: jupyterhub
-   serviceName: "user-placeholder"
+   serviceName: user-placeholder
    template:
      metadata:
        labels:
...
default, user-scheduler, Deployment (apps) has changed:
...
          app: jupyterhub
          release: jupyterhub
        annotations:
-         checksum/config-map: f82eeb4c3ed6a8a01ee173c639536e3ec9005d25c97b1aa2b389476b3dd382b9
+         checksum/config-map: 9d4f8c4718ddb40a54356e746d1f3a52a8ebb4cd412499372d89b9a8ee9d7dd6
      spec:
        serviceAccountName: user-scheduler
        nodeSelector: {}
...
```

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -47,9 +47,9 @@ spec:
         {{- .Values.hub.extraVolumes | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
         {{- if eq .Values.hub.db.type "sqlite-pvc" }}
-        - name: hub-db-dir
+        - name: pvc
```
yuvipanda, line 50: This is a possibly breaking change, btw. If you are using initContainers that also mount `/srv/jupyterhub`, you'll end up with this error:

```
Error: UPGRADE FAILED: cannot patch "hub" with kind Deployment: Deployment.apps "hub" is invalid: spec.template.spec.containers[1].volumeMounts[0].name: Not found: "hub-db-dir"
```

We should at least note this somewhere.

consideRatio: Ah good point

PR #1925, opened 2020-11-30
bump oauthenticator to 0.12.2

### LINKED ISSUE #1914: Authenticator not honoring the auth.whitelist.users key
### Bug description

The user whitelist is not being honored when using the Gitlab OAuthenticator.

#### Expected behaviour

Only users in the whitelist should be able to login to the JupyterHub.

#### Actual behaviour

Any (Gitlab) user is able to login and use the JupyterHub instance.

### How to reproduce

Try logging in with a account not in the whitelist.

### Your personal set up

```yaml
auth:
  whitelist:
    users:
      - user1
      - user2
  admin:
    access: true
    users:
      - <REDACTED>
  type: gitlab
  gitlab:
    clientId: <REDACTED>
    clientSecret: <REDACTED>
    callbackUrl: <REDACTED>
  scopes:
    - "read_user"
```
 - OS: Linux
 - Version: Chart 0.10.2
 - Configuration: See above.

minrk:
fixes issue with deprecated Authenticator config being ignored instead of preserved

closes #1914 (which was also fixed in release 0.10.6)

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -38,7 +38,7 @@ mako==1.1.3               # via alembic
 markupsafe==1.1.1         # via jinja2, mako
 mwoauth==0.3.7            # via -r requirements.in
 nullauthenticator==1.0.0  # via -r requirements.in
-oauthenticator==0.12.1    # via -r requirements.in
+oauthenticator==0.12.2    # via -r requirements.in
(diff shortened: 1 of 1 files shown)


consideRatio:
(Linting failed because of a thing I pushed directly to master without pleasing pre-commit linting, I'll fix that)

PR #1926, opened 2020-11-30
Changelog for 0.11.0

consideRatio:
This references the jupyterhub/oauthenticator security advisory, but it leaves some open questions to users still. 

There are some open questions for users below that I failed to address before and lost momentum on the releases.
- [ ] How to identify users that should be deleted because they got access when they shouldn't have?
- [ ] About the need to clean up resources created for the users, such as PVCs that should be deleted as well.

Any help with concrete suggestions, a LGTM, or added commits etc greatly appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,11 +26,12 @@ See [the published security
 advisory](https://github.com/jupyterhub/oauthenticator/security/advisories/GHSA-384w-5v3f-q499)
 for more information, and refer to [this forum
 post](https://discourse.jupyter.org/t/collaboration-to-mitigate-issues-of-security-advisory-in-oauthenticator/7520)
-to share experiences and insights that can be useful to others.
+to share insights that can be useful to others.
 
 #### Breaking changes
 
-- __`auth` configuration moves to `hub.config` - [#1943](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1943) ([@consideRatio](https://github.com/consideRatio))__
+- __`auth` configuration moves to `hub.config` - [#1943](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1943)__
+
   Helm chart configuration under `auth` is now no longer supported. If you make
   a `helm upgrade` using `auth` configuration, the upgrade will abort before any
   changes are made to the k8s cluster and you will be provided with the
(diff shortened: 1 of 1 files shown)


minrk (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,28 @@ Here you can find upgrade changes in between releases and upgrade instructions.
 
 ## [0.10]
 
+### [0.10.7] - 2020-11-30
+
+This release contains the patched version of jupyterhub/oauthenticator which contained a security issue that the 0.10.6 release of this Helm chart worked around. See https://github.com/jupyterhub/oauthenticator/security/advisories/GHSA-384w-5v3f-q499 for more information.
+
+Please don't use versions 0.10.0 - 0.10.5 and upgrade to 0.10.6 or later. If any users have been authorized during usage of 0.10.0 - 0.10.5 who should not have been, they must be deleted via the API or admin interface, [per the documentation](https://jupyterhub.readthedocs.io/en/1.2.2/getting-started/authenticators-users-basics.html#add-or-remove-users-from-the-hub).
```
yuvipanda, line 11:
```suggestion
Please don't use versions 0.10.0 - 0.10.5 and upgrade to 0.10.6 or later. If you are using OAuthenticator, please check your list of users and [delete any unauthorized users who may have logged in since your upgrade to 0.10.0 - 10.10.5](https://jupyterhub.readthedocs.io/en/1.2.2/getting-started/authenticators-users-basics.html#add-or-remove-users-from-the-hub). 
```

yuvipanda:
>  How to identify users that should be deleted because they got access when they shouldn't have?

Made a suggested edit for this.

Re: PVCs, not sure! Maybe we should add a section in our docs about cleaning up user PVCs?

minrk:
> Re: PVCs, not sure! Maybe we should add a section in our docs about cleaning up user PVCs?

I think that's a general topic we could have more coverage for: what happens when you delete a user, what gets deleted and what doesn't, and how to clean up what is leftover.

consideRatio:
I let this go stale and now I don't want to work to fix a 0.10.7 release but instead go for 0.11.0 directly.

choldgraf:
This LGTM - the only feedback I have is that some of the changelog items are a bit minor to explicitly list in the changelog (e.g., https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1926/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR95). I don't think this should block the PR though

consideRatio:
@choldgraf yeah - I'm a bit afraid of defining where to draw a line of what to include that isn't very distinct.

One could exclude docs/ci changes for example though.

PR #1933, opened 2020-12-07
CI: fix of intermittent netpol test failure

### LINKED ISSUE #1927: CI: Intermittent failure error code 4
Our CI system fails intermittently like this.

image

---

The error code is from wget apparently: `4       Network failure`.

consideRatio:
This PR fixes a issue in our tests caused by jupyter.org added a new IP address, so when we tried to access jupyter.org which was assumed be sent to one of two ips, it could sometime be sent to another IP.

This PR Closes #1927.

In this PR I also bump helm to 3.4.1 from 3.4.0, and starts using manics/actions-k3s-helm of the latest version so we get a more modern version of Calico for example. These changes only influence the test suite and shouldn't hurt having as part of this PR in my mind, so I'm saving myself the trouble of creating a dedicated PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,13 +90,10 @@ jobs:
       # kubectl and helm
       #
       # ref: https://github.com/manics/action-k3s-helm/
-      - uses: manics/action-k3s-helm@v0.2.1
+      - uses: manics/action-k3s-helm@dev
         with:
           k3s-version: ${{ matrix.k3s-version }}
-          # FIXME: Let's rely on the latest helm version by omitting this when
-          #        we can bump beyond manics/action-k3s-helm@v0.2.1 that has
-          #        such support.
-          helm-version: v3.4.0
+          helm-version: v3.4.1
           metrics-enabled: false
           traefik-enabled: false
           docker-enabled: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,7 @@ jobs:
       # kubectl and helm
       #
       # ref: https://github.com/manics/action-k3s-helm/
-      - uses: manics/action-k3s-helm@dev
+      - uses: manics/action-k3s-helm@main
         with:
           k3s-version: ${{ matrix.k3s-version }}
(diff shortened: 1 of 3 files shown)


betatim:
Looks like a reasonable thing to do.

What is confusing is that mybinder.org is listed as a blocked URL. But the error message says "Access to {blocked_url} should have worked". Which seems odd. Surely we want access to the blocked URL to be blocked?

consideRatio:
@betatim ah nice catch, the test made is correct but the printed message is wrong I think.

This change did not solve the issue apparently, because one of the test failures still fails in the same way. I think that the network policy enforcement (for us Calico) is acting up. I wonder if it is related to calico versions in general or something like that, perhaps there is a bugfix in calico available for the version we currently use stemming from the k3s github action developed by manics.

consideRatio:
FINALLY! I tracked it down to the fact that jupyter.org can return other IPs than those we have allowed explicitly. To mitigate this issue in the future, I've made us test to open a TCP connection to the specific whitelisted ip without involving DNS lookups in that test. Ability to make DNS lookups are tested earlier.

betatim:
Which IP are you using? I think the reason jupyter.org returns different IPs at different times is because it is hosted on a CDN. I would not be surprised if the CDN removes/adds IPs "at random", so our test will fail again when that happens.

Unfortunately I don't know an IP that we can make a HTTP request to where the IP is going to be stable for "super long times" :-/

consideRatio:
Two jupyter.org ips like before where the netpol allows one but not the other, and we test the allowed ip and the blocked ip without using the domain name jupyter.org at that point.

This can be compared to the old test which was allowing two jupyter.org IPs and none from mybinder.org, and testing against jupyter.org as allowed and mybinder.org as a blocked.

---

What I know couls fail in the future, is if the blocked ip is removed. Then it will fail as wanted for a different reason than the blocked ip is blocked. I dont want to put in my own time to make it more reliable against this risk though. I tried adding a nslookup of jupyter.org to verify both IPs were still around, but that could fail also for reasons probably like you guess about CDN, so then i surrendered on covering that situation.

PR #1938, opened 2020-12-11
Disable PDBs for hub/proxy, add PDB for autohttps, and relocate config proxy.pdb to proxy.chp.pdb

### LINKED ISSUE #1934: Deployments with 1 replica and a PodDisruptionBudget enabled by default prevent node drain operations from succeeding
### Bug description
Installing the chart with the default values results in workloads such as the hub created with a hardcoded `replica: 1` and a PodDisruptionBudget with `minAvailable: 1`.
```
kubectl get poddisruptionbudget --all-namespaces
NAMESPACE    NAME               MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
jupyterhub   hub                1               N/A               0                     191d
jupyterhub   proxy              1               N/A               0                     191d
jupyterhub   user-placeholder   0               N/A               0                     191d
jupyterhub   user-scheduler     1               N/A               0    
```

This prevents operations such as draining the node with kured (see https://github.com/weaveworks/kured/issues/82) which require the workload to be redeployed to a different node, killing that single replica in the process. Any operation that needs to kill that single replica will conflict with the PodDisruptionBudget that is defined by default.

You can see in the kured's output the following error repeated again and again:
```
time="2020-12-07T11:59:35Z" level=info msg="evicting pod jupyterhub/hub-6dddf4fdc7-mv8jg" cmd=/usr/bin/kubectl std=out
time="2020-12-07T11:59:35Z" level=warning msg="error when evicting pod \"hub-6dddf4fdc7-mv8jg\" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget." cmd=/usr/bin/kubectl std=err
time="2020-12-07T11:59:40Z" level=info msg="evicting pod jupyterhub/hub-6dddf4fdc7-mv8jg" cmd=/usr/bin/kubectl std=out
time="2020-12-07T11:59:40Z" level=warning msg="error when evicting pod \"hub-6dddf4fdc7-mv8jg\" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget." cmd=/usr/bin/kubectl std=err
```

The only way to redeploy such workload and automated processes such as node drains or kured reboots is by manually removing the PDB. (See https://kubernetes.io/docs/tasks/run-application/configure-pdb/#think-about-how-your-application-reacts-to-disruptions)

#### Expected behaviour
I would expect the PodDisruptionBudgets to be disabled by default (or to include a `minAvailable: 0` by default). This way installing Jupyter using its default values will not prevent nodes from being drained (or similar operations that might require redeploying the workload)

#### Actual behaviour
by default jupyter will deploy both the `hub` and `proxy` workloads with one replica and a PodDisrutionBudget that includes `minAvailable: 1`. This prevents maintenance operations like a controller kured reboot (which needs to drain the node) since any operation killing the container will fail.

### How to reproduce

Follow these steps:
1. Install and start minikube
2. Install kured using its official chart
    ```
    $ helm repo add kured https://weaveworks.github.io/kured
    $ helm install kured kured/kured --set configuration.period=1m
    ```
3. Install jupyter using this chart
    ```
    $ helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
    $ helm repo update
    $ helm install jupyter jupyterhub/jupyterhub --set proxy.secretToken=$(openssl rand -hex 32)
    ```
4. Run  `minikube ssh`
5. Run `sudo touch /var/run/reboot-required`
5. Check the logs of the kured container. Note how the drain operation is consistently being attempted and failing due to the default PodDisruptionBudget
    ```
    $ kubectl logs kured-8ndbm --tail 10 -f
    ...
    error when evicting pod "proxy-569d78d49c-wc6db" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
    evicting pod default/user-scheduler-8664cb5984-hctvs
    error when evicting pod "user-scheduler-8664cb5984-hctvs" (will retry after 5s): Cannot evict pod as it would violate the 
    pod's disruption budget.
    ...
    ```

If you want, cleanup your minikube environment with:
```
$ helm delete kured
$ helm delete jupyter
```
### Your personal set up

Minikube:
 ```
$ minikube version
minikube version: v1.15.1
```

Jupyter chart
```
$ helm show chart jupyterhub/jupyterhub
apiVersion: v1
appVersion: 1.2.2
description: Multi-user Jupyter installation
home: https://z2jh.jupyter.org
icon: https://jupyter.org/assets/hublogo.svg
kubeVersion: '>=1.14.0-0'
maintainers:
- email: erik@sundellopensource.se
  name: Erik Sundell
- name: Simon Li
  url: https://github.com/manics/
name: jupyterhub
sources:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s
version: 0.10.6
- kured chart
```

kured chart
```
$ helm show chart kured/kured
apiVersion: v1
appVersion: 1.5.1
description: A Helm chart for kured
home: https://github.com/weaveworks/kured
icon: https://raw.githubusercontent.com/weaveworks/kured/master/img/logo.png
maintainers:
- email: christian.kotzbauer@gmail.com
  name: ckotzbauer
- email: david@davidkarlsen.com
  name: davidkarlsen
name: kured
sources:
- https://github.com/weaveworks/kured
version: 2.2.1
```

### PR DESCRIPTION
**Pull request by consideRatio**
This PR follows #1937 and discussion in #1934. I suggest we continue the discussion about _what to do_ in #1934 unless it is about the lines of codes changed.

I've opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1951 to represent the wish for better documentation about HA so that the we are more explicit with assumptions.

## PR summary

- [x] Disabling hub/proxy PDB by default.
  This is done to close #1934, and motivated by them being non-HA applications. For more in-depth discussion, see #1934.
- [x] Added autohttps PDB, also disabled by default.
  This is done for completeness as it would generally make sense to enable PDBs for proxy/hub/autohttps all at once if either was enabled.
- [x] Renamed pdb configuration proxy.pdb to to proxy.chp.pdb.
  This was done to make the configuration more cohesive and easy to understand now that we have one for the proxy pod and the autohttps pod. This is in alignment of how we have mapped the other configuration. It will make it easier in the future if we replace the configurable-http-proxy with another implementation also that may merit from other defaults. The old name is deprecated with a `fail` to help users quickly spot the change and transition.

## Motivation for the new default

1. Most other Helm chart defaults to having PDBs disabled by default
   making our chart's behavior unexpected.

2. The idea of using a PDB is tightly coupled with applications that
   support high availability, in other words, to run in parallel to
   another. But, the hub/proxy/autohttps pod doesn't support it. As an
   indication of this, see the k8s docs on "configure-pdb" which has a
   "Before you begin" section that sais "You are the owner of an
   application running on a Kubernetes cluster that requires high
   availability.

   ref: https://kubernetes.io/docs/tasks/run-application/configure-pdb/

3. When evictions take place, such as when a node is drained, the PDBs
   will be respected, and since we have `minAvailable: 1` and cannot
   create another one, a manual pod deletion is required.

   It is my perception that the act of draining a node should come with
   intent and consideration typically. If it comes as part of an
   automated k8s upgrade of nodes, it should at least be scheduled at a
   acceptable time window etc. So, it wouldn't be helpful to add the
   need to run `kubect delete pod` alongside this.

4. If the wish is to avoid making the cluster autoscaler relocate a pod
   part of a deployment, then it is more direct and sensible to use the
   `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"`
   annotation.

   ref: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -562,10 +562,24 @@ properties:
             type: object
             description: |
               Kubernetes annotations to apply to the hub service.
-      pdb:
+      pdb: &pdb-spec
         type: object
         description: |
-          Set the Pod Disruption Budget for the hub pod.
+          Configure a PodDisruptionBudget for this Deployment.
+
+          These are disabled by default for our deployments that doesn't support
+          being run in parallel with multiple replicas. Only the user-scheduler
+          currently support being run in parallel with multiple replicas. If
+          they are enabled for a Deployment with only one replica, they will
+          block `kubectl drain` of a node for example.
+
+          Note that if you aim to block scaling down a node with the
+          hub/proxy/autohttps pod that would cause disruptions of the
+          deployment, then you should instead annotate the pod's of the
+          Deployment [as described
+          here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node).
+
+              "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
 
           See [the Kubernetes
           documentation](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/)
@@ -574,11 +588,13 @@ properties:
           enabled:
             type: boolean
             description: |
-              Whether PodDisruptionBudget is enabled for the hub pod.
+              Decides if a PodDisruptionBudget is created targeting the
+              Deployment's pods.
           minAvailable:
             type: integer
             description: |
-              Minimum number of pods to be available during the voluntary disruptions.
+              The minimum number of pods required to be available during
+              voluntary disruptions.
       existingSecret:
         type:
           - string
@@ -702,6 +718,7 @@ properties:
 
               For more information, see the [Kubernetes EnvVar
               specification](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#envvar-v1-core).
+          pdb: *pdb-spec
           nodeSelector: *nodeSelector-spec
           tolerations: *tolerations-spec
       secretToken:
@@ -886,23 +903,6 @@ properties:
               hosts:
                 - <your-domain-name>
               ```   
-      pdb:
-        type: object
-        description: |
-          Set the Pod Disruption Budget for the proxy pod.
-
-          See [the Kubernetes
-          documentation](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/)
-          for more details about disruptions.
-        properties:
-          enabled:
-            type: boolean
-            description: |
-              Whether PodDisruptionBudget is enabled for the proxy pod.
-          minAvailable:
-            type: integer
-            description: |
-              Minimum number of pods to be available during the voluntary disruptions.
       traefik:
         type: object
         description: |
@@ -952,6 +952,7 @@ properties:
 
               For more information, see the [Kubernetes EnvVar
               specification](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#envvar-v1-core).
+          pdb: *pdb-spec
           nodeSelector: *nodeSelector-spec
           tolerations: *tolerations-spec
     required:
@@ -1195,23 +1196,7 @@ properties:
               You can have multiple schedulers to share the workload or improve
               availability on node failure.
           image: *image-spec
-          pdb:
-            type: object
-            description: |
-              Set the Pod Disruption Budget for the user scheduler.
-
-              See [the Kubernetes
-              documentation](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/)
-              for more details about disruptions.
-            properties:
-              enabled:
-                type: boolean
-                description: |
-                  Whether PodDisruptionBudget is enabled for the user scheduler.
-              minAvailable:
-                type: integer
-                description: |
-                  Minimum number of pods to be available during the voluntary disruptions.
+          pdb: *pdb-spec
           nodeSelector: *nodeSelector-spec
           tolerations: *tolerations-spec
       podPriority:
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -567,15 +567,15 @@ properties:
         description: |
           Configure a PodDisruptionBudget for this Deployment.
 
-          These are disabled by default for our deployments that doesn't support
+          These are disabled by default for our deployments that don't support
           being run in parallel with multiple replicas. Only the user-scheduler
-          currently support being run in parallel with multiple replicas. If
+          currently supports being run in parallel with multiple replicas. If
           they are enabled for a Deployment with only one replica, they will
           block `kubectl drain` of a node for example.
 
           Note that if you aim to block scaling down a node with the
           hub/proxy/autohttps pod that would cause disruptions of the
-          deployment, then you should instead annotate the pod's of the
+          deployment, then you should instead annotate the pods of the
           Deployment [as described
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by manics** [comment]
LGTM 😄 

Does schema.yml also need an update?
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/dea8badd102ac6bd76e7c1df20b38afec83b0fb1/jupyterhub/schema.yaml#L889-L905

Also what do you think about copying your first commit message https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1938/commits/7ccfd50a5ae275f8d3a38de7ad664e1b15e5bdb3
into the description of this PR? I think it's a good explanation, and it means if someone comes to the PR in future they'll see the rationale in the PR description without digging into the commits or reading through the linked issues.

**Comment by consideRatio** [comment]
@manics thanks for your time reviewing this!!! :heart: :tada:

I copied the commit message over to the PR and updated schema.yaml by relocating the proxy.pdb config to proxy.chp.pdb as well as adding one for traefik. I refactored away the repeating text by using a YAML anchor as well.

File: jupyterhub/schema.yaml [hunk 1]
```diff
@@ -562,10 +562,24 @@ properties:
...
             description: |
               Kubernetes annotations to apply to the hub service.
-      pdb:
+      pdb: &pdb-spec
         type: object
         description: |
-          Set the Pod Disruption Budget for the hub pod.
+          Configure a PodDisruptionBudget for this Deployment.
+
+          These are disabled by default for our deployments that doesn't support
```
manics, line 570:
```suggestion
          These are disabled by default for our deployments that don't support
```

jupyterhub/schema.yaml
```diff
@@ -562,10 +562,24 @@ properties:
...
+          being run in parallel with multiple replicas. Only the user-scheduler
+          currently support being run in parallel with multiple replicas. If
```
manics, line 572:
```suggestion
          currently supports being run in parallel with multiple replicas. If
```

jupyterhub/schema.yaml
```diff
@@ -562,10 +562,24 @@ properties:
...
+          they are enabled for a Deployment with only one replica, they will
+          block `kubectl drain` of a node for example.
+
+          Note that if you aim to block scaling down a node with the
+          hub/proxy/autohttps pod that would cause disruptions of the
+          deployment, then you should instead annotate the pod's of the
```
manics, line 578:
```suggestion
          deployment, then you should instead annotate the pods of the
```

manics (commented):
A few typos but otherwise LGTM!

Off topic ... but since schema.yaml has come up, do you know anything about `values.schema.json`  https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/?

consideRatio:
:heart: your help with review matters a lot to me, thank you again!! :tada: :heart:

I find my focus bottlenecking if I have too many open active issues/prs, so I really appreciate helping moving them to a decisive outcome :)

consideRatio:
I've opened #1951 to represent the wish for better documentation about HA so that the we are more explicit with assumptions.

I suggest to merge this and not also do those docs update as part of this PR.

manics (approved):
I agree with merging this now. Worst case we can always revert this before release.

consideRatio:
:tada: 

I'll document this change as part of the changelog explicitly.

meeseeksmachine:

PR #1943, opened 2020-12-14
add hub.config passthrough and use it for all auth config

### LINKED ISSUE #1379: Documentation update for LDAPAuthenticator
Spent today getting LDAP working on my server. It was...frustrating. The first issue is the mapping of zero-to-jupyterhub values to actual jupyterhub config values is not always clear. It is-

```
    c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
    c.LDAPAuthenticator.server_address = get_config('auth.ldap.server.address')
    set_config_if_not_none(c.LDAPAuthenticator, 'server_port', 'auth.ldap.server.port')
    set_config_if_not_none(c.LDAPAuthenticator, 'use_ssl', 'auth.ldap.server.ssl')
    set_config_if_not_none(c.LDAPAuthenticator, 'allowed_groups', 'auth.ldap.allowedGroups')
    set_config_if_not_none(c.LDAPAuthenticator, 'bind_dn_template', 'auth.ldap.dn.templates')
    set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn', 'auth.ldap.dn.lookup')
    set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn_search_filter', 'auth.ldap.dn.search.filter')
    set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn_search_user', 'auth.ldap.dn.search.user')
    set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn_search_password', 'auth.ldap.dn.search.password')
    set_config_if_not_none(c.LDAPAuthenticator, 'lookup_dn_user_dn_attribute', 'auth.ldap.dn.user.dnAttribute')
    set_config_if_not_none(c.LDAPAuthenticator, 'escape_userdn', 'auth.ldap.dn.user.escape')
    set_config_if_not_none(c.LDAPAuthenticator, 'valid_username_regex', 'auth.ldap.dn.user.validRegex')
    set_config_if_not_none(c.LDAPAuthenticator, 'user_search_base', 'auth.ldap.dn.user.searchBase')
    set_config_if_not_none(c.LDAPAuthenticator, 'user_attribute', 'auth.ldap.dn.user.attribute')
```

Which I figured out by reading the config.py file inside the running pod. I am not allowed to make PRs from work or I would just go add it somewhere myself.

### LINKED ISSUE #1871: Rework of Authenticator configuration (General Python class traitlets configuration)
After looking through all the issues in this repo, the most recurring issue related to configuration of the JupyterHub Authenticator inside the `hub` pod where JupyterHub run's through the Helm chart's template values.

## How it works now
This is the current flow of configuration.

1. Helm chart users update their config.yaml
2. Helm chart renders a k8s Secret resource with the users config.yaml merged with the Helm chart's default values.yaml
3. The hub pod mounts the merged values as a .yaml file
4. The hub pod starts jupyterhub with a jupyterhub_config.py file that in turn will parse the values in the .yaml file
5. One configuration option at the time, the passed values will update configuration of Python classes like KubeSpawner or GitHubOAuthenticator.

## Outcome of that
[Issue 0](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/521), [Issue 1](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/644#issuecomment-703857662), [Issue 2](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/673), [Issue 3](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/968), [Issue 4](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1166), [Issue 5](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1167), [Issue 6](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1171), [Issue 7](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1349), [Issue 8](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1789), [Issue 9](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html#authenticating-with-ldap), [Issue 10](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1435), [Issue 11](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1551#issuecomment-705104894), [Issue 12](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1600)

## Underlying issue diagnosis
There are many issues stemming from step 5.
1. The config.yaml -> Authenticator class configuration mapping require constant maintenance.
2. The config.yaml -> Authenticator class configuration mapping has documented examples which also need constant maintenance.
3. Since the config.yaml -> Authenticator class configuration mapping is camelCase (YAML in general) to snake_case (Python class), we cannot just refer to the Authenticator's own documentation (clientSecret -> client_secret)
4. Since the config.yaml -> Authenticator class configuration mapping also nests configuration, we cannot just refer to the Authenticator's own documentation ([examples here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/8d5503db66a8c8b16ee36d57e18ac1326736ef91/jupyterhub/files/hub/jupyterhub_config.py#L356-L368)).

## Solution idea
We can make a X:X mapping of the configuration that won't need maintenance instead, like described in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1396.

## Action points
- [x] Suggest a configuration syntax
- [x] Iterate and come to an agreement on the configuration syntax
- [x] Implement the configuration syntax in parallel to the old configuration under `auth` and let it start emit deprecation warnings when using `auth` in NOTES.txt.

consideRatio:
Closes #1871.
Closes #1379 by updating the documentation examples to be directly mapped to LDAP config.

## PR Status

Review -> merge remain, I don't know of anything I'd like to do further at this point for this PR.

- [x] The deprecation logic is tested against all the auth config examples in the docs we had, and it worked out fine. Automating testing seems overkill as we shouldn't make changes to it over time.
- [x] An e2e test of the deprecation and that the new configuration is valid has been done successfully.

## Review suggestions

For me, the key part to review is the deprecation mechanisms. Secondary, we can focus on the documentation language etc, but I think we should iterate on that later as this PR is required for a release that I'd like to make. Below are some review points anyhow.

- [x] 1. Updates to schema.yaml in 5d5778f
- [x] 2. Changes in jupyterhub_config.py e4b64ce 72e3967
- [x] 3. Adding of null safe lookup utility in ee323db
- [x] 4. Handling of auth.state.cryptoKey in 9ed09c0
- [x] 5. Transform of configuration in authentication.md 245e86c
  @minrk I replaced replaced env var based configuration as well, because I recall your refactoring in oauthenticator made us no longer need to use env vars + traitlet config but could instead rely on only traitlet config of token_url etc. I have also verified this worked for a deployment of mine. Does it sound okay to you?
- [x] 6. The documentation added before the "Configuring authenticator classes" section - [PR preview](https://zero-to-jupyterhub--1943.org.readthedocs.build/en/1943/administrator/authentication.html)

## Summary
This Helm chart's configuration of authenticator classes such as GitHubOAuthenticator were as observed by the influx of issues not so straightforward for users, and it wasn't so straightforward for maintainers either. We had to keep up with adding mappings between the helm charts configuration and the authentication classes configuration.

In this PR, we leave that system behind in favor of a more general system. In this transition users having configured authentication will fail directly when they run `helm upgrade` and be provided with an useful error message to help with the transition.

### Transition message

__config.yaml__

```
auth:
  key1: val1
  key2:
    key3: val3
  type: github
  google:
    clientSecret: test-secret
    clientId: test-id
    callbackUrl: callback-url
  scopes:
    - profile
    - email
  state:
    enabled: true
  admin:
    access: true
    users:
      - cat
  whitelist:
    users:
      - kitten1
  allowedUsers: # maps to the same as whitelist.users
    - kitten2
```

__helm upgrade__

```
The JupyterHub Helm chart's auth config has been reworked and requires changes.

The new way to configure authentication in chart version 0.11.0+ is printed
below for your convinience. The values are not shown by default to ensure no
secrets are exposed, run helm upgrade with --set global.safeToShowValues=true to
show them.

hub:
  config:
    Authenticator:
      admin_users: '***'
      allowed_users: '***'
      enable_auth_state: '***'
    JupyterHub:
      admin_access: '***'
      authenticator_class: oauthenticator.github.GitHubOAuthenticator
    OAuthenticator:
      client_id: '***'
      client_secret: '***'
      oauth_callback_url: '***'
      scope: '***'
    WarningUnrecognizedConfig:
      key1: '***'
      key2.key3: '***'

For more details, please see the updated auth config documentation at:
https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html
```

__helm upgrade --set global.safeToShowValues=true__

```
...

hub:
  config:
    Authenticator:
      admin_users:
      - cat
      allowed_users:
      - kitten1
      enable_auth_state: true
    JupyterHub:
      admin_access: true
      authenticator_class: oauthenticator.github.GitHubOAuthenticator
    OAuthenticator:
      client_id: test-id
      client_secret: test-secret
      oauth_callback_url: callback-url
      scope:
      - profile
      - email
    WarningUnrecognizedConfig:
      key1: val1
      key2.key3: val3

...
```


doc/source/conf.py
```diff
@@ -162,7 +162,7 @@ def parse_schema(d, md=[], depth=0, pre=""):
         depth += 1
         # Create markdown headers for each schema level
         for key, val in d["properties"].items():
-            md.append("(schema:%s)=" % (pre + key))
+            md.append("(schema_%s)=" % (pre + key))
```
consideRatio, line 165: @choldgraf I think there may be one or two issues to report in the MyST ecosystem, do you have suggestions on what repo's to report this in?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
+(authentication)=
+
+# Authentication and authorization
+
+Authentication is about identity, while _authorization_ is about access rights.
+In this section you will learn how to configure both. For example, you can
+configure to authenticate GitHub users and authorize those part of a specific
+GitHub organization.
```
manics, line 8: Not sure, but I think you're trying to say
"You can configure authentication using GitHub accounts, but restrict what users are authorised to do (or what users can log in?) based on membership of a GitHub organiisation"
?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Useful understanding
+
+### Authenticator classes
+
+As JupyterHub can't know how to authenticate all kinds of users, it expose
+configuration to extend its capabilities to rely on a dedicated [_authenticator
+class_](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html).
+Several such classes are already made available through [installed Python
+packages](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/hub/requirements.txt).
```
manics, line 20: Is there any magic in MyST that would allow the tag to appear here instead of `master`? So that the stable docs corresponding to a release could link to the correct set of packages, whereas the dev/master docs could link to `master`?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+### The Authenticator base class
+
+JupyterHub provides a base class,
+[`Authenticator`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/auth.py),
+that all other authenticator classes are supposed to derive from. By configuring
+this base class, we influence the behavior of the derived class as well.
+
+### The configuration system
+
+We configure the JupyterHub to use our chosen authenticator class and the
```
manics, line 31:
```suggestion
We configure JupyterHub to use our chosen authenticator class and the
```
Or maybe "the JupyterHub deployment"?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+
+### OAuth2 based authentication
+
+JupyterHub's
+[oauthenticator](https://github.com/jupyterhub/oauthenticator) has
+support for enabling your users to authenticate via a third-party OAuth
+provider, including GitHub, Google, and CILogon.
+
+Follow the service-specific instructions linked on the [oauthenticator
+repository](https://github.com/jupyterhub/oauthenticator) to generate
```
manics, line 71: There's also a readthedocs site https://oauthenticator.readthedocs.io/en/stable/ , does this also have the required information?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+your JupyterHub instance's OAuth2 client ID and client secret. Then
+declare the values in the helm chart (`config.yaml`).
+
+Here are example configurations for common authentication services. Note
+that in each case, you need to get the authentication credential
+information before you can configure the helm chart for authentication.
```
manics, line 77:
```suggestion
that in each case, you need to get the client ID and secret
before you can configure the helm chart for authentication.
```
I think repeating this is clearer than introducing a new term "authentication credential information"

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+    GoogleOAuthenticator:
+      hosted_domain:
+        - your-university.edu
+      login_service: Your university
+    JupyterHub:
+      authenticator_class: oauthenticator.GoogleOAuthenticator
+```
+
+#### CILogon
+
```
manics, line 165: Could you add a couple of sentences on what CILogon is, since it's the only probvider that doesn't have any background.

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+      authenticator_class: oauthenticator.generic.GenericOAuthenticator
+```
+
+#### Auth0
+
+Auth0 is a popular commercial provider of identity management. The
+JupyterHub helm chart does not include support for Auth0 by default. To
+use Auth0, `extraEnv` and `extraConfig` must be configured as follows:
+
+Note that without the scope defined, authenticating to JupyterHub after
```
manics, line 272:
```suggestion
Note that without the scopes defined, authenticating to JupyterHub after
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+6.  Set "Authorized JavaScript origins" to be your hub's URL.
+7.  Set "Authorized redirect URIs" to be your hub's URL followed by
+    `/hub/oauth_callback`. For example,
+    `https://your-jupyterhub-domain/hub/oauth_callback`.
+8.  When you click "Create", the console will generate and display a
+    Client ID and Client Secret. Save these values.
+9.  Type "consent screen" in the search field at the top and click to
+    access the OAuth consent screen. Here you will customize what your
+    users see when they login to your JupyterHub instance for the first
+    time. Click Save when you are done.
```
manics, line 322: As an aside, the above steps would be really useful for non Z2JH users too! How do you feel about linking to this doc from other project docs?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+        state: state
+    JupyterHub:
+      authenticator_class: oauthenticator.generic.GenericOAuthenticator
+```
+
+#### Auth0
+
+Auth0 is a popular commercial provider of identity management. The
+JupyterHub helm chart does not include support for Auth0 by default. To
+use Auth0, `extraEnv` and `extraConfig` must be configured as follows:
```
manics, line 270: You've mentioned `extraEnv` and `extraConfig` but there's no sign of it below, maybe leftover from a copy and paste?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+    OAuthenticator:
+      client_id: your-client-id.apps.googleusercontent.com
+      client_secret: your-client-secret
+      oauth_callback_url: https://your-jupyterhub-domain/hub/oauth_callback
+    GoogleOAuthenticator:
+      hosted_domain:
+        - your-university.edu
+      login_service: Your university
+    JupyterHub:
+      authenticator_class: oauthenticator.GoogleOAuthenticator
```
manics, line 338: In other places you've used `oauthenticator.MODULE.XXXAuthenticator` instead of `oauthenticator.XXXAuthenticator`. Either works, but I think it'd be good to be consistent. Alternatively can an entrypoint be specified?

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,427 @@
...
+      lookup_dn_search_filter: ({login_attr}={login})
+      lookup_dn_search_password: secret
+      lookup_dn_search_user: ldap_search_user_technical_account
+      lookup_dn_user_dn_attribute: cn
+      server_address: ad.EXAMPLE.org
+      user_attribute: sAMAccountName
+      user_search_base: ou=people,dc=wikimedia,dc=org
+```
+
+#### Example Auth0 Configuration
```
manics, line 400: I'm a bit confused, there's an earlier ` #### Auth0` section that uses `oauthenticator.auth0.Auth0OAuthenticator` but this is different? Is to illustrate OpenID (and Auth0 just happens to be an example OpenID provider)?

manics (commented):
I made a first pass through the docs, I'll go through the code and schema changes later.

consideRatio:
@manics thank you for your review!

I want to clarify that I have not really changed the text or structure of the documentation about how to configure the specific authenticators, but I'll give it a pass given your review comments. The reason for it being listed as an addition is the rST -> markdown conversion I did to not need to work with rST.

doc/source/administrator/authentication.md
consideRatio:
```suggestion
Please see CyberInfrastructure Logon's [website](https://www.cilogon.org) for more information about what kind of identity is managed by CILogon.

```

doc/source/administrator/authentication.md
consideRatio: Ah, I removed extraEnv and perhaps also extraConfig without updating the text. I'll update the text to match.

doc/source/administrator/authentication.md
consideRatio: Sure why not. One answer would be to avoid locking in whats written exactly here by letting it be referenced by other projects.

I think i'd like to see some refactoring of this page in general, and preferably before referencing it since the structure may change a bit based on the refactoring.

As an example of refactoring, I think it could be nice to have a common section of the OAuthenticator configs, or link out to oauthenticator where we improve the docs of that etc. Not sure!

doc/source/administrator/authentication.md
consideRatio: @manics oh, I have just copied the jupyterhub_config.py files mapping. I assume this relates to the __init__.py file of the oauthenticator modules folder? I suggest we keep it short but not using `google` or similar as it hides the fact its related to the oauthenticator python package which I want us to highlight.

doc/source/administrator/authentication.md
consideRatio: I'm also confused, I decided to not care about this difference as part of this PR.

doc/source/administrator/authentication.md
consideRatio:
```suggestion
repository](https://oauthenticator.readthedocs.io/en/stable/) to generate
```

doc/source/administrator/authentication.md
consideRatio: https://github.com/jupyterhub/oauthenticator/blob/6f239bebecbb3fb0242de7f753ae1c93ed101340/oauthenticator/__init__.py#L1-L7

It seems that only a few are added there for backward compatebility. So, we can only make all longer. Let's make all longer.

consideRatio: Done.

doc/source/administrator/authentication.md
consideRatio: What do you think about 3ea9ded0?

doc/source/administrator/authentication.md
consideRatio: I did some refactoring. I think in general, it would be good if we let OAuthenticator maintain the docs on how to setup client id etc, as it is used by z2jh / tljh and vanilla JupyterHub's etc, while z2jh is just one distribution where its relevant.

doc/source/administrator/authentication.md
consideRatio: I went ahead and updated the entire authentication page where I located two examples under a GenericOAuthenticator section, where I took Auth0 and KeyCloak examples against OpenID connect.

consideRatio:
@manics I've done a full pass of the auth docs following the questions you raised.

doc/source/conf.py
choldgraf: The MyST parser repo or the Jupyter book repo both work!

doc/source/administrator/authentication.md
consideRatio: This is now resolved through https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1968/files#diff-d8d3ed25802824d15bf411f8e97416d8fc6f9247e821b0617f2b869dc584b99cR102-R109.

choldgraf:
I wonder if @yuvipanda would be particularly interested in this, since we're doing a lot of auth config in 2i2c and this would probably affect us, no?

jupyterhub/templates/_helpers-auth-rework.tpl
```diff
@@ -0,0 +1,223 @@
...
+
+    {{- /* UPDATE c dict authenticator_class */}}
+    {{- $_ := merge $c (dict "JupyterHub" (dict "authenticator_class" $class_long)) }}
+
+{{- /* Output a sensible error message */}}
+
+The JupyterHub Helm chart's auth config has been reworked and requires changes.
+
+The new way to configure authentication in chart version 0.11.0+ is printed
+below for your convinience. The values are not shown by default to ensure no
```
minrk, line 215:
```suggestion
below for your convenience. The values are not shown by default to ensure no
```

jupyterhub/values.yaml
```diff
@@ -22,6 +22,10 @@ imagePullSecrets: []
 # ConfigurableHTTPProxy speaks with the actual ConfigurableHTTPProxy server in
 # the proxy pod.
 hub:
+  config:
+    JupyterHub:
+      admin_access: true
+      authenticator_class: dummyauthenticator.DummyAuthenticator
```
minrk, line 28: shortcut works and is perhaps easier to understand:

```suggestion
      authenticator_class: dummy
```

We shouldn't be using dummyauthenticator anymore, as dummy ships with jupyterhub now.

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+Before configuring this, you should have [setup HTTPS](https).
+
+## Useful understanding
+
+### Authenticator classes
+
+JupyterHub are by itself unable to authenticate any kind of identity, but
+instead relies on _one_ chosen [_authenticator
+class_](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html)
+to do it. Several such classes are already made available through [installed
```
minrk, line 19:
```suggestion
JupyterHub by default ships with only one source of authentication:
PAM, the underlying unix authentication of the host system.
To use other sources of authentication, choose _one_ [_authenticator
class_](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html) to use.
Several such classes are already available in the hub image through [installed
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+
+### The configuration system
+
+We configure JupyterHub to use our chosen authenticator class and the
+authenticator class itself through this Helm chart's
+[`hub.config`](schema_hub.config) configuration.
+
+## General configuration
+
+As all authenticators classes derive from the `Authenticator` base class, they
```
minrk, line 36:
```suggestion
As all authenticator classes derive from the `Authenticator` base class, they
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+      allowed_users:
+        - user3
+        - user4
+    # ...
+    DummyAuthenticator:
+      password: a-shared-secret-password
+    JupyterHub:
+      authenticator_class: dummyauthenticator.DummyAuthenticator
+```
+
```
minrk, line 63: We can add some explanation to these configuration examples:

```suggestion

In the above configuration, we have configured three things:

1. JupyterHub is instructed to use the dummy authenticator to login (only appropriate for testing purposes),
2. anyone will be able to login with username `user1-4` and the password `a-shared-secret-password`
3. `user1` and `user2` will have admin permissions, while `user3` and `user4` will be regular users.

```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+### [auto_login](https://jupyterhub.readthedocs.io/en/latest/api/auth.html#jupyterhub.auth.Authenticator.auto_login)
+
+If you have configured authentication with GitHub for example, the page
+`/hub/login` will feature a single orange button that users are to press to
+login. If you want to bypass this screen and redirect users directly, you can
```
minrk, line 68:
```suggestion
login. If you want to bypass this screen and send users directly to GitHub login, you can
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+
+```yaml
+hub:
+  config:
+    OAuthenticator:
+      client_id: your-client-id
+      client_secret: your-client-secret
+      oauth_callback_url: https://your-jupyterhub-domain/hub/oauth_callback
+    JupyterHub:
+      authenticator_class: oauthenticator.github.GitHubOAuthenticator
```
minrk, line 143:
```suggestion
      authenticator_class: github
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+To restrict access to the members of one or more GitHub organizations, amend
+your previous configuration with these parts.
+
+```yaml
+hub:
+  config:
+    GitHubOAuthenticator:
+      allowed_organizations:
+        - my-github-organization
+    OAuthenticator:
```
minrk, line 157: While we *can* set every option on the highest possible class, it is not necessary. It's fine and probably easier to follow to only directly configure the class being used in many of these cases:

```suggestion
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+      client_id: your-client-id.apps.googleusercontent.com
+      client_secret: your-client-secret
+      oauth_callback_url: https://your-jupyterhub-domain/hub/oauth_callback
+    GoogleOAuthenticator:
+      hosted_domain:
+        - your-university.edu
+      login_service: Your university
+    JupyterHub:
+      authenticator_class: oauthenticator.google.GoogleOAuthenticator
```
minrk, line 210:
```suggestion
      authenticator_class: google
```

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+
+```yaml
+hub:
+  config:
+    OAuthenticator:
+      client_id: your-client-id
+      client_secret: your-client-secret
+      oauth_callback_url: https://your-jupyterhub-domain/hub/oauth_callback
+    JupyterHub:
+      authenticator_class: oauthenticator.cilogon.CILogonOAuthenticator
```
minrk, line 231:
```suggestion
      authenticator_class: cilogon
```

minrk (commented):
This is awesome work! I left some minor comments in the docs, but so happy for this to move forward.

jupyterhub/values.yaml
consideRatio: Ah! So I'll remove jupyterhub-dummyauthenticator from the requirements.txt, and we deprecate the repsitory?

minrk: Yes, I believe so. This change is still needed to avoid the current crash loop backoffs because you've removed the `dummyauthenticator` package but are still using it in config.

consideRatio: Resolved by 5b0a011

doc/source/administrator/authentication.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### CILogon
+
+Please see CyberInfrastructure Logon's [website](https://www.cilogon.org) for
+more information about what kind of identity is managed by CILogon.
+
+```yaml
+hub:
+  CILogonOAuthenticator:
+    CiOAuthenticator:
```
minrk, line 231: typo:

```suggestion
 config:
    CILogonOAuthenticator:
```

minrk (approved):

minrk:
Great work, @consideRatio!

consideRatio:
Thanks @minrk and @manics for your review efforts!!! :Heart: :heart: :heart:

meeseeksmachine:

bsikander:
Two cents from my side while upgrading to 0.11.1.

When I tried to upgrade, helm suggested the configuration changes as mentioned in the PR. So, I just copy pasted the output in my CLI and placed in my config. Hub kept of giving 500 internal server errors and after digging deep inside, i found that hub was not recognising my providing authenticator and its configuration and was using the dummy default authenticator. The root cause of the problem was that I had two `hub.*` tags in my yaml. One for `hub.extraConfig.* and other one that CLI suggested. When I manually merged the two seperate yaml sections into one like below, it started to recognize.

```
hub:
  config:
     ........
  extraConfig:
     ........
```

Apart from this, dynamically setting the configurations worked also
```
hub:
    extraConfig:
        authenticator:
            c.JupyterHub.authenticator_class = "oauthenticator.azuread.AzureAdOAuthenticator"
            c.JupyterHub.admin_access = True
            c.AzureAdOAuthenticator.client_id = ""
            c.AzureAdOAuthenticator.client_secret = ""
            c.AzureAdOAuthenticator.tenant_id = ""
            c.AzureAdOAuthenticator.oauth_callback_url = ""
            c.Authenticator.admin_users = {'usera', 'userb'}
```

PR #1948, opened 2020-12-17
ci: let dependabot update jupyterhub, replace JUPYTERHUB_VERSION with PIP_OVERRIDES

consideRatio:
I wanted to make dependabot also manage the jupyterhub version. To make that happen, I had to stop pinning it explicitly here and there.

Associated with this in the PR, is a removal of the JUPYTERHUB_VERSION --build-args. I added PIP_OVERRIDES instead as it would be more generally helpful.

## Action points
- [x] Improve inline docs of chartpress.yaml so that users understand how to learn more etc if they inspect it.
- [x] Help users make use of the new PIP_OVERRIDES argument by commenting in chartpress.yaml

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -1,3 +1,13 @@
+# This is the configuration for chartpress, a CLI for Helm chart management.
+#
+# chartpress is used to:
+# - Build images
+# - Update Chart.yaml (version) and values.yaml (image tags)
+# - Package and publish Helm charts to a GitHub based Helm chart repository
+#
+# For more information, see the projects README.md file:
+# https://github.com/jupyterhub/chartpress
+#
 charts:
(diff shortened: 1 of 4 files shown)


consideRatio:
I assigned this the 0.11.0 milestone as I'd like to use this opportunity to try if the automation works with a JupyterHub 1.3.0 bump which I want to be in the 0.11.0 release of the chart.

consideRatio:
It worked as it should, but I found that we have strict test to match the singleuser-sample version against the hub version, so we should also bump the singleuser-sample version preferably with dependabot. It probably needs to be two separate dependabot setups, but thats okay.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1959

PR #1949, opened 2020-12-17
ci: let dependabot update used GitHub action's versions

consideRatio:
It seems that dependabot can keep our workflows used GitHub actions' versions, so I figure why not!


.github/dependabot.yaml
```diff
@@ -1,15 +1,22 @@
...
+  # Maintain Python dependencies for the jupyterhub/k8s-hub image
+  - package-ecosystem: pip
+    directory: "/images/hub"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 4
+    versioning-strategy: lockfile-only
+    labels:
+      - maintenance
+      - dependencies
```
manics, line 13: This is overriding the default dependabot labels:
https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-dependency-updates#setting-custom-labels

Are we using them for a particular purpose, or could we remove this bit of the config and rely on the defaults?

consideRatio: Interesting to know! Hmm want github-activity to sort these PR quickly when generating a changelog, which based on current github-activity function means the maintenance label.

Id like a dedicated category for dependencies etc, but at this point its maintenance.

Can we add extra labels in a hardcoded way instead?

manics: Sure, that's fine 😀. Just wanted to check it's an intentional override.

.github/dependabot.yaml
```diff
@@ -1,15 +1,22 @@
...
+
+  # Maintain dependencies in our GitHub Workflows
+  - package-ecosystem: "github-actions"
+    directory: "/"  # This should be / rather than .github/workflows
+    schedule:
+      interval: daily
+    labels:
+      - maintenance
+      - dependencies
```
consideRatio, line 22:
```suggestion
      - dependencies

```

.github/dependabot.yaml
consideRatio: :) thanks for review @manics! Merge?

manics (approved):

PR #1950, opened 2020-12-19
Fix failure to block insecure metadata server IP

consideRatio:
#1805 softly deprecated a configuration, but the new configuration didn't work which made the user pods not startup with an initContainer blocking access to the metadata server which is known to be a security concern.


meeseeksmachine:

PR #1957, opened 2020-12-22
Adds a linebreak

arokem:
Bad things happen if you don't have this linebreak in place. In particular, it looks like helm is upgrading, but the values in your config file are not propagated and you get the wrong version onto your hub. Both of which can take hours to figure out (speaking from personal experience 😅)


consideRatio:
Thank you @arokem for fixing these things as you experience them! :heart: :tada: :sunflower: :1st_place_medal:

PR #1960, opened 2020-12-22
ci: configure dependabot for singleuser-sample

consideRatio:
Helps us update jupyterhub's version in this image as well, which should mirror the hub image version even though its often not important.


consideRatio:
This was supposed to be contained in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1961, but as I made a small change to a shared commit that I forgot to push here, this didn't automatically close.

PR #1967, opened 2020-12-25
ci: full_namespace_report function improved

consideRatio:
The full_namespace_report functioned defined in ci/common prints some useful output about the current k8s namespace using kubectl. This PR updates that function.

1. We now always run it, it only takes 1 second for a local k3s cluster.
2. We provide headers with nice colors, yellow / red depending on if the content to known to be associated with an error or not.
3. I refactored the function a bit, removing xargs as it was hard to only print headers when a condition was met without duplicating a lot of code.

full-namespace-report


consideRatio:
@manics btw, the beutify shell thing in pre-commit worked well to autoformat some indentation details just like black would!

PR #1978, opened 2021-01-09
ci: add another upgrade test and provide a template rendering diff

consideRatio:
Working on #1923 it became very relevant to inspect the changes in rendered helm templates. The [helm diff plugin](https://github.com/databus23/helm-diff) can provide a diff compared to an already installed helm chart, so in other words it is only sensible to add as part of the upgrade tests, diffing from the installed version and the version to be installed.

In this PR I add another upgrade tests, so instead of just going from the latest stable release, we try to upgrade also from the latest dev release. By upgrading from the latest dev release we can also smoothly inject a `helm diff` first that provides us with the specific details on what changed in the Helm template rendered output of this PR.

While this is a good start, I'm not sure about what happens if the PR isn't recently rebased. Will it then also show removal of added parts in the helm diff? I guess it comes down to how the checkout action works, does it checkout the exact state of the PR, or the state of the repo after PR would be merged?

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -80,7 +80,7 @@ jobs:
         # k3s-version: https://github.com/rancher/k3s/tags
         # k3s-channel: https://update.k3s.io/v1-release/channels
         include:
-          - k3s-channel: latest   # v1.20 as of 2020-12-31
+          - k3s-channel: v1.20
             test: install
           - k3s-channel: v1.19
             test: install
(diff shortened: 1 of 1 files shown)


manics:
The `pull_request` test will run against the PR merged into the upstream branch at the time the PR was last updated irrespective of whether the PR's base is up to date or not. This is why if there's an old PR it's often useful to close and re-open the PR to force an updated test run before merging in case upstream changes have invalidated the PR.

The `branch` status only looks at the branch. This is why it's useful to have both triggers in a workflow, as well as allowing tests to run without opening a PR you can also see if it passes on a branch but fails on the PR or vice-versa, indicating there's a mismatch somewhere.

PR #1993, opened 2021-01-16
Seed secrets (proxy.secretToken, etc) so they don't have to be manually generated

### LINKED ISSUE #1910: Generate passwords etc. into a dedicated k8s Secret that can be re-used
### Proposed change

Split hub-secret into two different secrets, one containing the ` values.yaml ` key and generated from the Helm config, the other would hold the `proxy.token` and  `cookie.secret` and can be set via ` existingSecret `.

Following commit 8fab275f1400f1f1c603ca06d621b25c2082a064 , the values.yaml are pulled into the secret. That’s an annoyance if you are trying to manage secrets encrypted in version control via Mozilla SOPS or Bitnami SealedSecrets. Ideally, you would like not to have to re-encrypt every single innocuous change to the ` values.yaml ` and instead keep it free of any sensitive content, so it can go into version control as clear text. 
That’s also a potential breaking change for users who upgrade the chart, since the `existingSecret` needs to be replaced. It is also annoying not be able to have direct access to your values in git.

Anyone using the existingSecret option / using similar Gitops tools to manage Secrets.

See proposed changes

Take [this file](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/secret.yaml):

- always create no matter what a `hub-secret` Secret  with lines 10-16 and if and only if `existingSecret ` doesn’t exist add lines 18-35 to that secret (move line 1 to line 17)
- switch the templates secretKeyRef to ` {{hub-secret}} ` and only use ` {{ .Values.hub.existingSecret | default "hub-secret" }} ` for `proxy.token` and other keys from lines 18-35
- switch the templates condition;  instead of using only ` hub.db.password ` add a `.Values.hub.db.existingSecretHoldsPassword`  Boolean flag. A line like https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/605cebc07dca36aa55cdf8e68c6bc811dcd02cda/jupyterhub/templates/hub/deployment.yaml#L171 becomes `    {{- if or .Values.hub.db.password   .Values.hub.db.existingSecretHoldsPassword}} ` 
- if you still want to keep the possibility to encrypt the values.yaml, add a new Boolean option `Values.hub.existingSecretHoldsEncryptedValues` and use that as the ‘ if condition’ on line 1. Users should then call their secret `hub-secret`, or better the Chart could use a user supplied variable allowing other secret name than `hub-secret`

consideRatio:
Wouldn't it be great if we could automatically seed the passwords like `proxy.secretToken`, `hub.cookieSecret`, and `hub.config.CryptKeeper.keys`, so that the user doesn't have to generate and set them manually?

The reason we haven't before is that we had no mechanism to re-use previous generated passwords, but that changed with Helm 3.1 that introduced the `lookup` function which we can make use of following #1994.

The strategy in this PR is to...

1. Use an explicitly provided password (backward compatebility!)
2. Use an existing password in k8s Secret (helm3's new `lookup` function!)
3. Use a new randomly generated password

With this, we close #1910.

## Action points
- [x] Update documentation to reflect we don't have to set proxy.secretToken any more
- [x] Update schema.yaml if there are such entries
- [x] CI: Provide some extra --set flags to the upgrade steps were we pass proxy.secretToken explicitly so it won't keep failing until next release.
- [x] Provide feedback to users in a NOTES.txt note about no longer needing to pass an explicit value for the secrets we now can autogenerate.

## Notes for changelog
Note that users should make the upgrade once and then remove what they previously set explicitly in order to avoid a key rotation.

## Bonus feature for another PR
- [ ] Support setting "rotate" or similar to generate a new secret.


manics:
I like it, I'll test it locally if I have time.

I guess we'll need to update the Z2JH docs too?

doc/source/administrator/debug.md
```diff
@@ -173,48 +173,3 @@ And now we see that we have a running user pod!
...
-   ```
-   proxy:
-      secretToken: '<output of `openssl rand -hex 32`>'
-   ```
-
-3. Redeploy the helm chart:
-
-   ```
-   helm upgrade --cleanup-on-fail jhub jupyterhub/jupyterhub -f config.yaml
-   ```
```
consideRatio, line 220: There is an open issue to improve the debugging section, so instead of finding an example with diagnosis etc that wasn't correct any more, I opted to delete the section.

doc/source/jupyterhub/installation.md
```diff
@@ -2,54 +2,45 @@
...
-   It is common practice for Helm and Kubernetes YAML files to indent using
-   two spaces.
-
-4. Save the `config.yaml` file. In the nano editor this is done by pressing **CTRL+X** or
-   **CMD+X** followed by a confirmation to save the changes.
-
-<!---
-Don't put an example here! People will just copy paste that & that's a
-security issue.
--->
```
consideRatio, line 52: There were significant differences here, so I ended up updating some leading paragraphs as well.

consideRatio (commented):
PR ready for review / merge!

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -372,6 +372,11 @@ def camelCaseify(s):
     c.Spawner.debug = True
 
 
+# load seeded secrets
+c.JupyterHub.proxy_auth_token = get_secret_value("JupyterHub.proxy_auth_token")
```
minrk, line 376: To clarify: the change here is that these secrets are no longer optional in the chart, and cannot possibly be left unspecified or null, is that correct? The previous logic allowed these to be unspecified, in which case the existing default behavior would occur, but explicitly setting them to null or empty values is not the same thing.

consideRatio: It is correct that we will now always provide a value to these three configuration options, if you search for "assert hack" you will find an assertion I make about that.

Any falsy value for proxy.secretToken, hub.cookieSecret, hub.config.CryptKeeper.keys would lead to automatic generation of a new passwords.

Is this indirectly influencing the behavior of JupyterHub?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -372,6 +372,11 @@ def camelCaseify(s):
...
+c.JupyterHub.cookie_secret = a2b_hex(get_secret_value("JupyterHub.cookie_secret"))
+c.CryptKeeper.keys = get_secret_value("CryptKeeper.keys").split(";")
+
 # load hub.config values
 for section, sub_cfg in get_config("hub.config", {}).items():
     c[section].update(sub_cfg)
```
consideRatio, line 382: @minrk when we write `c.JupyterHub.cookie_secret = "asdf"` in lines above, will that make loading of a `hub.config.JupyterHub.cookie_secret` entry here fail?

Is the practice of assigning directly to a traitlet problematic if you want to work with it later like this?

minrk: assigning directly to a traitlet is fine, but should only be done if there really is a value to assign (assigning None is **not** the same as not assigning). However, as you say, if hub.config.JupyterHub.cookie_secret is still defined, we are assigning to the same value, possibly with two different values (i.e. first, `a2b_hex(hex_cookie_secret)` and second the original (wrong) hex `hub.config.JupyterHub.cookie_secret`, in which case the above assignment will have no effect, as it will be replaced with the original value here the second time the same traitlet is assigned. If you want an to have higher priority, it must be done *last*, as it is assignment just like any other code:

```python
a = a2b_hex(value)
a = value
print(a) # will be `value`
```

jupyterhub/templates/NOTES.txt
```diff
@@ -20,6 +20,17 @@ Note that this is still an alpha release! If you have questions, feel free to
   3. File issues at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues
 
 
+{{- if or .Values.proxy.secretToken .Values.hub.config.JupyterHub.proxy_auth_token }}{{ println }}
+NOTE: proxy.secretToken or hub.config.JupyterHub.proxy_auth_token was explicitly set, this is no longer needed.
```
minrk, line 24: I don't think we need to discourage setting these values. While it is no longer necessary, that doesn't mean it is a bad practice (values are stable across full helm delete & redeploy, for instance).

jupyterhub/templates/hub/secret.yaml
```diff
@@ -16,7 +16,9 @@ data:
   values.yaml: {{ $values | toYaml | b64enc | quote }}
 
   # Used to mount CONFIGPROXY_AUTH_TOKEN on hub/proxy pods for mutual trust
-  proxy.token: {{ .Values.proxy.secretToken | required "Proxy token must be a 32 byte random string generated with `openssl rand -hex 32`!" | b64enc | quote }}
+  JupyterHub.proxy_auth_token: {{ include "jupyterhub.config.JupyterHub.proxy_auth_token" . | required "assert hack 1" | b64enc | quote }}
```
minrk, line 19: I assume these 'required' messages are leftover debug messages?

jupyterhub/templates/NOTES.txt
consideRatio: I see a value in it this encouragement because this is the only secret that we need outside of the hub container afaik which increases the complexity of solving #2009 for example.

If people remove it from their configs, the examples they copy paste in forums etc will not propegate the practice either.

Would you consider it a better practice to let this be managed explicitly by a user? I figure this is ephemeral configuration that doesn't require any state change other than a restart of the helm chart managed pods, and it better to not set it explicitly as a user.

yuvipanda:
This is great! 

Resetting / rotating these values is important when you believe they're compromised. Access to proxy secret basically gives you full access to every user's traffic, and access to cookie secret is terrible too. Rotating cookie secret logs everyone out (I think) and that's also very useful.

So as long as these *can* be set explicitly, I'm <3. But, *great* that it isn't required!

jupyterhub/templates/hub/secret.yaml
consideRatio: I did not work to provide a user readable formulation of the messages, but I like the idea of having a hard failure during template rendering if this assumption would break for some reason because that would then cause runtime errors hard to debug anyhow.

Do you suggest a change?

consideRatio:
Thanks @minrk and @yuvipanda for your feedback!!

> Resetting / rotating these values is important when you believe they're compromised.

While it is a bit of a feature creep of this PR, do you have a suggestion on what API to use in order to do this?

Passing `-`, `reset`, or `rotate` as a signal to reset and generate a new secret? A more complicated solution would be to pass `reset-<time>` or similar which would only reset if it was changed rather than over and over.

yuvipanda:
> While it is a bit of a feature creep of this PR, do you have a suggestion on what API to use in order to do this?

If these properties are set explicitly, they should take precedence over the autogenerated values. I think that's the best way to do this. So if we are only autogenerating values when the fields are unset, that's good enough :)

consideRatio:
> If these properties are set explicitly, they should take precedence over the autogenerated values. I think that's the best way to do this. So if we are only autogenerating values when the fields are unset, that's good enough :)

Okay yeah then we are good to go - that is what this PR currently does.

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Ah, this would actually also fail unless we ensure to transform hub.config.JupyterHub.cookie_secret is triggered on modifying the traitlet value I figure.

I pushed 25bf8a0 to resolve this and created https://github.com/jupyterhub/jupyterhub/pull/3343 to avoid future struggles with cookie_secret in general.

jupyterhub/templates/NOTES.txt
minrk: I don't think it's better to not set it, or necessarily better to set it. I think it's a big improvment to not *have* to set it, but that's not the same as the user shouldn't be *able* to manage it in configuration, or even discouraged.

I think for the most part, behavior should be completely unchanged *except* in the event of the value being unset, in which case it has this 'persistent' generated default value. Certainly setting it manually should have precedence.

jupyterhub/templates/hub/secret.yaml
minrk: If it's required, it should have a message saying it's required. It doesn't need to be perfect or verbose, but it shouldn't be an opaque "hack 1" which to me communicates that a TODO item wasn't finished. Even "this shouldn't happen" is more meaningful, as it communicates that it's asserting that a branch has been taken that shouldn't be. Something like "the chart should not allow this value to be undefined" might best communicate what's wrong.

consideRatio: I'll update the text! Note though that this is the output of the named template, which will render the explicitly set value, previous value, OR the new generated value.

jupyterhub/templates/NOTES.txt
consideRatio: I'll remove this note as part of this PR.

Note that explicitly set values have precedence.

consideRatio: Resolved by b8c60ef

jupyterhub/templates/hub/secret.yaml
consideRatio: Resolved by b94fec5

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: `get_secret_value("JupyterHub.proxy_auth_token")` returns the hub k8s Secret's key's value, which in turn is set by a named template, which in turn has priority logic to:

1. return any explicitly set value
2. return the previous k8s Secret's value
3. return autogenerated value

consideRatio:
I added a randHex function and made the generated length 64 hex characters, which aligns with documentation in JupyterHub for cookie secret and how we previously instructed users to generate their passwords.

That was the last itch I had with regards to this PR.

minrk:
Awesome! So great to have this resolved.

consideRatio:
Wieeee thanks for review / merge everyone! :tada: :heart:

jupyterhub/templates/hub/_helpers-passwords.tpl
```diff
@@ -0,0 +1,67 @@
...
+
+    In practice, it generates up to 100 randAlphaNum strings
+    that are filtered from non-hex characters and augmented
+    to the resulting string that is finally trimmed down.
+*/}}
+{{- define "jupyterhub.randHex" -}}
+    {{- $result := "" }}
+    {{- range $i := until 100 }}
+        {{- if lt (len $result) . }}
+            {{- $rand_list := randAlphaNum . | splitList "" -}}
```
yuvipanda, line 21: For those curious - randAlphaNum *is* cryptographically safe too - http://masterminds.github.io/sprig/strings.html.

yuvipanda:
Also, you can rotate these secrets with `k -n <namespace> delete secret hub` and then doing a redeploy. New keys!

PR #1994, opened 2021-01-16
Require Helm 3 to allow for enhancements

consideRatio:
Helm 2 reached end of life in November 2020, so for future releases of
this Helm chart, just like many other Helm charts have opted to do, we
drop support for Helm 2 and allow ourselves to depend on Helm 3. By
updating this Chart.yaml to apiVersion 2 we will require Helm 3.

By relying on Helm 3.1+, we can solve some long standing issues, for
example the need to manually specify proxy.secretToken.

### DIFF
--- a/jupyterhub/Chart.yaml
+++ b/jupyterhub/Chart.yaml
@@ -1,16 +1,14 @@
-# Chart.yaml v1 reference: https://v2.helm.sh/docs/developing_charts/#the-chart-yaml-file
 # Chart.yaml v2 reference: https://helm.sh/docs/topics/charts/#the-chartyaml-file
-apiVersion: v1
+apiVersion: v2
 name: jupyterhub
 version: 0.0.1-set.by.chartpress
 appVersion: 1.3.0
 description: Multi-user Jupyter installation
+keywords: [jupyter, jupyterhub, z2jh]
 home: https://z2jh.jupyter.org
-sources:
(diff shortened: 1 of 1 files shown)


manics:
I haven't checked the yaml changes in Chart.yaml, but I support dropping Helm 2 and K8s 1.16.

minrk:
I definitely support dropping helm 2. +/- 0 on k8s 1.16 since it may cause us a headache on the mybinder.org ovh cluster. That's already out of commission, though, and has been since November due to Docker rate limit and bugs on the OVH side that make it impossible to work around. Requiring a version upgrade to stay in the federation is not unreasonable, I think.

yuvipanda:
1.16 is still the 'stable' version listed on GKE - https://cloud.google.com/kubernetes-engine/docs/release-notes. I think we shouldn't drop support for it until 1.17 is the stable version. 

Definitely stop helm 2 support though.

consideRatio:
> I think we shouldn't drop support for it until 1.17 is the stable version.

The stable channel includes k8s 1.15 to k8s 1.17 where k8s 1.17 was introduced 8th Jan. I felt that it was important to wait for it to be available in the stable channel, but not to wait for it to be the default or a version that everyone automatically would get upgraded to.

I think it's likely they will bump to k8s 1.17 as the new default in GKE stable next week based on a basic manual extrapolation.

consideRatio:
I transformed the PR to not bump k8s requirements. Merge?

PR #1995, opened 2021-01-16
Remove deprecated user-scheduler config

consideRatio:
We have maintained two separate versions of how we configure the user-scheduler, one for k8s 1.16 and below, and one for k8s 1.17 and above. This is because the version of kube-scheduler required us to do so.

With #2005 merged, we can now drop the <=1.16 version of how we configure kube-scheduler to become our user-scheduler and only debug / focus on the new configuration.


consideRatio:
This LGTM after a month, I'll go for the merge.

Issue #1899, opened 2020-11-05
What to execute with pre-commit for a good developer experience?


manics:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/44a531531a1f6a4a45e86243508a54b183ef56bb/ci/common#L1-L2
has clear instructions to verify the script on https://www.shellcheck.net/ 
It currently shows lots of warnings 😀

Shall we keep it or get rid of it? If we're keeping it we could enforce it's use with pre-commit using https://github.com/shellcheck-py/shellcheck-py#as-a-pre-commit-hook

consideRatio:
I appreciate using it manually to catch things, so +1 for retaining the comment.

But, do we want to do more things with pre-commit that as a whole makes it worth introducing? Hmm...
- spellcheck sh/bash scripts
- helm lint helm chart
- black python scripts
- yamllint values.yaml file

I'd be open to introducing pre-commit, but I'm quite scarred from past installation failures :P I struggled so much to get it to work before, but now I have not had issues with it for a while.

minrk:
I think pre-commit is most useful for auto-formatting. I love it when it fixes issues for me, I get annoyed when it tells me something I need to fix (*especially* style) unless it's a very basic clear error (e.g. invalid syntax, missing import). So I'd be +1 for black, and I've been googling for a bash autoformatter and found https://pypi.org/project/beautysh/ but haven't tried it. The furthest I would probably go on the rest is syntax-checking real yaml files (not templates). If shellcheck has an auto-fix function, I'd probably be for it.

consideRatio:
> If shellcheck has an auto-fix function, I'd probably be for it.

I have used shellcheck a lot on my sh/bash scripts and know they have suggestions, but those suggestions are often not right, and a lot of warnings are not relevant - but, there are errors that so far always have been relevant. I'm +1 for using shellcheck to spot errors, but not to output information about warnings or suggestions due to the too often irrelevant warnings/errors.

manics:
Thanks for your feedback. I'll start by enabling some autoformatters (black, maybe prettier) in a PR, and leave the potentially more controversial style configuration for a follow-up.

consideRatio:
Thank you for your work on this @manics! :heart:

[...]

PR #1996, opened 2021-01-17
Autoformat bash scripts, yaml files, and markdown files with pre-commit

manics:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1899

For discussion:
- [x] shellcheck (first 3 commits): I think it should be useful, and I've configured it to ignore the current warnings from `ci/common`. It's  not automatically installed though by pre-commit (in contrast to the other autoformatters where running pre-commit will auto-download it for you), so I've put it in a separate configuration file and added a second step to `test_chart.yaml` to run it.

    ```yaml
          - uses: pre-commit/action@v2.0.0
            with:
              extra_args: --config .pre-commit-shellcheck.yaml
    ```
   It picked up one minor error- `ci/common` uses bash not sh syntax. Should we keep shellcheck, make it part of the default `pre-commit-config.yaml`, or drop it?
- [x] helmlint: should be fine
- [x] prettier (last three commits): I've excluded `jupyterhub/templates/` (not proper yaml files) and `*.md` (it makes the diff hard to review so probably best left for a separate commit after this is reviewed). There was minor one error, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/662f5e84012ad0e76494b991716e0c0cd567f5e9 . The other thing to note is that had to override it's default of double quotes to use single quotes instead, this is because chartpress uses single quotes when resetting `jupyterhub/values.yaml`. Should we go ahead and auto-format markdown files too?


consideRatio:
> The other thing to note is that had to override it's default of double quotes to use single quotes instead, this is because chartpress uses single quotes when resetting jupyterhub/values.yaml.

Arrrgh, I'd like to avoid having both double quote and single quote enforced in different contexts. I remember this issue with chartpress, doh! Perhaps we can coerce chartpress with a new behavior instead? I believe I have tried in the past unsuccessfully though.

> Should we go ahead and auto-format markdown files too?

Sure, they are prone to formatting errors. We are using MyST markdown, I wonder if there are some linting and or autoformatting or similarly for MyST specifically btw. I'll investigate.

consideRatio:
@manics chartpress will not disrupt us after https://github.com/jupyterhub/chartpress/pull/115 is merged and released. It will properly preserve no / single / double quotes for Chart.yaml / values.yaml then.

consideRatio:
## Requested action points
- [ ] Get https://github.com/jupyterhub/chartpress/pull/115 merged -> cut chartpress release
- [ ] Remove the prettier config so we use the default of double quotes
- [ ] Update prettier pre-commit hook to use a non-deprecated entry - see comment in https://github.com/prettier/prettier/blob/master/.pre-commit-hooks.yaml

## Bonus action points
- [x] Decide on mdformat / prettier for markdown.
  I'm since just now in favor of prettier as we use it for our dedicated yaml files etc.
- [ ] Make use of prettier for markdown also, it takes care of the yaml/json examples in our code blocks and doesn't cause problems even if we use MyST it seems.

consideRatio:
@manics is it okay if I --force push to this branch?
- I omitted f000c75 specifically and added commits on top of it
- Rebased on master
- Configured pre-commit for markdown, yaml double quotes, bumped non-deprecated version of prettier, bump shellcheck version.

What would remain then:
- [ ] Get jupyterhub/chartpress#115 merged -> cut chartpress release
- [ ] Actually run pre-commit and push those changes

manics:
Go ahead and push

consideRatio:
@manics this LGTM now, what do you think?

I did a visual inspection of the markdown and is generally happy about the kind of formatting changes there, as well as in the YAML now that we have double quotes which is the default. With prettier, we also automatically get indentation within the yaml syntax highlighted code blocks! We don't get python black formatting with in the python code blocks though.

manics (commented):
Seems fine to me! Thanks for finishing this off.

consideRatio:
:tada: thanks for starting it up!

consideRatio:
@manics it has been a great success to use prettier I think! :D I catch several syntax errors because it fails to autoformat invalid python code! :heart:

PR #2000, opened 2021-01-19
Specify prometheus.io/port for hub service

yuvipanda:
Without this, the hub metrics don't seem to be
scraped.

### DIFF
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -11,6 +11,9 @@ metadata:
     {{- if not (index .Values.hub.service.annotations "prometheus.io/path") }}
     prometheus.io/path: {{ .Values.hub.baseUrl }}hub/metrics
     {{- end }}
+    {{- if not (index .Values.hub.service.annotations "prometheus.io/port") }}
+    prometheus.io/port: "8081"
+    {{- end }}
     {{- with .Values.hub.service.annotations }}
     {{- . | toYaml | nindent 4 }}
     {{- end }}


consideRatio:
Hmmm, I'm quite confident the default behavior of prometheus in the past was to use the first / or only available port of the k8s service in the past.

It would be nice to track down if this has changed or similarly, because if it has there needs to be changes like this in many other charts I figure.

Do you know if this failure relates to a recent update of prometheus? Could this have to do with network policies rather than the port?

yuvipanda:
I spent a day or so poking around here, and am pretty sure it wasn't networkpolicies. I agree this used to work, but unfortunately it hasn't for a while.

Do you know other installations that are scraping hub metrics? That would be very helpful to check.

consideRatio:
@yuvipanda okay let's merge this to avoid issues, but in case it becomes relevant to isolate the reason for this in the future can you provide information of...
- What version of z2jh / prometheus helm chart was used when this fix was required?
- Was the hub metrics configured to be exposed with or without authentication?

For reference, the [neurohackademy deployment](https://github.com/neurohackademy/nh2020-jupyterhub) was functional without this, and the specs were:
- JupyterHub version: 0.9.0-n150.h1ce17ed
- Prometheus version: [11.6.0, appVersion: 2.19.0](https://github.com/prometheus-community/helm-charts/blame/329feabf35042c12e8a425ef2fac9a3b0de9fb3d/charts/prometheus/Chart.yaml)
- `c.JupyterHub.authenticate_prometheus = False`

yuvipanda:
Thanks for the merge, @consideRatio! I agree that being explicit is good anyway, regardless of default behavior. 

Unfortunately, this happened a while ago and I've made further changes to the infra since :( I'll report back if I run into stuff like this again, and make sure to note versions.

Thanks!

Issue #1962, opened 2020-12-22
ci: intermittent k8s 1.19 and 1.20 test failures


consideRatio:
I saw this issue in the CI system, it was showing up in a k8s 1.20 test that was just added. I'm not sure what is the cause yet, but it would be nice to print more info than the status code as a first step and then debug further.

```
    @pytest.fixture(scope="function")
    def jupyter_user(api_request):
        """
        A temporary unique JupyterHub user
        """
        username = "testuser-" + str(uuid.uuid4())
        r = api_request.post("/users/" + username)
        assert r.status_code == 201
        yield username
        r = api_request.delete("/users/" + username)
>       assert r.status_code == 204
E       assert 400 == 204
E         +400
E         -204
```

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1961/checks?check_run_id=1594410192

## Update

The failure is intermittent, about 50% of the time in our ci tests. I added some logging and such allowing the failure to be inspected [in this run](https://github.com/consideRatio/zero-to-jupyterhub-k8s/runs/1596599338?check_suite_focus=true).

The conclusion I draw is that there is some pod termination issues in k8s 1.20, and it sais...

```
    State:          Waiting
      Reason:       PodInitializing
    Last State:     Terminated
      Reason:       ContainerStatusUnknown
      Message:      The container could not be located when the pod was deleted.  The container used to be Running
      Exit Code:    137
      Started:      Mon, 01 Jan 0001 00:00:00 +0000
      Finished:     Mon, 01 Jan 0001 00:00:00 +0000
```

To me, this is very confusing, but googling for k8s 1.20 and this message, I ended up in a k8s 1.20 issue. For reference, we are using k3s 1.20.0+k3s2 currently.

https://github.com/kubernetes/kubernetes/issues/97288

consideRatio:
I think this will close with a future patch version of  k8s 1.20, see https://github.com/kubernetes/kubernetes/pull/97998.

consideRatio:
Yikes! Now it is even worse!!!!

At least k8s 1.19 and 1.20 is acting up now. I assume a patch release introduced the same issue in k8s 1.19 also.

dhirschfeld:
I've just started seeing this error too :(

consideRatio:
[...]

PR #2003, opened 2021-01-20
refactor: systematically prefer use of with in templates

consideRatio:
This PR transforms statements like this using `if`...

```yaml
        {{- if .Values.hub.annotations }}
        {{- .Values.hub.annotations | toYaml | trimSuffix "\n" | nindent 8 }}
        {{- end }}
```

Into statements like this using `with`...

```yaml
        {{- with .Values.hub.annotations }}
        {{- end }}
```

This is a practice that has become popular and common practice among Helm charts. Currently our chart use a mix of these, but not after this PR where we only use the `with` variation.

### Any actual changes?

No, As can be seen in [the rendered helm template diff](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2003/checks?check_run_id=1737818655), this didn't change anything. The test failure is intermittent k8s 1.20 stuff discussed in #1962.

### DIFF
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -9,11 +9,9 @@ metadata:
   name: {{ include "jupyterhub.ingress.fullname" . }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
-  {{- if .Values.ingress.annotations }}
+  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- range $key, $value := .Values.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
+    {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
   {{- end }}
 spec:
   rules:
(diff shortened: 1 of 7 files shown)


jupyterhub/templates/ingress.yaml
```diff
@@ -9,11 +9,9 @@ metadata:
   name: {{ include "jupyterhub.ingress.fullname" . }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
-  {{- if .Values.ingress.annotations }}
+  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- range $key, $value := .Values.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
+    {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
```
consideRatio, line 14: Actually, there could be a breaking changes here if someone has misused ingress.annotations having used a boolean value such as `true` instead of the string `"true"`.

If there is any such user, they would get an error from the k8s api-server about an invalid annotation.

manics: I think that's fine, if we're dropping helm 2 I think that should count as a breaking release anyway

PR #2005, opened 2021-01-21
Require k8s 1.17+ to reduce complexity

consideRatio:
By relying on k8s 1.17+, we can cut away a hard to maintain
user-scheduler implementation for k8s 1.16 and earlier that is different
from the implementation we use for k8s 1.17+. See #1995.

## Action points
- [x] Await merge criteria.
  @yuvipanda suggested holding off until GKE's stable release channel defaults to k8s 1.17, which I think will be in a week or two.
  > 1.16 is still the 'stable' version listed on GKE - https://cloud.google.com/kubernetes-engine/docs/release-notes. I think we shouldn't drop support for it until 1.17 is the stable version.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -77,8 +77,6 @@ jobs:
             test: install
           - k3s-channel: v1.17
             test: install
-          - k3s-channel: v1.16
-            test: install
 
           # We run two upgrade tests where we first install an already released
           # Helm chart version and then upgrades to the version we are now
--- a/jupyterhub/Chart.yaml
+++ b/jupyterhub/Chart.yaml
@@ -8,7 +8,7 @@ keywords: [jupyter, jupyterhub, z2jh]
 home: https://z2jh.jupyter.org
 sources: [https://github.com/jupyterhub/zero-to-jupyterhub-k8s]
 icon: https://jupyter.org/assets/hublogo.svg
-kubeVersion: ">=1.14.0-0"
+kubeVersion: ">=1.17.0-0"
 maintainers:
   # Since it is a requirement of Artifact Hub to have specific maintainers
   # listed, we have added some below, but in practice the entire JupyterHub team


manics:
The failing test error
```
Run . ./ci/common
  . ./ci/common
  UPGRADE_FROM_VERSION=$(curl -sS https://jupyterhub.github.io/helm-chart/info.json | jq -er '.jupyterhub.dev')
  echo "UPGRADE_FROM_VERSION=$UPGRADE_FROM_VERSION" >> $GITHUB_ENV
  
  echo ""
  echo "Installing already released jupyterhub version $UPGRADE_FROM_VERSION"
  
  # FIXME: We change the directory so jupyterhub the chart name won't be
  #        misunderstood as the local folder name.
  #
  #        https://github.com/helm/helm/issues/9244
  cd ci
  helm install jupyterhub --repo https://jupyterhub.github.io/helm-chart/ jupyterhub --values ../dev-config.yaml --version=$UPGRADE_FROM_VERSION
  
  echo ""
  echo "Installing Helm diff plugin while k8s resources are initializing"
  helm plugin install https://github.com/databus23/helm-diff
  shell: /usr/bin/bash -e {0}
  env:
    KUBECONFIG: /home/runner/.kube/config
    pythonLocation: /opt/hostedtoolcache/Python/3.8.6/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.6/x64/lib

Error: chart "jupyterhub" version "0.11.1-n058.h9dcf8f80" not found in https://jupyterhub.github.io/helm-chart/ repository
Installing already released jupyterhub version 0.11.1-n058.h9dcf8f80
Error: Process completed with exit code 1.
```
sounds suspiciously similar to https://github.com/jupyterhub/mybinder.org-deploy/issues/1706

consideRatio:
I think of two reasons for this happening:

1. We run helm repo update at one point in time, and then later we check the available versions which may now contain something newer. In this repo, we don't use `helm repo add` + `helm repo update`, but use the `--repo` flag during upgrades.
2. We run with `--repo` two times nad helm is using a cache in the second lookup. I doubt this though.
3. GitHub pages serving our Helm chart repo is using a cache of some kind. In z2jh we first check the version against a template generated .json file and then later use `helm upgrade --repo ...` against the index.yaml file. In case the index.yaml file is cached, we can get this issue.

I suspect this is caused by step 3.

yuvipanda:
Thanks for keeping this separate, @consideRatio! I'd like us to consider also making a point release before we bump to k8s 1.17. That would make life easier for folks, since they wouldn't have to move to helm 3 and a k8s version bump at the same time.

consideRatio:
@yuvipanda 1.17 is now the default in GKE and users of 1.16 are now being automatically upgraded to 1.17.

After considering this for a while I'd like to just get rid of old logic and reduce complexity sooner rather than later.

The key point is that it would give me peace of mind when users report issues with the user-scheduler to know that if they use a specific version a specific user-scheduler implementation rather than one depending on the k8s version. There have been some issues with it lately as reported in #2025 for example and some before it also.

Okay to merge this?

manics (commented):
```
$ git grep 1\\.16

CHANGELOG.md:than 1.16.
CHANGELOG.md:- Fixed a compatibility issue with Kubernetes 1.16+
CHANGELOG.md:- Kubernetes 1.16 compatibility achieved
jupyterhub/templates/scheduling/user-scheduler/deployment.yaml:          image: {{ .Values.scheduling.userScheduler.image.name }}:v1.16.15
jupyterhub/templates/scheduling/user-scheduler/rbac.yaml:  #       1.16 and in 1.17, but unchanged in 1.18 and 1.19.
```

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/35ca170e1709542f6f3bae8efab51093cc48beda/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml#L57
is removed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1995

Does the comment in
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/35ca170e1709542f6f3bae8efab51093cc48beda/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml#L21-L24
need to be updated or removed? As it is I don't understand the it's significance.

Regardless, :+1: for dropping 1.16

yuvipanda:
Thanks for patiently waiting, @consideRatio! This is good to go now!

Issue #1987, opened 2021-01-14
allow extending jupyterhub_config.py from files


scauglog:
### Proposed change

It would be nice if we can extend jupyterhub_config by adding file in a folder like https://tljh.jupyter.org/en/latest/topic/escape-hatch.html#extending-jupyterhub-config-py

### Who would use this feature?

This could be used to pass sensitive information like password or token used in jupyter_config.py without setting them during chart installation otherwise the entire values.yaml should be encrypted if we want to persist it in a git repository. It will allow to provide those sensitive value from secret already present in kubernetes that we can mount as folder.

consideRatio:
Hi @scauglog! I agree, this would be useful! I'll go ahead and close this in favor of the issue I just referenced this issue from (see link above this comment).

Thanks for chiming in!

Issue #1266, opened 2019-05-07
Smooth JupyterHub template customization


consideRatio:
I've tried figuring out a good pattern to configure the z2jh's charts templates and considered many options, perhaps we can document a decent option like below?

# Not good options
- If we would mount a persistent disk with the templates we would need to have an init container to make the jupyterhub user be able to access it.
- If we would use a script to copy content from a git repo or so, we would copy perhaps a bit much and need to provide credentials to do so if the git repo was private, as well as write such script etc to be started.
- If we would override the Docker image, we would need to manage building and keeping up to date with new builds etc of Docker images, not fun and slow.

# One limited but decent option
We configure JupyterHub to look for templates from k8s ConfigMap resources that we mount on the hub pod. If we change this ConfigMap, it will be reflected in the pod without restart, and JupyterHub will within a minute actually use the latest template.

Woops now that I've written this much, I realize that we won't be able to provide images if we want to, only text. Doh. UPDATE: we can provide images as well, but they must fit within ConfigMap's after base64 encoding, which limits the image size to about ~2MB at most I think.

### Chart configuration
```yaml
hub:
  extraConfig:
    templates: |
      ## Configure JupyterHub to look for the templates we inject with a ConfigMap
      c.JupyterHub.template_paths.insert(0, "/etc/jupyterhub/templates")
  extraVolumes:
    ## Make the hub pod be able to access a ConfigMap volume
    - name: hub-templates
      configMap:
        name: hub-templates
  extraVolumeMounts:
    ## Make the hub pod hub container mount that ConfigMap volume
    - mountPath: /etc/jupyterhub/templates
      name: hub-templates
```

### Manually added k8s ConfigMap

Create a `configmap.yaml` file with the following content and `kubectl apply -n <your-jhub-namespace> configmap.yaml`.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: hub-templates
data:
  # NOTE: helm will render content inside {{` here `}} as a
  # string no matter what curly braces etc. it will find there.
  # So if this yaml is first to be rendered by helm before parsed
  # as a k8s resource, you can wrap the entire section below with
  # that. Like below for example:
  # {{`
  # page.html |
  #   my template content with lots of curly braces that could conflict with
  #   helm rendering...
  # `}}
  spawn.html: |
    {% extends "page.html" %}
    {% if announcement_spawn %}
      {% set announcement = announcement_spawn %}
    {% endif %}

[...]

PR #2006, opened 2021-01-22
Allow extraFiles to be injected to hub / singleuser pods and automatically load config in /usr/local/etc/jupyterhub_config.d

### LINKED ISSUE #1990: Enable injection of files to the singleuser pods and the hub pod
## Motivation
This is meant to support a quite wide range of applications where mounting a file is relevant.

- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1987 and https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1580 suggests to mount files while also making hub.extraConfig load them. I suggest this is done addressed by mounting to a specific folder.
- The ability to mount a `jupyter_notebook_config.json` specifically to /etc/jupyter for user pods, see: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1691
- The ability to mount templates for jupyterhub, see: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1266

## Chart configuration idea
We expose configuration both under hub and singleuser, for example under `extraFiles`. We make it an dictionary just like `hub.extraConfig`, and let that dictionary's keys be arbitrary names and let the values be dictionaries with the file information.

```yaml
hub:
  extraFiles:
    kitten.png:
      mountPath: /usr/local/share/jupyterhub/static/kitten.png
      # A base64 encoding representing the binary content indirectly
      binaryData: |
        c29tZSBiaW5hcnkgY29udGVudCBiYXNlNjQgZW5jb2RlZAo=

singleuser:
  extraFiles:
    jupyter_notebook_config.json:
      mountPath: /etc/jupyter/jupyter_notebook_config.json
      # After this value is parsed from YAML (including JSON), we must render it back to a string.
      # At this point, we can opt to render toJson, toYaml, or toToml based on the file name extension.
      data:
        NotebookApp:
          allow_origin: '*'
          tornado_settings:
            trust_xheaders: true
          shutdown_no_activity_timeout: 600
    gitconfig:
      mountPath: /etc/gitconfig
      mode: 0644
      # A string representing the file content directly
      stringData: |
        [alias]
        # git l - an alias for a condensed view of commits
        # CREDITS: https://ma.ttias.be/pretty-git-log-in-one-line/
        l = log --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
```

## Implementation
The general idea is to add a volume and a volumeMount to the pod in question, but there are some deliberations along the way.

- __Use subPath?__
  Using subPath allow you to mount a specific key from a k8s configmap/secret into a specific file, rather than mounting all keys of a configmap/secret into a directory. By using subPath, the mount will no longer update on changes to the configmap/secret, but by not using subPath you must will erase all other content in the mount directory.
  - [x] Decision
    Rely entirely on usage of subPath, supporting dynamic updates for files injected into a folder is just a bit too much overhead and requires initContainers etc.
- __Use ConfigMaps or Secrets?__
  - ConfigMaps are easier to inspect by being written in plain text, while Secrets are base64 encoded.
  - Secret's base64 encoding will make content take 4/3 of the space.
  - Secret's base64 encoding will allow for binary file mounts such as images for hub templates.
  - Secret's is a safer bet from a permission standpoint.
  - [x] Decision
    Rely entirely on one or more k8s Secrets, don't complicate it by sometimes using a k8s ConfigMap.
- __How many Secrets should be used to pass the files?__
  - One single, one per file, or one default + a dynamic amount of additional ones? The benefit of having multiple is that more data can be fit into them. Using a single file, we are limited to about ~3 MB of data due to k8s / etcd constraints.
  - [x] Decision
    At this point, assume a single which won't shut the door from allowing a dynamic amount of additional ones later.
- __What k8s Secret resources for hub/singleuser should be used?__
  - Option A: a new dedicated k8s Secret containing both hub/singleuser extraFiles?
  - Option B: two new dedicated k8s Secrets containing hub and singleuser extraFiles repsectively?
  - Option C: re-use the hub k8s Secret to contain both hub/singleuser extraFiles?
  - Option D: re-use the hub k8s Secret and create a new dedicated k8s Secret for singleuser.
  - [x] Decision
    I'm going towards Option D for now.
- __Configuration option name?__
  - Is extraFiles good? I like that it includes files in the name, but perhaps being more explicit and saying extraFileMounts is worth considering, hinting it will influence volumeMounts (and volumes) of a pod spec also.
  - [x] Decision
    hub.extraFiles / singleuser.extraFiles it is!
  - Is mountPath / data / stringData / binaryData good?
  - [x] Decision
    I think so, let's go with it for now.
- __Implicit and/or explicit data format?__
  - When the user passes `data` as a data structure, we will in the end need to render it to a string. Helm provides us with `toYaml`, `toToml`, and `toJson` to do that. We could decide what data structure to render to using the filename extension, but perhaps there is an edge case where we need to support an explicit data format? If we need an explicit format, we could name it `dataFormat` where `yaml`, `json`, `toml` would be the valid options.
  - [x] Decision
    For now let's only support the implicit format and let Helm template rendering fail if a filename is passed without a supported filename extension when `data` is used.
- __Expose mode?__
  When declaring a SecretVolumeSource as a Volume you can set `defaultMode` such as 0644, and you can set an explicit map of files and `mode` for them. It can be critical for some files, such as SSH keys.
  - [x] Decision
    I'll try implement `mode` on a file to file basis and excluding `defaultMode` which I think will be fine as it is, and would cause complications with the decision to re-use the hub k8s Secret.
- __Explicit name notation?__
  As extraFiles declare files in many different folders, the having the name as a key could lead to name conflicts. Due to that, it would make sense to have a name field next to `mountPath` allowing multiple files of the same name to be mounted to different folders.
  - [x] Decision
    I'll implement it! For reference it's worth considering [hub/singleuser.extraEnv](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-extraenv) also provides this.
    
### On passing files
Its straightforward on how to pass dictionary like configuration, but, how to go about passing a large `.py` file or a binary image file?

__A standalone text file__
Just pass `--set-file singleuser.extraFiles.my_file_key.stringData=./my_file.py` to the helm upgrade command.

__A standalone binary file__
With binary files like an image, using `--set-file` wouldn't work as it doesn't know how to convert it to a string, which it must be. This is a challenge.

There is a PR open for `--include-file` etc which would make files accessible to the Helm template rendering through `.Files` which would save the day I think. It is being considered for Helm v3.6 at the earliest.

It would be possible for users to base64 encode their files and then use --set-file on the base64 encoded version and then passing `--set-file singleuser.extraFiles.my_image.binaryData=./my_image_base64_encoded.txt`.

__Decision__
For now, we don't create some suggested solution to mounting standalone binary files directly but instead await if for example https://github.com/helm/helm/pull/8841 becomes included in a future version of Helm.

## References
- Kubernetes API reference: [Secret](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#secret-v1-core), [ConfigMap](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#configmap-v1-core), [SecretVolumeSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#secretvolumesource-v1-core), [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#volume-v1-core), [VolumeMount](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#volumemount-v1-core).

consideRatio:
# Summary

- Added `hub.extraFiles` and `singleuser.extraFiles` allowing the chart users to inject files to the pods in specific locations with specific permissions (644, 400, etc).
- Added automatic loading of jupyterhub config files in `/etc/jupyterhub.d`.
- Closes #1990.

## Motivation
Injecting files to the hub / singleuser pods have a wide range of applications:

- To manage an external jupyterhub config file instead of embedding it to `hub.extraConfig` that. Related: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1987, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1580.
- To mount configuration files to the user environments, such as `jupyter_notebook_config.json`, `gitconfig`, etc. Related: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1691
- To mount templates for jupyterhub for easier customization of its user interface. Related: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1266.

## Usage documentation
See the [documentation preview](https://zero-to-jupyterhub--2006.org.readthedocs.build/en/2006/resources/reference.html#hub-extrafiles), or if its already merged the [latest documentation](https://z2jh.jupyter.org/en/latest/resources/reference.html#hub-extrafiles).

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -235,10 +235,10 @@ properties:
                 # filename extension that are required to be either .yaml, .yml,
                 # .json, or .toml.
                 #
-                # If your content is YAML, JSON, or TOML, it can be useful to
-                # use data over stringData, because then you can update parts of
-                # the files content from different files as compared to opting
-                # for one or another string.
+                # If your content is YAML, JSON, or TOML, it can make sense to
+                # use data to represent it over stringData as data can be merged
+                # instead of replaced if set partially from separate Helm
+                # configuration files.
                 #
                 # Both stringData and binaryData should be set to a string
                 # representing the content, where binaryData should be the
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -215,8 +215,9 @@ properties:
       extraFiles: &extraFiles
         type: string
         description: |
-          A dictionary with extra files that should be injected into the pod's
-          container on startup.
+          A dictionary with extra files to be injected into the pod's container
+          on startup. This can for example be used to inject: configuration
+          files, custom user interface templates, images, and more.
 
           ```yaml
           hub:
@@ -262,6 +263,8 @@ properties:
                 mode: <file system permissions>
           ```
 
+          **Using --set-file**
+
           To avoid embedding entire files in the Helm chart configuration, you
           can use the `--set-file` flag during `helm upgrade` to set the
           stringData or binaryData field.
@@ -288,6 +291,36 @@ properties:
               --set-file hub.extraFiles.my_image.binaryData=./my_image.png.b64 \
               --set-file hub.extraFiles.my_config.stringData=./my_jupyterhub_config.py
           ```
+
+          **Common uses**
+
+          1. **JupyterHub template customization**
+
+             You can replace the default JupyterHub user interface templates in
+             the hub pod by injecting new ones to
+             `/usr/local/share/jupyterhub/templates`. These can in turn
+             reference custom images injected to
+             `/usr/local/share/jupyterhub/static`.
+
+          1. **JupyterHub standalone file config**
+
+             Instead of embedding JupyterHub python configuration as a string
+             within a YAML file through
+             [`hub.extraConfig`](schema_hub.extraConfig), you can inject a
+             standalone .py file into `/etc/jupyterhub.d` that is automatically
+             loaded.
+
+          1. **Flexible configuration**
+
+             By injecting files, you don't have to embed them in a docker image
+             that you have to rebuild.
+
+             If your configuration file is a YAML/JSON/TOML file, you can also
+             use `data` instead of `stringData` which allow you to set various
+             configuration in separate Helm config files. This can be useful to
+             help dependent charts override only some configuration part of the
+             file, or to allow for the configuration be set through multiple
+             Helm configuration files.
       baseUrl:
         type: string
         description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -272,9 +272,14 @@ properties:
           ```yaml
           hub:
             extraFiles:
+              # Explicitly set the name to avoid having the file key include a .
+              # as it can hamper our ability to use --set-file.
               my_image:
                 name: my_image.png
                 mountPath: /usr/local/share/jupyterhub/static/
+
+              # Files in /etc/jupyterhub.d are automatically loaded just like
+              # hub.extraConfig entries.
               my_config:
                 name: my_jupyterhub_config.py
                 mountPath: /etc/jupyterhub.d
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -264,6 +264,32 @@ properties:
                       - 1
                       - 2
           ```
+
+          To avoid embedding entire files in the Helm chart configuration, you
+          can use the `--set-file` flag during `helm upgrade` to set the
+          stringData or binaryData field.
+
+          ```yaml
+          hub:
+            extraFiles:
+              my_image:
+                name: my_image.png
+                mountPath: /usr/local/share/jupyterhub/static/
+              my_config:
+                name: my_jupyterhub_config.py
+                mountPath: /etc/jupyterhub.d
+          ```
+
+          ```bash
+          # --set-file expects a text based file, so you need to base64 encode
+          # it manually first. Note that -w0 ensures no line breaks are
+          # introduced.
+          base64 -w0 my_image.png > my_image.png.b64
+
+          helm upgrade <...> \
+              --set-file hub.extraFiles.my_image.binaryData=./my_image.png.b64
+              --set-file hub.extraFiles.my_config.stringData=./my_jupyterhub_config.py
+          ```
       baseUrl:
         type: string
         description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -275,11 +275,11 @@ properties:
               my_image:
                 mountPath: /usr/local/share/jupyterhub/static/my_image.png
 
-              # Files in /etc/jupyterhub.d are automatically loaded in
-              # alphabetical order of the final file name when JupyterHub
-              # starts.
+              # Files in /usr/local/etc/jupyterhub/jupyterhub_config.d are
+              # automatically loaded in alphabetical order of the final file
+              # name when JupyterHub starts.
               my_config:
-                mountPath: /etc/jupyterhub.d/my_jupyterhub_config.py
+                mountPath: /usr/local/etc/jupyterhub/jupyterhub_config.d/my_jupyterhub_config.py
           ```
 
           ```bash
@@ -307,8 +307,9 @@ properties:
              Instead of embedding JupyterHub python configuration as a string
              within a YAML file through
              [`hub.extraConfig`](schema_hub.extraConfig), you can inject a
-             standalone .py file into `/etc/jupyterhub.d` that is automatically
-             loaded.
+             standalone .py file into
+             `/usr/local/etc/jupyterhub/jupyterhub_config.d` that is
+             automatically loaded.
 
           1. **Flexible configuration**
 
@@ -371,10 +372,10 @@ properties:
 
           ```{warning}
           By replacing the entire configuration file, which is mounted to
-          `/etc/jupyterhub/jupyterhub_config.py` by the Helm chart, instead of
-          appending to it with `hub.extraConfig`, you expose your deployment for
-          issues stemming from getting out of sync with the Helm chart's config
-          file.
+          `/usr/local/etc/jupyterhub/jupyterhub_config.py` by the Helm chart,
+          instead of appending to it with `hub.extraConfig`, you expose your
+          deployment for issues stemming from getting out of sync with the Helm
+          chart's config file.
 
           These kind of issues will be significantly harder to debug and
           diagnose, and can due to this could cause a lot of time expenditure
@@ -396,7 +397,7 @@ properties:
             args:
               - "jupyterhub"
               - "--config"
-              - "/etc/jupyterhub/jupyterhub_config.py"
+              - "/usr/local/etc/jupyterhub/jupyterhub_config.py"
               - "--debug"
               - "--upgrade-db"
           ```
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -221,26 +221,14 @@ properties:
           ```yaml
           hub:
             extraFiles:
-              <implicit file name>:
-                # name is an optional way to override the implicit file name
-                # that must be unique.
-                name: <explicit file name>
-
-                # mode is by default 0644 and you can optionally override it
-                # either by octal notation (example: 0400) or decimal notation
-                # (example: 256).
-                mode: <file system permissions>
-
-                # mountDir is required and must be the directory you want to
-                # mount the file.
-                mountDir: <directory path>
+              # The file key is just a reference that doesn't influence the
+              # actual file name.
+              <file key>:
+                # mountPath is required and must be the absolute file path.
+                mountPath: <full file path>
 
                 # Choose one out of the three ways to represent the actual file
                 # content: data, stringData, or binaryData.
-                #
-                # Both stringData and binaryData should be set to a string
-                # representing the content, where binaryData should be the
-                # base64 encoding of the actual file content.
                 #
                 # data should be set to a mapping (dictionary). It will in the
                 # end be rendered to either YAML, JSON, or TOML based on the
@@ -252,9 +240,10 @@ properties:
                 # the files content from different files as compared to opting
                 # for one or another string.
                 #
-                binaryData: aGVsbG8gd29ybGQhCg==
-                stringData: |
-                  hello world!
+                # Both stringData and binaryData should be set to a string
+                # representing the content, where binaryData should be the
+                # base64 encoding of the actual file content.
+                #
                 data:
                   config:
                     map:
@@ -263,6 +252,14 @@ properties:
                     list:
                       - 1
                       - 2
+                stringData: |
+                  hello world!
+                binaryData: aGVsbG8gd29ybGQhCg==
+
+                # mode is by default 0644 and you can optionally override it
+                # either by octal notation (example: 0400) or decimal notation
+                # (example: 256).
+                mode: <file system permissions>
           ```
 
           To avoid embedding entire files in the Helm chart configuration, you
@@ -272,18 +269,14 @@ properties:
           ```yaml
           hub:
             extraFiles:
-              # Explicitly set the name to avoid having the file key include a .
-              # as it can hamper our ability to use --set-file.
               my_image:
-                name: my_image.png
-                mountDir: /usr/local/share/jupyterhub/static/
+                mountPath: /usr/local/share/jupyterhub/static/my_image.png
 
               # Files in /etc/jupyterhub.d are automatically loaded in
               # alphabetical order of the final file name when JupyterHub
               # starts.
               my_config:
-                name: my_jupyterhub_config.py
-                mountDir: /etc/jupyterhub.d
+                mountPath: /etc/jupyterhub.d/my_jupyterhub_config.py
           ```
 
           ```bash
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -321,6 +321,25 @@ properties:
              help dependent charts override only some configuration part of the
              file, or to allow for the configuration be set through multiple
              Helm configuration files.
+
+          **Limitations**
+
+          1. File size
+
+             The files in `hub.extraFiles` and `singleuser.extraFiles` are
+             respectively stored in their own k8s Secret resource. As k8s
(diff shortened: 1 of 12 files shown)


consideRatio:
## Feature planning
The general idea is to add a volume and a volumeMount to the pod in question, but there are some deliberations along the way.

- __Use subPath?__
  Using subPath allow you to mount a specific key from a k8s configmap/secret into a specific file, rather than mounting all keys of a configmap/secret into a directory. By using subPath, the mount will no longer update on changes to the configmap/secret, but by not using subPath you must will erase all other content in the mount directory.
  - [x] Decision
    Rely entirely on usage of subPath, supporting dynamic updates for files injected into a folder is just a bit too much overhead and requires initContainers etc.
- __Use ConfigMaps or Secrets?__
  - ConfigMaps are easier to inspect by being written in plain text, while Secrets are base64 encoded.
  - Secret's base64 encoding will make content take 4/3 of the space.
  - Secret's base64 encoding will allow for binary file mounts such as images for hub templates.
  - Secret's is a safer bet from a permission standpoint.
  - [x] Decision
    Rely entirely on one or more k8s Secrets, don't complicate it by sometimes using a k8s ConfigMap.
- __How many Secrets should be used to pass the files?__
  - One single, one per file, or one default + a dynamic amount of additional ones? The benefit of having multiple is that more data can be fit into them. Using a single file, we are limited to about ~3 MB of data due to k8s / etcd constraints.
  - [x] Decision
    At this point, assume a single which won't shut the door from allowing a dynamic amount of additional ones later.
- __What k8s Secret resources for hub/singleuser should be used?__
  - Option A: a new dedicated k8s Secret containing both hub/singleuser extraFiles?
  - Option B: two new dedicated k8s Secrets containing hub and singleuser extraFiles repsectively?
  - Option C: re-use the hub k8s Secret to contain both hub/singleuser extraFiles?
  - Option D: re-use the hub k8s Secret and create a new dedicated k8s Secret for singleuser.
  - [x] Decision
    I'm going towards Option D for now.
- __Configuration option name?__
  - Is extraFiles good? I like that it includes files in the name, but perhaps being more explicit and saying extraFileMounts is worth considering, hinting it will influence volumeMounts (and volumes) of a pod spec also.
  - [x] Decision
    hub.extraFiles / singleuser.extraFiles it is!
  - Is mountPath / data / stringData / binaryData good?
  - [x] Decision
    I think so, let's go with it for now.
- __Implicit and/or explicit data format?__
  - When the user passes `data` as a data structure, we will in the end need to render it to a string. Helm provides us with `toYaml`, `toToml`, and `toJson` to do that. We could decide what data structure to render to using the filename extension, but perhaps there is an edge case where we need to support an explicit data format? If we need an explicit format, we could name it `dataFormat` where `yaml`, `json`, `toml` would be the valid options.
  - [x] Decision
    For now let's only support the implicit format and let Helm template rendering fail if a filename is passed without a supported filename extension when `data` is used.
- __Expose mode?__
  When declaring a SecretVolumeSource as a Volume you can set `defaultMode` such as 0644, and you can set an explicit map of files and `mode` for them. It can be critical for some files, such as SSH keys.
  - [x] Decision
    I'll try implement `mode` on a file to file basis and excluding `defaultMode` which I think will be fine as it is, and would cause complications with the decision to re-use the hub k8s Secret.
- __Explicit name notation?__
  As extraFiles declare files in many different folders, the having the name as a key could lead to name conflicts. Due to that, it would make sense to have a name field next to `mountPath` allowing multiple files of the same name to be mounted to different folders.
  - [x] Decision
    I'll implement it! For reference it's worth considering [hub/singleuser.extraEnv](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-extraenv) also provides this.

consideRatio:
## Implementation work items
- [x] Define a named template named `jupyterhub.extraFiles.data` to render a passed context of hub.extraFiles or singleuser.extraFiles `binaryData` items to be rendered within a k8s Secret's data field.
- [x] Define a named template named `jupyterhub.extraFiles.stringData` to render a passed context of hub.extraFiles or singleuser.extraFiles `data` and `stringData` items to be rendered within a k8s Secret's data field.
- [x] Conditionally create a new dedicated k8s Secret for singleuser based on singleuser.extraFiles is empty or not
- [x] Populate hub / singleuser k8s Secret with associated extraFiles entries
- [x] Update hub Deployment's volumes to include all extraFiles with their respective `mode`
- [x] Update hub Deployment's volumeMounts to include all extraFiles with their respective `mountPath`
- [x] Update c.KubeSpawner.volumes to include all extraFiles with their respective `mode`
- [x] Update c.KubeSpawner.volume_mounts to include all extraFiles with their respective `mountPath`
- [x] Define a set of extraFiles for lint-and-validate-config.yaml and dev-config.yaml
  - [x] Use duplicated entries in lint-and-validate-config.yaml to verify template rendering can handle it
  - [x] Prepare for testing of modes, data, stringData, binaryData, mountPath
- [x] Support explicit name notation to support having multiple files with the same filename in different folders.
- [x] Let a CI test verify files are made available for the hub pod and have the right `mode`
- [x] Let a CI test verify files are made available for the singleuser pod and have the right `mode`
- [x] Update schema.yaml to describe this configuration option
- [x] Add basic validation to enforce not multiple data fields are provided
- [ ] Suggest how to pass external files

jupyterhub/schema.yaml
```diff
@@ -212,6 +212,89 @@ properties:
...
+                name: <explicit file name>
+
+                # mode is by default 0644 and you can optionally override it
+                # either by octal notation (example: 0400) or decimal notation
+                # (example: 256).
+                mode: <file system permissions>
+
+                # mountDir is required and must be the directory you want to
+                # mount the file.
+                mountDir: <directory path>
```
manics, line 236: Since `mountDir` is mandatory have you thought about also requiring the name in the same property (`mountPath`) and always ignoring the key, so there's a bit less logic to deal with?

jupyterhub/schema.yaml
```diff
@@ -212,6 +212,89 @@ properties:
...
+
+                # Choose one out of the three ways to represent the actual file
+                # content: data, stringData, or binaryData.
+                #
+                # Both stringData and binaryData should be set to a string
+                # representing the content, where binaryData should be the
+                # base64 encoding of the actual file content.
```
manics, line 243:
```suggestion
                # stringData must be a string representing the content.
                # binaryData must be the file content encoded as base64.
```

Do we need to worry about character encodings? Or it it assumed to be UTF8?

jupyterhub/schema.yaml
```diff
@@ -212,6 +212,89 @@ properties:
...
+                # either by octal notation (example: 0400) or decimal notation
+                # (example: 256).
+                mode: <file system permissions>
+
+                # mountDir is required and must be the directory you want to
+                # mount the file.
+                mountDir: <directory path>
+
+                # Choose one out of the three ways to represent the actual file
+                # content: data, stringData, or binaryData.
```
manics, line 239:
```suggestion
                # content: stringData, binaryData or data.
```
I think it helps to keep the ordering consistent with the following explanations.

consideRatio: I resolved this by updating the order below instead, I wanted to highlight the benefits of `data` so I let it be the first thing described.

jupyterhub/schema.yaml
consideRatio: Good point! I removed the explicit `name` configuration entirely, removed `mountDir` and added `mountPath` as a required config that should include the file name as well.

jupyterhub/schema.yaml
consideRatio: Hmm... Not sure about this in general and doesn't clearly see the steps where there could arise encoding issues.

- Can you pass different encodings of string Helm through values.yaml?
  ?
- Does YAML have some defaults in general?
  It seems UTF-8 is favored by default: https://stackoverflow.com/a/648335
  
Hmmm unless further advised towards a concrete action I'll drop the ball on this consideration. I think we'll be fine.

consideRatio:
## Review points
- [ ] Does /etc/jupyterhub.d/ make sense for us to put extra configs in that jupyterhub automatically loads?
- [ ] The default location for templates and static files are under `/usr/local/share/jupyterhub/`. Perhaps we should provide an example where we override a template to reference an image we mount in static using extraFiles?
  image
- [ ] Does this PR overall LGTY?

choldgraf:
I think that the documentation of this field looks good to me, though I am not 100% sure why this would be useful I'm assuming it'd be obvious to someone trying to deploy their JupyterHub. Is the main reason for this to add extra configuration within the hub that one might want for other purposes? If there is a quick one-liner to explain when it is useful, perhaps that could be added to the top of the section.

yuvipanda:
> Does /etc/jupyterhub.d/ make sense for us to put extra configs in that jupyterhub automatically loads?

This should be next to wherever `jupyterhub_config.py` is, and should be called `jupyterhub_config.d`. This matches what most debian packages do (I think), and is also what we do in [TLJH](https://tljh.jupyter.org/en/latest/topic/escape-hatch.html#extending-jupyterhub-config-py).

Alternatively, we could move `jupyterhub_config.py` to `/usr/local/etc`, but IMO it's not worth that. Also note that this should be in `/usr/local/etc` and not `/etc`, since `/etc` should be reserved for system (apt) packages.

An example on overriding templates would be wonderful!

consideRatio:
@yuvipanda thank you for your input!! We currently mount it in /etc/jupyterhub following https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1407.

```
              subPath: jupyterhub_config.py
              name: config
```

If done from scratch, you would mount files something like...

```
/usr/local/etc/jupyterhub/jupyterhub_config.py
/usr/local/etc/jupyterhub/jupyterhub_config.d/my_extra_config.py
```

Is that correct?

---

I think this is won't be breaking actually. Only those that use a custom command to start jupyterhub and have hardcoded the config location would need to make a change afaik. I suggest we update to the sensible structure. What do you think?

yuvipanda:
> I suggest we update to the sensible structure. What do you think?

I agree! Let's do it :)

minrk:
> though I am not 100% sure why this would be useful I'm assuming it'd be obvious to someone trying to deploy their JupyterHub

The key point is mentioned in this bit in the docs:

> To avoid embedding entire files in the Helm chart configuration

It eliminates the need to do things like inline python/js/html/etc. in yaml config files like we do [in binderhub](https://github.com/jupyterhub/binderhub/blob/81660eb84a68e49307bc3f379e0c52ca25b8dc1c/helm-chart/binderhub/values.yaml#L67-L123), which loses syntax highlighting, editor support, linters, etc. Instead, you can write regular files and get all your familiar tooling, and then use this config to mount those files into the container. A big help when you have nontrivial things in there!

jupyterhub/templates/singleuser/secret.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.singleuser.extraFiles }}
+kind: Secret
+apiVersion: v1
+metadata:
+  name: {{ include "jupyterhub.singleuser.fullname" . }}
+  labels:
+    {{- include "jupyterhub.labels" . | nindent 4 }}
+type: Opaque
+{{- with include "jupyterhub.extraFiles.data" .Values.singleuser.extraFiles }}
```
minrk, line 9: what's the max size for a secret? Is there a reasonable risk of exceeding the max secret size with this by putting all mounted files in one secret?

consideRatio: Yes, that would be worth a mention!

I'm not sure all k8s clusters will have the same limitations which relates to the k8s Secret storage which probably is etcd generally, but 3 MB is the upper bound I've experienced on GKE.

Also, base64 encoding leads to an increase of 33%. This is probably only a limitation worth mentioning for binaryData.

## Action point
- [ ] Create a note about the file size limitations (hub / singleuser get one k8s secret to store the files in base64 encoding, and they can maximally be ~3 MB in total)

consideRatio:
## Action points
- [ ] Mention the file size limitations in schema.yaml: hub / singleuser get one k8s secret to store the files in base64 encoding, and they can maximally be ~3 MB in total
- [ ] Relocate /etc/jupyterhub stuff like described in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2006#issuecomment-768130273.
  Note to myself: while doing this, verify that it seems sensible given the feedback I remember receiving about where to locate certain files so it isn't tied to a linux distribution. I think this was about python files and z2jh though and due to that not applicable.
- [ ] Rebase away the avoid merge conflict.

consideRatio:
## Action point
- [ ] Something is wrong it seems. `helm diff` between the latest dev install upgraded to the local install make it seem like keys were generated twice.

jupyterhub/templates/singleuser/secret.yaml
consideRatio: I've added a docs about this under a Limitation section.

minrk:
This is great! I'm not sure the FHS doc supports switching from /etc/jupyterhub to /usr/local/etc/jupyterhub in a container/k8s/helm context (are we the "system administrators" in question, or the host system providers? I'd argue we are closer to the host system providers, based on the discussion of which files change during software updates).

This is the relevant FHS statement on /usr/local:

> The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated.

Translating that to our context, I think that means we should be guaranteeing to our users that we never modify files in `/usr/local` across helm upgrades. the 'system administrator' vs 'host system provider' is a fuzzy distinction in container image providers.

Since the work is already done, I don't object to it, but based on my reading of FHS, I think it was probably more appropriate and less surprising to use `/etc/` than `/usr/local/etc`, though `/etc/opt` would be the most appropriate for us as an 'add-on' (with jupyterhub in /opt) per 3.7.4 / 3.13. This is getting a little too close to bikeshedding, so I'm happy to merge as-is as this point.

Great work @consideRatio!

PR #2016, opened 2021-01-27
Followup fixes to seed secrets PR (#1993)

consideRatio:
I'm debugging an issue with the seed secrets PR which seemingly misbehaves a bit.

## Action points
- [x] The randHex function gives never gives me a leading hex character with a-f, only 0-9. Whyyyyyyyyyyyyyy?
  Unless my algorithm is implemented wrong, it should mean that randAlphaNum first a-f letter is always showing up after its first 0-9 digit for some reason.
  Answer: it turns out `randAlphaNum 1` in helm/sprig always starts with a number. See 9f0889c.
- [x] `helm diff` in the CI system under the upgrade dev job make it seem like passwords were generated both during the installation of the latest dev release and the local chart upgrade instead of re-using the existing value from the k8s Secret. Why?
  Answer: The logic failed as I forgot to inspect the `data` value. See f243d3c
- [x] We also need to normalize the fetched value to a non-base64 encoded string. a6d8e32

### DIFF
--- a/jupyterhub/templates/hub/_helpers-passwords.tpl
+++ b/jupyterhub/templates/hub/_helpers-passwords.tpl
@@ -36,7 +36,7 @@
         {{- if hasKey $k8s_state.data "JupyterHub.proxy_auth_token" }}
             {{- index $k8s_state.data "JupyterHub.proxy_auth_token" | b64dec }}
         {{- else }}
-            {{- include "jupyterhub.randHex" 64 }}
+            {{- randAlphaNum 64 }}
         {{- end }}
     {{- end }}
 {{- end }}
--- a/jupyterhub/templates/hub/_helpers-passwords.tpl
+++ b/jupyterhub/templates/hub/_helpers-passwords.tpl
@@ -37,7 +37,7 @@
         {{- .Values.proxy.secretToken }}
     {{- else }}
         {{- $k8s_state := lookup "v1" "Secret" .Release.Namespace (include "jupyterhub.hub-secret.fullname" .) | default dict }}
-        {{- if and $k8s_state (hasKey $k8s_state "JupyterHub.proxy_auth_token") }}
+        {{- if hasKey $k8s_state "JupyterHub.proxy_auth_token" }}
             {{- index $k8s_state "JupyterHub.proxy_auth_token" }}
         {{- else }}
             {{- include "jupyterhub.randHex" 64 }}
@@ -50,7 +50,7 @@
         {{- .Values.hub.cookieSecret }}
     {{- else }}
         {{- $k8s_state := lookup "v1" "Secret" .Release.Namespace (include "jupyterhub.hub-secret.fullname" .) | default dict }}
-        {{- if and $k8s_state (hasKey $k8s_state "JupyterHub.cookie_secret") }}
+        {{- if hasKey $k8s_state "JupyterHub.cookie_secret" }}
             {{- index $k8s_state "JupyterHub.cookie_secret" }}
         {{- else }}
             {{- include "jupyterhub.randHex" 64 }}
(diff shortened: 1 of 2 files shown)


jupyterhub/templates/hub/_helpers-passwords.tpl
```diff
@@ -5,35 +5,43 @@
...
-    to the resulting string that is finally trimmed down.
+    to the resulting string that is finally trimmed down. We
+    do it multiple times as on average we filter out 4/15 of
+    the characters.
 */}}
 {{- define "jupyterhub.randHex" -}}
     {{- $result := "" }}
-    {{- range $i := until 100 }}
-        {{- if lt (len $result) . }}
-            {{- $rand_list := randAlphaNum . | splitList "" -}}
```
minrk, line 21: Maybe we should be generating random numbers and using `printf "%x"` to get hex chars instead. Something like this:

```
{{- $result := "" }}
{{- range $i := until . }}
{{- $randHexChar := mod (randNumeric 4 | atoi) 16 | printf "%x" }} 
{{- $result = print $result $randHexChar }}
{{- end }}
{{- $result }}
```

This line:

```
{{- $randHexChar := mod (randNumeric 4 | atoi) 16 | printf "%x" }} 
```

always generates a single random hex character:

- `randNumeric 4 | atoi` generates a random number in [0, 10^4) (a range evenly divisble by 16, but even if off by one, that last partial interval offsetting randomness is only 1 part in 625)
- `mod N 16` maps to the range 0-15
- `printf "%x"` represents a single number 0-15 as a single hex character

It's a little wasteful to sample 10^4 to generate a number in 2^4, but since we can only generate random numbers in base 10, this works.

jupyterhub/templates/hub/_helpers-passwords.tpl
```diff
@@ -5,35 +5,43 @@
...
         {{- .Values.proxy.secretToken }}
     {{- else }}
-        {{- $k8s_state := lookup "v1" "Secret" .Release.Namespace (include "jupyterhub.hub-secret.fullname" .) | default dict }}
-        {{- if and $k8s_state (hasKey $k8s_state "JupyterHub.proxy_auth_token") }}
-            {{- index $k8s_state "JupyterHub.proxy_auth_token" }}
+        {{- $k8s_state := lookup "v1" "Secret" .Release.Namespace (include "jupyterhub.hub-secret.fullname" .) | default (dict "data" (dict)) }}
+        {{- if hasKey $k8s_state.data "JupyterHub.proxy_auth_token" }}
+            {{- index $k8s_state.data "JupyterHub.proxy_auth_token" | b64dec }}
         {{- else }}
             {{- include "jupyterhub.randHex" 64 }}
```
minrk, line 46: fwiw, unlike the others, which are encryption keys of a particular size, proxy_auth_token is actually a just random string. We should be able to use `randAlphaNum` here.

jupyterhub/templates/hub/_helpers-passwords.tpl
consideRatio: Ah nice solution!

I considered using `%x` but ended up tunnel visioning away from it because of the issue you have estimated.

Given the complicated the path i ended up taking, as well as knowledge that `randAlphaNum 1` has issues even though perhaps `randAlphaNum <larger number>` doesn't, I figure it is a good compromise to go with this more aesthetically pleasing code that is still random enough.

consideRatio: Oh, there isn't even an issue, but if there were, it would be small! :heart:

jupyterhub/templates/hub/_helpers-passwords.tpl
consideRatio: :tada:

consideRatio: Addressed by 437ed29

jupyterhub/templates/hub/_helpers-passwords.tpl
consideRatio: Addressed by 482bbe4

minrk:
Great!

PR #2023, opened 2021-02-02
fix: load only .py files in jupyterhub_config.d folder

### LINKED ISSUE #2022: /usr/local/etc/jupyterhub/jupyterhub_config.d subdirectory made jupyterhub fail to start
I mounted extra configuration to the /usr/local/etc/jupyterhub/jupyterhub_config.d folder. Using a k8s configmap without `subPath` in the volumeMount, it meant that I not only added a my_config.py file, but k8s also created a hidden folder as it does.

This led to z2jh's jupyterhub_config.py crashed while loading files from the jupyterhub_config.d folder, as it detected a folder and got upset.

```
      File "/usr/local/etc/jupyterhub/jupyterhub_config.py", line 415, in <module>
        with open(f"{config_dir}/{file_name}") as f:
    IsADirectoryError: [Errno 21] Is a directory: '/usr/local/etc/jupyterhub/jupyterhub_config.d/..2021_02_02_10_04_22.654959900'
```

@minrk @yuvipanda do you think we should ignore or include content within subdirectories of the jupyterhub_config.d folder?

## Action point
- [x] Decide on how to act regarding directories under the jupyterhub_config.d folder - ignore and warn or load files recursively?
- [x] Implement decision

consideRatio:
Closes #2022 by the implementation suggested by @minrk.

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -1,3 +1,4 @@
+import glob
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @consideRatio!

consideRatio:
@yuvipanda thanks for helping out with review efforts! :heart:

PR #2027, opened 2021-02-09
Allow ingress.hosts to be omitted for a more generic rule

### LINKED ISSUE #1314: Ability to configure our Ingress resource with a default backend
Hello everyone,

I just started to deploy Jupyterhub and trying to set up some features.
At the moment I'm struggling configuring the ingress route. I don't want to specify an explicit host for jupyterhub, being accessible by IP, so far.
I already have an Ingress Controller in my k8s cluster, and I'm using the paths to route my applications. So I want to configure an ingress route for my pre-existing Nginx ingress controller.

I have the configuration on ingress as follows:

```yaml
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/ssl-redirect: "false"
    nginx.ingress.kubernetes.io/rewrite-target: /$1
  pathSuffix: '(.*)'
```

And when upgrading jupyterhub I get the following error:

```
UPGRADE FAILED
Error: failed to create resource: Ingress.extensions "jupyterhub" is invalid: spec: Invalid value: [] extensions.IngressRule(nil): either `backend` or `rules` must be specified
Error: UPGRADE FAILED: failed to create resource: Ingress.extensions "jupyterhub" is invalid: spec: Invalid value: []extensions.IngressRule(nil): either `backend` or `rules` must be specified
```

I understand that it's because I didn't assign any value to 'hosts' and 
Is there any way around this?

consideRatio:
## Review points
Feedback greatly appreciated if this seem to make sense as I've not thought it through so much myself!

1. __Complications__
   This won't cause complications for existing users, right?
1. __Customization__
   Do we need to let some part of this new section be configurable in any way, and if so, in what way and why?
1. __Implementation__
   Does the change technically seem to be correct?

## References
- [v1 Ingress](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressspec-v1-networking-k8s-io) `backend` -> {`serviceName`, `servicePort`}
- [v1beta1 Ingress](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressspec-v1beta1-networking-k8s-io) - `defaultBackend` -> `service` -> {`name`, `port` -> {`name`, `number`}}

Closes #1314

### DIFF
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -17,7 +17,7 @@ spec:
   {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
   defaultBackend:
     service:
-      name: proxy-public
+      name: {{ include "jupyterhub.proxy-public.fullname" . }}
       port:
         name: http
   {{- else }}
@@ -36,7 +36,7 @@ spec:
             pathType: Prefix
             backend:
               service:
-                name: proxy-public
+                name: {{ include "jupyterhub.proxy-public.fullname" $ }}
                 port:
                   name: http
             {{- else }}
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -22,9 +22,10 @@ spec:
         name: http
   {{- else }}
   backend:
-    serviceName: {{ include "jupyterhub.proxy-public.fullname" $ }}
+    serviceName: {{ include "jupyterhub.proxy-public.fullname" . }}
     servicePort: 80
   {{- end }}
+  {{- if .Values.ingress.hosts }}
   rules:
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host | quote }}
@@ -44,6 +45,7 @@ spec:
               servicePort: 80
             {{- end }}
     {{- end }}
+  {{- end }}
   {{- with .Values.ingress.tls }}
   tls:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -14,17 +14,6 @@ metadata:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
-  defaultBackend:
-    service:
-      name: {{ include "jupyterhub.proxy-public.fullname" . }}
-      port:
-        name: http
-  {{- else }}
-  backend:
-    serviceName: {{ include "jupyterhub.proxy-public.fullname" . }}
-    servicePort: 80
-  {{- end }}
   {{- if .Values.ingress.hosts }}
   rules:
     {{- range $host := .Values.ingress.hosts }}
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -14,11 +14,9 @@ metadata:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.ingress.hosts }}
   rules:
-    {{- range $host := .Values.ingress.hosts }}
-    - host: {{ $host | quote }}
-      http:
+    {{- range $host := .Values.ingress.hosts | default (list "") }}
+    - http:
         paths:
           - path: {{ $.Values.hub.baseUrl }}{{ $.Values.ingress.pathSuffix }}
             {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
@@ -33,8 +31,10 @@ spec:
               serviceName: {{ include "jupyterhub.proxy-public.fullname" $ }}
               servicePort: 80
             {{- end }}
+      {{- if $host }}
+      host: {{ $host | quote }}
+      {{- end }}
     {{- end }}
-  {{- end }}
   {{- with .Values.ingress.tls }}
   tls:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -14,6 +14,17 @@ metadata:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
   {{- end }}
 spec:
+  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
+  defaultBackend:
+    service:
+      name: proxy-public
+      port:
+        name: http
+  {{- else }}
+  backend:
+    serviceName: {{ include "jupyterhub.proxy-public.fullname" $ }}
+    servicePort: 80
+  {{- end }}
   rules:
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host | quote }}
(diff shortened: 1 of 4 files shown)


jupyterhub/templates/ingress.yaml
```diff
@@ -14,6 +14,18 @@ metadata:
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
   {{- end }}
 spec:
+  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
+  defaultBackend:
```
manics, line 18: https://kubernetes.io/docs/concepts/services-networking/ingress/#default-backend
> An Ingress with no rules sends all traffic to a single default backend. The defaultBackend is conventionally a configuration option of the Ingress controller and is not specified in your Ingress resources.
>
> If none of the hosts or paths match the HTTP request in the Ingress objects, the traffic is routed to your default backend.

Is this going to change the behaviour of k8s clusters running other applications?

manics: @MickeyShnaiderman-RecoLabs Following up from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1314#issuecomment-776144297 could you please explain why you think this should be configured by Z2JH as opposed to in the ingress controller?

MickeyShnaiderman-RecoLabs: I don't think it should. I misused the term "default backend" in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1314#issuecomment-776144297. I actually meant the Ingress controller itself, my mistake. What I really meant is that not specifying a host in Z2JH's values should just create a rule with no host, or, if the API version supports it, a wildcard host (i.e. `*`). This will cause the Ingress controller (mistakenly called "default backend") to accept all hosts on the given address to the created Ingress rule.

MickeyShnaiderman-RecoLabs: See https://kubernetes.io/docs/concepts/services-networking/ingress/#:~:text=If%20you%20create%20an%20Ingress,based%20virtual%20host%20being%20required.

consideRatio: __Somwhat off topic confusion form my end__

> or, if the API version supports it, a wildcard host (i.e. *).

I think v1beta1 didn't pin how the path specification should work, so different ingress controllers interpreted the same path rule differently. Hmmm, or rather, the default value in v1beta1 had a `pathType` of `ImplementationSpecific` which left it to the ingress controller to decide how to treat the path rule, while `v1` defaults `pathType` to `Prefix`.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#httpingresspath-v1-networking-k8s-io
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#httpingresspath-v1beta1-networking-k8s-io

---

> This will cause the Ingress controller (mistakenly called "default backend") to accept all hosts on __the given address__ to the created Ingress rule.

I'm a bit confused in general about having an ingress without a host. My mental model is that an ingress controller will proxy traffic based on multiple Ingress resources based on a host header. But, I guess it can be more vague than this, that for example an ingress controller in practice could allocate domain names to ingress resources without a domain name associated rule and then let all traffic to that specific domain name go there.

This touches the parts I don't know, and make me not confident about this PR in general. How will a default backend of an ingress typically influence ingress controllers? Is this declared in k8s specs or is it up to the ingress controller entirely without any guidance from k8s specifications?

I wonder for example, what happens if I have a nginx-ingress-controller and two separate Ingress resources with a default backend and no host associated rule - it won't know where to send the traffic arriving to the single IP allocated to it as an entrypoint to the k8s cluster.

consideRatio: @MickeyShnaiderman-RecoLabs do you understand if there is a difference between omitting the `host` section, and declaring a `defaultBackend`?

consideRatio: Oh!

The difference is that by omitting host in a rule, you can still specify a path, while default backend will target all hosts and all paths.

__Question:__ do we want a `defaultBackend` or a rule like beforebut where host is allowed to be omitted?

@MickeyShnaiderman-RecoLabs, do you explicitly need a `defaultBackend` Ingress rule, or would a rule without an explicit host do fine?

consideRatio: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressrule-v1-networking-k8s-io

> If the host is unspecified, the Ingress routes all traffic based on the specified IngressRuleValue.

consideRatio:
I removed the defaultBackend (f36d285), and added logic to omit `host` if `ingress.hosts` wasn't specified in (6bc30d4) to provide a practical example of the choice I see us have.

- Should we _always_, _conditionally_, or _never_, include `defaultBackend`?
- Should we _omit_ the `host` field of the current `http` rule if `ingress.hosts` is falsy?

jupyterhub/templates/ingress.yaml
MickeyShnaiderman-RecoLabs: @consideRatio, @manics: Thanks for finding this caveat! Didn't actually know that it is possible to make a default routing on a service. 

As for the difference, it seems you're right. As for @manics question, I think that **if** the Ingress controller doesn't allocate more than one external address for any of the Ingress objects in the cluster it **will** change the behaviour of the routing for the entire cluster. But, I guess that in my usage example from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1314#issuecomment-776144297 the Ingress controller would get a new address allocated by AWS on every Ingress object, so specifying a `defaultBackend` or just omitting the host would have the same effect, especially when there is Z2JH's proxy Pod to handle different paths.

So yes, a rule without an explicit host would do just fine and possibly prevent problems in other people's clusters 😅

MickeyShnaiderman-RecoLabs:
Works well on EKS!

More insight: unlike my previous assumption at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2027#discussion_r573524479, turns out you can't create two Ingress objects without a host and with a default path (`/`) on the same cluster if you have only one backend (in our case is the nginx server). I tried to deploy the chart in a new namespace while the official was still running with a modified Ingress object without a host. Got the following error:
```
Error: admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: host "_" and path "/" is already defined in ingress jhub/jupyterhub
```

So everything works as intended.

consideRatio:
Okay! I'm happy with the current implementation of the PR, except perhaps a look through the documentation for potential updates.

Does the code changes look good to you @manics ?

consideRatio:
I updated the ingress docs as well to reflect this.

jupyterhub/templates/ingress.yaml
```diff
@@ -15,23 +15,25 @@ metadata:
   {{- end }}
 spec:
   rules:
-    {{- range $host := .Values.ingress.hosts }}
-    - host: {{ $host | quote }}
-      http:
+    {{- range $host := .Values.ingress.hosts | default (list "") }}
```
manics, line 18: Can this default go in
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/07cd588cf5173403c245900d180eca3b3e020491/jupyterhub/values.yaml#L504
instead? Or would that break something?

consideRatio: It would have to be `[""]` if it went in the default value, and it would break if someone set it to `[]` as the logic is to force a empty-string entry to trigger at least one iteration at all time, and use the blank string entry as a sign to omit the host field.

I think I prefer it as it is because it is more robust setting `[""]` by default expose irrelevant complexity to the end user I think.

manics: OK!

PR #2033, opened 2021-02-13
values.schema.json ships with chart and configuration reference now covers all options

### LINKED ISSUE #1316: Helm 3: Values.yaml schema validations
Helm v3 should support schemas for validating the values file: https://github.com/helm/community/blob/master/helm-v3/001-charts.md#schematized-values-files

If this was added to Z2JH it might help catch simple errors such as typos or changes in property names. E.g.:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/421#issuecomment-504690708
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1310#issuecomment-503290959

### LINKED ISSUE #1829: schema.yaml: not all configuration options are documented
## What is schema.yaml? The configuration reference!
[schema.yaml](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/schema.yaml) is a JSONSchema specification of the values that the Helm chart accepts, we currently use it to generate documentation as can be seen in [Configuration reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html) section of the guide.

This documentation in schema.yaml should be kept up to date with the values the Helm chart template depends on when they render into valid k8s resources. The default values of the Helm chart, in [values.yaml](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml), should also contain a default entry for all of the options and is currently a more complete overview of what is configurable.

## Missing schema.yaml entries
If you find more missing entries, comment in this thread so we can add them to this list, and if you document them in schema.yaml, either directly update this list if you have permission to do so or comment so a maintainer can do it for you.

- [ ] singleuser.serviceAccountName
- [ ] singleuser.storage
- [ ] singleuser.defaultUrl
- [ ] hub.concurrentSpawnLimit (reported in #1725)

consideRatio:
# Summary
- schema.yaml and values.yaml are now in parity making the configuration reference complete, [preview here](https://zero-to-jupyterhub--2033.org.readthedocs.build/en/2033/) - closes #1829
- `helm` validation logic is packaged with the chart in [values.schema.json](https://helm.sh/docs/topics/charts#schema-files), generated from schema.yaml - closes #1316

### Breaking changes

If someone has a configuration that overrides a default value to null or similarly, then they can get validation errors before the k8s api-server has been contacted by the `helm` CLI.

# Implementation notes

### Helper scripts
- `tools/compare-values-schema-content.py`
  This new script isn't run as part of the CI system, but was helping me to spot if there were values in values.yaml that weren't in 
  schema.yaml and vice versa.

- `tools/validate-against-schema.py`
  This script was previously located in the chart folder and named validate.py, I've just moved it and made it validate with the lint- 
  and-validate-values.yaml file as well. This script now runs in a dedicated job part of the test-chart workflow, before it was undocumented and not part of the CI system.

- `tools/generate-json-schema.py`
  This new script reads schema.yaml, cleans it from descriptions that isn't relevant for the `helm` CLI's logic as far as I know, and emits jupyterhub/values.schema.json. This script runs in all chart tests and before we package and publish the Helm chart.

### Accept null or not?

I have considered the type enforcement of all values that require a string or integer and considered if they should allow null or not. The principle I concluded was that if the field is a Helm chart native configuration is should typically not allow null. But, if it is a passthrough configuration that isn't required by the Helm chart templates, then we should allow it be null in order to be able to use the default values of the actual software such as JupyterHub or KubeSpawner. For such values, null implies the default value used in the software rather than the default value of the Helm chart typically.

If we would support null for all values, we would risk introducing unneeded runtime errors helm template rendering errors that are far harder to interpret. Due to this, I have opted to avoid allowing null.

Sometimes the chart has given a blank string different meaning to null, such as for `fullnameOverride`. At this times null is also allowed.

# Notes on `helm` validation logic

I don't think it is possible to be very strict and warn for passed configuration not part of the schema which would be useful if a user has misspelled an option. What we can do, is to declare _required fields_ (`required`), and enforce the fields passed have _accepted data types_ (`type`).

In other words, this is not an end to the mistakes you can make, but it is a good improvement. While there may be some fields that are required, I have only used that keyword to require the root objects such as `hub` and `proxy` to be set.


jupyterhub/values.yaml
```diff
@@ -63,10 +63,10 @@ hub:
       accessModes:
         - ReadWriteOnce
       storage: 1Gi
-      subPath:
-      storageClassName:
-    url:
-    password:
+      subPath: ""
+      storageClassName: ""
```
yuvipanda, line 67: If it's set to empty rather than unset, will the resulting field be present? I think many fields in Kubernetes have adifferent effect when it's set to empty string - particularly, the defaults might not apply. 

For example, an empty storageClassName has different behavior from having it be not present (see [doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume)). 

So my instinct is to allow anything that can be a non-container type (so strings, numbers, bools) be nullable.

tools/compare-values-schema-content.py
```diff
@@ -0,0 +1,94 @@
...
+       containerSecurityContext, readiness- and livenessProbe's, and hub.config.
+"""
+
+import jsonschema
+import os
+import sys
+
+from collections.abc import MutableMapping
+
+import yaml
```
yuvipanda, line 24: Don't we use ruamel.yaml elsewhere, instead of pyyaml?

yuvipanda (approved):
I AM ABSOLUTELY AMAZED BY THE AMOUNT OF WORK AND CARE YOU PUT INTO THIS. WOW

jupyterhub/schema.yaml
```diff
@@ -474,9 +476,7 @@ properties:
               zhy270a
               ```
           pullPolicy:
```
minrk, line 478: I believe a null pull policy is valid, and will use kubernetes' default pull policy (i.e. Always if tag == latest, IfNotPresent otherwise).

minrk:
This is awesome! ❤️ 


> Accept null or not?

I think we need to be careful with this a bit. Setting to null and leaving unset are not always equivalent, but *often* are, and that seems to be the norm for helm/kubernetes, but *not* for traitlets. If they are interpreted by the helm chart with things like `with .Values.something` allowing null is usually fine as the result is to leave it unset. With passthrough config to traitlets, though, null and unset are *not* generally the same, and the schema should generally distinguish. Similarly, with kubernetes, setting many fields to null is the same as leaving it unset (imagePullPolicy is one, I believe, where we should allow null to select the cluster default behavior).

The exception is the subset of config that is *not* direct passthrough, and instead is specifically transformed by the chart from values.yaml to traitlets config (e.g. `singleuser.cmd`). For those, I agree that allowing explicit null to mean using the default is the right thing to do. This is what we've implemented with the `set_config_if_not_none` utility in juptyerhub_config.py. That utility function is required to implement this different interpretation - helm values null means apply no config, while None in traitlets config means explicit override default with None.

.github/workflows/test-chart.yaml
```diff
@@ -50,6 +50,39 @@ jobs:
...
+
+      # FIXME: We can probably emit a GitHub workflow warning if these fail
+      #        instead having them show as green without a warning or similar
+      - name: Helm lint --strict (values.yaml)
+        run: helm lint --strict ./jupyterhub
+        continue-on-error: true
+
+      - name: Helm lint --strict (lint-and-validate-values.yaml)
+        run: helm lint --strict ./jupyterhub
+        continue-on-error: true
```
minrk, line 84: These two steps are identical. Something missing?

tools/validate-against-schema.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+import jsonschema
+import os
+import sys
+import yaml
+
+# Change current directory to this directory
+os.chdir(os.path.dirname(sys.argv[0]))
```
minrk, line 8: We generally don't need to change directories in scripts. Instead, we can resolve paths with `os.path.join`, and friends, e.g.

```python
tools_dir = os.path.abspath(os.path.dirname(__file__))
chart_dir = os.path.join(tools_dir, os.pardir, 'jupyterhub')
templates_dir = os.path.join(tools_dir, os.pardir, 'templates')

with open(os.path.join(chart_dir, 'schema.yaml')) as f:
    ....
etc.
```

tools/compare-values-schema-content.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Change current directory to this directory
+os.chdir(os.path.dirname(sys.argv[0]))
```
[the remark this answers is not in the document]
minrk: Also see comment below in validate-against-schema  - I don't think `os.chdir` in scripts is usually something we want to do.

jupyterhub/values.yaml
minrk: I agree, I think null for unset is often the right thing to do.

jupyterhub/schema.yaml
consideRatio: Hmmm, the valid options for k8s are _unset_, `IfNotPresent`, `Always`, `Never`.

The validation logic for this Helm chart config will of course always accept the valid non-empty string values, the question is with regards to the _unset_ part. It must be able to accept the unset part, but should it accept a blank string to imply unset, a null value to imply unset, or both?

I think the right call is to accept both, so I'll update the type field below to include `"null"` so it would accept both a blank string and `"null"`, sounds okay?

jupyterhub/values.yaml
consideRatio: Ah yes storageClassName was one of those fields where we differentiate between unset and empty string!

```yaml
  {{- if typeIs "string" .Values.hub.db.pvc.storageClassName }}
  storageClassName: {{ .Values.hub.db.pvc.storageClassName | quote }}
  {{- end }}
```

I agree. Resolved by 1153aa35.

.github/workflows/test-chart.yaml
consideRatio: Yes! Resolved by e65f4cdf

jupyterhub/schema.yaml
consideRatio: Resolved by f3938781

tools/compare-values-schema-content.py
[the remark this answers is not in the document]
consideRatio: I'd like the script to function the same way independently of where they are executed from.

I guess the crux of using os.chdir is that it can influence the environment calling the script? Running it from my terminal didn't influence the terminal, but it could perhaps?

I'm not sure on the path forward here, suggestions?

tools/generate-json-schema.py
```diff
@@ -0,0 +1,63 @@
...
+import json
+import os
+import sys
+
+from collections.abc import MutableMapping
+
+import yaml
+
+# Change current directory to this directory
+os.chdir(os.path.dirname(sys.argv[0]))
```
consideRatio, line 21:
```suggestion
os.chdir(os.path.abspath(os.path.dirname(__file__)))
```

tools/compare-values-schema-content.py
consideRatio: Our dev-dependencies have pyyaml while I see ruamel.yaml in the images/hub dependencies script.

Do you favor one or the other? I currently don't favor one or the other based on not being aware of differences.

consideRatio:
@minrk regarding comments on _Accept null or not?_ I think we are in agreement and have arrived at the same conclusions, but I find the topic quite complicated in general.

Is it correct that you don't suggest changes other than those you have already explicitly pointed out?

tools/validate-against-schema.py
minrk: Or you can use [pathlib](https://docs.python.org/3/library/pathlib.html) if you want to be a Fancy Person.

tools/compare-values-schema-content.py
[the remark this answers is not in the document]
consideRatio: Resolved by cb0ae7d9

tools/validate-against-schema.py
consideRatio: Resolved by cb0ae7d9

tools/compare-values-schema-content.py
yuvipanda: Ah, I see. I generally favor ruamel.yaml - it has an 'rt' mode that can write out YAML formatted the same way it is read (including comments!). So chartpress, etc use it. Not critical though.

jupyterhub/values.yaml
```diff
@@ -35,11 +35,11 @@ hub:
     annotations: {}
     ports:
       nodePort:
-    loadBalancerIP:
+    loadBalancerIP: ""
```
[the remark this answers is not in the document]
yuvipanda: Trying to understand why some vars are empty string and some are unset. My intuition is that default should be either a value we want (like `uid: 1000`), unset, or empty container (`[]`, `{}`, etc)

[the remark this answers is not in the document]
consideRatio: The principle ive had while writing these evolved to be:

1: if the helm chart provide an explicit default value: set it.
2: else if the helm chart config is a passthrough config for
jupyterhub/kubespawner/authenticator etc: we set it to null to represent the intent for the helm chart to not change the default value of the underlying config, and a falsy value to represent a wish for it to be explicitly set to falsy value.
3: else if, null snd a falsy instance type both imply a unset value, set it to the falsy instance type
4: else set it to null

To me, the most controversial point is the third. The question in my mind boils down to: if both null and a falsy type ("", {}, [], 0) by the helm templates imply that a value wont be set, should we specify the values falsy type of null?

I lean towards specifying the falsy instance type rather than null if the config isn't passthrough.

I think we break this logic on point 2 for array/objects passthrough config, which i think is motivated by avoiding warnings and help helm templste logic.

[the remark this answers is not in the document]
yuvipanda: So with loadBalancerIP, I see the following code

```
  {{- with .Values.proxy.service.loadBalancerIP }}
  loadBalancerIP: {{ . }}
  {{- end }}
```

So if I understand correctly, the `loadBalancerIP` key in the Service object won't be present at all in the following circumstances:

1. `.Values.proxy.service.loadBalancerIP` is specified in helm chart values but not set (current master situation)
2. `.Values.proxy.service.loadBalancerIP` is set to `""` (as in the PR)

Is this understanding correct?

tools/compare-values-schema-content.py
consideRatio: Ah thats very imporitant for chartpress for examole, good to know!

jupyterhub/values.yaml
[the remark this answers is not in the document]
consideRatio: Yepp! The with keyword using a falsy value is like a do nothing statement, it has become a practice to use it throughout the helm ecosystem over if ... Then ... - probably to avoid duplicating the reference needed in both the conditional statement and within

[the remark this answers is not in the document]
consideRatio: I like it for not needing to write the same values reference twice, but dislike that it is less obvious on how it behaves.

tools/compare-values-schema-content.py
consideRatio: Lets switch if that ever cause trouble to not have!

jupyterhub/values.yaml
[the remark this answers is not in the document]
yuvipanda: Right. So if I see the `values.yaml` file, it looks like I'll actually get `loadBalancerIP: ""` in my Service object, which is confusing! Since k8s differentiates between unset values and `""` values, so should we - otherwise it can be pretty confusing. I think the mental model of `if a value is unspecified (null), it is not present in the target` is better than `if a value is falsey (by go template's definition of falsey) it will not be present in the target`. The latter also has exceptions, like storageClassName, making it even more confusing.

So ideally, I'd like to hold the mental model of 'if it is null in my values.yaml, it will not be present in the output Service object'.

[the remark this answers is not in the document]
consideRatio: @yuvipanda I can make it so for strings/integers without problem I think, but if we apply the same principle to `{}` and `[]` I believe there will be trouble such as helm emitting warnings, broken assumptions by helm template logic, and potentially a schema that fail to capture otherwise easy to capture issues.

I suggest we apply the principle anyhow to our scalar strings and numbers, but don't try to apply it on falsy values for object `{}` and array `[]`, does that sound okay to you?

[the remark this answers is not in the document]
yuvipanda: That's perfect! I totally agree, @consideRatio :)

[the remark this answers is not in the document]
consideRatio: Thanks for quickly iterating with me on this @yuvipanda!! I love it when I arrive at a concrete action point!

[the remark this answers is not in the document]
consideRatio: It was a quite hard rule principle to follow without making exceptions. I was able to allow many strings to be null so our default values are now `null` rather than `""` by default, but plenty of booleans remain required to be "boolean" rather than "boolean or null" still.

__Example of complexity arisen:__
If helm templates have logic making an `eq` comparison between a variable and a string, then it will error if it holds a null value. And, since helm template logic will evaluate Y in statements like like `if X and Y`, one is forced to do `if X then if Y then ...` or `if X and (Y | default "")` etc.

Anyhow, things work okay now, don't want to change back, I don't see a perfect tradeoff with regards to these considerations.

---

Resolved by 2c1cc57

consideRatio:
I think I've addressed all review points again and that this PR is potentially ready for a merge.

yuvipanda:
@consideRatio happy to hit merge after the merge conflict is resolved.

consideRatio:
@yuvipanda :tada: rebased to resolve simple merge conflict

yuvipanda:
\o/ THANK YOU, @consideRatio!

consideRatio:
Wieeee :heart: :tada: thank you @yuvipanda!!

PR #2036, opened 2021-02-16
hub image build: fix use of PIP_OVERRIDES arg

remche:
Default shell for Docker build is `sh`, the double bracket is not supported and pip overrides wont work.
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/runs/1804962086?check_suite_focus=true#step:8:1181

I used the `test` notation as few lines below.


consideRatio:
Thanks @remche!! ❤️ 🎉 Can you cleanup the three initial no-change commits also?

consideRatio:
🥳🥳🥳

PR #2038, opened 2021-02-17
fix: imagePullSecret.enabled to work alongside imagePullSecret.create

consideRatio:
`imagePullSecret.create` were to be the new `.enabled` which should still work, but it turns out that I've missed one entry making `.create` function while `.enabled` wouldn't function.


consideRatio:
Actually hmmm... schema.yaml doesn't represent `.create` / `.enabled` properly yet.

I'm considering removing the `.enabled` alias and forcing users to transform their config to have `.create` instead of `.enabled` to reduce the complexity.

I think it would be quite messy to implement a check for either enabled or create based on my brief experience of jsonschema. Anyone knows how to do that nicely?

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/a8432b26459e91d02ebc665efdd0e9a49b7f0087/jupyterhub/schema.yaml#L80-L88

consideRatio:
> Actually hmmm... schema.yaml doesn't represent `.create` / `.enabled` properly yet.

Well, this isn't a real problem. It won't raise an issue for the user if the user has used `.enabled` instead of `.create`.

PR #2039, opened 2021-02-17
pdb: default to maxUnavailable=1 instead of minAvailable=1

consideRatio:
If you want to lower the replicas of user-scheduler to 1, then you will get the same PDB issues we have had with hub/proxy (#1938) where they block upgrades etc because we have `minAvailable: 1` instead of `maxUnavailable: 1` in order to handle workloads.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -890,9 +890,12 @@ properties:
             description: |
               Decides if a PodDisruptionBudget is created targeting the
               Deployment's pods.
+          maxUnavailable:
+            type: [integer, "null"]
+            description: |
+              The maximum number of pods required to not be available during
+              voluntary disruptions.
           minAvailable:
-            # FIXME: support maxUnavailable alongside minAvailable, making it
-            #        make sense for both to be null also.
             type: [integer, "null"]
             description: |
               The minimum number of pods required to be available during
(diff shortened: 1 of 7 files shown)


jupyterhub/schema.yaml
```diff
@@ -890,9 +890,12 @@ properties:
             description: |
               Decides if a PodDisruptionBudget is created targeting the
               Deployment's pods.
+          maxUnavailable:
+            type: [integer, "null"]
+            description: |
+              The maximum number of pods required to not be available during
```
yuvipanda, line 896:
```suggestion
              The maximum number of pods that can be unavailable during
```

yuvipanda:
So if we have 2 replicas, it will still ensure there will at least be one available, right?

consideRatio:
> So if we have 2 replicas, it will still ensure there will at least be one available, right?

Yepp! Note btw that maxUnavailable was introduced in k8s 1.9 or something, which was later than the PDBs were introduced in the first place in z2jh.

consideRatio:
image

Hmmmmm, perhaps did a mistake with the logic... It should show maxUnavailable there.

consideRatio:
Omg, `kindOf` returns `"float64"` instead of `"int"`, woops.

consideRatio:
Okay finally got everything right, the tests does a great job now! :heart:

yuvipanda:
Thanks, @consideRatio!

consideRatio:
Thanks @yuvipanda for your help with review/merge!

PR #2042, opened 2021-02-19
Make use of hub.existingSecret sustainable

### LINKED ISSUE #2009: Hub existingSecret functionality not working as expected
### Bug description

Hi,  I am trying to update jupyterhub to helm chart 0.10.6 from 0.9.0, and with the recent change of moving values.yaml only to hub-secret and removing it from hub-config is breaking the Hub existingSecret functionality. In our existing deployment on EKS we use gitops/flux/helm operator for our deployments and we didn't want to add secrets to git, so we created hub-secret which only includes secrets(proxy token, db creds, crypto-key, cookie-secret, etc) using [external secrets operator](https://github.com/external-secrets/kubernetes-external-secrets)  and set the name as existingSecret. Now when the existingSecret is set, the secret is not created and values.yaml doesn't make it to the hub pod and it doesn't start properly.

#### Expected behaviour
The previous behaviour of existingSecret should continue to work without having to add values.yaml to existingSecret.

consideRatio:
# Reimplementation of hub.existingSecret

Use of hub.existingSecret hasn't been a good experience for users because they end up needing to manually update it whenever they change something passed to the hub pod through values.yaml to be read in jupyterhub_config.py.

This PR is solving this by allowing hub.existingSecret be mounted in parallel to the chart managed k8s Secret resource we make use of and then merging values.

Closes #2009.

## Documentation preview

[Available here.](https://zero-to-jupyterhub--2042.org.readthedocs.build/en/2042/resources/reference.html#hub-existingsecret)

## Breaking changes

- Users of `hub.existingSecrets` need to recreate their secrets and read the docs on the changes.
- Those using `hub.services` _and_ referencing the api token from the chart managed k8s secret, need to reference `hub.services.<service-name>.api_token` going onwards instead of `service.token.<service-name>`

## Implementation challenges

__`lookup` of hub.existingSecret?__
There are two exceptions to the general rule of merging the config from hub.existingSecret. These exceptions are for `hub.db.password` and `hub.config.JupyterHub.proxy_auth_token`, or `hub.config.ConfigurableHTTPProxy.auth_token` as the non-deprecated value is called.

These specific keys are exceptions as we need to know if they are declared in the chart managed or self managed k8s Secret during template rendering. This is because we reference one k8s Secret or the other as a source of an environment variable on the hub and/or proxy pod during template rendering. And, deciding this is a problem as I don't think we can or should rely on a `lookup` of a non-chart managed k8s resource. I'm actually not sure if this works or not, I just recall this to fail, but even if this recollection is wrong and it would work, it would be good to minimize `lookup` calls in general I think due to the unknown performance or ability to do so in all k8s clusters independent on various k8s Secret mechanisms in play.

I consider this challenge managed:
- We document that the user need to specify hub.db.password within the user managed k8s Secret if they use an external database and hub.existingSecret.
- We document that the user should `kubectl edit` the chart managed k8s secret in place if they want to update ConfigurableHTTPProxy.auth_token.

## Implementation steps

- [x] Let hub.existingSecret mount in parallel to the defaultSecret in the hub pod.
- [x] Merge in existingSecret values with defaultSecret values, and let top level  keys override defaultSecret values if they are defined. This can be handled in z2jh.py.
- [x] Create three separate _helpers-names.tpl templates and make use of them where appropriate.
  - `hub`                                used for: volume declaration hub pod, mounting env var from JupyterHub.proxy_auth_token
  - `hub-existing-secret`                used for: volume declaration hub pod (conditional)
  - `hub-existing-secret-or-default`     used for: mounting of hub.db.password env var
- [x] Update configuration reference entry on hub.existingSecret
- [x] Changelog strategy defined: we simply reference docs in the configuration reference for whoever use hub.existingSecret.

## Unrelated but part of PR
- [x] Replace references to JupyterHub.proxy_auth_token with ConfigurableHTTPProxy.auth_token as the previous is deprecated by JupyterHub. This won't matter but feels nice to do. This configuration value is indirectly set using an environment variable, both on the hub and proxy pod though, so perhaps it wouldn't even be a warning about it before or after.


meeseeksmachine:

https://discourse.jupyter.org/t/how-am-i-supposed-to-use-existingsecret/5097/5

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -420,14 +420,18 @@ def camelCaseify(s):
...
-c.JupyterHub.proxy_auth_token = get_secret_value("JupyterHub.proxy_auth_token")
+c.ConfigurableHTTPProxy.auth_token = get_secret_value(
+    "ConfigurableHTTPProxy.auth_token"
+)
 c.JupyterHub.cookie_secret = a2b_hex(get_secret_value("JupyterHub.cookie_secret"))
 c.CryptKeeper.keys = get_secret_value("CryptKeeper.keys").split(";")
 
 # load hub.config values, except potentially seeded secrets
 for section, sub_cfg in get_config("hub.config", {}).items():
     if section == "JupyterHub" and sub_cfg in ["proxy_auth_token", "cookie_secret"]:
```
minrk, line 431: What does this disallowed list do? To make sense, sub_config must always be a dict, so these `sub_cfg in [list]` conditions will always be False. Or am I misunderstanding something?

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -74,20 +74,23 @@
     {{- include "jupyterhub.fullname.dash" . }}hub
 {{- end }}
 
-{{- /* hub Secret */}}
-{{- define "jupyterhub.hub-secret.fullname" -}}
+{{- /* hub-existing-secret Secret */}}
+{{- define "jupyterhub.hub-existing-secret.fullname" -}}
```
minrk, line 78: Does it make sense to have a helper that just maps exactly onto a single value?

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -74,20 +74,23 @@
...
     {{- /* A hack to avoid issues from invoking this from a parent Helm chart. */}}
     {{- $existing_secret := .Values.hub.existingSecret }}
     {{- if ne .Chart.Name "jupyterhub" }}
         {{- $existing_secret = .Values.jupyterhub.hub.existingSecret }}
```
minrk, line 82: Should the chart name change where in values our secret is? I wouldn't think a nested .Values.jupyterhub should be allowed to exist in any event.

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -188,20 +197,23 @@ spec:
...
+                  #       lookup function on the user managed k8s Secret.
+                  name: {{ include "jupyterhub.hub.fullname" . }}
+                  key: ConfigurableHTTPProxy.auth_token
             {{- if .Values.hub.db.password }}
             {{- if eq .Values.hub.db.type "mysql" }}
             - name: MYSQL_PWD
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "jupyterhub.hub-secret.fullname" . }}
+                  name: {{ include "jupyterhub.hub-existing-secret-or-default.fullname" . }}
```
minrk, line 210: What if we set this env in juptyerhub_config instead of the pod? Could we get rid of this exceptional behavior?

jupyterhub/templates/_helpers-names.tpl
consideRatio: `.Chart.name` is based on the name in Chart.yaml which will always be `"jupyterhub"` unless this template was called from a parent helm chart. So this logic is an attempt to make this return a valid response even if the passed scope must now dig into `.Values.jupyterhub` instead of just `.Values` to find what's relevant.

The point of doing this is to help a Helm chart like DaskHub or similar not hardcode names of resources it wants to interact with in this chart.

jupyterhub/templates/_helpers-names.tpl
consideRatio: I'd like to provide that no matter what for consistency. These helpers are accessible from parent Helm charts. And, with nameOverride and fullnameOverride, they can be important to arrive at the right resource name.

I have a feeling that that always referencing resources using a named template is a helpful practice to stick to.

jupyterhub/templates/hub/deployment.yaml
consideRatio: Absolutely, If you think we can avoid that, it would be great! I has been assuming that we couldn't because it would be consumed by various tools outside our control.

Do you know how we would configure this in a way that covers all situations properly?

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Ah yes this logic is just malfunctional because I did the mental mistake of thinking sub_cfg was string representing the sub config key name (`"proxy_auth_token"`).

Resolved by fcea63f I think.

jupyterhub/templates/_helpers-names.tpl
minrk: But that's not how parent charts work, is it? The the parent passes .Values.$dependency-name *as* .Values to the dependency chart. A chart doesn't need to know its name to find its values, otherwise ~nothing in this chart would work in binderhub.

jupyterhub/templates/hub/deployment.yaml
minrk: We can set `os.environ` in jupyterhub_config.py and it should work the same as setting it at the pod level. That will occur soon enough that it will affect the relevant targets (postgres/mysql imports later on in the same process), and also subprocesses, if there are any (there shouldn't be). Only init containers wouldn't get this env, but I don't think that's an issue, and they can always access the same secret if they need to.

So this should work in jupyterhub_config.py (with the appropriate `if`s):

```python
os.environ["MYSQL_PWD"] = values.db.password
```

jupyterhub/templates/_helpers-names.tpl
consideRatio: All the charts templates will work fine and won't see `.Values.jupyterhub.hub` etc but instead `.Values.hub`. But, templates are globally scoped and can be used from a parent chart with the parent chart's scope.

So, in a daskhub template, you could do `{{ include "jupyterhub.proxy-public.fullname" . }}` thanks to this, because when that template gets the scope's `.Chart.Name` is `"daskhub"` and now have `.Values.jupyterhub.hub`, the globally accessible named template from the jupyterhub helm chart will render as if it was called with a JupyterHub scope even though it wasn't.

jupyterhub/templates/hub/deployment.yaml
consideRatio: Haha doh oh yes of course!!

manics:
It sounds like there's loads of expertise hidden in here :smile:

Is there somewhere we can document this knowledge (not necessarily in this PR)? The comments about supporting the use of Z2JH in other Helm charts, and how to configure it, sound particularly valuable to me.

consideRatio:
Okay I've updated this, but I have not tested it manually or made this covered by the k8s tests.

Any suggestion on how to test this? We would need to create a k8s Secret and reference it. I'd like to avoid adding complexity when running tests locally without the CI system. How to add a test creating a k8s secret and configuring hub.existingSecret and running a specific pytest in only certain situations?

I can solve this also, just need some thinking on how best to go about it.

consideRatio:
> The comments about supporting the use of Z2JH in other Helm charts, and how to configure it, sound particularly valuable to me.

@manics I created https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2057!

consideRatio:
Tests added, but not to verify the behavior of setting the database environment variables from jupyterhub_config.py though as it is done based on hub.db.type which would interfere with startup to change.

consideRatio:
- [x] I made sure that anyone using `hub.services` which are also exposed in the k8s secret would also work properly.
- [x] I renamed various fields in the k8s secret to consistently use the chart configuration names the fields associate with.

jupyterhub/templates/_helpers-names.tpl
minrk: > But, named templates defined in _helpers.tpl are globally scoped and can be used from a parent chart with the parent chart's scope.

Ah! This is the part I misunderstood, thank you! It's a specific issue to *exporting* helpers to be used by downstream charts. I never realized helpers were available to other charts. Do we need to think about what our public helper API is and how it's documented? I always assumed these were purely private. What would happen in  a case like mybinder.org, where jupyterhub is a dependency two hops away, i.e. `binderhub.jupyterhub.hub...`? Is that sustainable?

jupyterhub/templates/hub/deployment.yaml
consideRatio: Resolved!

jupyterhub/templates/_helpers-names.tpl
consideRatio: I responded in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2057!

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -60,10 +60,15 @@ def camelCaseify(s):
...
 db_type = get_config("hub.db.type")
+db_password = get_secret_value("hub.db.password", None)
 if db_type == "sqlite-pvc":
     c.JupyterHub.db_url = "sqlite:///jupyterhub.sqlite"
 elif db_type == "sqlite-memory":
     c.JupyterHub.db_url = "sqlite://"
+elif db_type == "mysql" and db_password != None:
+    os.environ["MYSQL_PWD"] = db_password
+elif db_type == "postgres" and db_password != None:
+    os.environ["PGPASSWORD"] = db_password
```
minrk, line 71: I don't think these want to be in the same if/elif branches for db_url. This code as it is would mean `hub.db.url` is not set if db_type is mysql or postgres with a password, which I think isn't correct (generally db_url *must* be set in these cases). A separate if/elif block for setting the password env should be right.

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -360,13 +365,17 @@ def camelCaseify(s):
...
+    # c.JupyterHub.services is a list of dicts, but
+    # hub.services is a dict of dicts to make the config mergable
     service.setdefault("name", name)
-    # handle camelCase->snake_case of api_token
-    api_token = service.pop("apiToken", None)
+
+    # As the api_token could be exposed in hub.existingSecret, we need to read
+    # it it from there or fall back to the chart managed k8s Secret's value.
+    service.pop("apiToken", None)
+    api_token = get_secret_value(f"hub.services.{service['name']}.api_token", None)
```
minrk, line 375: What happens if service.apiToken is defined in values as it normally is now?

I think perhaps what we want is this:

```suggestion
    api_token = service.pop("apiToken", None)
    api_token = get_secret_value(f"hub.services.{service['name']}.api_token", api_token)
```

so that we load it from values, and allow override from the secret.

Related: in .Values we expect camelCase `apiToken`, but in secret we expect snake_case `api_token`. Is that intentional?

Also related: is there a reason the custom secret values are not merged into the config so that `get_config` works with values from any source?

consideRatio: > What happens if service.apiToken is defined in values as it normally is now?

It will be written to the chart managed k8s Secret, and then read by get_secret_value() which will favor any values coming from the mounted self managed k8s Secret if hub.existingSecret was set.

So, all good!

---

When rendering the self managed k8s secret i made it be snake_case deliberately but I'm not sure what I think make sense. Perhaps camelCase after all? I guess hub.services.<name>.apiToken and hub.config.Jupyterhub.services.0.api_token is what I think would make sense to reference with for consistency. So yeah I'll go for apiToken in this case.

So, changing to apiToken!

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Yikes yes very true, happy about your sharp eyes catching this!

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Resolved by b0716dd

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: Resolved by 5f8fac2

jupyterhub/files/hub/jupyterhub_config.py
minrk: Yeah, I think the scheme for putting things in the secret, exactly matching the 'path' in .Values makes good sense. That way, there's only one thing to remember.

For converting to snake_case, I think that makes sense when we are talking directly about traitlets config, but I don't think that's what we have here, since the schema is defined by the helm chart (we can talk separately about whether that's right, but assuming it's not changing here).

This PR is already great, so I think we should land it. But I think maybe a subsequent PR can maybe even more fully bake in this consistency and do away with `get_secret_value` in favor of only having `get_config` read from multiple sources. Not sure if that's feasible, but I think it would be a nice simplification!

dev-config.yaml
```diff
@@ -48,6 +53,34 @@ hub:
...
+
+          print()
+          print("TEST LOGS: test_hub_existing_secret")
+          print(f"hub.existingSecret={get_config('hub.existingSecret')}")
+          print()
+          print(f"hub.db.type={get_config('hub.db.type')}")
+          print(f"MYSQL_PWD={os.environ.get('MYSQL_PWD', None)}")
+          print(f"PGPASSWORD={os.environ.get('PGPASSWORD', None)}")
+          print()
+          print(f"hub.services.test-hub-existing-secret.apiToken={get_secret_value('hub.services.test-hub-existing-secret.apiToken', None)}")
```
minrk, line 70: get_secret_value doesn't take a second 'default' argument, leading to [this error](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/runs/1959618243#step:21:576):

> TypeError: get_secret_value() takes 1 positional argument but 2 were given

 But since you expect it to, maybe it should! That's certainly standard for `getters`.

After that, all LGTM!

consideRatio: So to conlcude, the hub.extraConfig have a new entry in dev-config.yaml, which is meant to emit relevant information from the container enabling a test to decide if everything is as it should.

Since that is used for all tests, even the upgrade tests where we first install 0.11.1 (latest stable), and 0.11.1-n123.h1234567 (latest dev release), then it need to run successfully there also.

But, in the logic we make use of the function that was introduced after 0.11.1 and gained an optional parameter in this PR. To avoid issues for 0.11.1 I have an try/catch block for ImportErrors but not for any error such as passing a parameter that isn't defined.

So, no change is needed I think.

minrk: Gotcha, thanks!

minrk:
Wonderful!

consideRatio:
Thanks @minrk! https://readthedocs.org/ seems to be down btw if you also got an error message.

PR #2055, opened 2021-02-20
Support setting resources to null to omit them

consideRatio:
It wasn't possible to reset the resource requests such as `hub.resources` to `null` because of our helm template logic assumed an object. This PR makes it so that if a .resources chart config value is `null`, then the resources section will be omitted.


yuvipanda:
w00p, thanks @consideRatio!

PR #2061, opened 2021-02-23
Add option to disable http port on LoadBalancer service

tkislan:
This is to disable `http/80` LoadBalancer port when deploying with `offload` proxy service type.
Without this, even though the hub resides behind the https terminating proxy, `http/80` port is still available, and even reuses the cookies from `https/443` endpoint.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -857,7 +857,7 @@ properties:
             description: |
               Default `false`.
 
-              If `true` disables service HTTP traffic on port 80, and therefore forcing use of HTTPS.
+              If `true`, port 80 for incoming HTTP traffic will no longer be exposed. This should not be used with `proxy.https.type=letsencrypt` or `proxy.https.enabled=false` as it would remove the only exposed port.
           extraPorts:
             type: list
             description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -852,6 +852,12 @@ properties:
                 type: integer
                 description: |
                   The HTTPS port the proxy-public service should be exposed on.
+          disableHttpPort:
+            type: boolean
+            description: |
+              Default `false`.
+
+              If `true` disables service HTTP traffic on port 80, and therefore forcing use of HTTPS. 
           extraPorts:
             type: list
             description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -857,7 +857,7 @@ properties:
             description: |
               Default `false`.
 
-              If `true` disables service HTTP traffic on port 80, and therefore forcing use of HTTPS. 
+              If `true` disables service HTTP traffic on port 80, and therefore forcing use of HTTPS.
           extraPorts:
             type: list
             description: |
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -57,12 +57,14 @@ spec:
       nodePort: {{ . }}
       {{- end }}
     {{- end }}
+    {{- if eq .Values.proxy.service.disableHttp true }}
     - name: http
       port: 80
       targetPort: http
       {{- with .Values.proxy.service.nodePorts.http }}
       nodePort: {{ . }}
       {{- end }}
+    {{- end }}
     {{- with .Values.proxy.service.extraPorts }}
     {{- . | toYaml | trimSuffix "\n" | nindent 4 }}
     {{- end }}
--- a/jupyterhub/templates/proxy/service.yaml
+++ b/jupyterhub/templates/proxy/service.yaml
@@ -57,7 +57,7 @@ spec:
       nodePort: {{ . }}
       {{- end }}
     {{- end }}
-    {{- if eq .Values.proxy.service.disableHttp true }}
+    {{- if ne .Values.proxy.service.disableHttpPort true }}
     - name: http
       port: 80
       targetPort: http
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -175,7 +175,7 @@ proxy:
     nodePorts:
       http:
       https:
-    disableHttp: false
+    disableHttpPort: false
     extraPorts: []
     loadBalancerIP:
     loadBalancerSourceRanges: []
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -175,6 +175,7 @@ proxy:
     nodePorts:
       http:
       https:
+    disableHttp: false
     extraPorts: []
     loadBalancerIP:
     loadBalancerSourceRanges: []


tkislan:
Is it possible that the build fails are cause because of the fork PR build?
Because I've only made backward compatible changes, so it shouldn't affect the default behaviour

consideRatio:
I'm preliminary positive to this change.

Regarding errors: the logic is currently inversed. If disable, then it enables.

- Can you make the config name more explicit? Such as disableHttpPort?
- Can you also add an entry about it in schema.yaml?

tkislan:
@consideRatio Ah sorry my bad .. what a typo
I've fixed it, and also renamed and added schema description

consideRatio:
Can you elaborate why this is actually helping you practically?

It is relevant to know that we expose both port 443 and 80 when using `proxy.service.type=offload` only as a workaround for inflexible cloud services that can handle TLS termination (more discussion about this in #1811). The silly thing is that traffic to both port 443 and 80 on the proxy-public service when using `proxy.service.type=offload` will go to the `http` port of the proxy pod.

I think perhaps what you want, is to not use `proxy.service.type=offload` which is actually not doing anything but applying this workaround, and then use TLS termination which send traffic to the proxy-public http port. Hmmm, yeah I suspect this is a workaround to a workaround and you should not use `proxy.service.type=offload`.

If this is correct, I'd like to instead improve documentation rather than add this config option.

tkislan:
Can you elaborate more on "infexible cloud services"?

LoadBalancer Service creates cloud provider managed load balancer (AWS Classic load balancer in our case), which is configured terminate TLS on port 443, and forward upstream to `proxy` pod which I guess then just forwards it to `hub` pod.
Which is fine .. but using port 80 for the service also makes the load balancer listen on port 80, without TLS, and forward that as well .. which just creates a backdoor.

All of this was discussed in the linked PR. Cloud provider handles TLS termination on port 443, and forwards decrypted traffic  to upstream on any port (be it 80, 8000, or whatever)
And this works, because in offload case, the configurations says so as well:
`targetPort: http`
In this case, upstream proxy shouldn't even listen on port 443, because nothing will send any traffic to that port

The change that I'm proposing here, just disables 80 port listener for the load balancer, no not allow unencrypted traffic

I definitely want to use `proxy.service.type=offload`, because it does just that .. it offloads the TLS termination to a managed service. Why would I want jupyterhub handling TLS termination?

I actually have this deployed right now, with the `port: 80` part removed, and everything works .. it just doesn't redirect http traffic to https, but it's better than having the port open and sending unencrypted traffic right upstream ..

I still don't understand what the "internal" problem is, because I don't have enough insights into Jupyterhub Kubernetes setup, but my guess is, that most of this confusion comes from trying to handle TLS termination with some custom components ...

consideRatio:
__Correction__
I wrote `proxy.service.type=offload` but meant `proxy.https.type=offload`.

__Current situation__
There are _two ways_ to configure this Helm chart to not let it be involved with TLS termination (offload it).

1. `proxy.https.enabled=true` and `proxy.https.type=offload` -> port 80 and 443 are open.
2. `proxy.https.enabled=false` -> only port 80 open.

__This PR__
A third way is proposed.

3. `proxy.https.enabled=true`, `proxy.https.type=offload`, and `proxy.service.disableHttpPort=true` -> port 443 open.

__My goal__

To address the needs of users of this chart while not increasing complexity unless needed. In practice this would mean to try improve documentation, clarify existing options, and provide recommendations etc, before adding more configuration options that further increase the complexity for users to understand how to configure this chart.

@tkislan can you help me understand if you need option 3 over option 2, and if you do also clarify why?

I'm sorry for holding back on pressing merge, but it is a real challenge to keep up with documentation and followup questions unless things are understood very well.

tkislan:
Yes, I understand the 3 options, and the summary is correct

Option 2 doesn't handle https, so if you want to have jupyterhub publicly available, this is just unacceptable, because it creates public unencrypted endpoint.

Option 1 creates load balancer with TLS termination, but also leaves the 80 port open, and still allows unencrypted traffic .. hence the backdoor I'm mentioning

Option 3 does exactly the expected thing, it provides security layer over the Jupyterhub deployment .. even though it

But I agree that configuration values to achieve option 3 doesn't look nice, but I wanted to be backward compatible, and wanted this to be just a small change.

Just to provide more context .. I just followed the documentation to the point here:
https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#off-loading-ssl-to-a-load-balancer
But it is just wrong setup, if it leaves the unencrypted port open

consideRatio:
> Option 2 doesn't handle https, so if you want to have jupyterhub publicly available, this is just unacceptable, because it creates public unencrypted endpoint.

It does not, but something in front of it can do that. But, I think the key thing to understand that option 1 and 3 doesn't either.

> Option 1 creates load balancer with TLS termination, but also leaves the 80 port open, and still allows unencrypted traffic .. hence the backdoor I'm mentioning

There is no TLS termination here, just also opening port 443 and redirecting it just like port 80 to the same pod.

> Option 3 does exactly the expected thing, it provides security layer over the Jupyterhub deployment .. even though it

This option is exactly like Option 2, except the port is 443 exposed instead of port 80.

---

Thanks for linking to the relevant documentation section that have been your reference. I would say it is misleading as it doesn't mention option 2.

In my mind, option 2 should be the recommended solution in the documentation, and option 1 should be a workaround to use if the TLS termination proxy of choice isn't flexible enough to send traffic to port 80 on the k8s Service after it has performed the TLS termination.

Are you in a situation where you cannot configure your TLS termination proxy of choice to proxy traffic onwards to port 80?

tkislan:
As per linked documentation, using the AWS annotations for LoadBalancer, it creates AWS Classic Load Balancer, that has 2 listeners.
first listens on port 80, without TLS termination
second listens on port 443, with TLS termination
443 listeners forwards to port 80 of `proxy` pod

I don't understand where the confusion is coming from, that the proxy deployment needs to be running on port 443, for AWS load balancer to be able to forward from 443 port

image

image

As you can see, LoadBalancer correctly points to the service, which maps to `http` port of proxy pod, not `https`.

So it's not true, that there is no TLS termination if you explicitly ask for it.

But to circle back to the top, I don't want load balancer listener on port 80, because that just bypasses the security ...

I also don't understand, what workarounds you are talking about ... the way I'm describing is that Load Balancer correctly terminates the TLS, with certificate provided through AWS, and routes directly to the k8s pod

> send traffic to port 80 on the k8s Service
Service doesn't accept any traffic .. Service to put it simple, just describes a networking .. it doesn't represent any deployment
in this case Service IS the AWS Classic Load balancer

so it's not:
AWS Load balancer -> Service -> Proxy pod
but:
AWS Load balancer (Service) -> Proxy pod

jupyterhub/schema.yaml
```diff
@@ -852,6 +852,12 @@ properties:
                 type: integer
                 description: |
                   The HTTPS port the proxy-public service should be exposed on.
+          disableHttpPort:
+            type: boolean
+            description: |
+              Default `false`.
+
+              If `true` disables service HTTP traffic on port 80, and therefore forcing use of HTTPS.
```
consideRatio, line 860:
```suggestion
              If `true`, port 80 for incoming HTTP traffic will no longer be exposed. This should not be used with `proxy.https.type=letsencrypt` or `proxy.https.enabled=false` as it would remove the only exposed port.
```

consideRatio:
We are in agreement about a lot, but I lack knowledge about the constraints on TLS termination configuration set through annotations. I also fail to find good documentation about these annotations.

If you say that you are required to be able to expose a 443 port on the k8s Service that points to the 80 port, and you don't want the port 80 port at all and think the k8s Service resource must be influenced rather than annotations. It would be what I have been looking to learn.

I also don't know if the AWS LB could help with redirection of http:// traffic on some port to https:// on another port, and if that exists, if that also requires something from the k8s Service apart from the annotations, such as having or not having port 80 exposed. That is discussed in https://aws.amazon.com/premiumsupport/knowledge-center/elb-redirect-http-to-https-using-alb/.

While this change is technically harmless to add, I've been hesitant because of the wish to find a good solution. Adding more and more options lead to more and more choices, and my goal has been to try identify a good choice given the constraints. This has been hard because of unknown constraints.

I'd like to avoid investing our time further into this question though, so I'm fine with merging and moving on after the code suggestion is committed.

jupyterhub/schema.yaml
tkislan: > This should not be used with `proxy.https.type=letsencrypt`
I don't think this is true
Even though you would use letsencrypt and proxy to handle TLS termination, you still want public Load balancer to listen on port 443 (not 80) .. because that's the port the browser hits when you prefix it with `https://`
If you use letsencrypt, you just don't add SSL annotations on the Service, so that it just forwards the traffic, without decrypting it.

The only case when you don't want to disable port 80 is, when either you don't want to use HTTPS at all, or when the upstream proxy is correctly handling redirection from http to https (but disabling port on the load balancer, not the proxy itself .. proxy always listens on only one port, if it doesn't handle TLS termination)
https://aws.amazon.com/premiumsupport/knowledge-center/redirect-http-https-elb/

tkislan:
What you linked is relevant to Application Load balancer, but this helm chart opts for using older Classic load balancer, that doesn't handle http to https redirection

See this article
https://aws.amazon.com/premiumsupport/knowledge-center/redirect-http-https-elb/

I've elaborated on this more in the above comment

jupyterhub/schema.yaml
consideRatio: If you configure `proxy.https.type=letsencrypt` (assuming also `proxy.https.enabled`), then I'm very confident you are required to have port 80 exposed. If you don't, then the ACME procedure using a HTTP01 challenge to acquire a new certificate that we have setup as part of this Helm chart would fail.

tkislan: Ah you are right. Forgot about the http challenge
But if you leave the 80 port open .. does it also forward traffic to the jupyterhub application? because if so, in this case it should handling redirect to https

consideRatio:
@tkislan the point about cookies is a point I'd like to investigate separately.

> , and even reuses the cookies from https/443 endpoint.

I don't understand that clearly, but I see a security issue if sensitive cookies are set without a [secure / httponly](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies) attribute. Have you observed that? If so, could you help by describing the Helm chart version used and what was the name of those cookies?

tkislan:
cookie names are: `jupyterhub-hub-login` and `jupyterhub-session-id`
you are correct, that they are not marked as secure, so they are reused for http as well .. but that is just a part of the problem

Even if they were secure, and not reused for http, I do not want the users to even be able to open `http://` page, and login again .. because that's exactly what happens, if you enter domain name into the browser, without explicitly typing `https://` prefix

The main issue is, that when using TLS termination at LoadBalancer, there is no redirection happening from http to https
I'm not sure if this is the issue with letsencrypt configuration as well, but my guess would be that it is

consideRatio:
Thanks for the cookie insights, I'm raising that for discussion with JupyterHub developers.

> The main issue is, that when using TLS termination at LoadBalancer, there is no redirection happening from http to https
> I'm not sure if this is the issue with letsencrypt configuration as well, but my guess would be that it is

If you are using TLS termination at LoadBalancer, the Helm chart isn't aware if the loadbalancer is using LetsEncrypt or not. The `proxy.https.type=letsencrypt` implies that the Helm chart will setup a dedicated k8s Pod (named autohttps), running the software Traefik. Traefik in turn make use of LEGO to acquire TLS certificates from LetsEncrypt using a HTTP01 challenge.

With letsencrypt activated, I'm quite confident we are redirecting properly. Then traffic will go to Traefik (autohttps pod) and then to the proxy pod, but if the traffic was a HTTP request that didn't relate to an ACME challenge, it will not forward to the proxy pod but instead be redirected to HTTPS.

The PR #1811 tried to accomplish something similar implemented at the proxy pod instead of at the autohttps pod which is only started if `proxy.https.type=letsencrypt`

jupyterhub/schema.yaml
consideRatio: There will be a redirect from the `autohttps` pod unless the HTTP traffic was associated with an ACME challenge.

This setup is very complicated in general, and makes things hard to understand. The proxy-public k8s Service will redirect to the autohttps pod which will redirect to a k8s Service named `proxy-http` after TLS termination. So the same k8s Service can go to different pods depending on the configuration of the Helm chart. Either it goes directly to the `proxy` pod, or it goes through the `autohttps` pod first which in turn will redirect non ACME http requests to HTTPS.

consideRatio:
Thank you for iterating with me on this @tkislan! It is one of the parts of the Helm chart that is complicated to maintain because of its complexity caused by coupling to third party services.

Let's merge this, I don't see a better option on the horizon!

Issue #653, opened 2018-04-17
warning: cannot overwrite table with non table for extraConfig (map[])


jgerardsimcock:
Every time I `helm upgrade {deployment} {helm_chart} -f config.yml` I get this error
`warning: cannot overwrite table with non table for extraConfig (map[])` 

I am not sure what it means.

## Contents of `config.yaml`
```
# Rhodium-Hub configuration values
# --------------------


# Create and use roles and service accounts on an RBAC enabled cluster.
rbac:
  enabled: true


# Dependency configuration values
# -------------------------------
# To configure dependencies you must create a key with the name of the dependency
# and then configure as per the dependent chart's values.yaml. Don't forget to indent!

jupyterhub:
  singleuser:
    image:
      name: rhodium/notebook
      tag: 2018-04-17
    storage:
      capacity: 10Gi
    dynamic:
      storageClass: jupyterhub-user-ssd
    serviceAccountName: daskkubernetes
    cmd: ['start-singleuser.sh']
    defaultUrl: "/lab"
    storage:
      extraVolumes:
        - name: fuse
          hostPath:
            path: /dev/fuse
      extraVolumeMounts:
        - name: fuse
          mountPath: /dev/fuse
    cloudMetadata:
      enabled: true

  hub:
    extraConfig: |
      from kubernetes import client
      def modify_pod_hook(spawner, pod):
          pod.spec.containers[0].security_context = client.V1SecurityContext(
              privileged=True,
              capabilities=client.V1Capabilities(
                  add=['SYS_ADMIN']
              )
          )
          return pod
      c.KubeSpawner.modify_pod_hook = modify_pod_hook
      c.Authenticator.whitelist = {REDACTED}

    image:
      name: jupyterhub/k8s-hub
      tag: v0.6     
    
  cull:
    enabled: true
    users: false
    timeout: 12
```

rabernat:
We are getting this warning constantly in pangeo jupyterhub deployments, and we are worried it is the source of deeper problems with our config. Any update?

mrocklin:
It looks like the extraConfig entry in the helm chart changed recently:

```
mrocklin@carbon:~/workspace/pangeo/gce$ helm inspect jupyterhub/jupyterhub | grep extraConfig:
  extraConfig: ''
mrocklin@carbon:~/workspace/pangeo/gce$ helm inspect jupyterhub/jupyterhub --version v0.7-9c93c00 | grep extraConfig:
[...]

PR #2062, opened 2021-02-23
Fix spawner env injection example.

danielballan:
I think there are three separate problems with the documented example for injecting an environment variable into the user container environment.

```yaml
hub:
  extraConfig: |
    import time
    c.Spawner.environment += {
       "CURRENT_TIME": str(time.time())
    }
```

1. The documentation claims that `extraConfig` accepts either a dict of strings or a single string (of Python code), but it actually only accepts the first one. If you give it a single string, following the documented example, it warns
    ```
    coalesce.go:196: warning: cannot overwrite table with non table for extraConfig (map[])
    ```
    and it does not take effect. See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/653 which seems to observe the same issue.
2. The documentation uses `+=` to update the `Spawner.environment` configurable, but that is not a legal operation on a dict or on the corresponding Configurable object. The logs:
    ```
    Loading /etc/jupyterhub/secret/values.yaml
    Loading extra config: 00-spawner-env-config
    [E 2021-02-23 18:07:58.648 JupyterHub app:2859]
        Traceback (most recent call last):
          File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2856, in launch_instance_async
            await self.initialize(argv)
          File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2347, in initialize
            self.load_config_file(self.config_file)
          File "/usr/local/lib/python3.8/dist-packages/traitlets/config/application.py", line 87, in inner
            return method(app, *args, **kwargs)
          File "/usr/local/lib/python3.8/dist-packages/traitlets/config/application.py", line 775, in load_config_file
            for (config, filename) in self._load_config_files(filename, path=path, log=self.log,
          File "/usr/local/lib/python3.8/dist-packages/traitlets/config/application.py", line 737, in _load_config_files
            config = loader.load_config()
          File "/usr/local/lib/python3.8/dist-packages/traitlets/config/loader.py", line 616, in load_config
            self._read_file_as_dict()
          File "/usr/local/lib/python3.8/dist-packages/traitlets/config/loader.py", line 648, in _read_file_as_dict
            exec(compile(f.read(), conf_filename, 'exec'), namespace, namespace)
          File "/etc/jupyterhub/jupyterhub_config.py", line 412, in <module>
            exec(config_py)
          File "<string>", line 3, in <module>
        TypeError: unsupported operand type(s) for +=: 'LazyConfigValue' and 'dict'
    ```
3. Configuring `c.Spawner.environment` has no effect on the environment. Instead, `c.KubeSpawner.environment` must be used.

I have confirmed that the revised example works as expected with Helm Chart version 0.11.1.

### DIFF
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -105,9 +105,23 @@ Code. Some examples of things you can do:
 Unfortunately, you have to write your python _in_ your YAML file. There's no way
 to include a file in `config.yaml`.
 
-You must specify `hub.extraConfig` as a dictionary. This enables you to
-logically split your customizations when there is more than one. The code will
-be evaluated in alphabetical sorted order of the key.
+You can specify `hub.extraConfig` as a raw string (remember to use the `|` for multi-line
+YAML strings):
+
+```yaml
+hub:
+  extraConfig: |
+    import time
+    c.KubeSpawner.environment.update(
+        {
+            "CURRENT_TIME": str(time.time())
+        }
+    )
+```
+
+You can also specify `hub.extraConfig` as a dictionary, if you want to logically
+split your customizations. The code will be evaluated in alphabetical sorted
+order of the key.
 
 ```yaml
 hub:
@@ -118,6 +132,7 @@ hub:
           {
               "CURRENT_TIME": str(time.time())
           }
+      # some code
     10-second-config: |
       # some other code
 ```
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -114,7 +114,7 @@ hub:
   extraConfig:
     00-first-config: |
       import time
-      c.Spawner.environment.update(
+      c.KubeSpawner.environment.update(
           {
               "CURRENT_TIME": str(time.time())
           }
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -132,7 +132,7 @@ hub:
           {
               "CURRENT_TIME": str(time.time())
           }
-      # some code
+      )
     10-second-config: |
       # some other code
 ```
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -105,27 +105,19 @@ Code. Some examples of things you can do:
 Unfortunately, you have to write your python _in_ your YAML file. There's no way
 to include a file in `config.yaml`.
 
-You can specify `hub.extraConfig` as a raw string (remember to use the `|` for multi-line
-YAML strings):
-
-```yaml
-hub:
-  extraConfig: |
-    import time
-    c.Spawner.environment += {
-       "CURRENT_TIME": str(time.time())
-    }
-```
-
-You can also specify `hub.extraConfig` as a dictionary, if you want to logically
-split your customizations. The code will be evaluated in alphabetical sorted
-order of the key.
+You must specify `hub.extraConfig` as a dictionary. This enables you to
+logically split your customizations when there is more than one. The code will
+be evaluated in alphabetical sorted order of the key.
 
 ```yaml
 hub:
   extraConfig:
     00-first-config: |
-      # some code
+      import time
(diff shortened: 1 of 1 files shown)


manics:
Thanks for pointing this out. The `c.Spawner.environment` problem is surprising as KubeSpawner inherits the variable from the parent Spawner class. We should investigate this in case therer's a bug somewhere....

Would you mind sharing your full Z2JH config with secrets redacted so we can try and track down the bug(s)?

danielballan:
```yaml
proxy:
  secretToken: "REDACTED"
  https:
    enabled: true
    hosts:
      - aimm-jupyter.blueskyproject.io
    letsencrypt:
      contactEmail: REDACTED
hub:
  config:
    GlobusOAuthenticator:
      client_id: REDACTED
      client_secret: REDACTED
      oauth_callback_url: https://aimm-jupyter.blueskyproject.io/hub/oauth_callback
      identity_provider: bnl.gov
    JupyterHub:
      authenticator_class: globus
  extraConfig:
    00-spawner-env-config: |
      # This string will be accessible to any logged-in user.
      c.KubeSpawner.environment = {"MONGODB_URI": "REDACTED"}
      print(c.KubeSpawner.environment)
```

consideRatio:
### Point 1 - Chart config value override with another type

I branched off point 1 to #2063, so let's focus only on point 2 and point 3 here.

### Point 2 - Use of `+=` on config option Spawner.environment

```
    TypeError: unsupported operand type(s) for +=: 'LazyConfigValue' and 'dict'
```

This is just wrong. Spawner.environment is a dict, and `+=` doesn't work, but I think `c.Spawner.environment.update({"TEST_ENV": "sure"})` would.

### Point 3 - Spawner.environment vs KubeSpawner.environment

> Configuring c.Spawner.environment has no effect on the environment. Instead, c.KubeSpawner.environment must be used.

I'm also confused why. Related code is:
- [Spawner.get_env](https://github.com/jupyterhub/jupyterhub/blob/c3ca924ba896bf2da40cab29e2a08784850a945b/jupyterhub/spawner.py#L732)
- [KubeSpawner.get_env()](https://github.com/jupyterhub/kubespawner/blob/a16e400d69b2481863f3c6452273abbedb74f76a/kubespawner/spawner.py#L1897)
- [Spawner.environment](https://github.com/jupyterhub/jupyterhub/blob/c3ca924ba896bf2da40cab29e2a08784850a945b/jupyterhub/spawner.py#L459)
-

consideRatio:
@danielballan this PR is addressing point 1-3, but I'm not willing to address point 1 without further investigation. I'd be happy to approve this PR for the example change from c.Spawner to c.KubeSpawner and from `+=` to `.update()` though.

doc/source/administrator/advanced.md
```diff
@@ -125,6 +127,11 @@ order of the key.
 hub:
   extraConfig:
     00-first-config: |
+      import time
+      c.KubeSpawner.environment.update(
+          {
+              "CURRENT_TIME": str(time.time())
+          }
       # some code
```
consideRatio, line 135:
```suggestion
          }
      )
```

danielballan:
Thanks for the syntax fix.

I too am confused by Point (3). I'm sure this kind of configuration merging has worked for me in other contexts, and [the docs are really clear on this exact point](https://traitlets.readthedocs.io/en/stable/config.html#class-based-configuration-inheritance).

I just checked it again to be sure I wasn't getting confused. It is definitely the case that setting the config `Spawner` instead of `KubeSpawner` does not set the environment variable.

consideRatio:
@danielballan could you open an issue in https://github.com/jupyterhub/kubespawner/issues/new? You could reference point 3 in this issue from there, but it would be good to have you describe there also what version of the Helm chart you've used.

consideRatio:
Thank you for your work investigating, reporting, and contributing with a change @danielballan!! :heart: :tada:

PR #2066, opened 2021-02-25
Add prePuller.hook.pullOnlyOnChanges flag

### LINKED ISSUE #1814: Run hook-image-puller only if an image has changed
If we start to assume that users have Helm 3.1 or perhaps Helm 3.2, then we can make use of [the `lookup` function](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/). This function allows us to inspect the currently installed resources. This in turn allows us to be smart and only run the `hook-image-puller` logic conditionally if something has changed.

In practice I think we would do the following to accomplish this.

1. We create a new ConfigMap resource that contains things we may want to lookup, so it becomes easy to find that specific information. We could name the ConfigMap `latest-state` or similarly, and we could render a hash of an augmented list of images that we pull into a key there.
2. We only render the `hook-image-awaiter` Job resource in templates/image-puller/job.yaml and the hook DaemonSet if we evaluate this:

   ```yaml
   {{- $latest_state := lookup "v1" "ConfigMap" .Release.Namespace "latest-state" }}
   {{- $image_hash := $latest_state.data.prePullerImagesHash }}
   {{- if $image_hash is not equal to the new hash }}
   # render image-awaiter Job resource
   {{- end }}
   ```

## Action points
- [x] Decide if it's worth adding this complexity to reduce the delay caused by this step during `helm upgrade`, or it's just better for this project to skip it.
- [x] Develop and triel this experimentally
- [x] Decide when to require Helm 3 of a version that has the lookup function

## Related
- Another case to use the Helm3 `lookup` function: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1017

consideRatio:
This PR renders adds a flag to not startup the hook-image-puller DaemonSet and hook-image-awaiter Job unless a checksum of the previous hook-image-daemonset differs from the new. This is made possible by helm3's `lookup` function.

This PR should not cause any difference for anyone except those that explicitly enables the flag.

Closes #1814

### Implementation notes

- It made sense to split apart the daemonset.yaml file with both hook-image-puller and continuous-image-puller daemonset to avoid complexity of putting a yaml separator (`---`) between the two only if both daemonset are rendered.


consideRatio:
Intermittent test failures caused by https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1962...

yuvipanda:
I haven't looked at the patch itself, but this would be an awesome feature! \o/

jupyterhub/schema.yaml
```diff
@@ -2077,6 +2077,12 @@ properties:
         properties:
           enabled:
             type: boolean
+          pullOnlyOnChanges:
+            type: boolean
+            description: |
+              Pull only if  changes have been made to the images to pull, or
```
consideRatio, line 2083:
```suggestion
              Pull only if changes have been made to the images to pull, or
```

consideRatio (commented):
I made a self-review of this a week later, and it LGTM :)

minrk (approved):

minrk:
Looks great! I'm a little concerned about our growing number and complexity of helpers, but that should be a separate discussion.

consideRatio:
Thanks for the review/merge @minrk!

I'm doing my best to avoid the introduction of helpers. There are many challanges to working with helm templates that I often consider myself forced to make significant trade offs either way though, where the introduction of a helper often follow.

minrk:
Yes! I totally understand. It's often a tough trade-off. I think it's just something to revisit once in a while to see how we are doing.

yuvipanda:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2099 is possibly related

yuvipanda:
Also, if the daemonset isn't getting created, the awaiter should also not be. Otherwise, it'll fail like in #2099

PR #2077, opened 2021-03-04
Don't restart the proxy pod with each deploy

yuvipanda:
Restarting the proxy pod causes current user sessions to
drop until the pod restarts, the hub notices it, and sets up
all the routes again. On a hub with a few thousand users, this
can be minutes. Instead, we provide the proxy pod with just the
one key it needs, so it'll restart only if that changes.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -24,8 +24,9 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the secret changes!
-        checksum/hub-secret: {{ include (print $.Template.BasePath "/hub/secret.yaml") . | sha256sum }}
+        # This restarts the pod when the proxy secret changes.
+        # The proxy secret is *only* present when doing manual HTTPS
+        # certs, so restarting is required to get those changes.
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
         {{- with .Values.proxy.annotations }}
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
@@ -106,14 +107,13 @@ spec:
             {{- . | toYaml | trimSuffix "\n" | nindent 12 }}
           {{- end }}
           env:
+            # Restarting the proxy pod causes a temporary outage,
+            # so we should do it as few times as possible. Including
+            # the auth token here directly means the pod only restarts
+            # if the auth token changes, rather than when the hub-secret
+            # changes, since the latter happens all the time.
             - name: CONFIGPROXY_AUTH_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  # NOTE: References the chart managed k8s Secret even if
-                  #       hub.existingSecret is specified to avoid using the
-                  #       lookup function on the user managed k8s Secret.
-                  name: {{ include "jupyterhub.hub.fullname" . }}
-                  key: hub.config.ConfigurableHTTPProxy.auth_token
+              value: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | quote }}
             {{- include "jupyterhub.extraEnv" .Values.proxy.chp.extraEnv | nindent 12 }}
           {{- with .Values.proxy.chp.image.pullPolicy }}
           imagePullPolicy: {{ . }}
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -26,8 +26,8 @@ spec:
       annotations:
         # We want to restart proxy only if the auth token changes
         # Other changes to the hub config should not restart.
-+       # We truncate to 4 chars to avoid leaking auth token info,
-+       # since someone could brute force the hash to obtain the token
+        # We truncate to 4 chars to avoid leaking auth token info,
+        # since someone could brute force the hash to obtain the token
         checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
         {{- with .Values.proxy.annotations }}
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -25,8 +25,10 @@ spec:
         {{- end }}
       annotations:
         # We want to restart proxy only if the auth token changes
-        # Other changes to the hub config should not restart
-        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum }}
+        # Other changes to the hub config should not restart.
++       # We truncate to 4 chars to avoid leaking auth token info,
++       # since someone could brute force the hash to obtain the token
+        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
         {{- with .Values.proxy.annotations }}
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -24,8 +24,9 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the secret changes!
-        checksum/hub-secret: {{ include (print $.Template.BasePath "/hub/secret.yaml") . | sha256sum }}
+        # We want to restart proxy only if the auth token changes
+        # Other changes to the hub config should not restart
+        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum }}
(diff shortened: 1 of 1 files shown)


jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -104,11 +105,13 @@ spec:
...
+            # so we should do it as few times as possible. Including
+            # the auth token here directly means the pod only restarts
+            # if the auth token changes, rather than when the hub-secret
+            # changes, since the latter happens all the time.
             - name: CONFIGPROXY_AUTH_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  name: {{ include "jupyterhub.hub-secret.fullname" . }}
-                  key: JupyterHub.proxy_auth_token
+              value: {{ include "jupyterhub.config.JupyterHub.proxy_auth_token" . | quote }}
```
minrk, line 114: 👍 for reasoning, requesting review from @consideRatio to check that the value will still be right in all cases here, or if more logic is needed.

Or if we should duplicate this value to the proxy-secret as well so it's not directly in the pod manifest.

yuvipanda: Especially with the lookup logic for autogenerating them - I don't know if it'll get generated twice?

consideRatio:
I support this change idea but want to make sure we do a proper think to minimize unwanted consequences.

1. The `CONFIGPROXY_AUTH_TOKEN` should not be set explicitly from the rendered k8s Deployment spec but instead reference a k8s secret for security reasons.
2. The restarts can still be triggered by just including the relevant information in a annotation with a hashed value of what's relevant to react to.
3. Do we have a good enough overview on what the proxy should restart on that we avoid failing to restarts on relevant changes?
   - :+1: I don't think the hub relies on it to have certain state reset along with itself.
   - :+1: `Values.hub` isn't referenced in any templates associated with the proxy deployment.
   - :thinking: Changes to the hub k8s Secret previously triggered `Values.proxy` to restart on changes to .Values.proxy, but now that won't happen. I don't think so. :+1:
   - :thinking: Anything else?
   
Assuming 3 is fine, I suggest we simply replace the hub secret checksum with a checksum, such as...

```yaml
# checksum length truncated to 4 hex characters (65k permuatations)
checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
```

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -24,8 +24,9 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the secret changes!
-        checksum/hub-secret: {{ include (print $.Template.BasePath "/hub/secret.yaml") . | sha256sum }}
+        # This restarts the pod when the proxy secret changes.
+        # The proxy secret is *only* present when doing manual HTTPS
+        # certs, so restarting is required to get those changes.
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
```
consideRatio, line 30: I'm not sure about this comment, especially `so restarting is required to get those changes.`

Changes to `proxy.https.type=manual` and `proxy.https.manual.cert|key` will trigger a restart through this checksum by rendering the proxy-secret.

What do you think about removing the comment and updating the checksum annotation name to `checksum/manual-https-certs` or similar?

minrk: Since we want the value to come from a secret, is it sensible/reliable to copy the value to the proxy secret given the current existing-secret / lookup / random-if-not-set behavior?

If so, then:

1. the env value comes from a secret (required, I think)
2. proxy-secret always exists
3. proxy-secret hash always makes sense, following existing practice of 'any change here means it should restart' like we do for the hub

jupyterhub/templates/proxy/deployment.yaml
consideRatio: Hmmm, it will be a transient issue, causing one unwanted restart.

It will generate the same value as long as its not doing the first installation. So when the upgrade following a seeded installation happen it will align with what is within the hub k8s secret but then remain stable.

jupyterhub/templates/proxy/deployment.yaml
consideRatio: I don't see the mechanism to help us copy the value reliably.

If we actually render secret.yaml multiple times, like we actually do when setting the checksum annotation, we end up with multiple random seeds anyhow. The crux is that we don't have a way to memoize the random output which we want to get from multiple Helm templates during a chart rendering.

yuvipanda:
Ah, great suggestion @consideratio! I've modified this to only restart if that particular key changes. I didn't truncate it to 4 chars tho, to match the other checksums.

jupyterhub/templates/proxy/deployment.yaml
consideRatio: If we can memoize the random output, we can avoid any unwanted restarts.

If we would duplicate the CHP auth token to the proxy k8s Secret otherwise only used for manually provided http certs, we would still be forced to use the same Helm template helper which relies on the: explicit config, the hub k8s Secret value, or a separate random seed because we haven't memoized the helper.

My suggestion:
1. We accept _one_ unwanted restart following the initial CHP auth token generation. 
   Note that this is caused by syncing with whatever was generated separately from either the proxy k8s Secret or the proxy k8s Deployment annotation with what was generated for the hub k8s Secret.
2. We use an checksum annotation instead of creating a k8s Secret for the proxy pod. If we would, we would _need to not mount it_, as then we would have a mismatch when the password is generated causing communication failure.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -24,9 +24,9 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
-        # This restarts the pod when the proxy secret changes.
-        # The proxy secret is *only* present when doing manual HTTPS
-        # certs, so restarting is required to get those changes.
+        # We want to restart proxy only if the auth token changes
+        # Other changes to the hub config should not restart
+        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum }}
```
consideRatio, line 29: If someone has a random 8 character long token, the sha256sum could be recognized by a hash dictionary - that is why I wanted to trunc it to 4 characters (65k permutations still).

consideRatio: I don't see this as a risk for the provided length cert, key and misc YAML hashed in the proxy-secret checksum though.

yuvipanda: @consideRatio hmm, can you tell me what you mean by 'recognized by a hash dictionary'? I'm slightly confused :)

consideRatio: By using brute force you can hash all combinations of for characters up to X characters. Then, my concern is that such has would be recognizable as the hash within the pod spec and be used to lookup the actual auth_token.

consideRatio: Assuming the password was less than X in length, so not 32 characters or such but for example 10

yuvipanda: @consideRatio aaaaaaaaaaaaaah, that's very cool and clever - thank you. I'll add the trunc with an explanation.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -24,8 +24,11 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the secret changes!
-        checksum/hub-secret: {{ include (print $.Template.BasePath "/hub/secret.yaml") . | sha256sum }}
+        # We want to restart proxy only if the auth token changes
+        # Other changes to the hub config should not restart.
++       # We truncate to 4 chars to avoid leaking auth token info,
++       # since someone could brute force the hash to obtain the token
+        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
```
consideRatio, line 31:
```suggestion
        # We want to restart proxy only if the auth token changes
        # Other changes to the hub config should not restart.
        # We truncate to 4 chars to avoid leaking auth token info,
        # since someone could brute force the hash to obtain the token
        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
```

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -24,8 +24,11 @@ spec:
...
         {{- end }}
       annotations:
-        # This lets us autorestart when the secret changes!
-        checksum/hub-secret: {{ include (print $.Template.BasePath "/hub/secret.yaml") . | sha256sum }}
+        # We want to restart proxy only if the auth token changes
+        # Other changes to the hub config should not restart.
+        # We truncate to 4 chars to avoid leaking auth token info,
+        # since someone could brute force the hash to obtain the token
+        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 }}
         checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum }}
```
consideRatio, line 32:
```suggestion
        checksum/auth-token: {{ include "jupyterhub.hub.config.ConfigurableHTTPProxy.auth_token" . | sha256sum | trunc 4 | quote }}
        checksum/proxy-secret: {{ include (print $.Template.BasePath "/proxy/secret.yaml") . | sha256sum | quote }}
```

consideRatio (approved):
I fixed a syntax error where `+` signs had been introduced, and added `| quote` to avoid ending up with numerical YAML entries.

I'll go for a merge!

yuvipanda:
AWESOME, THANK YOU, @consideRatio!

consideRatio:
Thank you @yuvipanda for being such a positive person and this PR =)

PR #2079, opened 2021-03-04
allow override of CHP defaultTarget, errorTarget

minrk:
mybinder.org-deploy used to set the whole CHP command, but this is no longer allowed (not quite sure how I feel about that - pros and cons), making it impossible to override these values.

- defaultTarget ~never makes sense, but it felt weird to leave it out.
  It may make sense in the future if JupyterHub develops an option
  to register itself *only* as a handler for `/hub`.
  This hasn't come up yet, as that behavior only makes sense when
  another service is providing the default target
  because requests to `/user/foo` will not behave sensibly without something handling the default route.
- errorTarget can be useful in API-only deployments like BinderHub to reduce load on the Hub
  and/or show more friendly messages to users who shouldn't need to know they are using jupyterhub

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -67,13 +67,14 @@ spec:
         - name: chp
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
+          {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
           command:
             - configurable-http-proxy
             - "--ip=::"
             - "--api-ip=::"
             - --api-port=8001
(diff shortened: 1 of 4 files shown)


consideRatio:
There is a missing values.yaml entry, and proxy.chp.abcd is used in schema yaml (good IMO) but the deployment reference proxy.abcd (not good IMO).

minrk:
xref: https://github.com/jupyterhub/jupyterhub/pull/3373 adding the feature making defaultTarget override make sense

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -67,13 +67,14 @@ spec:
         - name: chp
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
+          {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
```
minrk, line 70: @consideRatio is there a reason not to use  {{$hubNameAsEnv}}_SERVICE_HOST since we use it for SERVICE_PORT?

consideRatio: Yepp, IPv6 compatebility

PR #2080, opened 2021-03-05
[DOC] Satisfy linkcheck

minrk:
- update a few links to follow redirects
- kubectl autocompletion no longer has a single link target, link to kubectl per-platform instructions instead
- use more consistent, standard doc/requirements.txt name rather than redundant doc/doc-requirements
- remove extra `--no-cache-dir` in pip install command, which isn't helpful on CI.


consideRatio (approved):
Wieee thanks!

PR #2086, opened 2021-03-08
upload chart as github artifact

### LINKED ISSUE #2043: ci: let PRs produce build artifacts of a packaged chart
Can we make a PR produce a packaged chart as a build artifact, which in turn someone can use to install chart by referencing with a web URL or by first downloading the file?

This can be helpful for some manual testing.

minrk:
- runs publish workflow for push and pull request
- moves credential-requiring conditions to individual steps instead of workflow as a whole
- on pull request, upload as artifact instead of publishing to the chart repository

This will only *fully* work if there are no changes to images in the PR, since image upload won't happen without credentials.

closes #2043

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,7 @@ name: Publish
 
 # Trigger the workflow's on pushed tags or commits to main/master branch.
 on:
+  pull_request:
   push:
     branches: ["main", "master"]
     tags: ["[0-9]+.[0-9]+.[0-9]+*"]
@@ -16,15 +17,15 @@ jobs:
   #
   # ref: https://github.com/jupyterhub/helm-chart
   # ref: https://hub.docker.com/orgs/jupyterhub
-  #
+
   publish:
-    if: github.repository == 'jupyterhub/zero-to-jupyterhub-k8s'
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
           # chartpress requires the full history
           fetch-depth: 0
+
       - uses: actions/setup-python@v2
         with:
           python-version: "3.8"
@@ -33,7 +34,7 @@ jobs:
         run: |
           . ./ci/common
           setup_helm
-          pip install --no-cache-dir chartpress pyyaml
+          pip install chartpress pyyaml
 
       - name: Setup push rights to jupyterhub/helm-chart
         # This was setup by...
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -85,6 +88,25 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        run: helm package jupyterhub
+
+      # ref: https://github.com/actions/upload-artifact
+      - uses: actions/upload-artifact@v2
+        if: github.event_name == 'pull_request'
+        with:
+          name: jupyterhub-${{ github.sha }}
+          path: "jupyterhub-*.tgz"
+          if-no-files-found: error
```
manics, line 112: Does this have to be restricted to `pull_request` only? If it was for everything anyone who's enabled CI on their forks could get artifiacts built for their own branches

minrk: Good point. I can make it exactly the inverse of the publish condition so any build that won't be published will get an artifact.

minrk: Done.

.github/workflows/publish.yml
```diff
@@ -85,6 +111,23 @@ jobs:
...
           # the jupyterhub/helm-chart repo.
           ./ci/publish
+
+      - name: Package helm chart as a CI artifact
+        if: steps.publishing.outputs.publishing == ''
+        run: helm package jupyterhub
+
+      # ref: https://github.com/actions/upload-artifact
+      - uses: actions/upload-artifact@v2
+        if: steps.publishing.outputs.publishing == ''
```
minrk, line 129: Weird possible bug in actions: cannot use simpler `if: !steps.publishing.outputs.publishing` which produces a syntax error, when I'm pretty sure it should be fine.

.github/workflows/publish.yml
```diff
@@ -85,6 +111,23 @@ jobs:
...
+        env:
+          GITHUB_REPOSITORY: "${{ github.repository }}"
+        run: |
           # Package the Helm chart and publish it to the gh-pages branch of
           # the jupyterhub/helm-chart repo.
           ./ci/publish
+
+      - name: Package helm chart as a CI artifact
+        if: steps.publishing.outputs.publishing == ''
+        run: helm package jupyterhub
```
minrk, line 125: Question here: this runs `helm package` and uses the resulting tgz as the artifact, which then gets re-archived as a .zip (actions doesn't support unmodified single-file artifacts, they are *always* zipped up).

Should we just make the `jupyterhub` directory itself the artifact without running `helm package`? I see that the result of `helm package` does rewrite Chart.yaml, but it appears to only be re-serializing the yaml (e.g. without comments) without any actual changes.

manics: My initial thoughts are that the closer the artifact is to the final published artifact the better, for example to cover the case if `helm package ...` has some weird behaviour, or does something different in future.

consideRatio (approved):
Nice LGTM!

consideRatio:
Thank you @minrk! :heart:

PR #2094, opened 2021-03-11
Add .Values.cull.ssl_enabled=true 

### LINKED ISSUE #2093: Support idle culling when TLS is in use for the hubhub only accepts TLS connections
### Bug description
When internal_ssl is enabled on the hub containers, culling is broken

#### Expected behaviour
Culler gets 2xx responses from hub

#### Actual behaviour
```
hub-f8d8b4c-tb7md hub [W 2021-03-11 16:58:23.702 JupyterHub iostream:1406] SSL Error on 10 ('10.40.3.130', 22208): [SSL: HTTP_REQUEST] http request (_ssl.c:1123)
hub-f8d8b4c-tb7md hub [W 2021-03-11 16:58:33.702 JupyterHub iostream:1406] SSL Error on 10 ('10.40.3.130', 22290): [SSL: HTTP_REQUEST] http request (_ssl.c:1123)
hub-f8d8b4c-tb7md hub [W 2021-03-11 16:58:43.702 JupyterHub iostream:1406] SSL Error on 10 ('10.40.3.130', 22446): [SSL: HTTP_REQUEST] http request (_ssl.c:1123)
```

### How to reproduce

Deploy hub containers with tls and enable the culler

### Your personal set up
z2jh@master, kops, aws, etc

</details>

- <details><summary>Logs</summary>

```
# paste relevant logs here, if any
```
</details>

thomasv314:
Fixes culling when the Hub container is served over https://

Fixes #2093

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -361,7 +361,9 @@ if get_config("cull.enabled", False):
         cull_proto = "https"
 
     base_url = c.JupyterHub.get("base_url", "/")
-    cull_cmd.append(f"--url={cull_proto}://localhost:8081" + url_path_join(base_url, "hub/api"))
+    cull_cmd.append(
+        f"--url={cull_proto}://localhost:8081" + url_path_join(base_url, "hub/api")
+    )
 
     c.JupyterHub.services.append(
         {
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -332,8 +332,6 @@ c.JupyterHub.services = []
 
 if get_config("cull.enabled", False):
     cull_cmd = ["python3", "-m", "jupyterhub_idle_culler"]
-    base_url = c.JupyterHub.get("base_url", "/")
-    cull_cmd.append("--url=http://localhost:8081" + url_path_join(base_url, "hub/api"))
 
     cull_timeout = get_config("cull.timeout")
     if cull_timeout:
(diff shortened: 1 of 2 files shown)


jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -357,6 +355,16 @@ def camelCaseify(s):
...
+    cull_proto = "http"
+    if get_config("cull.ssl_enabled", False):
+        cull_cmd.append("--ssl-enabled")
+        cull_proto = "https"
+
+    base_url = c.JupyterHub.get("base_url", "/")
+    cull_cmd.append(
+        f"--url={cull_proto}://localhost:8081" + url_path_join(base_url, "hub/api")
+    )
+
```
consideRatio, line 367: Thanks for considering this @thomasv314!

@thomasv314 is it correct that you have made use of both this Helm chart@master and internal_ssl, and want to make this change to support culling before internal_ssl is properly supported for the helm chart?

---

Below I'm just writing down things that pop into my mind for some transparency, I intend not to burden you with answering them but it rather a strategy for myself to keep track of thoughts while also being transparent to you as a PR author.

- What does --ssl-enabled really do?
- What does --url really do?
- Could it be times when it makes sense for --ssl-enabled to be set without having a --url with HTTPS proto?
- With internal_ssl as a official supported feature, we probably want to adjust the URL based on having internal_ssl enabled. With that in mind, how should we avoid causing issues for anyone using this in-between fix that is only relevant for what I think is those with custom internal_ssl support?
- Is this worth adding before internal_ssl? Maybe, not sure.

Issue #2088, opened 2021-03-08
Opt-out of `hub.jupyter.org/dedicated` tolerations


kafonek:
### Proposed change
Add an option in `config.yaml` to opt out of including `hub.jupyter.org/dedicated` / `hub.jupyter.org_dedicated` tolerations to various pods.

### Who would use this feature?
Users who are deploying Jupyterhub to a cluster that has a [PodTolerationRestriction admission-controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podtolerationrestriction) which they cannot control, and to which `hub.jupyter.org/dedicated` / `_dedicated` are not on the whitelist.  

Those Jupyterhub managers / cluster administrators would need to accept that they are losing out on potentially important [optimizations](https://zero-to-jupyterhub.readthedocs.io/en/stable/administrator/optimization.html#using-a-dedicated-node-pool-for-users) to the Jupyterhub deployment.

PR #2101, opened 2021-03-15
Enable opt-out of hub.jupyter.org/dedicated tolerations

kafonek:
## Final PR description (EDIT by @consideRatio)

With a PodTolerationRestriction in the cluster, the default tolerations that we set across all user and core pods could cause trouble and be useful to configure. Due to this, this PR makes them configurable under `scheduling.corePods|userPods.tolerations`.

## Original PR description

EDIT: Fixes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2088

@consideRatio at a high level this PR does four things - 

- Add new entry in `values.yaml` for `hub.defaultTolerations.enabled` (default `true`)
- New `_helpers-default-tolerations.tpl` file that defines the `hub.jupyter.org/dedicated` / `_dedicated` tolerations
- Replace hardcoded tolerations in other yaml/tpl files with an `{{- include "jupyterhub.defaultTolerations" . }}`
- Add a check in `jupyterhub_config.py` for kubespawner to not include those labels

I've half system-tested this on our environment.  It works for us when we set `hub.defaultTolerations.enabled: false` and fails as expected with Pod toleration problems when we set to true.  I don't have a good dev environment handy for setting up a cluster without the PodTolerationRestriction controller to run `pytest` or make sure this doesn't break normal deployments though.  

I added one blurb in the optimizations.md page.  I don't think I understand how https://zero-to-jupyterhub.readthedocs.io/en/stable/resources/reference.html is built in order to add a note there as well.  I don't have a `doc/source/resources/reference.md` file in my fork of z2jh, is that built in a special way?


consideRatio:
Thank you @kafonek!

The configuration reference is built based on documentation in schema.yaml.

I can't merge this as it is because of a terrible boring technical thing. It is not acceptable that we risk generating a k8s YAML manifest that looks like...

```yaml
# this is not acceptable, it can make us error
tolerations:
  # nothing

# this is acceptable I think but ugly and equally hard as the option below typically
tolerations: {}

# this is preferred
# (tolerations not even specified)
```

This is the kinds of technical crap I dislike most about working with Helm templates, solving these kinds of thing can be really ugly and add a lot of complexity. When given time I'll go ahead and push commits to fix this where needed.

I'm also not sure about nesting the configuration under `hub` as the tolerations will be applied on many pods not just under hub, so I think perhaps the setting should be on the root level rather than under `hub`. Perhaps even better it should be nested under `scheduling.corePods` and `scheduling.userPods`? See [this section in the config reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#scheduling-corepods) about that btw.

I think it may take a full week before i can focus on this as I've planned some days vacation and have some work in the queue already, but to summarize what remains in my mind pre-merge are:

### Action points
- [ ] Ensure valid k8s manifests are always generated (don't set `tolerations: <blank>`)
- [ ] Make a decision of the config option, my current thinking is that perhaps scheduling.core|userPods.defaultTolerations should be used where we let these have default values being the actual tolerations we apply, and opting out means to set this to a blank list, null, or changing them.
  @yuvipanda perhaps you can opine on this as well? I think it would solve a problem for you where you wanted to be able to specify the toleration used.
- [ ] Update schema.yaml with documentation on the options
- [ ] Ensure we have a values.yaml entry and lint-and-validate-config.yaml entry preferably

consideRatio:
Btw! I also want to clarify I support this kind of change as motivated by reasons described in the linked issue.

kafonek:
@consideRatio @yuvipanda how does the top level `defaultTolerations` look in this new values.yaml?  I also moved the definitions out to values.yaml instead of having them defined in the helper function. 

I am now realizing that having the helper function return `{}` won't work if the defaultTolerations are disabled but other tolerations (like `.Value.prePuller.hook.tolerations`) are included, so I need to do more work here.  It tries to create something like,
```
tolerations:
  {}
  - foo: bar
```

kafonek:
@consideRatio I learned about some behavior that was unexpected but very useful.  They don't seem to go into detail about using a `with` scope where the output is nil in https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with, but it seems to do exactly what we want and the `helm create` vanilla templates are replete with syntax like,

```
{{- with .Values.annotations }}
annotations:
  {{- toYaml . | nindent 2 }}
{{- end }}
```

Where the `annotations` line is not rendered if there is no `.Values.annotations`.  I am using the same approach to the tolerations now.  The `jupyterhub.tolerations` function returns nil or an array of tolerations.  Yaml files should call it like,

```
{{- with include "jupyterhub.tolerations" (dict "root" . "extraTolerations" .Values.hub.tolerations) }}
  tolerations:
    {{- . | nindent 8 }}
{{- end }}
```

If `defaultTolerations.enabled` is false and there are no `.Values.hub.tolerations`, the `tolerations` line is not rendered.


- Added info in `schema.yaml`
- Added matching entries in `values.yaml` and `lint-and-validate-config.yaml` although I don't really know what I'm doing with the latter
- Need particular scrutiny on jupyterhub/templates/image-puller/_helpers-daemonset.tpl.  I think it was unintentionally pulling in `singleuser.extraTolerations` in addition to default tolerations and `.prePuller.extraTolerations` but I'm not sure what the intent was before.

jupyterhub/templates/_helpers-tolerations.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- with .root.Values.defaultTolerations.tolerations }}
+{{- . | toYaml | nindent 0 }}
+{{- end }}
+{{- end }}
+{{- if .extraTolerations }}
+{{- with .extraTolerations }}
+{{- . | toYaml | nindent 0 }}
+{{- end }}
+{{- end }}
+{{- end }}
```
consideRatio, line 25: Working with helm templates is horrible, there is a bug here if you would have both extraTolerations and defaultTolerations active at the same time, because of whitespace-chomping will concatenate the tolerations too tightly.

There are solutions, but they are messy.

jupyterhub/templates/image-puller/_helpers-daemonset.tpl
```diff
@@ -50,12 +50,11 @@ spec:
...
-      tolerations:
-        {{- include "jupyterhub.userTolerations" . | nindent 8 }}
-        {{- with .Values.prePuller.extraTolerations  }}
-        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
-        {{- end }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
+      {{- with include "jupyterhub.tolerations" (dict "root" . "extraTolerations" .Values.prePuller.extraTolerations) }}
+      tolerations:
+        {{- . | nindent 8 }}
+      {{- end }}
```
consideRatio, line 57: This adds the wrong toleration, it should have `value: user` instead of `value: core`.

jupyterhub/schema.yaml
```diff
@@ -2220,6 +2216,37 @@ properties:
...
 
+  defaultTolerations:
+    type: object
+    required: [enabled]
+    description: |
+      Default [Toleration objects](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#toleration-v1-core)
+      applied to all Pods.  If your cluster has a [PodTolerationRestriction 
+      admission-controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podtolerationrestriction)
+      then you may need to configure it to accept these tolerations, 
+      or disable the defaultTolerations.
```
consideRatio, line 2227: There are two sets of tolerations, one with `value: core` and one with `value: user`.

User pods are the user servers and the image puller pods part of the image-puller daemonsets.

All other pods are considered core pods.

jupyterhub/schema.yaml
```diff
@@ -2220,6 +2216,37 @@ properties:
...
+          ```
+          - key: hub.jupyter.org/dedicated
+            operator: Equal
+            value: user
+            effect: NoSchedule 
+          - key: hub.jupyter.org_dedicated
+            operator: Equal
+            value: user
+            effect: NoSchedule
+          ```
```
consideRatio, line 2244: I suggest we split this in two, one named userTolerations and one named coreTolerations.

consideRatio (commented):
@kafonek I've started working to make some changes! Hope to finish them shortly

consideRatio:
## Key changes
1. Relocation of config from defaultTolerations on the root level to scheduling.userPods|corePods.tolerations.
   There was a need to dinstinguish the user pod tolerations from the core pod tolerations. What core pods and user pods are respectively is described in the configuration reference. [See the docs PR preview here.](https://zero-to-jupyterhub--2101.org.readthedocs.build/en/2101/resources/reference.html#scheduling-corepods)
2. Updated logic to have a template for coreTolerations and userTolerations respectively instead of just tolerations which implied coreTolerations.

## Misc changes
3. Updated lint-and-validate-values.yaml to make it easier to debug changes with `helm template jh ./jupyterhub --values tools/templates/lint-and-validate-values.yaml | grep -A 10 tolerations`
4. Updated jupyterhub_config.py that configures KubeSpawner, in practice duplicating some logic user for the other pods considered to be user pods.
5. Updated docs

kafonek:
Thanks for all the tweaks to this PR @consideRatio !

consideRatio:
@kafonek and thank you for your thorough work up to this point!

This kind of change had to manage a lot of the Helm associated complexity: Helm template functions, whitespace chomping details, and more.

Does this LGTY? If so, perhaps we can merge and then you could trial it?

jupyterhub/templates/image-puller/_helpers-daemonset.tpl
```diff
@@ -50,12 +50,11 @@ spec:
...
-      tolerations:
-        {{- include "jupyterhub.userTolerations" . | nindent 8 }}
-        {{- with .Values.prePuller.extraTolerations  }}
-        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
-        {{- end }}
       nodeSelector: {{ toJson .Values.singleuser.nodeSelector }}
+      {{- with include "jupyterhub.userTolerations" (dict "root" . "extraTolerations" .Values.prePuller.extraTolerations) }}
+      tolerations:
+        {{- . | nindent 8 }}
+      {{- end }}
```
consideRatio, line 57: @kafonek you asked about this section before. The hook-image-puller and continuous-image-puller pods part of these daemonsets are considered "user pods" so they should get the user associated tolerations, but they should also have their dedicated additional tolerations if set: prePuller.extraTolerations

The need for being able to do this is described in #1883.

consideRatio:
I inspected the rendered helm template diff which is available as part of the dev upgrade test, and it seems OK.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2101/checks?check_run_id=2154134808

kafonek:
@consideRatio all looks good to me.  Given that we would turn off tolerations by setting `scheduling.userPods.toleration: []` rather than `enabled: false`, does it make sense to just skip using a `define` altogether and do the concat / if check in each yaml file?  For instance,

```
# jupyterhub/templates/hub/deployment.yaml
      {{- with include "jupyterhub.coreTolerations" (dict "root" . "extraTolerations" .Values.hub.tolerations) }}
      tolerations:
        {{- . | nindent 8 }}
      {{- end }}
```

could instead be,

```
# jupyterhub/templates/hub/deployment.yaml
      {{- $tolerations := concat .Values.scheduling.userPods.tolerations .Values.hub.tolerations }}
      {{- if $tolerations }}
      tolerations:
        {{- toYaml $tolerations | nindent 8 }}
      {{- end }}
```

consideRatio:
> could instead be,
> 
> ```
> # jupyterhub/templates/hub/deployment.yaml
>       {{- $tolerations := concat .Values.scheduling.userPods.tolerations .Values.hub.tolerations }}
>       {{- if $tolerations }}
>       tolerations:
>         {{- toYaml $tolerations | nindent 8 }}
>       {{- end }}
> ```

Hmm this could instead be:

```
      {{- with concat .Values.scheduling.userPods.tolerations .Values.hub.tolerations }}
      tolerations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

I'll consider if that is reasonable across the examples, if we can do something like that consistently without issues I'm for it. I'll see if I think that's possible

kafonek:
That is slick @consideRatio !

consideRatio:
@kafonek wiee nice that was very viable now with all our tolerations available as arrays rather than a mix of arrays and strings! NICE!

@minrk time for celebration! One less named Helm template helper that obfuscated our template logic! :D

consideRatio:
@kafonek I could verify while making these changes that there wasn't a single characters difference in the output. There is a final piece I want to ensure, and that is that we don't accept null values for our tolerations, but require empty arrays. One minute...

consideRatio:
@kafonek do you agree to merge?

kafonek:
@consideRatio I've deployed this branch into our cluster and can confirm that we get the same pod toleration errors when not using a custom `config.yaml`, and that the deployments all work as expected when we add in:

```
# config.yaml
scheduling:
  corePods:
    tolerations: []
  userPods:
    tolerations: []
```

LGTM for merging.

consideRatio:
:tada: :heart: Thanks @kafonek for your thorough work on this and concratulations to the first merged PR in this project!

PR #2112, opened 2021-03-23
[NetPol] Fix notebook port in singleuser network policy

mhwasil:
The commit from #1670  breaks network policy in singleuser server.
Previously in version `0.9.1`, the default notebook port was set to `8888`, while from that commit onward, it was changed to `notebook-port`, which is not defined anywhere in the helm values. 
```
ingress:
  # allowed pods (hub.jupyter.org/network-access-singleuser) --> singleuser-server
  - ports:
      - port: notebook-port
```
compared to the one in `0.9.1`
```
ingress:
  - ports:
      - port: 8888
```

Is it supposed to be set via helm or was it just a mistake?

Cheers,

### DIFF
--- a/jupyterhub/templates/singleuser/netpol.yaml
+++ b/jupyterhub/templates/singleuser/netpol.yaml
@@ -40,7 +40,7 @@ spec:
 
     # allowed pods (hub.jupyter.org/network-access-singleuser) --> singleuser-server
     - ports:
-        - port: notebook-port
+        - port: 8888 
       from:
         # source 1 - labeled pods
         - podSelector:


yuvipanda:
notebook-port is set in kubespawner, [here](https://github.com/jupyterhub/kubespawner/blob/ba42e0d1a4208e7de6ba55113740e8d5233a5fa2/kubespawner/objects.py#L454). I think referencing it like this lets users change the notebook port (if they want to?) while the network policy continues to apply, since `notebook-port` will point to the new port.

Can you tell us what has broken in your install?

mhwasil:
Hi yuvi,
The default network policy generated by helm looks like this:
```yaml
kind: NetworkPolicy
metadata:
  annotations:
    meta.helm.sh/release-name: e2x-dev
    meta.helm.sh/release-namespace: e2x-dev
  creationTimestamp: "2021-03-16T12:35:19Z"
  generation: 1
  labels:
    app: jupyterhub
    chart: jupyterhub-0.11.1
    component: singleuser
    heritage: Helm
    release: e2x-dev
  managedFields:
  - apiVersion: networking.k8s.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:meta.helm.sh/release-name: {}
          f:meta.helm.sh/release-namespace: {}
        f:labels:
          .: {}
          f:app: {}
          f:app.kubernetes.io/managed-by: {}
          f:chart: {}
          f:component: {}
          f:heritage: {}
          f:release: {}
      f:spec:
        f:egress: {}
        f:ingress: {}
        f:podSelector:
          f:matchLabels:
            .: {}
            f:app: {}
            f:component: {}
            f:release: {}
        f:policyTypes: {}
    manager: Go-http-client
    operation: Update
    time: "2021-03-16T12:35:19Z"
  name: singleuser
  namespace: e2x-dev
  resourceVersion: "25628"
  selfLink: /apis/networking.k8s.io/v1/namespaces/e2x-dev/networkpolicies/singleuser
  uid: 3034b510-6da1-4bb6-9058-1fbef1e2be22
spec:
  egress:
  - ports:
    - port: 8081
      protocol: TCP
    to:
    - podSelector:
        matchLabels:
          app: jupyterhub
          component: hub
          release: e2x-dev
  - ports:
    - port: 53
      protocol: UDP
    - port: 53
      protocol: TCP
  ingress:
  - from:
    - podSelector:
        matchLabels:
          hub.jupyter.org/network-access-singleuser: "true"
    ports:
    - port: notebook-port
      protocol: TCP
  podSelector:
    matchLabels:
      app: jupyterhub
      component: singleuser-server
      release: e2x-dev
  policyTypes:
  - Ingress
  - Egress

```
as you can see under `ingress` that the notebook port  is a string `notebook-port`, while in [the kubernetes network policy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) this has to be a port number .

If this is supposed to be set in the KubeSpawner config, I could not find any clue where I can set this.

I am new to helm and k8s stuffs, but should `notebook-port` in [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/e14d686fea782482b1f7d388118bf772a6ab5be7/jupyterhub/templates/singleuser/netpol.yaml#L43) be a helm variable which starts and ends with curly brackets i.e. `{{ notebook-port }}`? If this is intended to be reconfigurable. CMIIW though :-).

As I mentioned in the previous comment, the singleuser netpol works after I changed `notebook-port` to `8888`.

Stay healthy.

yuvipanda:
hmm, interesting. This has worked fine for me as is without issues - k8s resolves what port 'notebook-port' is at runtime. Maybe @consideRatio has ideas?

consideRatio:
@mhwasil if this was the case, our tests would fail I think. Does it say somewhere it has to be a port number rather than a name? What network policy controller are you using in your k8s cluster?

mhwasil:
Hi @consideRatio 

The deployment was still successful and I was able to spawn the singleuser server. As you can see from the my config, I restricted both ingress and egress, but since the netpol was broken, I was still able to connect to internal services as well as to internet.
How to debug if the netpol is working?
* I did `wget google.com`, and my singleuser server was connected
* I did the same thing with internal services, and it was also working.
The netpol did not block both ingress and egress communication.

Then I edited my netpol and replaced it with the notebook port, and everything worked. 

I use WeaveNet for my CNI and based on [this](https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/) k8s netpol is supported in WeaveNet.

consideRatio:
> I use WeaveNet for my CNI and based on this k8s netpol is supported in WeaveNet.

NetworkPolicy is said to be supported by many CNIs, but there have been many malfunctions. What version of WeaveNet are you using etc?

There are tests setup here that checks network policies, but if you want to test this specific rule, you could `kubectl exec` into the hub pod that has the label of relevance for the rule, and probe the port of a user server.

For reference, here are some of the tests, but not one where this ingress rule is directly checked, but I'm quite confident there would be other failures if the hub or proxy pod couldn't speak with singleuser-server.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/tests/test_spawn.py#L137-L175

---

It may be that this is a workaround, I open to accept a workaround PR to be merged, but I'd really want to understand what we are working around so it can be mentioned explicitly that we can't use this otherwise common practice to my knowledge.

Perhaps there is some networking trick making the actual traffic bound for the pod proxy through a sidecar container with a port lacking the name?

mhwasil:
> 
> NetworkPolicy is said to be supported by many CNIs, but there have been many malfunctions. What version of WeaveNet are you using etc?
> 
WeaveNet version: `weave-kube:2.7.0`
K8s version: `v1.18.16`
z2jh version: `0.11.1`
 
>
> There are tests setup here that checks network policies, but if you want to test this specific rule, you could `kubectl exec` into the hub pod that has the label of relevance for the rule, and probe the port of a user server.
> 
I did this via singleuser notebook terminal. 

>
> For reference, here are some of the tests, but not one where this ingress rule is directly checked, but I'm quite confident there would be other failures if the hub or proxy pod couldn't speak with singleuser-server.
> 
In this case, since the netpol was broken, all singleuser servers were able to communicate to the internal services including proxy and hub.

> 
> Perhaps there is some networking trick making the actual traffic bound for the pod proxy through a sidecar container with a port lacking the name?
>
I am not sure about this. How can I check this?

I just did another test on GKE, and the problem persists.
I deployed a simple nginx in the same namespace, and did `wget <my_nginx_ip>:<port>`, and I was still able to connect to my nginx.

consideRatio:
z2jh 0.9.1, 0.10.2, and 0.11.1 has been mentioned, you are using 0.11.1?

I think a good debugging technique is to try establish a TCP connection from a pod to a pod, where the pods invoöved have certain labels and reside in certain namespaces so that are rules of the netpols you debug apply.

I would also make sure to identify ALL netpols in the namespace of relevance and see what netpols target the pods in question.

Netpols is the hardest configuration in k8s i think =\ i dont feel i can help much more than this, but the merge criteria i have is a better understanding why this is done, and if its a CNI specific workaround i think we should include another rule rather than change the rule we have to avoid breaking assumptions for people.

mhwasil:
> z2jh 0.9.1, 0.10.2, and 0.11.1 has been mentioned, you are using 0.11.1?
> 
I have debugged this with:
* 0.9.1 : netpol works fine and the default port was set to `8888` in the helm chart
* 0.10.2: I tried this on GKE and our in-house k8s cluster and the netpol does not work due to this issue.
* 0.11.1: I tried this on our in-house k8s cluster, and the netpol does not work. 

> 
> Netpols is the hardest configuration in k8s i think =\ i dont feel i can help much more than this, but the merge criteria i have is a better understanding why this is done, and if its a CNI specific workaround i think we should include another rule rather than change the rule we have to avoid breaking assumptions for people.

Should I just move this to Issues?

consideRatio:
> Should I just move this to Issues instead?

Oh I would appreciate that, it is a bit of work for you though.

Btw, I scanned the referenced PR that you describe having a commit that broke made a difference to your functionality. If you have that, you can try render templates and see what the actual differences are and track the actual change that is of relevance.

What confuse me is that I don't see any change with regards to `notebook-port` / `8888` in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1842

mhwasil:
> 
> EDIT: Ah but i found [consideRatio@ca9d4db](https://github.com/consideRatio/zero-to-jupyterhub-k8s/commit/ca9d4dbefabb03f763c5e904f65738da62ecb1b0) as part of #1670
> 
Yes, this commit. Sorry, it was wrong commit number

> UPDATE: This example says it's fine to use port names or numbers: https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/09-allow-traffic-only-to-a-port.md, hmm.. I would debug and inspect the user pods, what port definitions are used for the pod's containers? So `kubectl get pod -o yaml jupyter-<some-username>`. Does it say that a port is named notebook-port, or does it only mention the number 8888?

It does have name `notebook-port` though
```
    name: notebook
    ports:
    - containerPort: 8888
      name: notebook-port
      protocol: TCP

```

consideRatio:
Googling on wavenet, netpol. and named port, i found: https://github.com/weaveworks/weave/issues/3032

Hmm... I'm not sure about working around this, it seems like a quite significant workaround to start doing systematically because this is just one of many situations it will become relevant then I think.

mhwasil:
So it's the netpol issue.
If someone else encounters the same issue, the workaround here hopefully should work.

Issue #2119, opened 2021-03-28
Publish ARM64 Docker images


manics:
### Proposed change
As described in https://discourse.jupyter.org/t/ztjh-on-a-raspberry-pi-k8s-cluster/3043/16 and https://github.com/sakuraiyuta/zero-to-jupyterhub-k8s/commit/d22290b2ad6b0c69a7f682ce7ddc9b2b98a6a473 by @sakuraiyuta it's possible to run Z2JH on ARM systems with some changes. I think we should look at supporting ARM64 in Z2JH.

Work on ARM64 started in TLJH (issue: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/62, WIP PR from @yuvipanda https://github.com/jupyterhub/the-littlest-jupyterhub/pull/674).

### Alternative options
Do nothing.

### Who would use this feature?
AWS EC2 ARM64 users.
Raspberry pi users.

### Suggest a solution
This will require work on the Docker images and the GitHub workflows. Based on https://github.com/sakuraiyuta/zero-to-jupyterhub-k8s/commit/d22290b2ad6b0c69a7f682ce7ddc9b2b98a6a473 I think we could make the following changes:

#### Docker images

1. Add a build-arg in all Dockerfiles to allow the architecture to be switched.
For example, replace [`FROM ubuntu:20.04`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/e14d686fea782482b1f7d388118bf772a6ab5be7/images/hub/Dockerfile#L1) with `FROM $ARCH/ubuntu:20.04`, see https://github.com/docker-library/official-images#architectures-other-than-amd64
2. Allow chartpress to accept command line build arguments, e.g. `chartpress --build-arg ARCH={amd64|arm64v8}` mirroring [Docker's `--build-arg`](https://docs.docker.com/engine/reference/commandline/build/#options)
3. Figure out what to do with [singleuser-sample/Dockerfile](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/e14d686fea782482b1f7d388118bf772a6ab5be7/images/singleuser-sample/Dockerfile#L1) since `jupyter/base-notebook` doesn't support other architectures.
   e.g. we could build our own image from `$ARCH/python`, or perhaps ignore it since it's just an example.

[...]

PR #2125, opened 2021-04-01
Publish Arm64 compatible images

manics:
Part 1 of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2119#issuecomment-812101837: Minor changes to get Docker images building on ARM64
- `hub`: `psycopg2-binary` is not available as a precompiled binary, so `libpq-dev` needs to be installed for compilation. This increases the build time and the image size. An alternative would be to omit PostgreSQL support from the arm64 image, what do you think?
- `secret-sync`: `tini` for the correct architecture needs to be downloaded.

- [x] This contains a temporary workflow just for testing, this will be removed. In the meantime you can pull the test images from `docker pull ghcr.io/manics/z2jh-{hub,image-awaiter,network-tools,secret-sync}:dev`. Not yet tested in a deployment

### DIFF
--- a/.github/workflows/test-docker-build.yaml
+++ b/.github/workflows/test-docker-build.yaml
@@ -25,25 +25,21 @@ on:
 
 jobs:
   # TODO: this is just a quick test to check the arm64 docker images
-  # this should instead be done using chartpress
   # Based on
   # https://github.com/docker/build-push-action/blob/v2.3.0/docs/advanced/local-registry.md
   # https://github.com/docker/build-push-action/blob/v2.3.0/docs/advanced/multi-platform.md
   build_images:
     runs-on: ubuntu-20.04
-    strategy:
-      # Keep running even if one variation of the job fails
-      fail-fast: false
-      matrix:
-        image:
-          - hub
-          - image-awaiter
-          - network-tools
-          - secret-sync
-          # - singleuser-sample
-
     steps:
       - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      # https://github.com/jupyterhub/chartpress/pull/124
+      - name: Install chartpress buildx-platforms-dev version
+        run: pip install --no-cache-dir chartpress@git+https://github.com/manics/chartpress.git@buildx-platforms-skip
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
@@ -60,9 +56,8 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build and push to local registry
-        uses: docker/build-push-action@v2
-        with:
-          context: images/${{ matrix.image }}
-          platforms: linux/amd64,linux/arm64
-          push: true
-          tags: ghcr.io/${{ github.repository_owner }}/z2jh-${{ matrix.image }}:dev
+        run: >-
+          chartpress --push --tag dev
+          --builder docker-buildx
+          --platform linux/amd64 --platform linux/arm64
+          --image-prefix ghcr.io/${{ github.repository_owner }}/z2jh-
--- a/.github/workflows/test-docker-build.yaml
+++ b/.github/workflows/test-docker-build.yaml
@@ -37,9 +37,8 @@ jobs:
         with:
           python-version: "3.8"
 
-      # https://github.com/jupyterhub/chartpress/pull/124
-      - name: Install chartpress buildx-platforms-dev version
-        run: pip install --no-cache-dir chartpress@git+https://github.com/manics/chartpress.git@buildx-platforms-skip
+      - name: Install chartpress
+        run: pip install chartpress
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
--- a/.github/workflows/test-docker-build.yaml
+++ b/.github/workflows/test-docker-build.yaml
@@ -46,17 +46,8 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
-      # https://github.com/docker/login-action/tree/v1.8.0#github-container-registry
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v1
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Build and push to local registry
+      - name: Build a multiple architecture Docker image
         run: >-
-          chartpress --push --tag 0.0.1-dev
+          chartpress
           --builder docker-buildx
           --platform linux/amd64 --platform linux/arm64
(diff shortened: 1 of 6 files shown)


images/hub/Dockerfile
```diff
@@ -4,7 +4,13 @@ FROM ubuntu:20.04
 
 ENV DEBIAN_FRONTEND=noninteractive \
     LANG=C.UTF-8
-RUN apt-get update && \
+
+# psycopg2-binary in requirements.txt is not compiled for linux/arm64
+# TODO: Use build stages to compile psycopg2-binary separately instead of
+# bloating the image size
+RUN EXTRA_PACKAGES=; \
```
consideRatio, line 11:
```suggestion
# psycopg2-binary in requirements.txt is not compiled for linux/arm64
# TODO: Use build stages to compile psycopg2-binary separately instead of
# bloating the image size
RUN EXTRA_PACKAGES=; \
```

What do you think about naming this EXTRA_APT_PACKAGES?

.github/workflows/test-docker-build.yaml
```diff
@@ -0,0 +1,68 @@
...
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push to local registry
+        uses: docker/build-push-action@v2
+        with:
+          context: images/${{ matrix.image }}
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ghcr.io/${{ github.repository_owner }}/z2jh-${{ matrix.image }}:dev
```
consideRatio, line 68: Ohhh wow how nice and clean these steps are.

We have some complexity if we want to use this instead of build args and a dynamic tag based on chartpress.yaml and git inspection logic in chartpress. This looks great though.

images/hub/Dockerfile
manics: Done!

.github/workflows/test-docker-build.yaml
manics: This is just a quick and easy way to expose the built images for download. I should have a chartpress PR to replace some of this soon :smile_cat:

manics:
Remaining requirements:
- [ ] https://github.com/jupyterhub/chartpress/pull/124
- [ ] https://github.com/jupyterhub/configurable-http-proxy/pull/304

.github/workflows/test-docker-build.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      # https://github.com/jupyterhub/chartpress/pull/124
+      - name: Install chartpress buildx-platforms-dev version
+        run: pip install --no-cache-dir chartpress@git+https://github.com/manics/chartpress.git@buildx-platforms-skip
+
```
consideRatio, line 43:
```suggestion
```

Chartpress 1.1.0 is released, and we don't have it pinned.

.github/workflows/test-docker-build.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push to local registry
+        run: >-
+          chartpress --push --tag dev
```
consideRatio, line 59:
```suggestion
          chartpress --push --tag 0.0.1-dev
```

manics:
@consideRatio Do you want me to keep this new workflow? Or get rid of it and add the flags to:
- the existing chartpress test workflows https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/08faa6f197e478ca28ce23d083287d37096cbb36/.github/workflows/test-chart.yaml#L182
- the publish workflow https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/08faa6f197e478ca28ce23d083287d37096cbb36/ci/publish#L29-L33

consideRatio:
I'm not sure, I'd like the idea of having the logic to publish images in a single workflow file I think. What is your take?

manics:
Just realised the `test-chart.yaml` workflow won't work because you can't load multiple architectures into the host Docker daemon. I don't see a reason to have a separate push workflow though, so what probably makes sense is:
-  a workflow that builds but doesn't push all architectures, this can be either a new job in `test-chart.yaml` or this existing workflow with `--push` removed
- add `--push` to `ci/publish`

consideRatio:
Thinking about this a bit more now...
- When we publish, we publish all images.
- When we test the chart we normally test image build + chart install/startup tests
- With the new arm64 addition, we are currently only able to test the build of arm64 images.
- With a arm64 runner, we can test the image build + chart install/startup tests just and hopefully avoid breaking DRY doing so by generalizing the test image build + chart install/startup test

I'm very afraid of duplicating the test suite and really want to avoid breaking DRY in this case. I'm hoping we can have a arm64 based (self-hosted / circle-ci hosted) github actions runner or similar. If we assume that is plausible to accomplish, I'd be leaning towards:

1. Embed the publish parts of arm64 images to the publish workflow
2. We can retain this separate workflow test-arm64-image-build workflow until we have a arm64 based runner that can run the chart tests as well, at which point we let this logic merge into the test-chart workflow that will execute on both amd64 and arm64 runners and be able to adjust based on this.
3.

consideRatio:
> * a workflow that builds but doesn't push all architectures, this can be either a new job in `test-chart.yaml` or this existing workflow with `--push` removed

Our test-chart workflow as it is already contains a build through chartpress, and it won't rebuild if it finds an existing image - but, the existing image will be in docker.io/jupyterhub/k8s-... rather than a local registry. Hmm...

I would be positive to have a dedicated build job if we don't just build in that dedicated build job AND the test-chart job that also builds images unless they are existing already.

I guess the ideal outcome given the complexity tradeoff doesn't make it non-ideal would be to:
- have a dedicate image build step
- have a dedicated test chart step using already built images
- (not sure, but maybe...) have a dedicated publish step that re-use images tested rather than re-builds them again

consideRatio:
@manics nice work on this! The CHP images are published like a charm and I love the new actions job to extract relevant tags!

image

manics:
I've updated the publish workflow to (hopefully) build and push the multiple architectures, and removed the push flag from the docker build workflow.

I don't think configuring circle-ci is too bad. The aim of the arm64 workflow will be to test the images _not_ the helm chart, so we can skip most of the complexity. I've made a start in https://github.com/manics/zero-to-jupyterhub-k8s/blob/arm64-circleci/.circleci/config.yml

Removing pebble and network policies removes a lot of the complex setup. It'll require a flag in the tests to allow non-https requests to be made, but I think that's worthwhile anyway since it makes it easier to developers to test locally. If we want we could use this as an opportunity to test some other config that's not already tests.

consideRatio:
Wieee nice work on this @manics! What an effort!

Issue #2102, opened 2021-03-16
Can't connect to jupyterhub managed services w/ url


kafonek:
### Bug description
I am having trouble deploying a [Jupyterhub-managed service](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#hub-managed-services) that exposes a `url`.  The configurable-http-proxy is unable to proxy to hosts `localhost`, `127.0.0.1`, or `0.0.0.0`.

### How to reproduce
Run a simple web server that copies syntax from [Tornado docs](https://www.tornadoweb.org/en/stable/#hello-world) and [`jupyterhub/examples/service-whoami`](https://github.com/jupyterhub/jupyterhub/tree/master/examples/service-whoami):

```
# config.yaml
hub:
  services:
    hello:
      admin: false
      command: ["python3", "hello.py"]
      url: 'http://127.0.0.1:8888'

  extraFiles:
    hello:
      mountPath: /srv/jupyterhub/hello.py
      stringData: |
        import tornado.ioloop
        import tornado.web

        class MainHandler(tornado.web.RequestHandler):
            def get(self):
                self.write("Hello, world")

        def make_app():
            return tornado.web.Application([
                (r"/?", MainHandler),
            ])

        if __name__ == "__main__":
            app = make_app()
            app.listen(8888)
            tornado.ioloop.IOLoop.current().start()
```

The "hello" service shows up in the Jupyterhub web UI.  Kubectl logs for `deploy/hub` and `deploy/proxy` show the service being spawned, connected, and proxy routes set.

```
# kubectl logs deploy/proxy
12:28:51.378 [ConfigProxy] info: Adding route /services/hello -> http://127.0.0.1:8888
12:28:51.378 [ConfigProxy] info: Route added /services/hello -> http://127.0.0.1:8888
12:28:51.379 [ConfigProxy] info: 201 POST /api/routes/services/hello 
```

Curling `http://127.0.0.1:8888` while exec'd into the Hub pod works.  Trying to curl `http://proxy-public/services/hello` or opening the service in the web UI shows errors such as:

```
#kubectl logs deploy/proxy
12:36:32.721 [ConfigProxy] debug: PROXY WEB /services/hello to http://127.0.0.1:8888
12:36:32.722 [ConfigProxy] error: 503 GET /services/hello connect ECONNREFUSED 127.0.0.1:8888
```
### Alternative attempt - 0.0.0.0
[...]

PR #2132, opened 2021-04-06
Support for Hub-managed services

kafonek:
Fix https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2102

@yuvipanda @consideRatio I'll add documentation if you support the idea of a new `k8s_managed_port` option for Jupyterhub Services in values.yaml/config.yaml (or want to name it something else).  This PR would look for that value in `templates/hub/netpol.yaml` and `templates/hub/service.yaml` to add the appropriate ports, as well as a change in `jupyterhub_config.py` to translate the port into the appropriate service url.

@minrk on a somewhat related note, I think https://github.com/kafonek/fastapi_service_example is at a place where I can PR to jupyterhub examples, if you want to review it first.  The gif below is an example of using that as a Hub-managed service in z2jh.

service_example

### DIFF
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -40,10 +40,10 @@ spec:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
-          # Support for Hub-managed services in k8s
+        {{- /* Also accept traffic to ports for additional services running in hub pod /*}}
         {{- range $name, $config := .Values.hub.services }}
         {{- if hasKey $config "k8s_managed_port" }}
-        - port: {{ get $config "k8s_managed_port" }}
+        - port: {{ $config.k8s_managed_port }}
         {{- end }}
         {{- end }}
       from:
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -40,7 +40,7 @@ spec:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
-        {{- /* Also accept traffic to ports for additional services running in the hub pod /*}}
+        {{- /* Also accept traffic to ports for additional services running in the hub pod */}}
         {{- range $name, $config := .Values.hub.services }}
         {{- if hasKey $config "k8s_managed_port" }}
         - port: {{ $config.k8s_managed_port }}
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -40,6 +40,12 @@ spec:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
+        # Support for Hub-managed services in k8s
+        {{- range $name, $config := .Values.hub.services }}
+        {{- if hasKey $config "k8s_managed_port" }}
+        - port: {{ get $config "k8s_managed_port" }}
+        {{- end }}
+        {{- end }}
       from:
         # source 1 - labeled pods
         - podSelector:
--- a/jupyterhub/templates/hub/netpol.yaml
+++ b/jupyterhub/templates/hub/netpol.yaml
@@ -40,7 +40,7 @@ spec:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
-        # Support for Hub-managed services in k8s
+          # Support for Hub-managed services in k8s
         {{- range $name, $config := .Values.hub.services }}
         {{- if hasKey $config "k8s_managed_port" }}
         - port: {{ get $config "k8s_managed_port" }}
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -31,11 +31,12 @@ spec:
       {{- with .Values.hub.service.ports.nodePort }}
       nodePort: {{ . }}
       {{- end }}
-      # Support for Hub-managed services in k8s
+    {{- /* Expose ports for additional services running in the hub pod /*}} 
     {{- range $name, $config := .Values.hub.services }}
     {{- if hasKey $config "k8s_managed_port" }}
-    - name: {{ $name | replace "_" "-"}}
-      port: {{ get $config "k8s_managed_port" }}
-      targetPort: {{ get $config "k8s_managed_port" }}
+    {{- $k8s_compliant_name := $name | lower | replace "_" "-" | trunc 63 | trimSuffix "-"}}
+    - name: {{ $k8s_compliant_name }}
+      port: {{ $config.k8s_managed_port }}
+      targetPort: {{ $config.k8s_managed_port }}
     {{- end }}
     {{- end }}
--- a/jupyterhub/templates/hub/service.yaml
+++ b/jupyterhub/templates/hub/service.yaml
@@ -31,7 +31,7 @@ spec:
       {{- with .Values.hub.service.ports.nodePort }}
       nodePort: {{ . }}
       {{- end }}
-    # Support for Hub-managed services in k8s
+      # Support for Hub-managed services in k8s
     {{- range $name, $config := .Values.hub.services }}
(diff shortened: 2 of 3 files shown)


yuvipanda:
I'll look at the PR itself later, but would <3 for a PR to examples with fastapi :) Was just bugging @bollwyvl  to make a fastapi example...

kafonek:
Linking https://github.com/jupyterhub/jupyterhub/pull/3403 to this PR @yuvipanda @bollwyvl

jupyterhub/templates/hub/netpol.yaml
```diff
@@ -40,6 +40,12 @@ spec:
     # allowed pods (hub.jupyter.org/network-access-hub) --> hub
     - ports:
         - port: http
+          # Support for Hub-managed services in k8s
+        {{- range $name, $config := .Values.hub.services }}
+        {{- if hasKey $config "k8s_managed_port" }}
+        - port: {{ get $config "k8s_managed_port" }}
+        {{- end }}
+        {{- end }}
```
consideRatio, line 48: I made the comment made into a helm comment to avoid confusion if inspecting the rendered template without any additional ports listed. I also suggest another syntax to accomplish the same thing as what I assume `get` does for consistency with the rest of the chart.

```suggestion
        {{- /* Also accept traffic to ports for additional services running in the hub pod /*}}
        {{- range $name, $config := .Values.hub.services }}
        {{- if hasKey $config "k8s_managed_port" }}
        - port: {{ $config.k8s_managed_port }}
        {{- end }}
        {{- end }}
```

jupyterhub/templates/hub/service.yaml
```diff
@@ -25,8 +25,17 @@ spec:
...
       nodePort: {{ . }}
       {{- end }}
+      # Support for Hub-managed services in k8s
+    {{- range $name, $config := .Values.hub.services }}
+    {{- if hasKey $config "k8s_managed_port" }}
+    - name: {{ $name | replace "_" "-"}}
+      port: {{ get $config "k8s_managed_port" }}
+      targetPort: {{ get $config "k8s_managed_port" }}
+    {{- end }}
+    {{- end }}
```
consideRatio, line 41: Same kind of changes suggested as in the other code suggestion.

```suggestion
    {{- /* Expose ports for additional services running in the hub pod /*}}
    {{- range $name, $config := .Values.hub.services }}
    {{- if hasKey $config "k8s_managed_port" }}
    - name: {{ $name | replace "_" "-" }}
      port: {{ $config.k8s_managed_port }}
      targetPort: {{ $config.k8s_managed_port }}
    {{- end }}
    {{- end }}
```

consideRatio: The name has the following criteria on it to be valid according to [the k8s api ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#serviceport-v1-core).

Name must either be an empty string or pass DNS_LABEL validation:
- must be no more than 63 characters long.
- must consist of lower case alphanumeric characters or '-'.
- must start and end with an alphanumeric character.

consideRatio (commented):
TL;DR I'm still processing this and isn't sure what makes sense from a user perspective and a long term maintenance perspective.

## Considerations

I think this is a PR approaching something we should handle to some degree, but I'm not sure to what degree. I fully support having a way for users to configure a jupyterhub managed service with a `command` and needs for a port to be opened up etc, but I'm not sure about letting the user configure it all via `hub.services`.

1. The user should be allowed to configure a `command` for the [jupyterhub managed service](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#properties-of-a-service)).
   - [x] They already can as hub.services is a passthrough config!

2. The user should be allowed to configure a port number to expose the jupyterhub managed service outside the container.
   1. The hub Pod needs to expose the hub container's port

      There is no way to do this at the moment, and this PR doesn't address it yet.
   2. The hub Service needs to expose the hub Pod's port

      There is no way to do this at the moment, but this PR address it with `hub.services.*.k8s_managed_port`.
   3. The hub NetPol needs to accept incoming and outgoing network traffic for this port.

       This is a way to configure this with `hub.netpol.ingress` and `hub.netpol.egress` already, but this adds another way specifically for ingress to the specified port from any pod with a specific label for hub network access that we already use.

3. The user should be allowed to configure the hub's networkpolicy to access that port

   Should we bundle such access with the current network access label? If so, this PR has the correct approach, but otherwise we should let the user configure it manually by setting up an additional ingress rule in `hub.netpol.ingress`. I'm not sure what I think about it.
   
## My influences

### Intertwining the chart config with software config
I like it when the Helm chart's config can be clearly separated from configuration of the software running in pods, such as JupyterHub. In this case we have `hub.services` that almost directly maps to `c.JupyterHub.services`. If we add config options there like `k8s_managed_port` we intertwining z2jh config with JupyterHub config.

I don't exclude this an option, but I like it when we can make a quite clear separation of what influences the chart and what influences the chart hosted software.

### Convenience configuration

Some configuration options could be exposed in multiple locations, but it would raise the complexity as well. It would be a convenience to configure a port number under hub.services but it would do quite a bit of things from a single location that may also need to be adjusted from other locations - so then perhaps we should just document how to get it done using the general way of configuring things?

jupyterhub/templates/hub/service.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       {{- with .Values.hub.service.ports.nodePort }}
       nodePort: {{ . }}
       {{- end }}
-    {{- /* Expose ports for additional services running in the hub pod /*}}
+    {{- /* Expose ports for additional services running in the hub pod */}}
```
consideRatio, line 34: Woops thanks!

kafonek:
@consideRatio thanks for the commits, the checklist, and insight into your general mindset on the PR.  

Quick question on 2.i, `The hub Pod needs to expose the hub container's port`.  Do you mean I should add something to `templates/hub/deployment.yaml` in the ports section (basically copied over from `service.yaml`) ?  As a sanity check for myself, including a `containerPort` entry in a deployment is purely informational besides giving you an alias to use in Service object `targetPort`'s (`targetPort: http` vs `targetPort: 8081`), right?   As long as a process is listening to `0.0.0.0:10101`, I can reach `10101` regardless of whether it's exposed.  Analogous to docker's `EXPOSE` verb.


I understand and appreciate what you mean by keeping chart config and software config separate.  I reviewed other sections of `values.yaml` and `jupyterhub_config.py`, and I don't see any examples that are doing as much "magic" as the `k8s_managed_port` implementation here.  

Not having a way to update the `hub/service.yaml` (2.iii in your post) in a `config.yaml` is the real blocker for my initial ticket https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2102.  If you think I should consider where to allow plugin configuration to that instead of this implementation, I will restructure the PR.  I suppose in that approach it may look something like the following?

```
# config.yaml
hub:
  services:
    foo:
      url: http://hub:8181
      command: ["uvicorn", "app:app", "--host", "0.0.0.0", "--port", "8181"]
  networkPolicy:
    allowedIngressPorts:
      - 8181
  service:
    extraPorts:
      - name: foo
        port: 8181
        targetPort: 8181 
```

```
# service.yaml
...
spec: 
  ports:
    - name: http
      port: 8081
      targetPort: http
    {{- with .Values.hub.service.extraPorts }}
    {{ . | nindent 6 }}
    {{- end }}
```

kafonek:
closing in favor of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2148

consideRatio:
> As a sanity check for myself, including a containerPort entry in a deployment is purely informational besides giving you an alias to use in Service object targetPort's (targetPort: http vs targetPort: 8081), right?

Ohhhh hmmm well if it works... I was thinking you were required to declare them in the pod as a pod can have multiple containers having services using the same ports etc, and that you may have a container listening to a lot of things but you perhaps only wanted to expose a selection of that to the k8s networking.

I'll ping you if you ping me when we learn how this fits together properly! :)

Thanks for your persistence and thorough work when contributing, you put in a lot of effort to pursue an approach that feels good for the reviewer. Thank you!

PR #2133, opened 2021-04-06
update k8 version for DO to currently available version

RyanQuey:
Not using the version that was there gets error: 

```
Error: POST https://api.digitalocean.com/v2/kubernetes/clusters: 422 validation error: invalid version slug
```

For latest errors, calling `doctl kubernetes options versions` shows: 

```
=> doctl kubernetes options versions
Slug            Kubernetes Version
1.20.2-do.0     1.20.2
1.19.6-do.0     1.19.6
1.18.14-do.0    1.18.14
```

### DIFF
--- a/doc/source/kubernetes/digital-ocean/step-zero-digital-ocean.md
+++ b/doc/source/kubernetes/digital-ocean/step-zero-digital-ocean.md
@@ -31,7 +31,7 @@ If you prefer to use the Digital Ocean portal see the [Digital Ocean Get Started
 
    ```
    export DIGITALOCEAN_ENABLE_BETA=1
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for opening this, @RyanQuey!

Is there a way to leave the version unspecified in the doctl command? We could do that. If not, we can bump it to 1.20

RyanQuey:
Yeah, according to [docs](https://docs.digitalocean.com/reference/doctl/reference/kubernetes/cluster/create/) it defaults to `latest`. I'll update

yuvipanda:
Thanks, @RyanQuey!

PR #2134, opened 2021-04-06
fix cluster name for DO installation instructions

RyanQuey:
keep cluster name consistent throughout the instructions


yuvipanda:
Thanks, @RyanQuey!

PR #2144, opened 2021-04-13
publish workflow: build amd64 and arm64 prerequisites added

consideRatio:
This is a complementary PR for #2125 that adds QEMU / Docker buildx before chartpress tries to build and publish multi-arch images.


manics (approved):
Good point!

manics:
Great! 🎉 
image

I'll update and close https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2119 with testing instructions.

PR #2162, opened 2021-04-21
ci: update network tests as jupyter.org IPs changed

consideRatio:
It appears that the IPs we use to test the charts network policies have become outdated. They were IPs associated with jupyter.org's webhosting, so, the new IPs are the new IPs I got from doing `nslookup jupyter.org`.


consideRatio:
GitHub runners seem to have some issues at the moment. The tests are not actually failing if they are successfully getting run.

PR #2163, opened 2021-04-21
ci: precautions for security, update github_token permissions, pin actions

consideRatio:
This PR is a followup of having the JupyterHub org's permission level of the available secrets.GITHUB_TOKEN now only have read access to the repo by default.

Along with this, I'm also pinning github actions referenced to specific hashes in the jobs requesting write permissions for the GITHUB_TOKEN.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -58,10 +58,14 @@ jobs:
           print(f"::set-output name=publishing::{publishing}")
 
       - name: Set up QEMU (for docker buildx)
-        uses: docker/setup-qemu-action@v1
+        # NOTE: Updates to the actions commit hash should go hand in hand with
+        #       updates in the test-docker-build workflow.
+        uses: docker/setup-qemu-action@25f0500ff22e406f7191a2a8ba8cda16901ca018 # associated tag: v1.0.2
 
       - name: Set up Docker Buildx (for chartpress multi-arch builds)
-        uses: docker/setup-buildx-action@v1
+        # NOTE: Updates to the actions commit hash should go hand in hand with
+        #       updates in the test-docker-build workflow.
+        uses: docker/setup-buildx-action@2a4b53665e15ce7d7049afb11ff1f70ff1610609 # associated tag: v1.1.2
 
       - name: Install chart publishing dependencies (chartpress, helm)
         run: |
--- a/.github/workflows/test-docker-build.yaml
+++ b/.github/workflows/test-docker-build.yaml
@@ -10,11 +10,13 @@ on:
   pull_request:
     paths:
       - "images/**"
-      - "**/test-docker-build.yaml"
+      - "chartpress.yaml"
+      - ".github/workflows/test-docker-build.yaml"
   push:
     paths:
       - "images/**"
-      - "**/test-docker-build.yaml"
+      - "chartpress.yaml"
+      - ".github/workflows/test-docker-build.yaml"
     branches-ignore:
       - "dependabot/**"
   workflow_dispatch:
--- a/.github/workflows/test-docker-build.yaml
+++ b/.github/workflows/test-docker-build.yaml
@@ -5,29 +5,24 @@ name: Test docker multiarch build
 
 # Trigger the workflow's on all PRs and pushes so that other contributors can
 # run tests in their own forks. Avoid triggering these tests on changes to
-# documentation only changes.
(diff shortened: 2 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -58,10 +58,14 @@ jobs:
           print(f"::set-output name=publishing::{publishing}")
 
       - name: Set up QEMU (for docker buildx)
-        uses: docker/setup-qemu-action@v1
+        # NOTE: Updates to the actions commit hash should go hand in hand with
+        #       updates in the test-docker-build workflow.
+        uses: docker/setup-qemu-action@25f0500ff22e406f7191a2a8ba8cda16901ca018 # associated tag: v1.0.2
```
manics, line 63: Dependabot is already configured on this repo for github actions: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/eb4dd3b2becd33b3db0d2b09daafd9d66c26e970/.github/dependabot.yaml#L38-L47

Is the `# associated tag: v1.0.2` going to cause problems?

consideRatio: I'm not sure, but i would guess not except that it may not update it with the annotation.

I think we shouldn't update these much in general and lean towards thinking we only do it when we see a need for it rather than when dependabot suggests it.

I suggest we wait and learn what happens and make a decision down the line about things in more detail at that point

manics: OK, would you mind adding a comment along those lines? I'm thinking about this fropm the perspective of a new or infrequent maintainer, there's nothing to say don't merge PRs by dependabot if they change this line.

consideRatio: Hmmm, I'm a bit confused - could you push or create a code suggestion?

I don't see it as a problem if we end up with a `# associated tag: v1.0.2` being stripped if dependabot updates this line of code. Are you thinking about if it would retain the comment, and it now points to an outdated tag because dependabot managed to retain the version comment but update the hash?

- Confused about what to expect from dependabot when pinning a hash
- Confused about what your concern is out of several concerns I can imagine (comment outdated, stripped, action out of sync, perhaps some more)

manics: Typically dependabot updates are OK to merge if they're non-breaking and CI passes. You've made an alternative suggestion:
> I think we shouldn't update these much in general and lean towards thinking we only do it when we see a need for it rather than when dependabot suggests it.

But that's going to be lost as soon as this PR is merged.

manics: What policy do you prefer? Merge dependabot PRs as usual in which case this can be merged as is and if the comment causes a problem we can fix it later, or document a new policy of only merging when it's needed?

consideRatio: @manics hmmm this became a bit complicated for me to grasp fully, so my answer below is vague because I'm a bit lost now.

I'm still uncertain about aspects like if dependabot will even try to bump a commit-pinned action, and if so, to what version it will bump. Whenever a new release is made, whenever a new commit is made, or never?

Since I don't know that, and I feel that the inline comment about keeping versions in sync is around as a warning that something could go wrong if they get out of sync, and I'm very uncertain it would risk becoming out of sync by having dependabot - i lean towards not taking any action.

With regards to a policy for bumping the version of the QEMU action (like "we shouldn't update the much in general" or "only when found to be needed" or "always keep updated!" or "when you have time to check for changes upstream" etc): I think I'm too vaguely opined to motivate we follow a certain policy.

consideRatio: +1 for merge as is and +1 to approve/merge any suggested policy or change you think make sense - I myself feel too vaguely opined and weakly informed about how dependabot will work to suggest something concretely.

.github/workflows/test-docker-build.yaml
```diff
@@ -5,29 +5,24 @@ name: Test docker multiarch build
...
 on:
   pull_request:
-    paths-ignore:
-      - "doc/**"
-      - "**/test-docs.yaml"
-      - "**.md"
-      - "**/schema.yaml"
+    paths:
+      - "images/**"
+      - "**/test-docker-build.yaml"
```
manics, line 13: Do we need `chartpress.yaml` in here too?

.github/workflows/publish.yml
manics: I'm happy with the status quo of merging all updates :smile:

PR #2165, opened 2021-04-24
Update NOTES.txt, including removing "alpha" designation

manics:
Also includes direct commands for extracting IPs/NodePort (not sure if this is too complicated? though it makes it easier for people who don't immediately understand how to parse `kubectl get svc proxy-public`)

TODO:
- [x] ingress?

### DIFF
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -6,16 +6,17 @@ You can check whether the hub and proxy are ready by running:
 
  kubectl --namespace={{ .Release.Namespace }} get pod
 
-and watching for both those pods to be in status 'Running'.
-{{/* Force an empty newline */}}
+and watching for both those pods to be in status 'Running'.{{ println }}
 
 {{- if eq .Values.proxy.service.type "LoadBalancer" }}
-You can find the public IP of JupyterHub by running:
+You can find the public (load-balancer) IP of JupyterHub by running:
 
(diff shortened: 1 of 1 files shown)


jupyterhub/templates/NOTES.txt
```diff
@@ -1,26 +1,56 @@
...
+  kubectl --namespace={{ .Release.Namespace }} get svc {{ include "jupyterhub.proxy-public.fullname" . }}
+
+If you have questions, please:
+
+  1. Read the guide at https://z2jh.jupyter.org
+  2. Ask for help or chat to us on https://discourse.jupyter.org/
+  3. If you find a bug please report it at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues{{ println }}
 
-{{- if hasKey .Values.singleuser.cloudMetadata "enabled" }}{{ println }}
+{{- if hasKey .Values.singleuser.cloudMetadata "enabled" }}
```
consideRatio, line 53: Assuming someone would copy paste the deprecation logic, they would need `{{ println }}` in it at least assuming two deprecation messages would trigger - in order to avoid no new blank line between them.

consideRatio: Whenever there is a `fail` it is not relevant with later messages as it would stop the rendering, but if it is a warning - multiple warnings can trigger.

Below there is a deprecation warning like this that includes println still while this doesn't any more. I suggest we consistently use println within each.

consideRatio (approved):
This LGTM, small whitespace detail I think could be addressed.

consideRatio:
@manics thank you so much for maintaining, well, everything !! :heart:

PR #2168, opened 2021-04-30
images/hub - a regular run of script: hub/images/dependencies freeze --upgrade

consideRatio:
Dependabot only updates what we have in requirements.in, but leaves the dependencies of these libraries. So, I ran `./dependencies freeze --upgrade` from the /images/hub folder to do that.


consideRatio:
Test failure unrelated, will make it more reliable in a separate PR.

Issue #2085, opened 2021-03-08
Document resource requests for core components


minrk:
We used to set default resource requests on most pods, but after #2034 we no longer set any. This means it falls to *documentation* to cover resource reservations and limits on our core pods, and what reasonable values might be, and when they are a good idea.

Tasks:

- [ ] explain a bit of background about resource reservations and when they are useful
- [ ] identify all components that might want resource requests
- [ ] identify reasonable values for key pods (mainly hub and proxy) and ranges based on load (e.g. using [mybinder.org as a reference point for high load](https://grafana.mybinder.org/d/GYEYQm7ik/components-resource-metrics?orgId=1&refresh=1m))

consideRatio:
While writing on the changelog for 1.0.0 I realize I think we must address this issue as https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2034 set stopped setting default requests on containers.

I'm still very ambiguous about this in general, but I think not having them set add quite a bit of complexity that at least need to be mitigated by having a guiding documentation and suggestions.

consideRatio:
In #2034 the default requests were removed which prompts some action points:

- [ ] Documentation about the new resource requests/limits added/updated
- [ ] Default values re-considered again before we lock something in.
  - [ ] The key motivation for the choices to be clarified

## Overview of pods and their requests

Configuration | pod | cpu/memory requests before 1.0.0 | Note
-|-|-|-
hub.resources | hub | 200m, 510Mi | JupyterHub and KubeSpawner runs here. Can manage with small resources but could peak up to 1 CPU during very heavy load of simultaneous users starting and stopping servers.
proxy.chp.resources | proxy | 200m, 510Mi | The container runs `configurable-http-proxy`. Will require small amounts of resources.
proxy.traefik.resources | autohttps | - | The container performs TLS termination only. Will require small amounts of resources.
proxy.secretSync.resources | autohttps | - | The sidecar container is a watchdog, watching a file for changes and updates a k8s Secret with those changes. Will require minimal resources.
scheduling.userScheduler.resources | user-scheduler | 50m, 256Mi | The container runs a `kube-scheduler` binary with custom configuration to schedule the user pods. Will require a small amount of resources.
[...]

PR #2175, opened 2021-05-05
Add changelog for 1.0.0-beta.1

consideRatio:
### Action points
- [x] Agree on security communication (https://github.com/orgs/jupyterhub/teams/jupyterhubteam/discussions/8)
- [x] Backport documentation (#2227)
- [x] Remove deprecation logic and force new config options (action point: merge #2211)
- [x] Add hub.lifecycle passthrough config and a fix to the schema for singleuser.lifecycleHooks (action point: merge #2223)
- [x] Resource requests documentation (#2085, action point: merge #2226)
- [x] Finalize this changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,20 @@
 
 Here you can find upgrade changes in between releases and upgrade instructions.
 
-## [12.0] - 2021-05-FIXME
+## [1.0]
 
-### [12.0.0]
+### [1.0.0] - 2021-05-FIXME
 
-FIXME: summarize release
+FIXME: Summary
 
-- note 12.0.0 bump from 0.11.1
+- note 1.0.0 bump from 0.11.1
+- arm64
+- pullOnlyOnChanges
+- extraFiles
+- fullnameOverride
+- seeding of secrets
+
+  NOTE about the need to do one upgrade before removing secrets.
 
 #### Security announcement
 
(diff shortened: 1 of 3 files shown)


manics:
Are we definitely going for 12.0.0 not 1.0.0? I understand the reasoning for that change in lower level components which are often hidden from end-users, but for a top-level distribution if we go straight to 12.0.0 I think we'll need more explanation.

Also worth noting that up until https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2165 we were saying Z2JH was still an alpha release, so in that sense we could justify a proper 1.0.0 release.

consideRatio:
Haha yes i agree @manics, it is worth celebrating - no matter what but 1.0.0 feels more worthy of a celebration in a way which is also the crux in my mind.

These are my emotional response to the differences:
- less pressure on cutting 12.0.0 than 1.0.0, we dont need to meet all the dreams and wishes we have and get it over with easier
- feels more acceptable/less pressure to cut next major version version for 12.0.0 to 13.0.0 than 1.0.0 to 2.0.0

I'm very open to any choice among 12.0.0 and 1.0.0 but not very open to delaying a release or cutting 0.12.0 instead of a major version. 

Thanks for raising this discussion @manics, i'll go ahead and ping @jupyterhub/jupyterhubteam - 12.0.0 vs 1.0.0 ?

(Kubespawner and Dockerspawner went from a shift like this)

choldgraf:
I think that as long as we are explicit in how this team defines "1.0.0" then that is the right path forward to avoid confusion. We should document what the decision means (even if we say something like "this isn't a particularly special release other than we are switching to a more standard interpretation of semver, and this 1.0 status doesn't mean we promise long term support or anything" etc. Does that make sense?

willingc:
Thanks @consideRatio for taking this on.

CHANGELOG.md
```diff
@@ -2,6 +2,182 @@
...
+
+- note 12.0.0 bump from 0.11.1
+
+#### Security announcement
+
+FIXME: describe AKS cluster setup documentation issue.
+
+#### Breaking changes
+
+- **Kubernetes 1.17+ and Helm 3.1+ is now required**
```
willingc, line 19:
```suggestion
- **Kubernetes 1.17+ and Helm 3.1+ are now required**
```

willingc (commented):
Great detail on the release note. Thanks @consideRatio

consideRatio:
Thumb emoji democracy have made it clear - we go for a 1.0.0 release!

### Action points and decisions to make
- JupyterLab 3 as default https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776?
  
  Should we do it for this major release, and if so, what is the practical change required and its consequences? Is it to set `singleuser.cmd` to `jupyter-labhub`, and would that work with JupyterLab 2 and 3?
  
  I wonder if perhaps @manics or @yuvipanda have some suggestions about this, I'm a bit lost in the complexities of:
  jupyter_server, JupyterLab version compatibility, `jupyter-labhub` as a entrypoint script and availability of it, and consequences of having `KubeSpawner.cmd` unset in KubeSpawner that stops passing `KubeSpawner.default_url` information `cmd` isn't set.
- KubeSpawner delete_forever feature inclusion: https://github.com/jupyterhub/kubespawner/pull/475
  - I've already asked Min to look at it and give input regarding if a JupyterHub 1.4.1 release is required or not for the feature to work.
  - A new release of KubeSpawner is required
- Bumping versions of various dependencies in Z2JH: CHP, KubeSpawner, etc.

manics:
When we made the decision to switch to JupyterLab 3 I thought it'd be easy, just set `JUPYTER_ENABLE_LAB=1` so docker-stacks images would pick it up, and custom images would just ignore it. It turns out that doesn't work because it's not enabled for JupyterHub in docker-stacks, and even if it was it wouldn't work because Z2JH overrides the command and ignores the startup scripts.

My next thought was to take Z2JH out of the equation and delegate everything to the Docker image, which would make future changes easier to manage. Unfortunately that ran into a whole load of other problems as detailed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2138. Having been burnt by that I'm now weary of introducing another override of the image defaults (`defaultUrl`) in case it causes problems for us in future.

My final concern is that we're introducing a breaking change quite late in the release process. JupyterHub 1.4.1 is out which means https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2138 can be released, so I think we're pretty much ready to release Z2JH, and the [current list of breaking changes](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pulls?q=is%3Apr+is%3Amerged+label%3Abreaking+closed%3A%3E%3D2021-01-13+) doesn't look too major compared to switching to JupyterLab 3.

Having said all that I won't block the change to `defaultUrl` if everyone else is in favour of it :smile_cat:

damianavila:
>  UPDATE: We wait!

+1 on this one :wink:
I do think there is still some time needed for the new server story to consolidate.
I have seen issues with server extensions supporting the new jupyter server but not the old one and cryptic failures when you have started the legacy server and your extension no longer work.

yuvipanda:
I'm happy to wait too :)

consideRatio:
The "UPDATE: we wait" was written before @yuvipanda suggested just modifying to a `/lab` singleuser.defaultUrl.

I'm delegating this to anyone else, but if we have a decision to wait with that also, then it would unblock a z2jh release process.

damianavila:
>  @damianavila did input to wait apply to a suggestion of just setting the defaultUrl to /lab as well, or was it to the more complicated discussion?

@consideRatio, my input was related to using the new server, specifically.
Sorry to not be clear enough about that!

> do we wait with ALL kinds of changes related to JupyterLab as default?

I think that is a sensible approach.

Setting up the defaultURL to /lab is less "potentially" problematic if we really want to do that... but I could imagine people seeing Lab as default experience and installing server extensions **only** compatible with the new jupyter server and somehow they would end up in a broken state because their extensions were not properly loaded.
Maybe, I am being over precautions here... but that usually helps with projects widely used as this one :wink:

CHANGELOG.md
```diff
@@ -2,6 +2,245 @@
...
+
+  NOTE about the need to do one upgrade before removing secrets.
+
+#### Security announcement
+
+FIXME: describe AKS cluster setup documentation issue.
+
+#### Breaking changes
+
+- **Kubernetes 1.17+ and Helm 3.1+ are now required**
```
manics, line 26:
```suggestion
- **Kubernetes 1.17+ and Helm 3.5+ are now required**
```

CHANGELOG.md
```diff
@@ -2,6 +2,245 @@
...
+
+  Helm 3 (3.5+) is now required. Helm 2 reached end of life last year and we
+  have started relying on Helm 3 specific features.
```
manics, line 29:
```suggestion
  have started relying on Helm 3.5 specific features.
```

manics (commented):
Noticed a couple of typos. Let me know when you want a full review.

consideRatio:
I marked this as ready for review, as it is read for accepting contributions in any form. I updated a checklist in the original post of what remains in my mind before we can cut a release.

CHANGELOG.md
```diff
@@ -2,6 +2,252 @@
...
+    expected a string.
+
+- default requests:
+  https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2034
+
+  prePuller.resources (0, 0)
+  prePuller.hook.resources (0, 0)
+  scheduling.userScheduler.resources (50m, 256Mi)
+  hub.resources (200m, 512Mi)
+  proxy.resources (200m, 512Mi)
```
manics, line 71:
```suggestion
- **Default resource requests are no longer set** [#2034](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2034)

  Z2JH now follows Helm chart best practice in not setting default resource limits.
  If you want to restore the old behaviour you can set:
  ```yaml
  hub:
    resources:
      requests:
        cpu: 200m
        memory: 512Mi

  proxy:
    chp:
      resources:
        requests:
          cpu: 200m
          memory: 512Mi

  scheduling:
    userScheduler:
      resources:
        requests:
          cpu: 50m
          memory: 256Mi

  prePuller:
    resources:
      requests:
        cpu: 0
        memory: 0
    hook:
      resources:
        requests:
          cpu: 0
          memory: 0
```
As mentioned on https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2085#issuecomment-846462083 I don't think we need to block this release on having full documentation for resources

CHANGELOG.md
```diff
@@ -2,6 +2,252 @@
...
+
+- **KubeSpawner and deletion of PVCs** ([jupyterhub#3337](https://github.com/jupyterhub/jupyterhub/pull/3337), [kubespawner#475](https://github.com/jupyterhub/kubespawner/pull/475))
+
+  Deleting a user in JupyterHub's admin interface (/hub/admin) or removing a
+  named server will now lead to the deletion of the user's or named server's
+  dynamically created PVC resource if there was one.
+
+  To opt out of this behavior and retain the current behavior where dynamically
+  created PVC resources will remain, configure KubeSpawner's `delete_pvc`
+  configuration to be false.
```
manics, line 81:
```suggestion
  created PVC resources will remain, set `KubeSpawner.delete_pvc` to `false`.
```

CHANGELOG.md
```diff
@@ -2,6 +2,252 @@
...
+
+  ```yaml
+  hub:
+    config:
+      KubeSpawner:
+        delete_pvc: false
+  ```
+
+  Note that this feature relies on both KubeSpawner 1.0.0+ and JupyterHub 1.4.1+
+  which is now used as part of this release.
```
manics, line 91:
```suggestion
  which are included in this release.
```

manics (commented):
I'm worried this CHANGELOG is going to become unreadable other than for the current release, it's length already makes it difficult to scroll through to find older information.

Not a blocker for this release but worth thinking about, e.g. could we archive the changelog for previous major releases into a different file? Or have a table of contents at the top?

CHANGELOG.md
consideRatio: I updated the suggestion to use four backticks on the suggestion block so that the three backticks within the suggestion wouldn't break the suggestion block.

consideRatio:
> I'm worried this CHANGELOG is going to become unreadable other than for the current release, it's length already makes it difficult to scroll through to find older information.

I opened #2221 to represent this concern and my suggestion for addressing it concretely.

CHANGELOG.md
```diff
@@ -2,6 +2,297 @@
...
+| [ltiauthenticator](https://github.com/jupyterhub/ltiauthenticator)               | 1.0.0             | 1.0.0             | [Changelog](https://github.com/jupyterhub/ltiauthenticator/blob/master/CHANGELOG.md)        | Run in the `hub` pod               |
+| [nativeauthenticator](https://github.com/jupyterhub/nativeauthenticator)         | 0.0.6             | 0.0.7             | [Changelog](https://github.com/jupyterhub/nativeauthenticator/blob/master/CHANGELOG.md)     | Run in the `hub` pod               |
+| [jupyterhub-idle-culler](https://github.com/jupyterhub/jupyterhub-idle-culler)   | 1.0               | 1.1               | -                                                                                           | Run in the `hub` pod               |
+| [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy) | 4.2.2             | 4.3.2             | [Changelog](https://github.com/jupyterhub/configurable-http-proxy/blob/master/CHANGELOG.md) | Run in the `proxy` pod             |
```
manics, line 153:
```suggestion
| [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy) | 4.2.2             | 4.4.0             | [Changelog](https://github.com/jupyterhub/configurable-http-proxy/blob/master/CHANGELOG.md) | Run in the `proxy` pod             |
```

CHANGELOG.md
```diff
@@ -2,6 +2,297 @@
...
+        delete_pvc: false
+  ```
+
+  Note that this feature relies on both KubeSpawner 1.0.0+ and JupyterHub 1.4.1+
+  which are included in this release.
+
+- **hub.existingSecret is reworked** ([#2042](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2042))
+
+  See [the documentation](http://z2jh.jupyter.org/en/latest/resources/reference.html#hub-existingsecret) and [pull request #2042](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2042) for more details.
+
```
manics, line 141:
```suggestion

 - **configurable-http-proxy statsd metrics replaced by prometheus metrics** ([#2231](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2231))
 
   [statsd metrics have been removed and support for Prometheus metrics added](https://github.com/jupyterhub/configurable-http-proxy/pull/314) in configurable-http-proxy.
```

manics (commented):
Suggestions for https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2231 (feel free to edit)

CHANGELOG.md
consideRatio: The software supports it, if enabled. But, the Helm chart doesn't expose it properly yet etc so there are some struggles still.

I opened #2232 to chart integration with the the CHP metrics to properly expose it to prometheus running in a k8s cluster.

manics, line 141:
```suggestion

 - **configurable-http-proxy statsd metrics removed** ([#2231](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2231))
 
   [statsd metrics have been removed](https://github.com/jupyterhub/configurable-http-proxy/pull/314) in configurable-http-proxy. This will only affect administrators who have overridden the CHP command line arguments as statsd is not supported in the Helm chart. Support for Prometheus metrics will be added in a future release.
```
How about this, so we don't block the release?

PR #2179, opened 2021-05-06
enable prePuller.hook.pullOnlyOnChanges by default

consideRatio:
By enabling this feature by default, those with `prePuller.hook.enabled=true` will not need to wait ~10 seconds on `hook-image-awaiter` and `hook-image-puller` pods to startup and shut down if no new image has been configured.

In practice, this feature works by inspecting the rendered Helm template of the hook-image-puller DaemonSet for a change to anything besides labels since the last `helm upgrade` was made.

This feature also relies on the ability for us to lookup a value in a ConfigMap using the Helm 3 feature called `lookup`.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -79,6 +79,7 @@ jupyterhub_version = chart["appVersion"]
 #        run into this issue:
 #        https://github.com/executablebooks/MyST-Parser/issues/282
 kube_version = chart["kubeVersion"].split("-", 1)[0][2:]
+helm_version = "3.5"  # minimum helm cli version
 
 # These substitution variables only work in markdown contexts, and does not work
 # within links etc. Reference using {{ variable_name }}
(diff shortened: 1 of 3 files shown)


yuvipanda:
Yay!

What's the minimum helm version for this? I had to update from 3.2 to get this to work - but not sure what's the minimum version

consideRatio:
Oh helm 3.1 introduced the lookup function.

yuvipanda:
oh, hmm - not sure why I needed at least 3.3... Either way, do we mention that somewhere? Happy to merge this once we have that...

consideRatio:
Hmmm i have mentioned 3.3, but there is no clear signal on whats required besides at least 3.1.

If you could conclude 3.3 was required that is good for us to know. Note that you will always need to make one upgrade with the feature bugfix, and only after that this flag will make a difference. (A checksum must be updated correctly and stored in the hub configmap, which later can be found to be the same)

yuvipanda:
@consideRatio yeah, with 3.2 I got an error in rendering the daemonset helper. I migraetd to latest (3.5) and it went away. Unfortunately I don't have those logs anymore

consideRatio:
@yuvipanda I've concluded that 3.4.X doesn't work but 3.5.0 works. As upgrading to a recent `helm` 3 binary is easy, I suggest we accept a requirement on helm 3.5 and above. I'm making sure this is documented and we test against the minimally known supported helm version going onwards.

yuvipanda:
+1 on a minimum helm 3.5 requirement.

yuvipanda:
yay! :)

PR #2180, opened 2021-05-06
maint: revert a workaround to make our priorityclass resources helm hooks

### LINKED ISSUE #1035: Our PriorityClass resources are Helm hooks, but shouldn't be any more
I made the PriorityClass resources Helm hooks to ensure they are installed before any pod would reference them, but that may not be needed any more as Helm may have added logic to always prioritize the creation of these anyhow by now. I don't know though.

If that is the case, we should stop installing them as Helm hooks, since this causes some trouble. For example, if you would disable `scheduling.podPriority.enabled` then there would be resources remaining...

We should also consider using the `.Capabilities` function to evaluate if K8s 1.11 or higher is installed, since that is required for `scheduling.podPriority.enabled` and fail in a nicer manner instead.

consideRatio:
At one point in time, `helm` installed PriorityClass resources too late - after they were referenced. Due to that, we added a workaround to make them _helm hooks_ - resources that could be installed _pre-upgrade_. The need for this workaround is long gone though, and the resources can just as well be normal resources. This PR makes them normal resources after having been helm hook resources.

I wasn't sure this would work without name collisions, but it seems like `helm` 3 manages this fine as our CI tests making a `helm` upgrade didn't have any issues.

Closes #1035.

---

Note that the commit 62b41b5 just addressed an intermittent unrelated harmless CI failure.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -280,7 +280,7 @@ jobs:
         uses: jupyterhub/action-k8s-await-workloads@v1
         with:
           timeout: 150
-          max-restarts: 0
+          max-restarts: 1
 
       - name: Await local chart cert acquisition
         run: |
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks, @consideRatio

PR #2186, opened 2021-05-10
ci: fix permissions of PR creating action

consideRatio:
```yaml
    # Write permissions granted for the peter-evans/create-pull-request action
    # to push to a branch and create/update a PR
    permissions:
      contents: write
      pull-requests: write
```

Both ability to push to a branch and create/update pull requests is required for an action in the vuln-scan workflow. This PR fixes that. It also removes some inline comments that would become outdated quickly.

### DIFF
--- a/.github/workflows/vuln-scan.yaml
+++ b/.github/workflows/vuln-scan.yaml
@@ -18,8 +18,11 @@ jobs:
   trivy_image_scan:
     if: github.repository == 'jupyterhub/zero-to-jupyterhub-k8s'
     runs-on: ubuntu-20.04
+    # Write permissions granted for the peter-evans/create-pull-request action
+    # to push to a branch and create/update a PR
     permissions:
       contents: write
+      pull-requests: write
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/vuln-scan.yaml
```diff
@@ -171,7 +174,7 @@ jobs:
       # ref: https://github.com/jacobtomlinson/gha-find-replace
       - name: Update VULN_SCAN_TIME in Dockerfile
         if: steps.analyze.outputs.proceed == 'yes'
-        uses: jacobtomlinson/gha-find-replace@f9e200cf233bcde71011fa5e4178037881764379 # associated tag: 0.1.3
+        uses: jacobtomlinson/gha-find-replace@f9e200cf233bcde71011fa5e4178037881764379
```
manics, line 177: Is this action just doing an in-place sed?

manics (approved):
LGTM

.github/workflows/vuln-scan.yaml
consideRatio: Yepp, it can certainly be replaced a `sed` script

consideRatio: Do you agree that we should stop using a dedicated action for this and use a `sed` script or similar instead, that doesn't involve a dedicated action that could contain a vulnerability etc?

PR #2200, opened 2021-05-15
schema: catch typos in template values using JSONSchema's `additionalProperties`

### LINKED ISSUE #2199: Schema validation to catch typos etc!
I just learned about JSONSchema's `additionalProperties`! It allows us to declare that we don't accept values besides those declared in the schema for a certain object with declared properties.

This can enable what we have wanted for a very long time, something that spots typos! Since we now have a values.schema.json file that covers almost all properties on all declared objects, we can actually set `additionalProperties: false` on those objects.

## Example

If a helm chart contained the following values.schema.json file, and the user provided the values.yaml file below, it would raise an error because it detected a unknown property - the misspelled `fullnameOveride`.

```yaml
# a YAML representation of the values.schema.json file to ship with a Helm chart
# and used to validate all the Helm chart's passed values
#
$schema": http://json-schema.org/draft-07/schema#
type: object
additionalProperties: false
required:
  - fullnameOverride
properties:
  fullnameOverride:
    type: [string, "null"]
```

```yaml
# a values.yaml file to be validated by the schema, as automated by `helm upgrade` etc
fullnameOveride: my-test-name
```

## Action points

What do you think about making this part of the 1.0.0 release where we introduce schema validation in the first place? I think it would be good to introduce this together with the other schema validation logic that only catches setting invalid data formats on known properties etc.

- [x] Go through the schema.yaml file and set `additionalProperties: false` wherever we feel confident we cover all properties.
- [ ] Trial this out in a 1.0.0 release candidate

## Related
- #2033

## Technical notes
- Someone passes values to access them from the hub pod will be required to do it under `custom`. A sensible practice to enforce, it is documented to be used for that already.
- I'm able to debug loads of issues by using lint-and-validate-values.yaml and the validation script in `tools`. I realize quickly we must allow additionalProperties whenever we have something like `labels`, `annotations`, `nodeSelector`, `resources`, etc where there can show up user specific values or in other cases when we don't have a schema representing all of the k8s details such as for the `livenessProbes` and such.
- I consider this a breaking change because it will make upgrades fail etc if they have typos.

consideRatio:
We have a `schema.yaml` file that we render into `values.schema.json` and ship with our Helm chart as supported by Helm 3. This file is a JSONSchema that covers _all our charts configuration options_ since #2033. `helm` will use it during `helm template` and `helm install|upgrade` to catch values that the schema considers invalid.

What this PR does is to start make use of a JSONSchema feature called [`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#properties). With `additionalProperties: false` we can make the schema invalidate any `object` type field that declare a property that isn't declared in the schema.

Since we have declared almost all properties, we can set `additionalProperties: false` on all those `object` type fields and catch all kinds of typos _before_ `helm install`, `helm upgrade`, or `helm template` would start requesting changes to the k8s api-server.

Closes #2199

## Suggested action points

- [x] We test it against deployments we control - _any failures will be caught before k8s api-server is contacted!_
- [ ] We merge this without much delay
- [ ] We release 1.0.0-beta.1
- [ ] We release 1.0.0

## Example UX

```
$ helm template jh jupyterhub --set singleuser.image.tagg=hello --set proxxy.secretToken=hello 

Error: values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- (root): Additional property proxxy is not allowed
- singleuser.image: Additional property tagg is not allowed
```

## Local testing

```shell
# By generating the values.schema.json file and
# testing against charts default values
tools/generate-json-schema.py
helm template jh jupyterhub --set proxxy.secretToken=hello

# By using the charts default values but combined with the
# tools/templates/lint-and-validate-values.yaml file
tools/validate-against-schema.py
```

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -12,7 +12,6 @@ on:
       - "doc/**"
       - "**/test-docs.yaml"
       - "**.md"
-      - "**/schema.yaml"
       - ".github/workflows/*"
       - "!.github/workflows/test-chart.yaml"
   push:
@@ -20,7 +19,6 @@ on:
       - "doc/**"
       - "**/test-docs.yaml"
       - "**.md"
-      - "**/schema.yaml"
       - ".github/workflows/*"
       - "!.github/workflows/test-chart.yaml"
     branches-ignore:
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -180,8 +180,8 @@ if image:
 # Combine imagePullSecret.create (single), imagePullSecrets (list), and
 # singleuser.image.pullSecrets (list).
 image_pull_secrets = []
-if get_config("imagePullSecret.automaticReferenceInjection") and (
-    get_config("imagePullSecret.create") or get_config("imagePullSecret.enabled")
+if get_config("imagePullSecret.automaticReferenceInjection") and get_config(
+    "imagePullSecret.create"
 ):
     image_pull_secrets.append(get_name("image-pull-secret"))
 if get_config("imagePullSecrets"):
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -849,8 +849,9 @@ properties:
       uid:
         type: [integer, "null"]
         minimum: 0
+        deprecated: true
         description: |
-          DEPRECATED: Use hub.containerSecurityContext instead.
+          DEPRECATED: Use [`hub.containerSecurityContext`](schema_hub.containerSecurityContext) instead.
 
           The UID the hub process should be running as.
 
@@ -1110,6 +1111,12 @@ properties:
         type: object
         additionalProperties: false
         description: *jupyterhub-native-config-description
+      imagePullSecret: &deprecated-imagePullSecret
+        type: object
+        additionalProperties: true
+        deprecated: true
+        description: |
+          DEPRECATED: Use [`imagePullSecret`](schema_imagePullSecret) instead.
 
   proxy:
     type: object
@@ -1573,6 +1580,12 @@ properties:
           containerSecurityContext: *containerSecurityContext-spec
           image: *image-spec
           resources: *resources-spec
+      pdb:
+        type: object
+        additionalProperties: true
+        deprecated: true
+        description: |
+          DEPRECATED: Use [`proxy.chp.pdb`](schema_proxy.chp.pdb) instead.
 
   singleuser:
     type: object
@@ -1940,6 +1953,7 @@ properties:
           sudo rights for some of your users. This can be done by starting up as
           root, enabling it from the container in a startup script, and then
           transitioning to the normal user.
+      imagePullSecret: *deprecated-imagePullSecret
 
   scheduling:
     type: object
@@ -2417,7 +2431,7 @@ properties:
 
   global:
     type: object
-    additionalProperties: false
+    additionalProperties: true
     properties:
       safeToShowValues:
         type: boolean
@@ -2425,3 +2439,10 @@ properties:
           A flag that should only be set to true temporarily when experiencing a
           deprecation message that contain censored content that you wish to
           reveal.
+
+  auth:
+    type: object
+    additionalProperties: true
+    deprecated: true
+    description: |
+      DEPRECATED: The Authenticator should be configured using [`hub.config`](schema_hub.config) instead.
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -851,7 +851,9 @@ properties:
         minimum: 0
         deprecated: true
         description: |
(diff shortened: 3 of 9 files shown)


yuvipanda:
Awesome :)

How do we test this against github.com/2i2c-org/pilot-hubs/ and github.com/berkeley-dsep-infra/datahub/? Their config is already public...

consideRatio:
@yuvipanda the easiest that I suggest is that we merge this and iterate after merge.

The easiest form of test is to `helm template` referencing all your values and the helm chart with a bundled `values.schema.json`.

How to reference a non-merged chart?
- The chart is packaged and made available here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/suites/2750369571/artifacts/60820128. This is an artifact for the publish job when triggered as part of a pull request. I think one can point to a packaged file like this.
- The `values.schema.json` file needs to be generated if you clone the repo and reference a local version.

How to test specifically against github.com/berkeley-dsep-infra/datahub/ for example?
- I suggest you create a workflow that has a job running `helm template` once for each deployment. Then you will get feedback for every deployment without aborting a job midway.

consideRatio:
@yuvipanda I tried this locally and the template validation worked fine, but when actually doing an upgrade I ran into a breaking change form a PR that I didn't think would be breaking.

This is being addressed in #2201.

consideRatio:
@yuvipanda I brainfarted and didn't realize I could trial this myself. I tried the basehub Helm chart in pilot-hubs using the default values.

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- (root): Additional property cloudResources is not allowed
- singleuser: Additional property admin is not allowed
```

You can keep passing information to the hub pod through `custom.anything-you-choose-here`, but to inject variables in the root context or singleuser isn't allowed. The added options could in practice be typo's so I think it is correct for the schema to complain here.

---

I also tried the `hub` Helm chart in berkeley-dsep-infra/datahub using its default values. No validation errors at all there!

consideRatio:
@yuvipanda this is now tested successfully against 2i2c (basehub, daskhub, + manual inspection of deployments configs), datahub's default config, and a deployment i manage for Sandvik.

I'll do some work today that remains before we can cut 1.0.0-beta.1 but besides that work this is the remaining piece that I'd like to have merged for such beta release.

manics:
Sounds good to me! I don't have anywhere suitable to test this, but I trust @yuvipanda's review :smile:

yuvipanda:
I'm very happy to have this!

This is a breaking change though, since config that deployed earlier will no longer deploy. Can we document this as a breaking change? Happy to merge after that

consideRatio:
@yuvipanda I plan to put this in the CHANGELOG.md which is drafted in another PR currently. Would you like me to make a stub in the changelog from here and then merge them?

consideRatio:
Preview of WIP of the changelog regarding breakign changes for this PR: https://github.com/consideRatio/zero-to-jupyterhub-k8s/blob/pr/12.0.0-release/CHANGELOG.md#breaking-changes

yuvipanda:
The test failure looks unrelated. THIS IS AWESOME, THANKS @consideRatio!

consideRatio:
Thank you @yuvipanda and @manics! :heart:

mriedem:
I'm working on upgrading from z2jh 0.11.1 to 1.2.0 and hit this error:

```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):

jupyterhub:

- (root): Additional property profiles is not allowed

- (root): Additional property secrets_checksum is not allowed

- (root): Additional property values_checksum is not allowed
```

I _think_ I can get around the `profiles` one which is just a JSON string we pass through `hub.extraEnv` and is read by our custom authenticator, it eventually sets `c.KubeSpawner.profile_list`. Why we don't just do this using `hub.config.KubeSpawner.profile_list` I'm not sure, except our custom authenticator does some logic based on the profile list. 🤷 

The bigger issue I have is our deployment process runs `helm upgrade` with calculated `secrets_checksum` and `values_checksum` values which are hashes of the (encrypted) secrets and (unencrypted) values files which we use. The idea being we trigger an upgrade if the checksum changes because something in one of the values files changed. Specifying the extra value on the command line is triggering the jsonschema validation error.

Is the solution to just specify those in [`custom`](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/advanced.html#custom-configuration), e.g.:

```yaml
custom:
  secrets_checksum: ''
  values_checksum: ''
```

It seems like this might be a decent advanced sub-topic for https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/extending-jupyterhub.html#applying-configuration-changes.

consideRatio:
> Is the solution to just specify those in [`custom`](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/advanced.html#custom-configuration), e.g.:
> 
> ```yaml
> custom:
>   secrets_checksum: ''
>   values_checksum: ''
> ```

Correct! All things not part of the chart, but you want to pass to as values so they can be accessed by the hub pod, should go there.

PR #2208, opened 2021-05-18
fix: remove no longer needed safeguard for enable_auth_state

consideRatio:
This fixes a bug where we could not properly use a seeded cookie secret because we had a failsafe checking that we explicitly had set a cookie secret. Since we now seed those secrets if they aren't explicitly set, this is no longer relevant.

### DIFF
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -115,12 +115,6 @@ directly instead.
 {{ "HARD DEPRECATION: proxy.pdb has renamed to proxy.chp.pdb" | fail }}
 {{- end }}
 
-{{- if include "jupyterhub.digToTrue" (list "hub.config.Authenticator.enable_auth_state" .Values) }}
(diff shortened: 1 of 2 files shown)


jupyterhub/templates/NOTES.txt
```diff
@@ -115,12 +115,6 @@ directly instead.
 {{ "HARD DEPRECATION: proxy.pdb has renamed to proxy.chp.pdb" | fail }}
 {{- end }}
 
-{{- if include "jupyterhub.digToTrue" (list "hub.config.Authenticator.enable_auth_state" .Values) }}
-{{- if not (include "jupyterhub.digToTrue" (list "hub.config.CryptKeeper.keys" .Values)) }}
-{{ "Using Authenticator.enable_auth_state, you also need to CryptKeeper.keys to a list with at least one key in it." | fail }}
-{{- end }}
-{{- end }}
-
```
consideRatio, line 123: Note that `jupyterhub.digToTrue` and `jupyterhub.digToJson` that it in turn relied on, is now no longer needed thanks to a Helm 3.5 requirement. Helm 3.5 has a version of _sprig_ (helper library) that includes `dig`.

PR #2211, opened 2021-05-19
Remove deprecated logic and emit clear messages

### LINKED ISSUE #1412: System to deprecate Helm value config
We need a way to systematically deprecate helm chart configuration. Here is [one approach I implemented in mattermost-team-edition](https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-team-edition/templates/_deprecations.tpl).

1. We render individual deprecation messages in individual templates.
2. We render all these individual deprecation messages inside a collection template.
3. We conditionally fail the rendering of the chart templates if the collection template isn't empty. using the [fail directive](http://masterminds.github.io/sprig/flow_control.html), which is part of the [sprig library](http://masterminds.github.io/sprig/) available for use in Helm templates.

Further, we could also consider having two collection templates, one for _hard deprecations_ and one for _soft deprecations_ or _warnings_. This was also [something I added](https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-team-edition/templates/_warnings.tpl) to mattermost-team-edition. A soft deprecation or warning would be something that the user would be notified about but that would not stop the upgrade process. A hard deprecation on the other hand would be something that if noticed would trigger the helm chart rendering to fail.

## Useful technical know how
- If you render a template, its output is always a string, even though its a valid YAML string. I've not realized this myself until very recently when @yuvipanda used it in #1409, but if a template has rendered a string you can use the `fromYaml` function to go back to a object you can inspect. The `fromYaml` function that Helm [is injected](https://github.com/helm/helm/blob/47ee538590e02854d595a44e6cf1ccd876084a5d/pkg/engine/engine.go#L79-L85) from [their chartutil GO class](https://godoc.org/k8s.io/helm/pkg/chartutil#FromYaml)) for use in Helm templates. We could potentially use this to combine the output of individual deprecations smarter, for example if they would render to a YAML structure that defined if it should be treated as a soft deprecation/warning, hard deprecation.
- [Helm's documentation relating to _partials_ and _templates_](https://helm.sh/docs/chart_template_guide/#partials-and-files).

## PR work to do
- [ ] Add a `_errors_and_warnings.tpl` file or similar.
- [ ] Document and support a demonstrative example for both a warning and an error and how to add new ones.
- [ ] Ensure to run a `fail` template directive if there is hard deprecations / errors. Note that if you don't render the template logic deciding if the `fail` directive will render, you won't stop an helm upgrade process.
- [ ] Add a place in NOTES.txt rendering the collection of soft deprecations / warnings and the collection of hard deprecations / errors.

### LINKED ISSUE #2157: Visually standing out warnings / deprecations etc in NOTES.txt
When you make an upgrade using the Prometheus Helm chart and they want to convey something, they do it like this. It really grabs the attention, so I think perhaps we should aim to do something similar for various deprecations.

```
The Prometheus server can be accessed via port 80 on the following DNS name from within your cluster:
prometheus-server.prometheus.svc.cluster.local


Get the Prometheus server URL by running these commands in the same shell:
  export POD_NAME=$(kubectl get pods --namespace prometheus -l "app=prometheus,component=server" -o jsonpath="{.items[0].metadata.name}")
  kubectl --namespace prometheus port-forward $POD_NAME 9090


#################################################################################
######   WARNING: Pod Security Policy has been moved to a global property.  #####
######            use .Values.podSecurityPolicy.enabled with pod-based      #####
######            annotations                                               #####
######            (e.g. .Values.nodeExporter.podSecurityPolicy.annotations) #####
#################################################################################



For more information on running Prometheus, visit:
https://prometheus.io/
Context "gke_ds-platform_europe-west4_ds-platform" modified.
Removing secret-values.yaml
```

Their NOTES.txt is defined [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/NOTES.txt).

consideRatio:
### Summary

- [x] Forcing deprecated values to become unset during template rendering and providing a clear error message if they aren't.
- [x] Ensuring all messages are shown together to help users fix all failures at once instead of one at the time.
- [x] Removes no longer needed logic to handle deprecated values

### Motivation for breaking changes

- __Complexity reduction of templates by removal of deprecation logic.__
- __Future complexity reduction of schema.__
  Our deprecation messages are triggered by detecting deprecated values being passed, which requires them to be accepted by the schema. So by ensuring all users have stopped using deprecated values when they upgrade to 1.0.0, we can in 2.0.0 remove them from the schema.

### Breaking changes

All of the breaking changes will lead to clear error messages emitted by a `fail` statement in `NOTES.txt`. This happens before any communication to the k8s api-server has been made.

- Previously deprecated options that were required to be falsy, are now required to be _unset_.
  - `auth` -> `hub.config`
  - `hub.imagePullSecret` -> `imagePullSecret` (new chart wide config)
  - `singleuser.cloudMetadata.enabled` -> `singleuser.cloudMetadata.blockWithIptables` (renamed / inverted value)
  - `singleuser.imagePullSecret` -> `imagePullSecret` (new chart wide config)
  - `proxy.containerSecurityContext` -> `proxy.chp.containerSecurityContext` (relocated)
  - `proxy.networkPolicy` -> `proxy.chp.networkPolicy` (relocated)
  - `proxy.pdb` -> `proxy.chp.networkPolicy` (relocated)
- Previously deprecated options that were still functional are now _required_ to be unset.
  - `hub.extraConfigMap` is now required to unset and configured using `custom` (renamed)
  - `hub.uid` is now required to be unset and configured using `hub.containerSecurityContext.runAsUser` (coalesced into k8s native config)

Closes #2157 by providing more visually standing out messages for warnings and breaking changes.
Closes #1412 by making NOTES.txt failures grouped into a single message, which I think is sufficient for now.

### DIFF
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -80,12 +80,8 @@ WARNING: Configuring proxy.https without setting proxy.https.enabled to true is 
   Deprecation messages that can be removed in 2.0.0
 */}}
 
-{{- if .Values.hub.extraConfigMap }}{{ println }}
-DEPRECATION: hub.extraConfigMap is deprecated in jupyterhub chart 0.8.
-Use top-level `custom` instead:
-
-custom:
-{{- (merge dict .Values.custom .Values.hub.extraConfigMap) | toYaml | nindent 2 }}
+{{- if hasKey .Values.hub "extraConfigMap" }}
+{{- fail "\n\nHARD DEPRECATION: hub.extraConfigMap has been renamed to custom" }}
 {{- end }}
 
 
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -106,7 +106,7 @@ WARNING: Configuring proxy.https without setting proxy.https.enabled to true is 
 
 
 {{- if hasKey .Values.hub "uid" }}
-{{- print "\n\nDEPRECATION: hub.uid is deprecated in jupyterhub chart 0.9. Set the hub.containerSecurityContext.runAsUser value directly instead." }}
+{{- fail "\n\nHARD DEPRECATION: hub.uid must as of 1.0.0 be configured using hub.containerSecurityContext.runAsUser" }}
 {{- end }}
 
 
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -86,7 +86,7 @@ If you have questions, please:
 
 
 {{- /*
-  Breaking changes in 1.0.0
+  Breaking changes.
 */}}
 
 {{- $breaking := "" }}
@@ -152,11 +152,3 @@ If you have questions, please:
 {{- if $breaking }}
 {{- fail (print $breaking_title $breaking) }}
 {{- end }}
-
-
-
-
-
-{{- /*
-  Deprecation messages that can be removed in 3.0.0
-*/}}
--- a/jupyterhub/templates/NOTES.txt
+++ b/jupyterhub/templates/NOTES.txt
@@ -77,55 +77,64 @@ WARNING: Configuring proxy.https without setting proxy.https.enabled to true is 
 
 
 {{- /*
-  Deprecation messages that can be removed in 2.0.0
+  Deprecation messages that can be removed in 2.0.0 because they have been
+  strictly enforced in 1.0.0.
 */}}
 
+{{- $hard_deprecations := "" }}
+
+
 {{- if hasKey .Values.hub "extraConfigMap" }}
-{{- fail "\n\nHARD DEPRECATION: hub.extraConfigMap has been renamed to custom" }}
+{{- $hard_deprecations = print $hard_deprecations "\n\nHARD DEPRECATION: hub.extraConfigMap has been renamed to custom" }}
 {{- end }}
 
 
 {{- if hasKey .Values "auth" }}
 {{- if .Values.auth }}
-{{- fail (include "jupyterhub.authDep.remapOldToNew" .) }}
+{{- $hard_deprecations = print $hard_deprecations (include "jupyterhub.authDep.remapOldToNew" .) }}
 {{- else }}
-{{- fail "\n\nHARD DEPRECATION: Please remove the empty 'auth' config" }}
+{{- $hard_deprecations = print $hard_deprecations "\n\nHARD DEPRECATION: Please remove the empty 'auth' config" }}
 {{- end }}
 {{- end }}
 
 
 {{- if hasKey .Values.proxy "containerSecurityContext" }}
-{{- fail "\n\nHARD DEPRECATION: proxy.containerSecurityContext has been renamed to proxy.chp.containerSecurityContext" }}
+{{- $hard_deprecations = print $hard_deprecations "\n\nHARD DEPRECATION: proxy.containerSecurityContext has been renamed to proxy.chp.containerSecurityContext" }}
 {{- end }}
 
 
 {{- if hasKey .Values.proxy "pdb" }}
-{{ fail "\n\nHARD DEPRECATION: proxy.pdb has renamed to proxy.chp.pdb" }}
+{{- $hard_deprecations = print $hard_deprecations "\n\nHARD DEPRECATION: proxy.pdb has renamed to proxy.chp.pdb" }}
 {{- end }}
 
 
 {{- if hasKey .Values.proxy "networkPolicy" }}
-{{- fail "\n\nHARD DEPRECATION: proxy.networkPolicy has been renamed to proxy.chp.networkPolicy" }}
+{{- $hard_deprecations = print $hard_deprecations "\n\nHARD DEPRECATION: proxy.networkPolicy has been renamed to proxy.chp.networkPolicy" }}
 {{- end }}
 
 
 {{- if hasKey .Values.hub "uid" }}
-{{- fail "\n\nHARD DEPRECATION: hub.uid must as of 1.0.0 be configured using hub.containerSecurityContext.runAsUser" }}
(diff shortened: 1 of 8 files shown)


consideRatio:
### Verification of deprecation logic

This deprecation logic could fail in various ways, for example by the schema validation erroring before the NOTES.txt provides a more helpful message on how to migrate the configuration.

I've used the following to test these changes.

```
helm template ./jupyterhub \
    --set hub.extraConfigMap.dummy=dummy \
    --set auth.dummy=dummy \
    --set proxy.containerSecurityContext.dummy=dummy \
    --set proxy.pdb.dummy=dummy \
    --set proxy.networkPolicy.dummy=dummy \
    --set hub.uid=0 \
    --set hub.imagePullSecret.dummy=dummy \
    --set singleuser.imagePullSecret.dummy=dummy \
    --set singleuser.cloudMetadata.enabled=true

Error: template: jupyterhub/templates/NOTES.txt:137:4: executing "jupyterhub/templates/NOTES.txt" at <fail $hard_deprecations>: error calling fail: 

HARD DEPRECATION: hub.extraConfigMap has been renamed to custom

The JupyterHub Helm chart's auth config has been reworked and requires changes.

The new way to configure authentication in chart version 0.11.0+ is printed
below for your convenience. The values are not shown by default to ensure no
secrets are exposed, run helm upgrade with --set global.safeToShowValues=true
to show them.

hub:
  config:
    WarningUnrecognizedConfig:
      dummy: '***'

For more details, please see the updated auth config documentation at:
https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html

HARD DEPRECATION: proxy.containerSecurityContext has been renamed to proxy.chp.containerSecurityContext

HARD DEPRECATION: proxy.pdb has renamed to proxy.chp.pdb

HARD DEPRECATION: proxy.networkPolicy has been renamed to proxy.chp.networkPolicy

HARD DEPRECATION: hub.uid must as of 1.0.0 be configured using hub.containerSecurityContext.runAsUser

HARD DEPRECATION: hub.imagePullSecret has renamed to imagePullSecret

HARD DEPRECATION: singleuser.imagePullSecret has renamed to imagePullSecret

HARD DEPRECATION: singleuser.cloudMetadata.enabled must as of 1.0.0 be configured using singleuser.cloudMetadata.blockWithIptables with the opposite value.

Use --debug flag to render out invalid YAML
```

consideRatio:
@yuvipanda @manics I've updated the PR summary to be a bit quicker and easier to read.

manics:
For clarity what do you think about changing `HARD DEPRECATION:` to `BREAKING CHANGE:` or `CHANGED:` as there doesn't seem to be a clear definition of "hard deprecation" e.g.  https://www.google.com/search?q=%22hard+deprecation%22

consideRatio:
@manics absolutely - _hard deprecation_ is not easy to understand or well defined! Perhaps we want to write a title of "BREAKING CHANGES", and then we write individual messages with "REMOVED" for example?


Btw, do you agree on this understanding of the word "Deprecated" - do we use that word whenever we mean something that still will remain functional but is an option that we mark as something that users should transition away from?

manics:
For me "deprecated" always means it still works for existing uses, but you probably shouldn't use it for anything new, and should consider updating existing uses to a replacement. Adding an adjective in front conveys (in an ill-defined manner since there's no standard terminology 😃) how soon you should plan to move to something else.

consideRatio:
@manics okay updated! See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2211#issuecomment-843672562 for a rendered output as it is now.

minrk (approved):
LGTM, but will give @maincs a chance to merge based on prior review.

jupyterhub/templates/NOTES.txt
```diff
@@ -58,17 +58,28 @@ If you have questions, please:
...
 
 
 {{- if eq .Values.proxy.https.enabled false }}
 {{- if or (not (eq .Values.proxy.https.type "letsencrypt")) (not (eq (.Values.proxy.https.letsencrypt.contactEmail | default "") "")) }}{{ println }}
-WARNING: Configuring proxy.https without setting proxy.https.enabled to true is no longer allowed.
+#################################################################################
+######   WARNING: proxy.https.enabled is set to false by default since      #####
+######            version 0.10.0. It is now set to false but proxy.https    #####
+######            seem to have been modified indicating you may want it     #####
+######            enabled.                                                  #####
```
manics, line 81:
```suggestion
######            has been modified indicating you may want it enabled.     #####
```

jupyterhub/templates/NOTES.txt
```diff
@@ -77,61 +88,69 @@ WARNING: Configuring proxy.https without setting proxy.https.enabled to true is
...
-{{- fail "\n\nHARD DEPRECATION: hub.imagePullSecret has renamed to imagePullSecret" }}
+{{- if hasKey .Values.hub "imagePullSecret" }}
+{{- $breaking = print $breaking "\n\nREMOVED: hub.imagePullSecret has been removed, but there is now a chart wide wide configuration named imagePullSecret" }}
 {{- end }}
 
 
-{{- if (.Values.singleuser | dig "imagePullSecret" "enabled" "") }}
-{{- fail "\n\nHARD DEPRECATION: singleuser.imagePullSecret has renamed to imagePullSecret" }}
+{{- if hasKey .Values.singleuser "imagePullSecret" }}
+{{- $breaking = print $breaking "\n\nCHANGED: singleuser.imagePullSecret has been removed, but there is now a chart wide wide configuration named imagePullSecret" }}
```
manics, line 145:
```suggestion
{{- $breaking = print $breaking "\n\nREMOVED: singleuser.imagePullSecret has been removed, but there is now a chart wide wide configuration named imagePullSecret" }}
```

manics (commented):
A couple of very minor suggestions, otherwise feel free to merge!

PR #2214, opened 2021-05-20
singleuser: pass along image pull policy from config

betamike:
The image pull policy from values was not being passed along.

### DIFF
--- a/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
+++ b/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
@@ -93,6 +93,9 @@ spec:
         {{- /* --- Pull default image --- */}}
         - name: image-pull-singleuser
           image: {{ .Values.singleuser.image.name }}:{{ .Values.singleuser.image.tag }}
+          {{- with .Values.singleuser.image.pullPolicy }}
+          imagePullPolicy: {{ . }}
+          {{- end }}
           command:
             - /bin/sh
             - -c


consideRatio:
Does it make sense to use the same policy in the image-puller daemonset pods? Hmmm, maybe maybe not.

The singleuser image pull policy should be applied to the spawned user pods, but this is configuring the image-puller daemonset that helps propare new nodes and nodes before the helm chart is upgraded

Can you describe what you want in terms of when an image is to be pulled and with what policy?

betamike:
> Can you describe what you want in terms of when an image is to be pulled and with what policy?

Sure, sounds like I'm applying this in the wrong place.

We have a custom base image we want to provide to users. To release a new version of the image we update a tag (something like `prod`). We found that after updating the tag, newly launched user pods still used the previous image (using 0.11.1). We'd like to set the pull policy to `Always` so whenever we update the tag, new pods launch with the new image.

I'll be honest that I have not fully wrapped my head around the architecture here, so maybe I'm missing some existing config that will achieve this. We currently have a block like this in our config:
```
singleuser:
  image:
    pullPolicy: Always
    name: <redacted>
    tag: prod
```

consideRatio:
Okay that should work, and that change is unrelated to the change suggested in this PR.

The user pods k8s manifests are created by KubeSpawner which is configured in the hub pod via the Helm chart configuration within jupyterhub_config.py mounted to the hub pod.

Can you verify your bug and open a bug report?
1. Apply the `singleuser.image.pullPolicy=Always` configuration
2. Spawn a user pod, then `kubectl get pod -o yaml <user pod's name>` and observe if the user pod did or didn't include `imagePullPolicy: Always` in the pod specification
3. Open a github issue describing this

consideRatio:
I'm closing this as I think the next action point is to write a bug report if the issue can be confirmed based on https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2214#issuecomment-846445811

PR #2224, opened 2021-05-24
Add hub.lifecycle configuration for postStart/preStop hooks

### LINKED ISSUE #2223: Add hub.lifecycle configuration
### Proposed change
Add `hub.lifecycle` config.

### Motivation
A k8s _postStart_ lifecycle hook of the _exec_ kind could for example `pip install <package>` on startup for use by a `pre_spawn_hook`.

I want to install the pip package `gql` that help make GraphQL requests to a GitLab instance's API. Using a lifecycle hook i can avoid maintaining my own custom hub pod image.

### Alternative options
Do nothing.

It is quite general purpose, but for example:
- on startup install some small package needed by a pre_spawn_hook
- on startup/shutdown send a web request annotating some grafana dashboard that the hub pod started or shutdown

### Reference
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#lifecycle-1

consideRatio:
Closes #2223.

Note that I named it `hub.lifecycle` as it apparently is called `lifecycle` in the container specification rather than `lifecycleHooks`. Our kubespawner config is called `lifecycle_hooks` which is not mimicing the k8s specification, which made us name our similar config `singleuser.lifecycleHooks` that instead of `singleuser.lifecycle`.


minrk:
Agreed re: `singleuser.lifecycleHooks`. I think we should try to move toward matching k8s as much as we can (`singleuser.lifecycle`), but at least we can adopt it in new config.

PR #2226, opened 2021-05-24
docs: add documentation about resource requests

### LINKED ISSUE #2085: Document resource requests for core components

consideRatio:
This documents the current situation on how to set resource requests and closes #2085.

- Is the location of this documentation section okay?
- Are there something missing from the documentation of relevance to convey?
- Is the language okay?

An action point list from #2085 made by @minrk, self-checked by me.
- [x] explain a bit of background about resource reservations and when they are useful
- [x] identify all components that might want resource requests
- [x] identify reasonable values for key pods (mainly hub and proxy) and ranges based on load (e.g. using [mybinder.org as a reference point for high load](https://grafana.mybinder.org/d/GYEYQm7ik/components-resource-metrics?orgId=1&refresh=1m))
  - [x] I've provided reasonable recommendations for requests for a reliable configuration.
  - [x] I've opted not to try provide a recommendation for a smaller deployment because it is a grey scale and I lack information about it.


doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,133 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+| prePuller.hook.resources                | hook-image-awaiter            | 0, 0                             | The container just polls the k8s api-server. Will require minimal resources.                                                                                                                                                                                                                                                       |
+| singleuser.cpu\|memory.guarantee\|limit | jupyter-username              | 0, 1G                            | The configuration syntax is different because it is native to the Spawner base class rather than Kubernetes. It is commonly useful to guarantee a certain amount of memory rather than CPU to help users share CPU with each other.                                                                                                |
+
+### Example resource requests for high reliability
+
+Here are some examples of resource requests reasonable for a larger deployment.
+If you collect metrics on the resource usage of pods in your k8s cluster
+(Prometheus) and have dashboards showing you its usage (Grafana), you could
+tweak these over time.
```
manics, line 471: What counts as a "larger deployment"? If it's mybinder.org you could say that, and also link to the relevant mybinder.org grafana dashboards?

manics (commented):
Looking good! I made a first pass over it.

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,133 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+
+   Check the namespace where you deploy JupyterHub for such resource by doing
+   `kubectl get limitrange --namespace <k8s-namespace>`.
+
+1. Setting only resource limits _will make_ k8s assume the same in resource
+   requests, but setting only resource requests _will not make_ k8s assume any
+   resource limits.
+
+1. Setting resource requests improves the reliability of the container's
+   execution. Setting resource limits improves the consistency of the execution
```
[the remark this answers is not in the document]
consideRatio: Yeah hahah i figured when I wrote that that it would be not so obvious what i meant, but I felt that it would be troublesome to explain it in detail as well.

By limiting your pod to a certain memory for example, you can consistently make it crash by trying to surpass the limit. But, for a higher reliability of the containers execution the limit could be removed. It is then less prone to crash due to memory limits, but less consistent in when it does.

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,133 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+   of the specific container and other containers on the nodes.
+
+1. Requested memory is reserved and unavailable for other containers on a node
+   to use, but any requested CPU can be used by other containers until its
+   required by the requesting container.
+
+1. A container running out of memory is typically terminated and restarted as a
+   process is killed by a _Out Of Memory Killer_ (OOMKiller). When this happens
```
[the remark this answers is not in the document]
consideRatio: Hmmmm, not sure if I understand your question, but...

1. A pod managed by a replicaset or statefulset could be recreated, for example if it is evicted from a node that is about to undergo maintenance of if another pod with higher priority evicts it to be able to be scheduled on a node.
2. A container can restart independent of having the entire pod be recreated, which is in other words something that can happen with a standalone pod. This is what I'm referring to here.

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,133 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+
+1. Software may not be able to make use of more than 1 CPU because it doesn't
+   support running code concurrently across multiple CPU cores. The `hub` pod
+   running JupyterHub in Python and the `proxy` pod running
+   ConfigurableHTTPProxy in NodeJS are such applications.
+
+Some additional more technical details are:
+
+1. A container requesting `0` CPU will still be granted a share of `0.002` CPU,
+   or `2m` CPU, it is a lower limit.
```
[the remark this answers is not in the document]
consideRatio: Yes I know =/ Well, i think it is out of scope for k8s to declare it, but the docker Container Runtime Interface (CRI) at least does this. I considered this as well and decided that it was better to not try to make a disclaimer about that it may not be true with different CRI or different configurations of that CRI etc.

I think its a relevant point to clarify that one can expect 0 0 requests to actually be granted some minimal share of a CPU. If you would schedule something to run with absolute zero CPU it wouldn't make sense at all anyhow so I think its safe to guess all CRI will provide some minimum.

doc/source/administrator/optimization.md
consideRatio: I'm not sure, I guess I meant "if you are so large that you don't care about allocating 1 CPU to the hub and proxy pod to ensure stability over minimizing cost".

doc/source/administrator/optimization.md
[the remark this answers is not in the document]
consideRatio: I updated "container engine" to "container runtime".

doc/source/administrator/optimization.md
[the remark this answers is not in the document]
consideRatio: A pod can have a restartPolicy, and this is a description on how containers should behave. Should they restart on failure / completion etc?

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

doc/source/administrator/optimization.md
consideRatio: I rewrote it as the following in 041d37f

> Here are some examples of resource requests that could be reasonable for adeployment where service reliability and performance matters more thanminimizing the required capacity on nodes.

doc/source/administrator/optimization.md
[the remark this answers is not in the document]
consideRatio: I attempted a new formulation in 1a7bc6d

> 1. Setting resource _requests_ guarantees a minimal level of resources for the
   container and therefore some minimal capacity for it to do its job. Setting
   resource _limits_ close to or equal to the resource requests can be useful to
   make the container operational performance more consistent.

consideRatio:
@manics thanks for your feedback, I've tried addressing it now! What do you think?

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,141 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+prePuller: # hook|continuous-image-puller pods (optional, running pause container)
+  resources:
+    requests:
+      cpu: 10m
+      memory: 8Mi
+  hook: # hook-image-awaiter pod (optional, running GoLang binary defined in images/image-awaiter)
+    resources:
+      requests:
+        cpu: 10m
+        memory: 8Mi
```
consideRatio, line 522: @yuvipanda does these suggestions look reasonable to you for a deployment under heavy load / stress test?

consideRatio: Looking at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2034#issuecomment-782262510 I realize that the requests set here are too low to ensure they get what they need for great performance.

The amount of memory and cpu varied a bit more than I thought. I'm really not sure what to suggest here... Hmm....

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,141 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+1. Setting only resource limits _will make_ k8s assume the same in resource
+   requests, but setting only resource requests _will not make_ k8s assume any
+   resource limits.
+
+1. Setting resource _requests_ guarantees a minimal level of resources for the
+   container and therefore some minimal capacity for it to do its job. Setting
+   resource _limits_ close to or equal to the resource requests can be useful to
+   make the container operational performance more consistent.
+
+1. Requested memory is reserved and unavailable for other containers on a node
```
minrk, line 413: Where did you find this info about memory requests being unavailable? I thought if you had two pods that each request 50% and limit 100% of a node's memory, they would each be able to allocate 100% of the node memory as long as it's not at the same time. But if this is accurate, if requests add up to 100%, no pod will be able to exceed its request at all.

My understanding has been that  'request' is really only an input to the scheduler, while 'limit' is the only input to the runtime (OOMKiller).

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,141 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
 We suggest **starting with a ratio of 2 to 1** and adjusting from there based on whether you run into problems on your hub.
 For example, if the limit is 10GB, start with a guarantee of 5GB.
 Use a service such as Prometheus + Grafana to monitor the memory usage over time, and use this to decide whether to adjust your ratio.
+
+## Explicit memory and CPU allocated to core pods' containers
+
+The Helm chart creates several k8s Pods that typically have one container
```
minrk, line 389: Probably worth linking to k8s own [docs on the matter](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) somewhere.

doc/source/administrator/optimization.md
```diff
@@ -383,3 +383,141 @@ Choosing the right _limit to guarantee ratio_ ratio is an art, not a science.
...
+   Pod's main containers are started, the effective requests/limits are
+   calculated as the highest of the init containers requests/limits and the sum
+   of the main containers requests/limits.
+
+1. Software may not be able to make use of more than 1 CPU because it doesn't
+   support running code concurrently across multiple CPU cores. The `hub` pod
+   running JupyterHub in Python and the `proxy` pod running
+   ConfigurableHTTPProxy in NodeJS are such applications.
+
+Some additional more technical details are:
```
minrk, line 437:
```suggestion
Some additional technical details:
```

minrk (approved):
Agreed with @manics that this need not be a blocker for 1.0, but this is already great info as it is, and I'm happy for it to be merged right now and refined.

Tweaking individual numbers for best practices / reasonable limits is something that is always going to evolve over time, and what's here seems like a great start.

doc/source/administrator/optimization.md
consideRatio: Oh oh hmmmmm, you are probably correct in your understanding.

I've not questioned this understanding much, I think it was formulated by myself based on how I figured it was unreasonable that a pod that have requested a set of memory would have to wait for memory to be freed up when it currently was used used by another container. At the same time, that process could be very quick and i bet the OOMKiller can take action before all memory is depleted as well.

manics:
How about take out any test we're uncertain about since it's better to say nothing than to give incorrect information, then merge? And refine later?

doc/source/administrator/optimization.md
minrk: Yeah, I think requests are a lot more basic than folks often expect - I think it's purely scheduler bookkeeping, and has no actual effect once pods are scheduled (except possibly in OOMKiller priority when it's time for something to be killed).

Maybe worth just deleting this point then. Then 👍 to merge.

minrk: I tested with k3d:

<details>
<description>pods.yaml</description>

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: memory-eater
spec:
  containers:
    - name: container
      image: python:3.9-alpine
      env:
        - name: PYTHONUNBUFFERED
          value: "1"
      command:
        - python3
        - -c
        - |
          import os
          import time
          MB = 1024 * 1024
          chunk_size = 100 * MB  # 100 MB
          chunks = []

          while True:
            chunk = bytearray(chunk_size)
            chunks.append(chunk)
            print(f"{len(chunks) * chunk_size // MB }MB")
            time.sleep(5)
      resources:
        requests:
          memory: 900Mi
        limits:
          memory: 1800Mi

---
apiVersion: v1
kind: Pod
metadata:
  name: memory-squatter
spec:
  containers:
    - name: container
      image: python:3.9-alpine
      env:
        - name: PYTHONUNBUFFERED
          value: "1"
      command:
        - sh
        - -c
        - while true; do sleep 30; done
      resources:
        requests:
          memory: 900Mi
        limits:
          memory: 1800Mi
```
</details>

on a node with 2Gi, and the memory eater is allowed to comfortably allocate almost all of the node's memory being killed. The presence or absence of the memory squatter's request for 900M doesn't have any effect on how much the eater is able to allocate before being killed.

If there is a point to be made here, I think it's that oversubscribing CPU results in things being slower than they could be, but usually not catastrophic, while oversubscribing memory results in hangs and crashes. The same actually goes for *Under*-provisioning: low limits on CPU means things might be slow, while low limits on memory means things will keep getting killed.

consideRatio: @minrk thank you!!! This is an excellent summary!

doc/source/administrator/optimization.md
consideRatio: Done

doc/source/administrator/optimization.md
consideRatio: I relied on this insight and made another pass

consideRatio:
I made another pass. Let's make refinements in separate PRs and get this going!

PR #2236, opened 2021-05-31
Add annotations to rbac serviceAccount

AndreaGiardini:
I am using this helm chart on GKE. The preferred method to authorize GKE workloads is using Google workload identity ( https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity )

In order to use workload identity it's necessary to add a specific annotation( `iam.gke.io/gcp-service-account` ) to the k8s serviceaccount. In this way we can link the k8s serviceaccount with google's service account. 

Adding annotations to the hub's serviceAccount was not possible with the current helm chart so I implemented it in this PR.

Let me know if you have any question

### DIFF
--- a/jupyterhub/templates/hub/rbac.yaml
+++ b/jupyterhub/templates/hub/rbac.yaml
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "jupyterhub.hub.fullname" . }}
+  {{- with .Values.hub.serviceAccount.annotations }}
+  annotations:
+    {{- . | toYaml | nindent 4 }}
+  {{- end }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 ---
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -263,6 +263,8 @@ proxy:
       enabled: false
       maxUnavailable:
       minAvailable: 1
+    serviceAccount:
+      annotations: {}
   secretSync:
     containerSecurityContext:
       runAsUser: 65534 # nobody user
@@ -433,6 +435,8 @@ scheduling:
       maxUnavailable: 1
       minAvailable:
     resources: {}
+    serviceAccount:
+      annotations: {}
   podPriority:
     enabled: false
     globalDefault: false
@@ -498,6 +502,8 @@ prePuller:
     nodeSelector: {}
     tolerations: []
     resources: {}
+    serviceAccount:
+      annotations: {}
   continuous:
     enabled: true
   pullProfileListImages: true
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -134,6 +134,8 @@ hub:
     failureThreshold: 1000
     timeoutSeconds: 1
   existingSecret:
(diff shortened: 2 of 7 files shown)


jupyterhub/templates/hub/rbac.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "jupyterhub.hub.fullname" . }}
+  annotations:
+    {{- with .Values.rbac.serviceAccount.annotations }}
```
consideRatio, line 7:
```suggestion
    {{- with .Values.rbac.serviceAccount.annotations }}
```

jupyterhub/templates/hub/rbac.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: v1
...
+    {{- . | toYaml | nindent 4 }}
+    {{- end }}
```
consideRatio, line 9: This would render to...

```
annotations:
labels:
  # ...
```

If rbac.serviceAccount.annotations is falsy, and that would be invalid k8s YAML specification. This on the other hand would be okay.

```suggestion
  {{- with .Values.rbac.serviceAccount.annotations }}
  annotations:
    {{- . | toYaml | nindent 4 }}
  {{- end }}
```

jupyterhub/values.yaml
```diff
@@ -137,6 +137,8 @@ hub:
 
 rbac:
   enabled: true
+  serviceAccount:
+    annotations: {}
```
consideRatio, line 141: There are more service accounts because we have more pods that all need som permissions against the k8s-apiserver, and I think we would need annotations specific to the service accounts.

So, something related to hub must be part of the config of this annotation, so its not applied to all serviceaccounts

consideRatio (changes requested):
Feel free to come with suggestions, but I need more time to think myself still

consideRatio:
I think I'd like our configuration to follow the configuration locations suggested here: https://helm.sh/docs/chart_best_practices/rbac/

So, `hub.serviceAccount.annotations` would be the configuration location of relevance then I figure.

jupyterhub/values.yaml
AndreaGiardini: I will follow your suggestion. I added it to the `rbac` section because it was it changes files in the `rbac` yml file but I have nothing against moving it to the `hub` section

AndreaGiardini:
@consideRatio Thank you for the feedback. I implemented the changes you requested. Can you have a second look?

AndreaGiardini:
@consideRatio sorry for pinging again, I would love to see this in v1.0.0 :) let me know if there is something else missing

consideRatio:
@AndreaGiardini thanks for your work, I'll help get it merged!

This is what remain in my mind and I figure I'll push a commit for this later today if you haven't already:

- [ ] schema.yaml entry about this configuration
- [ ] lint-and-validate-values.yaml entry for this configuration

consideRatio (approved):

AndreaGiardini:
Many thanks @consideRatio !

consideRatio:
Thank you @AndreaGiardini!

PR #2248, opened 2021-06-11
Add participation in study notice to readme

sgibson91:
This PR adds a notice to the readme informing users that the repo is participating in an open source sustainability study, see https://github.com/jupyterhub/team-compass/issues/417 for details

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,20 +16,6 @@ This repo contains a _Helm chart_ for JupyterHub and a guide to use it. Together
 they allow you to make a JupyterHub available to a very large group of users
 such as the staff and students of a university.
 
----
-
-Please note that this repository is participating in a study into sustainability
-of open source projects. Data will be gathered about this repository for
-approximately the next 12 months, starting from 2021-06-11.
-
-Data collected will include number of contributors, number of PRs, time taken to
-close/merge these PRs, and issues closed.
-
-For more information, please visit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,20 @@ This repo contains a _Helm chart_ for JupyterHub and a guide to use it. Together
...
+approximately the next 12 months, starting from 2021-06-11.
+
+Data collected will include number of contributors, number of PRs, time taken to
+close/merge these PRs, and issues closed.
+
+For more information, please visit
+[our informational page](https://sustainable-open-science-and-software.github.io/) or download our [participant information sheet](https://sustainable-open-science-and-software.github.io/assets/PIS_sustainable_software.pdf).
+
+---
+
```
consideRatio, line 32: Could we put this under a `## ` topic between `## The Helm chart` and `## History` perhaps? I'd like to avoid having it take too much attention from introducing the repo to the reader.

consideRatio: I looked in the description [here](https://sustainable-open-science-and-software.github.io/readme_notice) about the research notice, and it seems they only ask that it is placed in the README without detailing it need to be front and center for example.

sgibson91: New commit pushed :)

consideRatio: Thanks @sgibson91!

PR #2251, opened 2021-06-15
Fix fullnameOverride for Ingress & PriorityClass resources

### LINKED ISSUE #2250: Ingress object cannot be created when full-name is overridden
### Bug description

When the default naming is overridden, helm release fails because `Ingress` object has an invalid name.

#### Expected behaviour
I can override the default naming for jupyterhub releases with `Ingress` object.

#### Actual behaviour

The ingress object has a naming pattern with a _minus sign_ at the end, like `{name-override}-`, which is not a valid dns name and kubernetes rejects it.

The rest of jupyterhub kubernetes objects append a string after `jupyterhub.fullname.dash`, while ingress object does not have it: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/_helpers-names.tpl#L177

### How to reproduce

Try to release helm chart with parameter `fullnameOverride: jupyterhub-1`

### Your personal set up

```
Kubernetes: v1.19.7
Helm: v3.6.0
Jupyterhub chart: 0.11.1-1.0.0
```

v1r7u:
Closes #2250, which blocks fullname-overrides with Ingress objects

### DIFF
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -173,8 +173,8 @@
 
 {{- /* Ingress */}}
 {{- define "jupyterhub.ingress.fullname" -}}
-    {{- if (include "jupyterhub.fullname.dash" .) }}
-        {{- include "jupyterhub.fullname.dash" . }}ingress
+    {{- if (include "jupyterhub.fullname" .) }}
+        {{- include "jupyterhub.fullname" . }}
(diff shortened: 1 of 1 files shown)


consideRatio:
Thank you @v1r7u for catching this!

I pushed a commit fixing another resource as well that had a similar mistake in its naming even though it wasn't critical like it was for the ingress resource.

Issue #2253, opened 2021-06-16
Z2JH on Raspberry PI Cluster exec format error on pre puller


JamesCarscadden:
### Bug description
I am attempting to set up the Z2JH chart on a Raspberry PI (arm64) based cluster and am encountering some problems with the pre Puller job/pods. Specifically we get an exec format error (unsurprising since that container does not appear to have an arm64 version. 

#### Expected behaviour
These containers should have arm64 versions

#### Actual behaviour
exec format error due to no arm64 version

### How to reproduce
Use base Z2JH chart (v1.0.0) on an arm64 based cluster node

### Your personal set up
Kubernetes v1.21.1 running on Raspberry PI 4 (8Gb) 4 node cluster.
Base OS is current version of Raspbian arm64 distribution
Using Helm to deploy the jupyterhub chart from https://jupyterhub.github.io/helm-chart/

This issue can be worked around by setting

```yaml
prePuller:
    hook:
      enabled: false
    continuous:
      enabled: false
```

in the helm values

manics:
I suspect the problem is that
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/1a2ff6816303a9843e1a0e26da51e26d7bc9eff8/images/image-awaiter/Dockerfile#L13
isn't multi-arch.

JamesCarscadden:
I don't believe the image-awaiter is the actual problem in this case -- looking closer at the logs the initial problem (if I remove the values configuration mentioned in the issue), is that the 'hook-image-puller' pod fails with 'exec format error' actually on 'image-pull-singleuser' init container (which is probably related to #2145.)

If I add 
  prePuller:
    hook:
      enabled: false

to my values, then I encounter the same problem with 'continuous-image-puller' pod. And then finally if I also add the continuous->enabled->false as well then I get past this problem onto other issues. So that's my bad for not looking at this as closely as I should have. Likely this can be considered a duplicate of #2145

JamesCarscadden:
My initial bug report I made the assumption that since the values were 'prePuller' related, that the exec format error was related to the jupyterhub/k8s-pre-puller container, but that doesn't actually appear to be used anymore. Sorry.

manics:
[...]

PR #2255, opened 2021-06-16
image-awaiter: use busybox (multiarch)

manics:
scratch doesn't look like it's a multi-arch image, busybox is quite small.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2253


manics:
@JamesCarscadden Would you mind building the image-awaiter locally on your ARM64 system and checking it runs?
```
cd images/image-awaiter/
docker build -t tmp .
docker run -it --rm tmp /image-awaiter
```

JamesCarscadden:
I've tried the build as suggested and the resulting image does work. However, the image also works if built from main branch. The image-awaiter is not the actual problem for #2253 -- I've commented further on the bug. This PR can probably be abandoned -- sorry

PR #2262, opened 2021-06-18
hub image: add sqlalchemy-cocroachdb dependency

weisdd:
Based on jupyterhub's docs, it should be generally compatible with any sqlalchemy provider.

In our environment, we're relying on cocroachdb due to easier clusterization and scalability.
Aside from a tiny defect in jupyterhub (jupyterhub/jupyterhub#3484, it was not part of 1.4.1), which might not even be specific to cocroachdb (rather to the dummy authenticator), everything seems to work well.

Currently, we have to dynamically install `sqlalchemy-cockroachdb` upon container start. It'd be great to have it included in the base image.

P.S. If I got it right, I had to run the `dependencies freeze --upgrade` script to make sure all dependencies are added to the requirements.txt. That's why you can see two files updated instead of just one (requirements.in).

### DIFF
--- a/images/hub/requirements.in
+++ b/images/hub/requirements.in
@@ -28,6 +28,7 @@ jupyterhub-kubespawner
 pymysql  # mysql
 psycopg2-binary  # postgres
 pycurl  # internal http requests handle more load with pycurl
+sqlalchemy-cockroachdb # cocroachdb
 statsd  # statsd metrics collection (TODO: remove soon, since folks use prometheus)
 
 # The idle culler service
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
@minrk I think it is reasonable to add this python library to the hub image while still considering that we want to avoid increasing complexity and such, do you agree?

minrk:
I think this is fine to do for db backends that folks use, since there can't be many. Authenticators are more of a slippery slope, so I opened #2265 so we can have a better story for less-used dependencies than always submitting patches to the base image.

PR #2268, opened 2021-06-21
added AWS EKS cluster scaling/auto-scaling documentation for z2jh

cdibble:
This is an unsolicited PR but I figured I'd add what I've learned about scaling on AWS EKS, since that was not yet included in the documentation for `Expanding and contracting the size of your cluster`.


doc/source/jupyterhub/customizing/user-resources.md
```diff
@@ -202,3 +202,40 @@ az aks show --name <YOUR-CLUSTER-NAME> --resource-group <YOUR-RESOURCE-GROUP>
 
 It may take some time for the new cluster nodes to be ready.
 You can use `kubectl get node` to report the current `Ready/NotReady` status of all nodes in the cluster.
+
+### Amazon Web Services Elastic Kubernetes Service (EKS)
+
+To scale an existing nodegroup using the `eksctl` command line tools:
```
consideRatio, line 208: Is this a valid step for anyone having an EKS cluster, no matter if it was setup with `kops` or `eksctl` initially? If not, I think it should be clarified that this is scoped specifically for AWS EKS clusters setup with `eksctl`.

cdibble: Good point. This is EKS specific and doesn't apply to either `kops` or a "manual" k8s installation in AWS. I'll clarify.

cdibble:
I think it's clear now that this is EKS specific. You don't technically need to use `eksctl` to create the cluster (it can be done in the AWS console), but you have to install `eksctl` and configure it to point to your cluster. I've added that point as well.

consideRatio:
Thanks you @cdibble!

PR #2273, opened 2021-06-22
Remove pre-commit from GHA

minrk:
runs on pre-commit.ci now

leaves shellcheck in there, which requires additional dependency and is run separately

It would be nice to bring shellcheck in so that it works like everything else. It can run in docker, but also needs the entrypoint wrapper in https://github.com/gruntwork-io/pre-commit/blob/HEAD/hooks/shellcheck.sh

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -43,7 +43,6 @@ jobs:
           pip install chartpress yamllint
 
       - uses: pre-commit/action@v2.0.3
-      - uses: pre-commit/action@v2.0.3
         with:
           extra_args: --config .pre-commit-config-shellcheck.yaml
 


consideRatio:
Why do you want it on pre-commit instead? Without a benefit, i figure it would just increase the complexity without benefit by becoming another way to inspect a failure and inspect the CI logic etc.

minrk:
Sorry, forgot to link https://github.com/jupyterhub/team-compass/issues/379 . The pre-commit [action](https://github.com/pre-commit/action) is also officially deprecated in favor of using pre-commit.ci. I don't expect that to be an issue, as running pre-commit manually instead of via a custom action is not a problem.

consideRatio:
No worries @minrk! So far so good with using pre-commit.ci, I had to disable the `helmlint` pre-commit hook we used as that relies on `helm` being available in the system which pre-commit.ci wouldn't have. I also had to make use of the jupyterhub/chartpress defined pre-commit hook as that made sure chartpress was installed and wasn't assumed to be available in the system already.

minrk:
I wonder if we should make our own pre-commit-hooks repo on this org so that we can have pre-commit hooks that ensure things like helmlint, etc. are installed. Then we'd have a place to make shellcheck run its auto-fixer as I did [here](https://github.com/minrk/simula-summer-school/blob/main/shellcheck-fix)

consideRatio:
The key thing to answer to accomplish that I think is: how can we make a pre-commit hook defined in some repo referenced from some pre-commit-config elsewhere make a binary like `helm` or `shellcheck` be installed - and not only a python or similar package.

Alternatively, if not any kind of binary can be installed, then perhaps we can find a way to install the specific binaries we want, like `helm` which is a GoLang binary at least.

minrk:
Lots of pre-commit repos vendor files, as the simplest route. Works well for single binaries like most stuff in the go/kube/helm world. See, for example [shellcheck-py](https://github.com/shellcheck-py/shellcheck-py) a repo created by the pre-commit author specifically for the purpose of easy env installation of shellcheck.

PR #2276, opened 2021-06-23
Corrected scheduler rbac custom naming

### LINKED ISSUE #2275: User-scheduler role-binding uses hardcoded service-account name
### Bug description
User-scheduler role-binding has hardcoded `user-scheduler` in the Service-Account name. If the name is overridden - the scheduler cannot connect to kube-api.

#### Expected behaviour
Jupyterhub should be able to work from the box with customized names.

#### Actual behaviour
No user pods could be scheduled, because scheduler is not authorized to connect to kube-api

### How to reproduce
Try to release helm chart with parameter `fullnameOverride: jupyterhub-1`

### Your personal set up
```
Kubernetes: v1.19.7
Helm: v3.6.0
Jupyterhub chart: 0.11.1-1.0.0
```

v1r7u:
Closes #2275

### DIFF
--- a/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml
@@ -203,7 +203,7 @@ metadata:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 subjects:
   - kind: ServiceAccount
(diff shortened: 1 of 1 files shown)


consideRatio:
Nice catch!! Thanks @v1r7u!!!

PR #2287, opened 2021-06-24
Add changelog for 1.0.1

consideRatio:
This is the first release since 1.0.0, and now when we can bump both the minor and patch version I'm not so confident on what version makes sense to bump, but I think this should be considered a patch release as its mostly docs and bugfixes.

But for example, if it would bump kubespawner from some minor version to another, then we should perhaps bump the Helm chart a minor version as well? But, perhaps when nbgitpuller is bumped from 0.9 to 0.10 we don't, because its part of singleuser-sample rather than something that provides core functionality?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,44 @@ Here you can find upgrade changes in between releases and upgrade instructions.
 
 ## [1.0]
 
+### [1.0.1] - 2021-06-25
+
+#### Bugs fixed
+
+- Relax hub.db.type schema to accept unknown database types [#2285](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2285) ([@consideRatio](https://github.com/consideRatio))
+- templates: quote namespace in case they are only contain numbers [#2284](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2284) ([@consideRatio](https://github.com/consideRatio))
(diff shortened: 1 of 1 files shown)


minrk:
I don't think any changes in singleuser-sample should require a version bump, since that isn't used by actual deployments. It's just a reference example, and more akin to docs when it comes to API-sensitivity.

> if it would bump kubespawner from some minor version to another, then we should perhaps bump the Helm chart a minor version as well?

I think that's going to depend on the change. When we have a big bundle like this, I don't think we need to be extremely strict about semantic versioning, as it's only three numbers that distill *tons* of accumulated packages. If we've bumped kubespawner and it exposed an extra flag or two, I don't think we need to force a minor bump, patch is fine. If it's a truly new feature, especially one we expose and/or document *here*, then minor in the chart makes sense.

Taking a step back at semantic versioning to *why* we have three levels of changes:

- major: there's a good chance you need to change things to keep your deployment working, so read carefully, backup, and test before upgrading
- minor: there might be something new in here to look at, but anything that was working should keep working without modification
- patch: principally bugs fixed, nothing new *for most users*

I think it's also okay to think in terms of 'bigness'. A big change we don't *think* will break anything can still make a major release to signal its importance, etc.

minrk (approved):

consideRatio:
Thanks @minrk for reasoning aloud about this! I think what you say is very reasonable and i like the idea of allowing for some freedom of choice on the version rather than stricly commiting to act in a certain way based on some ruleset.

sgibson91 (approved):
LGTM! And I agree with @minrk's semver discussion above. Great work!

consideRatio:
Thanks @sgibson91! I'll go for a release!

PR #2293, opened 2021-06-29
Fix failure to set imagePullSecrets for user-placeholder pods (scheduling.userPlaceholder.image config added)

michaellzc:
### Updated summary by Erik

`prePuller.pause.image.name|tag` was used to set the user-placeholder pods image, but the user-placeholder pods didn't make use of the `prePuller.pause.image.pullPolicy|pullSecrets` or any other config to set imagePullPolicy or imagePullSecrets. With this PR the dedicated option `scheduling.userPlaceholer.image` and respective fields are added and respected for the user-placeholder pods.

### Original PR summary

`prePuller.pause.image.pullSecrets` was already defined in `values.yaml` but it was never actually used in the placeholder statefulset template.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -2181,6 +2181,7 @@ properties:
         properties:
           enabled:
             type: boolean
+          image: *image-spec
           replicas:
             type: integer
             description: |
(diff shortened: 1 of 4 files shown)


consideRatio:
It was quite weird that the prePuller.pause.image was used in userPlaceholder instead of scheduling.userPlaceholder.image, so I adjusted this PR that has a very valid point to instead of adding to the weird behavior - creating a dedicated configuration for the userPlaceholder image.

consideRatio:
Thanks @michaellzc! I wanted to resolve this in a slightly different way, does it make sense to you? I pushed a commit and update the PR description.

michaellzc:
> Thanks @michaellzc! I wanted to resolve this in a slightly different way, does it make sense to you? I pushed a commit and update the PR description.

lgtm 👍

PR #2294, opened 2021-06-29
Fix regression where hub.services require an apiToken to be specified

consideRatio:
This is not a regression in Z2JH itself, but with JupyterHub 1.4.0 and 1.4.1, but it is fixed in current main branch. See https://github.com/jupyterhub/jupyterhub/pull/3531


consideRatio:
This was not an issue in Z2JH, but in JupyterHub: https://github.com/jupyterhub/jupyterhub/pull/3531

Issue #2296, opened 2021-06-30
1.0.1 updated traefik from v2.4.8 to v2.4.9 and introduced an upstream regression


consideRatio:
### Bug description

I've reported an issue in Traefik about a regression to acquire new TLS certificates when using version 2.4.9 instead of 2.4.8 as we do in z2jh 1.0.1: https://github.com/traefik/traefik/issues/8240

#### Expected behaviour

The proxy.https.type=letsencrypt option to work properly... Bumping traefik minor versions haven't proven to be reliable... I should stop bumping traefik and only do it one every quarter or similar...

#### Actual behaviour
TLS cert acquisition fail like this in `autohttps` pods logs when having `debug.enabled=true` set.

```
time="2021-06-30T08:51:43Z" level=debug msg="legolog: [INFO] Deactivating auth: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/78629711"
time="2021-06-30T08:51:43Z" level=debug msg="legolog: [INFO] Unable to deactivate the authorization: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/78629711"
time="2021-06-30T08:51:43Z" level=error msg="Unable to obtain ACME certificate for domains \"hub.neurohackademy.org\" : unable to generate a certificate for the domains [hub.neurohackademy.org]: error: one or more domains had a problem:\n[hub.neurohackademy.org] acme: error: 400 :: urn:ietf:params:acme:error:connection :: Fetching http://hub.neurohackademy.org/.well-known/acme-challenge/xiBRKmK3WjN0E8P3KBq3jeh1BZzOv8YK2xqZDRmY-BE: Connection refused\n" providerName=default.acme
```

consideRatio:
I could not reproduce this with the lets encrypt staging server, and I could not keep trying with the production server.

PR #2297, opened 2021-06-30
fix: downgrade traefik in autohttps pod to avoid a regression about ACME TLS acquisition

consideRatio:
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2296.

Not sure what goes on... Debugging a lot... I really really really dislike intermittent ACME cert acquisition issues...

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -233,7 +233,13 @@ proxy:
       allowPrivilegeEscalation: false
     image:
       name: traefik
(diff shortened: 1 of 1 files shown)


consideRatio:
Cant wrap my head around this... It wasn't intermittent with LE staging, but it was with LE prod where I can't debug it properly without getting blocked so I stopped debugging that there...

Issue #2256, opened 2021-06-16
Add options to run hub and proxy pods on host network


mallman:
### Proposed change
Add configuration options to the helm chart to deploy the hub and/or proxy pods on the host network.

### Alternative options
The alternative is the current configuration, which does not put the either the hub or the proxy on the host network. Rather, it creates a proxy-public service, which by default is of type "load balancer". It can also be configured as a "node port". However, the current configuration is (to the best of my ability) incompatible with the scenario where the notebook pods are deployed on the host network (see below).

### Who would use this feature?
This will greatly simplify and ease deployment when the notebook pods are configured to run on the host network. We need to run the notebook pods on the host network for compatibility with Spark. Spark requires a routable client address. When Spark runs outside the kubernetes cluster, the client cannot send an address to Spark that's routable back to the notebook pod.

### (Optional): Suggest a solution
Add the configuration options to the chart and set the value of the 'hostNetwork' pod spec as appropriate. There will likely be more work to do beyond that, like reconfiguring the public proxy service to route to the host address of the proxy instead of an internal address.

manics:
Hi! Could you explain a bit more about why running the singleuser pods with host networking is incompatible with the standard services for the hub and proxy? It's possible the JupyterHub config would need to be modified to pass the external (add opposed to the internal k8s) address to the singleuser servers, but that should be a less drastic change than making everything use host networking.

mallman:
My understanding is that the hub must send its URL to the notebook server. What configuration setting should I set the external URL in? This is what I'm considering:

c.JupyterHub.hub_connect_url
c.JupyterHub.hub_ip
c.JupyterHub.ip
c.JupyterHub.port

mallman:
Let me expand on our situation. My understanding from https://jupyterhub.readthedocs.io/en/stable/reference/technical-overview.html#the-subsystems-hub-proxy-single-user-notebook-server is we need an external address for the browser. This is the proxy-public service. We have this set up with a NodePort because we are bare-metal and don't support the LoadBalancer configuration.

The proxy connects to the hub via its internal service connection. It also connects to the notebook server. Since we are running the notebook server pod on host networking, the proxy connects to its host address. The notebook server needs to connect to the Hub. Since it's on the host network, it needs a connection to an external address. So I suppose we set up the Hub service with a nodeport connection and configure the notebook servers to connect to the hub on that port.

[...]

PR #2298, opened 2021-06-30
Add configuration options for using host networking

mallman:
Add configuration options for using host networking for the hub and proxy pods. This PR relates to issue #2256.

As mentioned in #2256, this will greatly simplify and ease deployment when the notebook pods are configured to run on the host network. For example, notebook pod networking is used for compatibility with Spark. Spark requires a routable client address. When Spark runs outside the kubernetes cluster, the client cannot send an address to Spark that's routable back to the notebook pod unless it's on the host network.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -504,6 +504,19 @@ properties:
           # to generate a value, run
           openssl rand -hex 32
           ```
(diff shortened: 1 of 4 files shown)


mallman:
I'm going to re-submit this PR from a company github org.

PR #2300, opened 2021-07-01
Add configuration options for using host networking

mallman:
Add configuration options for using host networking for the hub and proxy pods. This PR relates to issue #2256.

As mentioned in #2256, this will greatly simplify and ease deployment when the notebook pods are configured to run on the host network. For example, notebook pod networking is used for compatibility with Spark. Spark requires a routable client address. When Spark runs outside the kubernetes cluster, the client cannot send an address to Spark that's routable back to the notebook pod unless it's on the host network.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -30,6 +30,10 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
     spec:
+      {{- if .Values.hub.hostNetwork }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- end }}
       {{- if .Values.scheduling.podPriority.enabled }}
       priorityClassName: {{ include "jupyterhub.priority.fullname" . }}
       {{- end }}
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -504,6 +504,19 @@ properties:
           # to generate a value, run
           openssl rand -hex 32
           ```
+      hostNetwork:
+        type: boolean
+        description: |
+          Use host networking for the hub pod. This is disabled by default.
+
+          Host networking is an advanced configuration for use in nonstandard
+          deployments. For example, host networking is useful in deployments where
+          notebook server pods are also deployed with host networking.
+
+          Host networking is intended for expert use where the implications are
+          well-understood. Host networking has its own class of security
+          vulnerabilities. Therefore, we recommend its use only in trusted
+          environments such as private data centers.
       image: &image-spec
         type: object
         additionalProperties: false
@@ -1165,6 +1178,19 @@ properties:
     type: object
     additionalProperties: false
     properties:
+      hostNetwork:
+        type: boolean
+        description: |
+          Use host networking for the proxy pod. This is disabled by default.
+
+          Host networking is an advanced configuration for use in nonstandard
+          deployments. For example, host networking is useful in deployments where
+          notebook server pods are also deployed with host networking.
+
+          Host networking is intended for expert use where the implications are
+          well-understood. Host networking has its own class of security
+          vulnerabilities. Therefore, we recommend its use only in trusted
+          environments such as private data centers.
       chp:
         type: object
         additionalProperties: false
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -504,10 +504,10 @@ properties:
           # to generate a value, run
           openssl rand -hex 32
           ```
-      hostNetwork:
-        type: boolean
+      hostNetwork: &hostNetwork-spec
+        type: [boolean, "null"]
         description: |
-          Use host networking for the hub pod. This is disabled by default.
+          Use host networking for this pod. This is disabled by default.
 
           Host networking is an advanced configuration for use in nonstandard
           deployments. For example, host networking is useful in deployments where
@@ -1178,19 +1178,6 @@ properties:
     type: object
     additionalProperties: false
     properties:
-      hostNetwork:
-        type: boolean
-        description: |
-          Use host networking for the proxy pod. This is disabled by default.
-
-          Host networking is an advanced configuration for use in nonstandard
-          deployments. For example, host networking is useful in deployments where
-          notebook server pods are also deployed with host networking.
-
-          Host networking is intended for expert use where the implications are
-          well-understood. Host networking has its own class of security
-          vulnerabilities. Therefore, we recommend its use only in trusted
-          environments such as private data centers.
       chp:
         type: object
         additionalProperties: false
@@ -1262,6 +1249,7 @@ properties:
           nodeSelector: *nodeSelector-spec
           tolerations: *tolerations-spec
           containerSecurityContext: *containerSecurityContext-spec
+          hostNetwork: *hostNetwork-spec
           image: *image-spec
           livenessProbe: *probe-spec
           readinessProbe: *probe-spec
@@ -1599,6 +1587,7 @@ properties:
                 type: integer
               preload:
                 type: boolean
+          hostNetwork: *hostNetwork-spec
           image: *image-spec
           resources: *resources-spec
           serviceAccount: *serviceAccount
@@ -2081,6 +2070,7 @@ properties:
             description: |
               You can have multiple schedulers to share the workload or improve
               availability on node failure.
+          hostNetwork: *hostNetwork-spec
           image: *image-spec
           pdb: *pdb-spec
           nodeSelector: *nodeSelector-spec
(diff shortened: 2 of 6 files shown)


mallman:
Incidentally, another approach to solving this problem would be to provide an "escape hatch" in the form of a YAML snippet in values.yaml that gets applied to the pod spec for the hub or proxy. This would allow users to introduce arbitrary customization to the pod specs.

mallman:
The linkcheck test should not have failed. Can I get a retest, please?

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -30,6 +30,10 @@ spec:
         {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
     spec:
+      {{- if .Values.hub.hostNetwork }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
```
consideRatio, line 35: I think I'm in support of adding configuration like this consistently throughout our workloads if you find it needed.

I think I'd like to avoid intertwining the configuration of `hub.hostNetwork` as a boolean the `dnsPolicy` setting though, perhaps someone wants it to be set to `"None"` rather than `"ClusterFirstWithHostNet"`? Perhaps someone assumes `hostNetwork` is actually just setting `hostNetwork` and then tries to configure dnsPolicy explicitly and end up confused by seeing an entry already.

### Change 1

Currently not feeling 100% confident this won't need to disentangle in the future or cause more problems than benefits by being a helpful black box, I would like this configuration of `hostNetwork` to be broken apart into separate `dnsPolicy` and `hostNetwork` configurations of the Helm chart.

### Change 2

For robustness, I'd like the following logic change.

```diff
-{{- if .Values.hub.hostNetwork }}
+{{- if not (typeIs "<nil>" .Values.hub.hostNetwork) }}
```

### Change 3

I'd like to be as consistent between the workloads we manage as possible. This PR currently modifies `hub` and `proxy.chp` (proxy pod), but not `proxy.traefik` (autohttps pod) or `scheduling.userScheduler` (user-scheduler pod) - can you add the configuration option for these pods as well?

### Change 4

In schema.yaml you can make use of YAML anchors to avoid repeating the same text multiple times. It should look something like...

```yaml
      # where you define hostNetwork configuration for the first time (under hub)
      hostNetwork: &hostNetwork-spec
        type: [boolean, "null"]
        description: *jupyterhub-native-config-description

      # like this where you define hostNetwork configuration for the three other workloads
      hostNetwork: *hostNetwork-spec
```

consideRatio (changes requested):
I've not read up on the discussion yet, but I'm just finishing this as a technical review assuming we add this option which I think is reasonable assuming someone - like you - want to use it.

jupyterhub/templates/hub/deployment.yaml
mallman: Hi @consideRatio.

Thank you for your feedback. I don't understand one of your requested changes:

```diff
-{{- if .Values.hub.hostNetwork }}
+{{- if not (typeIs "<nil>" .Values.hub.hostNetwork) }}
```

This is just checking that the `hub.hostNetwork` property exists, right? So if we set `hostNetwork: false` then we still include the template `hostNetwork: true`.

mallman: Given that a default value is set in schema.yaml, Is it possible for `hostNetwork` to be `null` unless the user specifically sets it that way? Perhaps we should set the type of that property in the schema to exclude `null` as a valid value? That's just coming from my basic understanding of json schema.

mallman: Regarding the `dnsPolicy`, I see where you're coming from. In setting pod `hostNetwork: true` I'm going by the advice of https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy, which states that for pods running on host network, we should explicitly set `dnsPolicy: ClusterFirstWithHostNet`. The documentation then presents an example of that configuration where `hostNetwork: true`.

I see the advantage of decoupling `hostNetwork` and `dnsPolicy` in the helm configuration. I see a disadvantage in introducing the complexity of two new configuration properties instead of one. Additionally, if the user sets `dnsPolicy: None`, they need to be able to configure `dnsConfig`. And according to the documentation they should be able to configure `dnsConfig` no matter what the value of `dnsPolicy`. So that introduces another complex configuration property in our json schema and helm chart.

Let me suggest an alternative. Given the strong recommendation that the user set `dnsPolicy: ClusterFirstWithHostNet` when `hostNetwork: true` (possibly a requirement—the documentation writes "should" not "must"), let's keep those settings bundled together in the helm chart as they currently are in this PR. Second, I'll add documentation to the schema's `hostNetwork` property that states that setting `hostNetwork: true` also sets `dnsPolicy: ClusterFirstWithHostNet`. I'll also include a link to the Kubernetes documentation where that configuration is stated. Setting `hostNetwork: true` is already a non-standard configuration. Setting `hostNetwork: true` and not setting `dnsPolicy: ClusterFirstWithHostNet` is a step beyond that.

Optionally, in tandem with this PR or instead of it, I can add a general "escape hatch" to the deployment configs that allows the user to set arbitrary configuration at the pod level for each of the four deployments. This would, of course, be an advanced option with the caveat that the user should only use this when the standard configuration does not provide the functionality they need. Further, they should be warned that their custom configuration may conflict with configuration introduced by future versions of the chart.

Thoughts?

consideRatio: > Given that a default value is set in schema.yaml, Is it possible for hostNetwork to be null unless the user specifically sets it that way? Perhaps we should set the type of that property in the schema to exclude null as a valid value? That's just coming from my basic understanding of json schema.

```yaml
# inside a helm template file
# we set the value if its explicitly set, either false/true,
# but we omit it if it's not explicitly set. This then assumes
# we set it to `null` (which is the same as <blank>)
{{- if not (typeIs "<nil>" .Values.hub.hostNetwork) }}
hostNetwork: {{ .Values.hub.hostNetwork }}
{{- end }}
```

```yaml
# this is how it should look in values.yaml
hub:
  # ...
  hostNetwork:
  # ...
```

---

Regarding creating a bundle option, I understand your point and I saw the documentation about how they are quite cuoupled, so that having hostNetwork: true may very well always end up being associated with that specific dnsPolicy -- but... Given a history of needing to revert such options, or clarify in the documentation, or explaining that its something documented in the documentation, I'm inclined to think the simplest end user experience is to let them do the configuration fully, and for us to provide both separate options mirroring the k8s native syntax.

Regarding escape hatch, hmmm ... It can be implemented in various ways. I would be in favor of that option if we had not already provided so many dedicated fields that we very seldom get requests to add another configuration field any more.

I'm quite strongly in favor of making our configuration as transparent and non-complicated as possible to avoid breaking anyones assumptions or forcing users to read details in the docs to get things right. I think having two separate options is the best call here, but if I would rewrite this Helm chart from scratch I think I'd favor an escape hatch for a lot of rarer options.

Hmmm...

If we would have an escape hatch it would be `hub.extraPodConfig` or `hub.extraPodSpec` or similar I think, as a dictionary which we render with `toYaml` and then append...

Hmmm hmmm... I'm a bit conflicted, not sure what I think about having `hub.extraPodSpec` or adding `hub.hostNetwork` + `hub.dnsPolicy`... Maybe `hub.extraPodSpec` after all?

What do you think at this point? Sorry for just writing from the top of my head, thanks for taking the time to think with me about this!

consideRatio: I'm currently leaning towards favoring having `<workload-config>.extraPodSpec` for the four workloads below, and that we let `hostNetwork`, `dnsPolicy` etc be configured via that.

- `hub` (hub deployment)
- `proxy.chp` (proxy deployment)
- `proxy.traefik` (autohttps deployment)
- `scheduling.userScheduler` (user-scheduler deployment)

The only thing that I'm feeling strongly opinionated about is to not seem like k8s native config but actually be a config that does more things than would be obvious - I've had a bit too much bad experiences with that both as a user and as a maintainer getting asked questions about such configs.

mallman: I agree an `*.extraPodSpec` is the better way to go. It gives the deployer an option to add their own special pod spec if necessary without ad-hoc patching and/or PRs. It should be clear this is for people who know what they're doing, and may conflict with current or future versions of the helm chart.

If we go the way of adding `hub.hostNetwork` + `hub.dnsPolicy`, we also need to add `hub.dnsConfig`, which is even more complex. And then we need to add that to the three other deployments. I think think a catch-all `extraPodSpec` is both sufficient to any special pod configuration needs while simple enough that it's basically a "pass-through" of k8s configuration.

I'll work on `*.extraPodSpec` for your review and update the title of this PR when I've pushed the new changes. Or I'll close this PR and add a new one... We'll see.

consideRatio: Thank you @mallman for your patience and effort into this! I feel very happy about where this is headed at the moment as we have deliberated on it and both agree it is reasonable and sustainable! :heart: :tada: :sunflower:

mallman:
I've created a new branch and PR #2306 to continue on with `extraPodSpec`. I will close this PR now.

Issue #2146, opened 2021-04-13
Add CI tests for arm64


manics:
#2119 adds support for arm64 in Z2JH. The Docker image builds are cross architecture but they're not run in any test.

- We could add this with CircleCI which has arm VMs available for free on request https://github.com/jupyterhub/team-compass/issues/396
- Or a self-hosted GitHub runner on arm

If we're unable to use the existing test workflow directly we can use a cut down workflow since the aim is to test the functionality of the docker images and _not_ the logic of the helm chart.

manics:
Circle CI never got back to me, back it looks like they've enabled ARM resources for everyone by default :smile:

PR #2302, opened 2021-07-02
MRG: Arm64 circleci test

manics:
Experimenting with a minimal CircleCI ARM64 test.

This is currently running in my own [CircleCI account](https://app.circleci.com/pipelines/github/manics/zero-to-jupyterhub-k8s?branch=arm64-circleci), it'll obviously need to be added in the JupyterHub CircleCI account if we go ahead with this.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2146

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,67 @@
+version: 2.1
+
+orbs:
+  python: circleci/python@0.2.1
+
+jobs:
+  # Testing on ARM64
+  # https://circleci.com/docs/2.0/arm-resources/#using-arm-resources
+  test-arm:
+    machine:
+      image: ubuntu-2004:202101-01
+    # TODO: resource_class is ignored and falls back to the standard x86_64 medium
+    # Need to apply to use arm
+    resource_class: arm.medium
+    steps:
+      - checkout
+
+      - run:
+          command: uname -a
+          name: Check architecture
+
+      - run:
+          command: >-
+            curl -sfL https://get.k3s.io |
+            INSTALL_K3S_CHANNEL=v1.20 sh -s -
+            --disable metrics-server
+            --disable traefik
+            --docker
+          name: Install K3S
+
+      - run:
+          command: |
+            mkdir -p ~/.kube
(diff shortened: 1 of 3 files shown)


manics:
To keep things simple I'm juts running `chartpress` in the arm64 VM without `docker buildx`- this also has the advantage of testing the non buildx setup for developers.

Unfortunately this means it tries to build the singleuser image... and fails...

.circleci/config.yml
```diff
@@ -42,9 +42,13 @@ jobs:
...
           name: Install dependencies
 
+      # Ideally we'd use standard docker build since we're building for the native
+      # platform. Unfortunately the singleuser image can't be built on arm64, and the
+      # easiest way to skip this image is to user chartpress's --platform option,
+      # which is only supported with docker-buildx engine
       - run:
           command: |
-            chartpress
+            chartpress -builder docker-buildx
```
consideRatio, line 51: `--builder`, right?

manics: Yes, along with needing to be explicit about the platform.

manics:
To my surprise this actually runs!

dev-config-arm.yaml
```diff
@@ -0,0 +1,9 @@
+# Override some values in dev-config.yaml
+proxy:
+  https:
+    enabled: false
```
consideRatio, line 4: Did this have to be disabled?

.circleci/config.yml
```diff
@@ -0,0 +1,70 @@
+version: 2.1
```
consideRatio, line 1: Can you document that we need CircleCI for their arm64 runners?

dev-config-arm.yaml
consideRatio: Is it disabled to avoid adding too much additional steps done in the GitHub action to setup k3s with calico that is troublesome to copy over to circle ci? That is a good reason!

.circleci/config.yml
consideRatio: Aaah well obvious enough based on job names etc.

consideRatio:
Nice work @manics!!! :heart: :tada:

PR #2305, opened 2021-07-06
Add ingress.pathType config

jtrouth:
This adds the ability to set the pathType field on ingress rules for situations where Prefix is not available or appropriate, such as on GKE 1.19+ with the native ingress controller which only accepts 'ImplementationSpecific' as the pathType.


consideRatio (approved):
Thanks @jtrouth and congrats on your first code contribution to this repo! :heart: :tada:

PR #2306, opened 2021-07-07
Add configuration for arbitrary extra pod spec

mallman:
Add configuration for arbitrary extra pod spec for the hub, chp, traefik and user scheduler deployments.

This change arose from a desire to deploy the hub and proxy on the host network. See PR #2300 for that discussion.

Following that conversation, I am pursuing an alternate strategy whereby we provide an escape hatch for the end user to specify arbitrary pod specification for the hub, chp, traefik and user scheduler deployments.

The extra pod specification can be provided in four locations in values.yaml:

  1. hub.extraPodSpec
  1. proxy.chp.extraPodSpec
  1. proxy.traefik.extraPodSpec
  1. scheduling.userScheduler.extraPodSpec

For example,

```yaml
hub:
  extraPodSpec:
    hostNetwork: true
    dnsPolicy: ClusterFirstWithHostNet
```

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1160,13 +1160,13 @@ properties:
             patternProperties: *labels-and-annotations-patternProperties
             description: |
               Kubernetes annotations to apply to the k8s ServiceAccount.
-      extraPodSpec: &jupyterhub-extra-pod-spec
+      extraPodSpec: &extraPodSpec-spec
         type: object
         additionalProperties: true
         description: |
           Arbitrary extra k8s pod specification as a YAML object. The default
           value of this setting is an empty object, i.e. no extra configuration.
-          The value of this property is merged into the pod specification as-is.
+          The value of this property is augmented to the pod specification as-is.
 
           This is a powerful tool for expert k8s administrators with advanced
           configuration requirements. This setting should only be used for
@@ -1293,7 +1293,7 @@ properties:
               or produce more informative error messages than the Hub's default,
               e.g. in highly customized deployments such as BinderHub.
               See Configurable HTTP Proxy for details on implementing an error target.
-          extraPodSpec: *jupyterhub-extra-pod-spec
+          extraPodSpec: *extraPodSpec-spec
       secretToken:
         type: [string, "null"]
         description: |
@@ -1614,7 +1614,7 @@ properties:
           image: *image-spec
           resources: *resources-spec
           serviceAccount: *serviceAccount
-          extraPodSpec: *jupyterhub-extra-pod-spec
+          extraPodSpec: *extraPodSpec-spec
       labels:
         type: object
         additionalProperties: false
@@ -2139,7 +2139,7 @@ properties:
                           type: integer
           resources: *resources-spec
           serviceAccount: *serviceAccount
-          extraPodSpec: *jupyterhub-extra-pod-spec
+          extraPodSpec: *extraPodSpec-spec
       podPriority:
         type: object
         additionalProperties: false
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1174,7 +1174,17 @@ properties:
           Misusing this setting can break your deployment and/or compromise
           your system security.
 
-          One use of this setting is to configure the hub pod with host networking:
+          This is one of four related settings for inserting arbitrary pod
+          specification:
+
+          1. hub.extraPodSpec
+          2. proxy.chp.extraPodSpec
+          3. proxy.traefik.extraPodSpec
+          4. scheduling.userScheduler.extraPodSpec
+
+          One real-world use of these settings is to enable host networking. For
+          example, to configure host networking for the hub pod, add the
+          following to your helm configuration values:
 
           ```yaml
           hub:
@@ -1183,7 +1193,8 @@ properties:
               dnsPolicy: ClusterFirstWithHostNet
           ```
 
-          To configure the proxy pod with host networking:
+          Likewise, to configure host networking for the proxy pod, add the
+          following:
 
           ```yaml
           proxy:
@@ -1193,10 +1204,11 @@ properties:
                 dnsPolicy: ClusterFirstWithHostNet
           ```
 
-          and so on...
+          N.B. Host networking has special security implications and can easily
+          break your deployment. This is an example—not an endorsement.
 
           See [PodSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec)
-          for the latest, complete definition of the pod specification.
+          for the latest pod resource specification.
 
   proxy:
     type: object
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1168,14 +1168,13 @@ properties:
           value of this setting is an empty object, i.e. no extra configuration.
           The value of this property is merged into the pod specification as-is.
 
-          This is a powerful tool for _expert_ k8s administrators with advanced
+          This is a powerful tool for expert k8s administrators with advanced
           configuration requirements. This setting should only be used for
-          configuration that cannot be accomplished through the other settings
-          in your values.yaml. Misusing this setting can break your deployment
-          and/or compromise your system security.
+          configuration that cannot be accomplished through the other settings.
+          Misusing this setting can break your deployment and/or compromise
+          your system security.
 
-          For example, to configure the hub pod with host networking, you could
-          include the following in your chart values:
+          One use of this setting is to configure the hub pod with host networking:
 
           ```yaml
           hub:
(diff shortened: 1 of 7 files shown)


mallman:
Hmmm... in reviewing the generated documentation, I think the examples should be customized for each particular setting. Any idea how to do that with a minimum of copy-pasta? As an alternative I could just remove the examples, but I think they're helpful.

jupyterhub/schema.yaml
```diff
@@ -1160,6 +1160,43 @@ properties:
             patternProperties: *labels-and-annotations-patternProperties
             description: |
               Kubernetes annotations to apply to the k8s ServiceAccount.
+      extraPodSpec: &jupyterhub-extra-pod-spec
+        type: object
+        additionalProperties: true
+        description: |
+          Arbitrary extra k8s pod specification as a YAML object. The default
+          value of this setting is an empty object, i.e. no extra configuration.
+          The value of this property is merged into the pod specification as-is.
```
consideRatio, line 1169:
```suggestion
          The value of this property is augmented to the pod specification as-is.
```

jupyterhub/schema.yaml
```diff
@@ -1160,6 +1160,43 @@ properties:
             patternProperties: *labels-and-annotations-patternProperties
             description: |
               Kubernetes annotations to apply to the k8s ServiceAccount.
+      extraPodSpec: &jupyterhub-extra-pod-spec
```
consideRatio, line 1163:
```suggestion
      extraPodSpec: &extraPodSpec-spec
```

This aligns with a not so obvious way of naming these anchors for whenever they reference a duplicated configuration. There are other anchors named differently when they are referencing something entirely different etc.

jupyterhub/schema.yaml
```diff
@@ -1160,6 +1160,43 @@ properties:
...
+
+          ```yaml
+          proxy:
+            chp:
+              extraPodSpec:
+                hostNetwork: true
+                dnsPolicy: ClusterFirstWithHostNet
+          ```
+
+          and so on...
```
consideRatio, line 1196: Since this will be showing up in the documentation four times, one for the hub, proxy.chp, proxy.traefik, and scheduling.userScheduler, I figure we can reduce the examples provided to only one.

```suggestion
          One use of this setting is to configure the hub pod with host networking:

          ```yaml
          hub:
            extraPodSpec:
              hostNetwork: true
              dnsPolicy: ClusterFirstWithHostNet
          ```
```

jupyterhub/schema.yaml
```diff
@@ -1256,6 +1293,7 @@ properties:
               or produce more informative error messages than the Hub's default,
               e.g. in highly customized deployments such as BinderHub.
               See Configurable HTTP Proxy for details on implementing an error target.
+          extraPodSpec: *jupyterhub-extra-pod-spec
```
consideRatio, line 1296:
```suggestion
          extraPodSpec: *extraPodSpec-spec
```

jupyterhub/schema.yaml
```diff
@@ -1576,6 +1614,7 @@ properties:
           image: *image-spec
           resources: *resources-spec
           serviceAccount: *serviceAccount
+          extraPodSpec: *jupyterhub-extra-pod-spec
```
consideRatio, line 1617:
```suggestion
          extraPodSpec: *extraPodSpec-spec
```

jupyterhub/schema.yaml
```diff
@@ -2100,6 +2139,7 @@ properties:
                           type: integer
           resources: *resources-spec
           serviceAccount: *serviceAccount
+          extraPodSpec: *jupyterhub-extra-pod-spec
```
consideRatio, line 2142:
```suggestion
          extraPodSpec: *extraPodSpec-spec
```

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -79,4 +79,7 @@ spec:
           securityContext:
             {{- . | toYaml | trimSuffix "\n" | nindent 12 }}
           {{- end }}
+      {{- if .Values.scheduling.userScheduler.extraPodSpec }}
+      {{ toYaml .Values.scheduling.userScheduler.extraPodSpec | nindent 6 }}
+      {{- end }}
```
consideRatio, line 84:
```suggestion
      {{- with .Values.scheduling.userScheduler.extraPodSpec }}
      {{- . | toYaml | trimSuffix "\n" | nindent 6 }}
      {{- end }}
```

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -163,3 +163,6 @@ spec:
           securityContext:
             {{- . | toYaml | trimSuffix "\n" | nindent 12 }}
           {{- end }}
+      {{- if .Values.proxy.chp.extraPodSpec }}
+      {{ toYaml .Values.proxy.chp.extraPodSpec | nindent 6 }}
+      {{- end }}
```
consideRatio, line 168:
```suggestion
      {{- with .Values.proxy.chp.extraPodSpec }}
      {{- . | toYaml | trimSuffix "\n" | nindent 6 }}
      {{- end }}
```

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -135,4 +135,7 @@ spec:
           securityContext:
             {{- . | toYaml | trimSuffix "\n" | nindent 12 }}
           {{- end }}
+      {{- if .Values.proxy.traefik.extraPodSpec }}
+      {{ toYaml .Values.proxy.traefik.extraPodSpec | nindent 6 }}
+      {{- end }}
```
consideRatio, line 140:
```suggestion
      {{- with .Values.proxy.traefik.extraPodSpec }}
      {{- . | toYaml | trimSuffix "\n" | nindent 6 }}
      {{- end }}
```

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -230,3 +230,6 @@ spec:
               path: {{ .Values.hub.baseUrl | trimSuffix "/" }}/hub/health
               port: http
           {{- end }}
+      {{- if .Values.hub.extraPodSpec }}
+      {{ toYaml .Values.hub.extraPodSpec | nindent 6 }}
+      {{- end }}
```
consideRatio, line 235:
```suggestion
      {{- with .Values.hub.extraPodSpec }}
      {{- . | toYaml | trimSuffix "\n" | nindent 6 }}
      {{- end }}
```

consideRatio (commented):
This looks great to me! I made some suggestions of changes - if you want to approve those changes please use "add suggestion to batch" and then add commit all in one go to avoid creating loads of commits.

Can you also add entries in tools/lint-and-validate-values.yaml with a concrete example such as `<workload>.extraPodSpec.hostNetwork: true` so we validate it renders correctly via our CI system?

jupyterhub/templates/hub/deployment.yaml
consideRatio: Note that `with` is conditional, if its a falsy value - it will not evaluate anything inside its block. `trimSuffix "\n"` is added because `toYaml` has a quirk of always adding a `\n` in the end which makes the rendered YAML look a bit ugly.

mallman:
> Can you also add entries in tools/lint-and-validate-values.yaml with a concrete example such as <workload>.extraPodSpec.hostNetwork: true so we validate it renders correctly via our CI system?

Will do.

jupyterhub/schema.yaml
mallman: I like the idea of having a concrete example in the documentation. However, if you look at the documentation for `proxy.chp.extraPodSpec`, it gives an example for "this setting" that is not for "this setting". See what I mean? I think it's a bit awkward. I'm going to think about this one a bit more.

consideRatio: Yes it is a bit awkward to not adjust the text based on setting =\ i dont know of a way that feels viable to fix that.

We can retain this example as it is also btw, not strongly opinionated about this. Let me know if you think the state is ok to merge from your end.

mallman:
> > Can you also add entries in tools/lint-and-validate-values.yaml with a concrete example such as .extraPodSpec.hostNetwork: true so we validate it renders correctly via our CI system?
> 
> Will do.

Done.

mallman:
@consideRatio I'm pretty happy with the latest documentation I pushed. What do you think?

consideRatio (approved):
Thank you for your thorough effort put into this PR @mallman - I appreciate the care you have had for the documentation part!! :heart: :tada: :sunflower:

consideRatio:
TIL what N.B. refers to ^^

mallman:
> TIL what N.B. refers to ^^

It's Latin for _note bene_ (note well).

https://en.wikipedia.org/wiki/Nota_bene

PR #2307, opened 2021-07-08
Remove deprecation logic for hub.extraConfig as a string

### LINKED ISSUE #2063: hub.extraConfig as a string is not supported - update docs and remove potential related logic
As reported by @danielballan in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2062#issue-578712676 (point 1), it may not be supported for us to set hub.extraConfig to a string.

> ```yaml
> hub:
>   extraConfig: |
>     import time
>     c.Spawner.environment += {
>        "CURRENT_TIME": str(time.time())
>     }
> ```
> 
> The documentation claims that extraConfig accepts either a dict of strings or a single string (of Python code), but it actually only accepts the first one. If you give it a single string, following the documented example, it warns
> coalesce.go:196: warning: cannot overwrite table with non table for extraConfig (map[])
> 
> and it does not take effect. See #653 which seems to observe the same issue.

## Action point
1. Reproduce and conclude that helm 3 and 0.11.1 that this doesn't work.

   ```yaml
   hub:
     extraConfig: |
       print("THIS SHOULD BE SEEN IN LOGS ON STARTUP, IS IT?")
   ```

consideRatio:
Closes #2063 by:
- updating the documentation to reflect it must be a dictionary as have been enforced by schema in JupyterHub 1.0.0 and was deprecated in 0.8.0.
- removing logic to handle hub.extraConfig being a string in our jupyterhub_config.py file that now assumes its a dictionary instead

### DIFF
--- a/doc/source/administrator/advanced.md
+++ b/doc/source/administrator/advanced.md
@@ -88,40 +88,24 @@ There are four properties you can set in your `config.yaml` to do this.
 
 ### `hub.extraConfig`
 
-The value specified for `hub.extraConfig` is evaluated as python code at the end
-of `jupyterhub_config.py`. You can do anything here since it is arbitrary Python
-Code. Some examples of things you can do:
+The value specified for `hub.extraConfig` is evaluated as Python code at the end
+of a `jupyterhub_config.py` file JupyterHub loads. You can do anything in
+`hub.extraConfig` since it is arbitrary Python code, while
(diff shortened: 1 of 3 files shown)


manics (commented):
LGTM

consideRatio:
Thanks for the review work @manics! :heart:

PR #2312, opened 2021-07-13
Seed hub.services' apiTokens

### LINKED ISSUE #2308: Autogenerate and set passwords for JupyterHub's registered services' api_tokens
### Proposed change

To help charts like dask-gateway and binderhub that depend on z2jh to no longer need to explicitly set an api_token via `jupyterhub.hub.services.dask-gateway.api_token` for example. To instead let them read an automatically created api_token for their service via a k8s Secret.

This is very similar to the pr to seed this Helm chart's internally used tokens as was done in #1993.

### Alternative options

To do nothing and to still manually require creating and managing a secret api_token for a JupyterHub registered service, which also needs to be passed to the service in undefined ways for trusted communication between the service and JupyterHub.

This is currently done like this for dask-gateway for example, as described [in its README.md file](https://github.com/dask/helm-chart/tree/main/daskhub#prepare-configuration-file).

```yaml
jupyterhub:
  hub:
    services:
      dask-gateway:
        apiToken: "<secret-token>"

dask-gateway:
  gateway:
    auth:
      jupyterhub:
        apiToken: "<secret-token>"
```

Any user of binderhub, dask-gateway, and other Helm charts as those currently require configuration like above. With this feature, those chart can make it not needed.

### Implementation ideas

No matter what, we should have a way to expose the api_token of registered services, but this is already done.

- [x] We automatically declare api_tokens in the hub k8s Secret under a key named `hub.services.<service-name>.apiToken` in the k8s Secret typically named `hub` but more generally named in a way that can be dynamically set via `{{ include "jupyterhub.hub.fullname" . }}`.
  ```yaml
  data:
    # Any JupyterHub Services api_tokens are exposed in this k8s Secret as a
    # convinience for external services running in the k8s cluster that could
    # mount them directly from this k8s Secret.
    {{- range $key, $service := .Values.hub.services }}
    {{- if or $service.api_token $service.apiToken }}
    hub.services.{{ $service.name | default $key }}.apiToken: {{ $service.api_token | default $service.apiToken | b64enc | quote }}
    {{- end }}
    {{- end }}
  ```

#### Strategy 1 - specific automation of generating hub.services' api_token (#2312)

- [x] We automatically create an api_token for all registered JupyterHub services that doesn't explicitly set one.
- [x] We document that we do this, and give an example on how that api_token can be accessed via the hub k8s Secret.

#### Strategy 2 - generic password generation and api_token configuration

- [ ] We create an option to generate passwords to be stored in a k8s Secret
- [ ] We create an option for api_tokens to reference the value in a k8s Secret's specific key

consideRatio:
Closes #2308 by automatically seeding `hub.services` entries that doesn't have apiToken or api_token set, but reuse explicitly set tokens if they are explicitly set.

### Review notes

- [ ] Agree to introduce non-essential breaking change motivated below?
- [ ] Agree to always generate and set an api_token that becomes exposed via the `hub` k8s Secret?
  Note that JupyterHub 1.4.0 introduced a regression that forces us to set it that has now been fixed in the main branch, but it isn't yet released. If we always generate a token, this will avoid problems with that. See https://github.com/jupyterhub/jupyterhub/pull/3531.
- [ ] Does the implementation look ok?
- [ ] Is it documented ok? (PR preview [here](https://zero-to-jupyterhub--2312.org.readthedocs.build/en/2312/resources/reference.html#hub-services))

### About non-essential breaking change introduced

I've introduced an isolated commit (925a4a6) with a very minor breaking change that is unlikely to influence people. The motivation is that it will be the name I think people expect as well as it will be something far more consistent in its naming by actually being the path to a certain value in the config.

It is a breaking change if _both_ the following conditions apply, which I find unlikely and not merit a 2.0.0 release (but 1.1.0) and still worth doing to reduce complexity etc.

1. You have configuration like this:
   ```yaml
   hub:
     services:
       # potentially breaking: you have a config key different from the name like below
       my-service-config-key:
         name: my-service-name
   ```
2. You reference the `hub` k8s Secret's key from something external to the JupyterHub Helm chart itself, previously named `hub.services.my-service-name.apiToken`, but is now named `hub.services.my-service-config-key.apiToken`.


yuvipanda (approved):
Thanks for working on this, @consideRatio.. I think the breaking change is fine and doesn't need a major version bump.

I'll leave this open for a few more days for others to review. I think this will be great for BinderHub!

consideRatio:
Thanks for your review @yuvipanda, does it feel okay to go for a merge at this point?

consideRatio:
I did a final self-review of this and it LGTM. I consider this to be evaluated well enough to be merged at this point.

PR #2343, opened 2021-08-05
additionalProperties should be set to True for templateVars in the schema

MridulS:
While trying to set `templateVars` we can't set additionalProperties, this PR fixes that

A minimal example
```
hub:
  templateVars:
    announcement: 'Some Text'
```

and deploying with
```
helm upgrade --install jhub jupyterhub/jupyterhub --version=1.1.1 --values config.yaml
```

fails the lint with the current schema.


consideRatio:
Thanks for also fixing the broken link! :tad: :heart:

consideRatio:
@MridulS I released 1.1.2 with this.

PR #2355, opened 2021-08-17
Test Arm64 CI

consideRatio:


manics:
Is this for your own self-hosted ARM64 cluster? 😀

How are you handling the [potential security issues](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories)?

consideRatio:
Yepp @manics! I fail to get my runners to pick up the work though for some reason.

Here are some notes about the bullet points listed, I'm not confident how I feel yet but is exploring before I propose these are actively adopted. I'm relying on the [actions-runner-controller/actions-runner-controller](https://github.com/actions-runner-controller/actions-runner-controller) Helm chart.

- _Malicious programs running on the machine._
  No actual protection for something to run within the container.
- _Escaping the machine's runner sandbox._
  Kubernetes pods, but if they would be able to control the node instead of the pod, that would be bad.
- _Exposing access to the machine's network environment._
  I've used a NetworkPolicy to forbid the pods to communicate to anything in the local network.
- _Persisting unwanted or dangerous data on the machine._
  The pods start with EmptyDir volumes only and restarts between job.

consideRatio:
So I've managed to get my runners to function, but doh... setup/python can't work no matter what image I use etc, and it feels like we can't reuse GitHub actions' workflows anyhow so well if we use runners in general so I've lost enthusiasm about this sadly.

PR #2358, opened 2021-08-21
refactor: remove redundant trimSuffix of new lines after toYaml

consideRatio:
In the past `toYaml` created a new line that we trimmed away - but I contributed a fix to that in helm/helm so now it doesn't in the modern versions of the `helm` CLI which the JupyterHub helm chart now require anyhow :)

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -322,9 +322,9 @@ limits:
         {{- if $file_details.data }}
             {{- $file_key | quote }}: |
               {{- if or (eq (ext $file_name) ".yaml") (eq (ext $file_name) ".yml") }}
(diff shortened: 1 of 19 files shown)


manics:
Is the minimum Helm version enforced in the chart?

consideRatio:
The minimum Helm version for this chart is 3.5.0, and it will error without that version, and helm 3.5.0 behaves in a way making this redundant. I'm not sure when in helm 3 this was fixed, but it was fixed in helm 3 after they first merged it into helm2 and concluded it had to wait for helm3.

I concluded also that our helm diff test step in the CI jobs making a upgrade from a previous version also didn't report any changes to the rendered outcome.

manics (approved):
Thanks for confirming. I was worried changes like this might lead to strange error messages for someone using an out of date Helm (as opposed to new syntax which would fail immediately) but it sounds like it's all taken care of 😄

PR #2376, opened 2021-09-04
Remove no longer needed arm test adjustment

consideRatio:
We have a arm compatible singleuser image at this point in time, no need to override it.


manics (approved):
LGTM! Any idea why circle CI isn't running?

consideRatio:
Probably github failing temporary or so

PR #2395, opened 2021-09-21
Tighten permissions for jupyterhub-idle-culler

consideRatio:


manics (approved):
Matches the doc https://github.com/jupyterhub/jupyterhub-idle-culler#permissions

I wonder if (in future) we can automatically test this somehow.....

minrk:
Testing would be awesome. We could launch a Hub with a short idle-interval and nullauth/simple spawner. Might be a little tricky, but I run these exact tests by hand locally with some toy config

PR #2397, opened 2021-09-21
Remove breaking change messages relevant for upgrading to 1.0.0

consideRatio:
By removing these, we can simplify our schema.yaml file and various helm
templates. The assumption is that users won't upgrade from z2jh `0.*` to
`>=2` skipping `1.*`, if they would do that they would get errors that
is less user friendly from the Helm schema validation logic.

If you are upgrading to Z2JH 2+, first upgrade to Z2JH 1+ - don't go
straight from `0.*`!


yuvipanda:
So after we merge this, if users run a `helm upgrade` with a new z2jh version, they'll get a proper failure from the schema validation failure, and will have to figure out what to do themselves. But it won't 'succeed' and be broken. Is that correct? If so, happy to merge :)

consideRatio:
> So after we merge this, if users run a helm upgrade with a new z2jh version, they'll get a proper failure from the schema validation failure, and will have to figure out what to do themselves. But it won't 'succeed' and be broken. Is that correct? If so, happy to merge :)

Fully correct, assuming they upgrade from `0.*` to `2.*` that is!

PR #2401, opened 2021-09-25
ci: use PVCs when testing upgrades

manics:
This ensures we're testing whether persistent data like the database is actually upgraded- `sqlite-memory` is not persisted when the hub restarts.

Best verified using https://github.com/jupyterhub/action-k8s-namespace-report/pull/11

e.g.
image

In future we could add tests to verify the persistent volumes, especially singleuser volumes, are correctly maintained across upgrades.


consideRatio:
@manics perfect! Nice catch!

Issue #2386, opened 2021-09-16
Release 2.0


consideRatio:
As JupyterHub 2.0 is coming up, we need to cut JupyterHub Helm chart 2.0 when we bump, because that bump will be a disruptive upgrade that forces user servers to restart etc.

As part of releasing Z2JH 2.0 we should try figure out what we want to get done.

### Things to consider get done for Z2JH 2.0

- [x] Get JupyterHub 2.0.0b0 released
- [x] Cleanup deprecation logic related to pre Z2JH 1.0.0
  #2397 
- [x] Consider if we need or should expose some options besides what is configurable via `hub.config`, for example c.JupyterHub.load_roles etc. It is currently an array, should we make it a dictionary like hub.services perhaps?
  #2405
  hub.config.JupyterHub.load_groups can be used, because that is a dictionary.
- [ ] jupyterhub-idle-culler related:
  - [x] Update the permissions we set based on https://github.com/jupyterhub/jupyterhub-idle-culler/issues/31
    #2395
  - [x] Get https://github.com/jupyterhub/jupyterhub-idle-culler/pull/30 merged
  - [x] Cut a jupyterhub-idle-culler release
  - [x] Bump jupyterhub-idle-culler in z2jh's hub image
  - [x] Cut a jupyterhub-idle-culler release 1.2.1
  - [x] Bump jupyterhub-idle-culler to 1.2.1 (#2442)
  - [ ] Setup generic passthrough of culler flags for long term sanity.
- [x] Use JupyterLab by default perhaps
  @manics wrote in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776#issuecomment-920230682
  > Shall we try and figure out something that works? If we can pass all the relevant options as environment variables that minimises breakage of other singleuser implementations.
- [x] Resolve https://github.com/jupyterhub/nativeauthenticator/issues/157
  - [ ] Release a new version of nativeauthenticator
  - [ ] Bump nativeauthenticator

### How to release?

I suggest we bump jupyterhub==2.0.0b0 within the hub image in the main branch, rather than that we create a feature branch and such. I consider this is a very technical choice on how to make maintenance easy. With regards to that, I'm feeling very comfortable backporting a few PRs to 1.1.x branches if we need to backport something before we cut 2.0.0.

Release steps:
- Z2JH Alpha for JupyterHub 2.0.0 beta
- Z2JH Beta for JupyterHub 2.0.0
[...]

PR #2405, opened 2021-09-25
Add hub.loadRoles configuration

consideRatio:
The motivation for adding a dedicated chart configuration is that
hub.config.JupyterHub.load_roles would end up being overridden easily if
configured from multiple config files. So, having a dictionary
configuration for the same thing can help.

An example for when this would be relevant is the BinderHub helm chart. It defines a hub.service and will want to define a hub.loadRole as well. But then what happens if a user wants to configure the BinderHub Helm chart? The user will end up having overridden BinderHub's config of the JupyterHub Helm chart.

I've deliberated a while on this, and some comments are made in #2386 about it.

### DIFF
--- a/dev-config-local-chart-extra-config.yaml
+++ b/dev-config-local-chart-extra-config.yaml
@@ -1,0 +1,26 @@
+# This config file is useful in the upgrade tests, where we upgrade from
+# either the latest stable chart or the latest dev release of the chart. This
+# config is only applied to the chart we upgrade to. It helps us handle
+# situations when we add new configuration options that would fail with a schema
+# validation error in the previous chart versions.
+#
+# Note that one could think that it would be possible to have dev-config.yaml
+# include this config and then pass --set hub.some-option=null to null it out
+# when it must not be passed, but that still triggers schema validation errors.
+#
+hub:
+  # FIXME: move loadRoles to dev-config.yaml after 2.0.0 is released.
+  loadRoles:
+    test-role-1:
+      description: Access to users' information and group membership
+      scopes: [users, groups]
+      users: [cyclops, gandalf]
+      services: [test]
+      groups: []
+    test-role-2-explicit-name:
+      name: test-role-2
+      description: Access to users' information and group membership
+      scopes: [users, groups]
+      users: [cyclops, gandalf]
+      services: [test]
+      groups: []
--- a/dev-config-local-chart-extra-config.yaml
+++ b/dev-config-local-chart-extra-config.yaml
@@ -11,13 +11,11 @@
 hub:
   # FIXME: move loadRoles to dev-config.yaml after 2.0.0 is released.
   loadRoles:
-    test-role-1:
-      description: Access to users' information and group membership
-      scopes: [users, groups]
-      users: [cyclops, gandalf]
-      services: [test]
-      groups: []
-    test-role-2-explicit-name:
+    test-scoped-access:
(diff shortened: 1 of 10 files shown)


dev-config-local-chart-extra-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+      users: [cyclops, gandalf]
+      services: [test]
+      groups: []
+    test-role-2-explicit-name:
+      name: test-role-2
+      description: Access to users' information and group membership
+      scopes: [users, groups]
+      users: [cyclops, gandalf]
+      services: [test]
+      groups: []
```
manics, line 26: Do you reckon we can test this in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/tests/test_hub.py by making a couple of API calls, one that should be allowed and one that should be blocked?

consideRatio: I'd like to avoid ending up testing jupyterhub's own functionality, that is already tested. But, this is a z2jh config, so it makes sense to verify it is actually loading this configuration properly.

We also have the idle-culler configured and setup, it is currently failing with permission issues. Having a test to verify that doesn't fail is on my radar. #2407

Okay to settle for just testing that our idle-culler works for now?

manics: The idle-culler problems was why I thought a test would be good. If we can get a full idle-culler test (running, and also ideally test the culling) that would be great!

consideRatio: This chart config isnt used by the idle culler, it directly adds to c.JuoyterHub.load_roles and services without doing anything here.

Hmmm, well i think perhaps doing a dedicated test is easier. Ill add a service and a key, then make a call to hub/api/info with a hub:read scoped token to be coupled with the service.

I note that to be failing currently for idle culler, so could be nice to verify separately as well.

manics: It's a change from what we've done in the past where most new config changes aren't explicitly tested, but I think it'd be good to try and do so in future, especially as the chart continues to grow in complexity. We could even use it as a guide for whether to accept a new feature- if it can't be tested in the chart then maybe it should be supported via a generic extension point rather than a config parameter?

consideRatio: Im open to ideas on new tests, but the implementation idea matters a lot - are they complicated or easy to understand is a key thing. But we are out of scope for this PR to discuss that here.

---

@manics ive added a test to verify a token is granted access to an endpoint correctly.

jupyterhub/schema.yaml
```diff
@@ -1201,6 +1201,40 @@ properties:
...
+          This is where you should define JupyterHub roles and apply them to
+          JupyterHub users, groups, and services to grant them additional
+          permissions as defined in JupyterHub's RBAC system.
+
+          Complement this documentation with [JupyterHub's
+          documentation](https://jupyterhub.readthedocs.io/en/latest/rbac/roles.html#defining-roles)
+          about `load_roles`.
+
+          Note that while JupyterHub's native configuration `load_roles` accepts
+          a list of role objects, this Helm chart only accept a dictionary where
```
manics, line 1217:
```suggestion
          a list of role objects, this Helm chart only accepts a dictionary where
```

manics (commented):
Thanks, one typo then this is good to merge!

consideRatio:
Thank you for your review efforts @manics! :heart: :tada:

Issue #2399, opened 2021-09-24
Upgrading k8s-hub image from 0.9->1.1.3 on AKS changes PVC syntax for named servers, thus creating new PV's and makes old storage unreachable by default


bjornarfjelldal:
### Bug description

Hi, guys! Thanks for a great tool and for your great work maintaining and developing it. I have encountered a bug:

When upgrading the k8s-hub image from version 0.9 to 1.1.3, the persistant volume claim syntax on named servers on Azure changes from "**claim-username-servername**" to "**claim-usernameservername**".

This causes new PVC's and corresponding PV's to be created, and the old ones are no longer mounted - giving uses the impression that all files are gone.

When downgrading the k8s-hub image, I verify that the old syntax is again used, and the old PV's are mounted again.

#### Expected behaviour

The same PVC should be used mounting the same PV as before

#### Actual behaviour

A new PVC is created, creating a new PV

### How to reproduce

1. Use the jupyterhub/k8s-hub image with the tag 0.9.0 on an Azure Kubernetes Service following the excellent zero-to-kubernetes guide.
2. Enable "named servers", and create a server with a given user logged in. The perisant volume claim name should then be "claim-username-servername"
3. Upgrade the k8s-hub image to 1.1.3 (I have no idea where this change was introduced, but this is the version I upgraded to) and re-deploy
4. Start the same server previously created in step 2. A new PVC is then created with the name "claim-usernameservername", and the old PVC/PV is unreachable without manual configuration.
5. 
### Your personal set up

Azure Kubernetes Service (AKS) and Azure Container Registry (ACR), following the zero-to-kubernetes guide.

k8s-hub image is locally built with a nativeauthenticator as per Dockerfile below, before pushed to ACR:
```

FROM jupyterhub/k8s-hub:1.1.3
USER root

RUN mkdir -p /srv/auth

RUN git clone https://github.com/jupyterhub/nativeauthenticator.git /srv/auth

RUN pip install --no-cache -e /srv/auth

USER jovyan
```

Then it is referred to in the config.yaml:

```
hub:
  image: 
    name: myacr.azurecr.io/jupyterhub/jupyterhub_native_auth
    tag: 1.1.3
```

And updated using the latest helm chart repo:

```
helm upgrade --cleanup-on-fail \
  --install jupyterhub jupyterhub/jupyterhub \
  --namespace jupyterhub \
  --create-namespace \
  --version=1.1.3 \
  --values config.yaml
[...]

PR #2410, opened 2021-09-26
Retrospectively add breaking change to changelog entry 0.10.0

consideRatio:


manics (approved):
Note this does not close #2399 which reports an absence of the `--` separator

PR #2421, opened 2021-09-30
Fix proxy pod's liveness/readiness probes to be fully configurable

### LINKED ISSUE #2415: proxy.chp.readinessProbe.failureThreshold is not used
### Bug description

We're using z2jh 0.11.1 and I was noticing chp probe failures in our deployment. One thing I noticed was the readiness probe config on the proxy pod does not match the failure threshold from the values file:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.11.1/jupyterhub/values.yaml#L202

This is from the pod:

>Readiness:  http-get http://:http/_chp_healthz delay=0s timeout=1s period=2s #success=1 #failure=3

It looks like this is because `failureThreshold` isn't used in the `proxy` deployment template:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.11.1/jupyterhub/templates/proxy/deployment.yaml#L127-L140

#### Expected behaviour

I expected the failure threshold from the values file to be set on the readiness probe in the proxy (if the readiness probe is enabled, which it is by default).

#### Actual behaviour

The failure threshold is the default 3 from k8s.

### How to reproduce

Deploy z2jh 0.11.1 with the chp proxy (default); looks like this is also an issue on the latest `main` branch commit.

### Your personal set up

Shouldn't really matter for this issue, but we're running on IBM Cloud Container Service.

----

Having said this, like the hub `readinessProbe`, since the proxy is a singleton having the readiness probe enabled is probably unnecessary and I'm guessing that's why the default `failureThreshold` is so high - because taking the proxy out of service doesn't help since it's the only one. So our workaround will likely just be to disable the chp proxy readiness probe (we also disable the hub readiness probe for the same reason).

consideRatio:
We had values and schema documenting these, but we didn't have an actual implementation for the proxy pod active.

Closes #2415.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -135,6 +135,8 @@ spec:
           livenessProbe:
             initialDelaySeconds: {{ .Values.proxy.chp.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.proxy.chp.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.proxy.chp.livenessProbe.timeoutSeconds }}
+            failureThreshold: {{ .Values.proxy.chp.livenessProbe.failureThreshold }}
             httpGet:
               path: /_chp_healthz
               {{- if or $manualHTTPS $manualHTTPSwithsecret }}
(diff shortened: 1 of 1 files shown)


consideRatio:
Going for a merge as it has been a while and we are low on review capacity. This is a quite basic thing as well.

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -135,6 +135,8 @@ spec:
           livenessProbe:
             initialDelaySeconds: {{ .Values.proxy.chp.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.proxy.chp.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.proxy.chp.livenessProbe.timeoutSeconds }}
+            failureThreshold: {{ .Values.proxy.chp.livenessProbe.failureThreshold }}
```
mriedem, line 139: This value doesn't exist in the values file, at least not with a default like the readinessProbe:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.11.1/jupyterhub/values.yaml#L194

(same on `main`)

Should that be added to the values file with a default like the readiness probe or at least be defined in the schema?

PR #2423, opened 2021-10-04
Add missing default values for proxy pod's probes

consideRatio:


mriedem:
> I opt to set the helm chart defaults to be a bit more tolerant though, hoping for a bit longer that the pod becomes responsive like we do for the hub pod - now also for the proxy pod.

The logic makes sense to me since the default proxy (chp) is also a singleton (like the hub) so we want to be a bit more conservative with the liveness probe so it's not killing the proxy all the time, especially under load - like a lot of users signing on at once during the beginning of some event.

I looked over our proxy liveness probe failure metrics for the last month and there were 3 instances where we would have tripped the liveness probe:

image

Those were also during a high load / user event and there were users complaining at the time of issues with accessing their notebooks (they were getting 5xx errors). As I looked more into that issue it seemed the chp proxy was the problem - response rates steadily increased and memory increased:

image

image

We're currently using chp version 4.4.0.

In our case the liveness probe restarting the proxy _might_ have helped but it's hard to tell since we're not in the middle of that issue anymore.

With this change the proxy liveness probe will only restart the container after 90 seconds of consecutive failures. Since we hit a max of 6 failures in the graph above I'm not sure if we would have hit the failure threshold of 30 set here, but I think that's a decent default since like I said we don't want to have the opposite problem of restarting the proxy too often since it's a singleton. So +1 from me.

mriedem:
> Those were also during a high load / user event and there were users complaining at the time of issues with accessing their notebooks (they were getting 5xx errors).

For more context users (and myself when I tried to recreate) were getting [522 errors from Cloudflare](https://support.cloudflare.com/hc/en-us/articles/115003011431-Troubleshooting-Cloudflare-5XX-errors#522error) and that was likely a result of the slow responses from the proxy in our case.

consideRatio:
@mriedem wow thank you for such a thought out and researched response!!!

What would make cloudflare give a 522 error at all, is it involved to proxy all incomming traffic? I guess it could be doing that as a protection from DDoS attacks or similar.

consideRatio:
@mriedem I'd love to learn about the object storage setup. Would you be willing to write something brief on discourse.jupyter.org about that or help me point to some references about backing up filesystem data on PVCs via object storage?

For you information, regarding the 503 errors, that relates very much to work by @yuvipanda and @minrk here:
- https://github.com/jupyterhub/jupyterhub/pull/3636
- https://github.com/jupyterhub/mybinder.org-deploy/pull/2043
  It is my understanding that mybinder.org has a proxy that filters out some traffic, 503 requests specifically. That is what the proxy-patches service is about I think.

Issue #2150, opened 2021-04-15
Make the autohttps pod automatically re-attempt TLS cert acquisition on failure during startup somehow


consideRatio:
Our autohttps pod that does TLS termination (encrypt/decrypt outbound/inbound HTTPS traffic) is using running the `traefik` docker image with the Traefik software configured to our needs. Treafik in turn relies on LEGO or the Let's Encrypt Go-lang ACME client to request/renew a TLS certificate.

The enhancement that is requested with this issue is to make sure that there is no need for a user to manually restart the autohttps pod in order to acquire a TLS certificate, which i think is needed if something goes wrong during when it attempts to do so during startup.

It can reduce the complexity to get this chart up and running with TLS if we help this become robust and automated, and it can help us avoid issues in CI tests that currently have steps to really ensure our ACME server is up and running before installing our helm chart so that it can succeed on first and only attempt.

### Implementation ideas
I think this can be a good situation to use a `startupProbe` or `livenessProbe`, but the challenge is that we need to have a clear signal to indicate if we succeeded or not to acquire the TLS certs, so we would probably need to use an `exec` probe (executes a command in the container) rather than a `http` probe (sends a web request) to do so.

It would be great if we could get a direct feedback from traefik on its status in the TLS cert acquisition rather than trying probing a file in the file system or similar though. I'm not sure what can be done.

### Action points to resolve this
- Propose a implementation strategy on how to resolve this
- Describe why you think it makes sense and what your deliberations are regarding the strategy
- Develop and open a PR

### Contribute with research!
- Verify if traefik still doesn't support re-attempting failed ACME challenges on startup.
- Explore if a command as executed form a terminal in the container traefik is running can detect if the initial ACME challenge has failed or not - it can be used as a k8s startupProbe to help us automatically restart the container on failure to re-attempt the ACME challenge.

pvanliefland:
I might be wrong here, but while debugging the issue (for a new deployment - from scratch- it seemed that the autohttps pod was trying to acquire the certificate "too soon" (the load balancer was not ready at the time?)

consideRatio:
> I might be wrong here, but while debugging the issue (for a new deployment - from scratch- it seemed that the autohttps pod was trying to acquire the certificate "too soon" (the load balancer was not ready at the time?)

[...]

PR #2431, opened 2021-10-08
Adds a livenessProbe to traefik for LE TLS

joraff:
Adds a livenessProbe for traefik that checks for a successful let's encrypt certificate issuance. No change in default behaviors with this PR, all new values are optional and disabled by default.

Adds documentation for the new values to the config reference.

Attempts to address #2150

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1727,14 +1727,14 @@ properties:
           extraPodSpec: *extraPodSpec-spec
           livenessProbe:
             type: object
-            additionalProperties: true
+            additionalProperties: false
             properties:
               enabled:
-                type: [boolean, "false"]
+                type: boolean
               initialDelaySeconds:
-                type: [integer, 30]
+                type: integer
               periodSeconds:
-                type: [integer, 15]
+                type: integer
             description: |
               This `livenessProbe` is provided to check for a successful certificate issuance from Let's Encrypt.
               You can enable this if you are having issues with race conditions in your infrastructure or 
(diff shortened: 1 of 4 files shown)


joraff:
I'll dive into the CI checks a bit later. At first glance they appear to be unrelated to this PR's changes.

consideRatio:
@joraff I've thought in length about this situation.

1. Traefik didn't want to implement a configurable delay to make their ACME request.
2. A livenessProbe can force the pod to restart, and this resolves your situation with a LoadBalancer not being setup fast enough, but doesn't solve people who have issues with the k8s networking being setup fast enough. This livenessProbe can also make you get blocked by Let's Encrypt as it will make the pod make repeated attempts that can fail. Due to this, it is quite sensitive and requires significant documentation warning about it etc.
3. By introducing an init-container instead that sleeps ~10 seconds, you have the downside of starting up a pod slower, but the upside of not risking being blocked and being able to resolve this for your use case and for the situations containers networking is slow to start as well.

Due to this, I'd like to avoid the approach of declaring a livenessProbe to force a restart, even if its opt-in.

Due to this, I'll go for a close of this PR for now. I'll work to ensure that we can configure an initContainer to do the work needed unless we already can't do that. I'll write about an example to handle this situation in #2150 when both something makes us need a delay for when the traefik pod attempts to request a certificate.

consideRatio:
For reference, note that `proxy.traefik.extraInitContainers` is available to allows you to sleep during startup until the ACME certificate should be ready to requested.

This issue also relates to #2601.

PR #2436, opened 2021-10-13
Minor fixes to the Microk8s documentation

mike-matera:
This patch fixes an indentation bug in a YAML sample from my last PR and a minor grammar bug. I noticed both on a re-read. The YAML error would be a problem for someone who copy-pasted my sample configuration. Thanks!


consideRatio:
Thank you!!! ❤️

PR #2446, opened 2021-10-19
idle-culler: fix the new restricted scopes to include read:servers

consideRatio:
Thank you @snickell for catching this and reporting it very clearly from https://github.com/jupyterhub/jupyterhub-idle-culler/issues/40!!! :heart: :tada:


snickell:
@consideRatio absolutely, I'm sitting riveted as my CI build progresses 😹 just a littttttttle longer until bed...... ;-)

PR #2449, opened 2021-10-21
Unset `singleuser.cmd`, previously `jupyterhub-singleuser`, to instead rely on the image's CMD by default

### LINKED ISSUE #776: JupyterLab - the default?
What we decide on to be the default matters a great deal. I think we should make it so that the default is to use jupyterlab.

Yay or nay?

minrk:
As discussed in #2386 

Leaving singleuser.cmd unset means relying on the image for the default command to run. This matches the default in KubeSpawner itself, which sets the default cmd to None, which is interpreted as using the CMD form the image definition. This simplifies things as there are now only two places for the command-to-run to come from:

1. the image itself (chosen by the user), or
2. `singleuser.cmd` in *the user's* `values.yaml`

In 1.x, the chart's own default was `jupyterhub-singleuser`

- Pro: `jupyterhub-singleuser` means it will always launch something that works without additional user config
- Con: If the image *does* have a custom CMD, it must be duplicated into `singleuser.cmd` because the image has been overridden by the chart defaults

The main con of the change is that one line of config is required for images that don't have a CMD that will launch `jupyterhub-singleuser`, which seems fine.

closes #776 which should already be closed by the upgrade to jupyterhub 2.0, which launches lab by default (if available) itself in `jupyterhub-singleuser`.

This is a reissue of #2138 which was withdrawn from 1.x because of the complications around passing cli args by default. The "full command" to launch is`singleuser.cmd` + `Spawner.get_args()`. This isn't an issue for DockerSpawner, which can inspect the image and extract its CMD for the default value if unspecified, KubeSpawner cannot assume this capability. So what it does is [ignore cli args][] if the image's CMD is used, meaning `singleuser.cmd` *must* be specified if KubeSpawner is to pass any CLI args.  This was a deal breaker, because JupyterHub 1.x sets common options like ip, port, and defaultUrl via CLI args. JupyterHub 2.0 promises to never communicate with itself via cli args to `jupyterhub-singleuser`, [only environment variables][]. This makes it much safer to ignore cli args in common cases, as CLI args will always come from user config, which can in turn specify `singleuser.cmd` to match.

[ignore cli args]: https://github.com/jupyterhub/kubespawner/pull/533
[only environment variables]: https://github.com/jupyterhub/jupyterhub/pull/3381

### DIFF
--- a/doc/source/jupyterhub/customizing/user-environment.md
+++ b/doc/source/jupyterhub/customizing/user-environment.md
@@ -242,6 +242,12 @@ FROM jupyter/minimal-notebook:177037d09156
 
 # install additional package...
 RUN pip install --no-cache-dir astropy
+
+# set the default command of the image,
+# if the parent image will not launch a jupyterhub singleuser server.
+# The JupyterHub "Docker stacks" do not need to be overridden.
+# Set either here or in `singleuser.cmd` in your values.yaml
+# CMD ["jupyterhub-singleuser"]
 ```
 
 ```{note}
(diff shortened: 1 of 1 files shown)


consideRatio:
@minrk does this mean that https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/images/singleuser-sample/Dockerfile should be updated to set CMD to jupyterhub-singleuser then?

consideRatio:
If someone wants to help with changelog writing, writing up a changelog for this is especially complicated in my mind. This relates to that far more people will now experience the bug that `default_url` etc are now going to be silently ignored. See https://github.com/jupyterhub/kubespawner/issues/493.

PR #2474, opened 2021-11-04
docs: fix failure to show correct version

consideRatio:


doc/source/conf.py
```diff
@@ -35,21 +35,6 @@ def setup(app):
 # -- Referencable variables --------------------------------------------------
 
 
-def _get_latest_tag():
```
consideRatio, line 38: This was used to set a variable, which in turn was never used. Let's instead rely on chartpress logic to extract something in a consistent manner.

PR #2489, opened 2021-11-16
Fix docs typo

mriedem:
Following [1] and other instances in the same doc
this fixes the typo on the env var to set to use
jupyter-server in the singleuser server pod.

Also make the retro example set the env var value
in a quotes since it's a string.

Also fix the docs linkcheck job failure while in here.

[1] https://jupyterhub.readthedocs.io/en/stable/reference/config-user-env.html#switching-to-jupyter-server


mriedem:
Found this the hard way by copying the example from the docs and then wondering why my test singleuser pod was still using the legacy notebook app. 😄

consideRatio:
:heart: thank you @mriedem for upstreaming these insights! :D

PR #2508, opened 2021-12-09
breaking: add ...networkPolicy.egressAllowRules and don't allow singleuser pods to access PrivateIPv4 addresses by default

yuvipanda:
## Updated technical change summary by Erik

`<hub|proxy.chp|proxy.traefik|singleuser>.networkPolicy.egressAllowRules` is introduced. They are meant to help adjust the created NetworkPolicy resource rules with a few convenient true/false values. They are all set to true by default besides for the singleuser NetworkPolicy governing the user pods. Using the default configuration, user pods are no longer granted access to [PrivateIPv4 addresses](https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses).

Also note that with the introduction of the `egressAllowRules`, the configuration to allow outbound network connections moved from `.egress` to `.egressAllowRules`. This means that if users have configured `.egress` previously to not be as permissive, they may experience a permission escalation by having `.egressAllowRules` introduced. So, for anyone configuring `.egress` and upgrading to version 2 of this chart can set all .egressAllowRules to false.

## Original PR description

This PR enables all traffic to the *internet* from user pods,
but stops allowing access to the private network where other
kubernetes pods or services may be running. Meant to be a
minimally intrusive defense in depth mechanism that prevents
access to unsecured endpoints running on VMs, in the kubernetes
cluster, etc. For example, users often run Prometheus without
any kind of network protection in the same cluster as the
users, and access to prometheus can leak a lot of sensitive
user data.

https://www.wiz.io/blog/chaosdb-explained-azures-cosmos-db-vulnerability-walkthrough
shows the immense security value of restricting access to internal
network from user written code. Most software projects do not have
their authz/authn layers, and rely on network segmentation for
protection. When network segmentation fails like it does in the Azure security incident,
there is often no secondary line of defense.

Most JupyterHub users talk to the internet, but by default I think
nobody really talks to the *internal network*. This also only affects
network connections from user code - stuff like mounting NFS shares
are done by the kubelet, and hence are not affected. Connections
to hub infrastructure (proxy, hub, etc) are also not affected,
as they are separately allowed by other NetworkPolicy rules.
Access to `http://proxy-public` for user code to talk in a hub agnostic
way to services, as well as access to the full hub URL when autohttps
is used are also added in this PR.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -664,15 +664,13 @@ properties:
 
                   Since not all workloads in the k8s cluster may have
                   NetworkPolicies setup to restrict their incoming connections,
-                  having this set to false can be a good defence against
+                  having this set to false can be a good defense against
                   malicious intent from someone in control of software in these
                   pods.
 
-                  To improve security we recommend only allowing the private IP CIDRs required
-                  <TODO: insert example>
-                  or namespaces
-                  <TODO: insert example>
-                  instead of enabling this.
+                  If possible, try to avoid setting this to true as it gives
+                  broad permissions that could be specified more directly via
+                  the [`.egress`](schema_singleuser.networkPolicy.egress).
           interNamespaceAccessLabels:
             enum: [accept, ignore]
             description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -575,19 +575,48 @@ properties:
           enabled:
             type: boolean
             description: |
-              Toggle the creation of the NetworkPolicy resource for this pod.
+              Toggle the creation of the NetworkPolicy resource targeting this
+              pod, and by doing so, restricting its communication to only what
+              is explicitly allowed in the NetworkPolicy.
           ingress:
             type: array
             description: |
-              Additional ingress rules to add except those that is known to be needed by the respective pods in the Helm chart.
+              Additional ingress rules to add besides those that are required
+              for core functionality.
           egress:
             type: array
             description: |
-              Additional egress rules to add except those that is known to be needed by the respective pods in the Helm chart.
+              Additional egress rules to add besides those that are required for
+              core functionality and those added via
+              [`.egressAllowRules`](schema_hub.networkPolicy.egressAllowRules).
 
-              The default value of this egress is to allow all traffic, except for the `singleuser.networkPolicy.egress`, which is also limiting access to a metadata server that can be exploited.
+              ```{versionchanged} 2.0.0
+              The default value changed from providing one very permissive rule
+              allowing all egress to providing no rule. The permissive rule is
+              still provided via
+              [`.egressAllowRules`](schema_hub.networkPolicy.egressAllowRules)
+              set to true though.
+              ```
 
-              If you want to restrict egress, you can override this permissive default to be an empty list.
+              As an example, below is a configuration that disables the more
+              broadly permissive `.privateIPs` egress allow rule for the hub
+              pod, and instead provides tightly scoped permissions to access a
+              specific k8s local service as identified by pod labels.
+
+              ```yaml
+              hub:
+                networkPolicy:
+                  egressAllowRules:
+                    privateIPs: false
+                  egress:
+                    - to:
+                        - podSelector:
+                            matchLabels:
+                              app: my-k8s-local-service
+                      ports:
+                        - protocol: TCP
+                          port: 5978
+              ```
           egressAllowRules:
             type: object
             additionalProperties: false
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -607,6 +607,10 @@ properties:
               with the other flags to a reduced set of rules to avoid a
               performance penalty.
               ```
+
+              ```{versionadded} 2.0.0
+              All `egressAllowRules` are new in JupyterHub Helm chart 2.0.0.
+              ```
             properties:
               cloudMetadataServer:
                 type: boolean
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -667,6 +667,12 @@ properties:
                   having this set to false can be a good defence against
                   malicious intent from someone in control of software in these
                   pods.
+
+                  To improve security we recommend only allowing the private IP CIDRs required
+                  <TODO: insert example>
+                  or namespaces
+                  <TODO: insert example>
+                  instead of enabling this.
           interNamespaceAccessLabels:
             enum: [accept, ignore]
             description: |
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -583,11 +583,86 @@ properties:
           egress:
             type: array
             description: |
-              Additional egress rules to add except those that is known to be needed by  the respective pods in the Helm chart.
+              Additional egress rules to add except those that is known to be needed by the respective pods in the Helm chart.
 
               The default value of this egress is to allow all traffic, except for the `singleuser.networkPolicy.egress`, which is also limiting access to a metadata server that can be exploited.
 
               If you want to restrict egress, you can override this permissive default to be an empty list.
+          egressAllowRules:
+            type: object
+            additionalProperties: false
+            description: |
+              This is a set of predefined rules that when enabled will be added
+              to the NetworkPolicy list of egress rules.
+
+              The resulting egress rules will be a composition of:
+              - rules specific for the respective pod(s) function within the
+                Helm chart
+              - rules based on enabled `egressAllowRules` flags
+              - rules explicitly specified by the user
+
+              ```{note}
+              Each flag under this configuration will not render into a
+              dedicated rule in the NetworkPolicy resource, but instead combine
+              with the other flags to a reduced set of rules to avoid a
+              performance penalty.
+              ```
+            properties:
+              cloudMetadataServer:
+                type: boolean
+                description: |
+                  Defaults to `false` for singleuser servers, but to `true` for
+                  all other network policies.
+
+                  When enabled this rule allows the respective pod(s) to
+                  establish outbound connections to the cloud metadata server.
+
+                  Note that the `nonPrivateIPs` rule is allowing all non Private
+                  IP ranges but makes an exception for the cloud metadata
+                  server, leaving this as the definitive configuration to allow
+                  access to the cloud metadata server.
+              dnsPortsPrivateIPs:
+                type: boolean
+                description: |
+                  Defaults to `true` for all network policies.
+
+                  When enabled this rule allows the respective pod(s) to
+                  establish outbound connections to private IPs via port 53.
+
+                  Note that we can't reliably identify the k8s internal DNS
+                  server due to variations between k8s clusters. Due to that,
+                  this rule which is critical for core functionality, can be
+                  disabled for a more refined custom rule.
+              nonPrivateIPs:
+                type: boolean
+                description: |
+                  Defaults to `true` for all network policies.
+
+                  When enabled this rule allows the respective pod(s) to
+                  establish outbound connections to the non-private IP ranges
+                  with the exception of the cloud metadata server. This means
+                  respective pod(s) can establish connections to the internet
+                  but not (say) an unsecured prometheus server running in the
+                  same cluster.
+              privateIPs:
+                type: boolean
+                description: |
+                  Defaults to `false` for singleuser servers, but to `true` for
+                  all other network policies.
+
+                  Private IPs refer to the IP ranges `10.0.0.0/8`,
+                  `172.16.0.0/12`, `192.168.0.0/16`.
+
+                  When enabled this rule allows the respective pod(s) to
+                  establish outbound connections to the internal k8s cluster.
+                  This means users can access the internet but not (say) an
+                  unsecured prometheus server running in the same cluster.
+
+                  Since not all workloads in the k8s cluster may have
+                  NetworkPolicies setup to restrict their incoming connections,
+                  having this set to false can be a good defence against
+                  malicious intent from someone in control of software in these
+                  pods.
           interNamespaceAccessLabels:
             enum: [accept, ignore]
             description: |
--- a/jupyterhub/templates/_helpers-netpol.tpl
+++ b/jupyterhub/templates/_helpers-netpol.tpl
@@ -1,0 +1,73 @@
+{{- define "jupyterhub.networkPolicy.standardEgressConfiguration" -}}
+{{- include "jupyterhub.networkPolicy.standardEgressConfiguration.withLeadingNewLine" . | trimPrefix "\n" }}
+{{- end }}
+
+{{- define "jupyterhub.networkPolicy.standardEgressConfiguration.withLeadingNewLine" -}}
+{{- $root := index . 0 }}
+{{- $netpol := index . 1 }}
+{{- if and ($netpol.egressAllowRules.dnsPortsPrivateIPs) (not $netpol.egressAllowRules.privateIPs) }}
+# Allow outbound connections to the DNS port in the private IP ranges
+- ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+  to:
+    - ipBlock:
+        cidr: 10.0.0.0/8
+    - ipBlock:
+        cidr: 172.16.0.0/12
+    - ipBlock:
+        cidr: 192.168.0.0/16
+{{- end }}
+
+{{- if $netpol.egressAllowRules.nonPrivateIPs }}
+# Allow outbound connections to non-private IP ranges
+{{- if $netpol.egressAllowRules.privateIPs }}
+# Allow outbound connections to private IP ranges (and their DNS ports)
+{{- end }}
+{{- if $netpol.egressAllowRules.cloudMetadataServer }}
+# Allow outbound connections to cloud metadata server
+{{- end }}
+- to:
+    - ipBlock:
+        cidr: 0.0.0.0/0
+        {{- if or (not $netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.cloudMetadataServer) }}
+        except:
+          {{- if not $netpol.egressAllowRules.privateIPs }}
+          # Don't allow outbound connections to private IP ranges
+          - 10.0.0.0/8
+          - 172.16.0.0/12
+          - 192.168.0.0/16
+          {{- end }}
+
+          {{- if not $netpol.egressAllowRules.cloudMetadataServer }}
+          # Don't allow access to the cloud metadata server
+          - {{ $root.Values.singleuser.cloudMetadata.ip }}/32
+          {{- end }}
+        {{- end }}
+{{- end }}
+
+{{- if and ($netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.nonPrivateIPs) }}
+# Allow outbound connections to private IP ranges (and their DNS ports)
+- to:
+    - ipBlock:
+        cidr: 10.0.0.0/8
+    - ipBlock:
+        cidr: 172.16.0.0/12
+    - ipBlock:
+        cidr: 192.168.0.0/16
+{{- end }}
+
+{{- if and ($netpol.egressAllowRules.cloudMetadataServer) (not $netpol.egressAllowRules.nonPrivateIPs) }}
+# Allow outbound connections to cloud metadata server
+- to:
+    - ipBlock:
+        cidr: {{ $root.Values.singleuser.cloudMetadata.ip }}/32
+{{- end }}
+
+{{- with $netpol.egress }}
+# Allow outbound connections based on user specified rules
+{{- . | toYaml }}
+{{- end }}
+{{- end }}
--- a/jupyterhub/templates/_helpers-netpol.tpl
+++ b/jupyterhub/templates/_helpers-netpol.tpl
@@ -3,10 +3,9 @@
   common configuration.
 
   It is rendering based on the `egressAllowRules` and `egress` keys of the
-  passed networkPolicy config object. In practice each flag under
-  egressAllowRules could be a dedicated egress rule, but for potential
-  performance impact we have increased the complexity of this rendering logic
-  and render them in a coalesced form.
+  passed networkPolicy config object. Each flag set to true under
+  `egressAllowRules` is rendered to a egress rule that next to any custom user
+  defined rules from the `egress` config.
 
   This named template needs to render based on a specific networkPolicy
   resource, but also needs access to the root context. Due to that, it
@@ -28,7 +27,7 @@
 {{- define "jupyterhub.networkPolicy.renderEgressRules" -}}
 {{- $root := index . 0 }}
 {{- $netpol := index . 1 }}
-{{- if and ($netpol.egressAllowRules.dnsPortsPrivateIPs) (not $netpol.egressAllowRules.privateIPs) }}
+{{- if $netpol.egressAllowRules.dnsPortsPrivateIPs }}
 # Allow outbound connections to the DNS port in the private IP ranges
 - ports:
     - protocol: UDP
@@ -46,33 +45,21 @@
 
 {{- if $netpol.egressAllowRules.nonPrivateIPs }}
 # Allow outbound connections to non-private IP ranges
-{{- if $netpol.egressAllowRules.privateIPs }}
-# Allow outbound connections to private IP ranges (and their DNS ports)
-{{- end }}
-{{- if $netpol.egressAllowRules.cloudMetadataServer }}
-# Allow outbound connections to cloud metadata server
-{{- end }}
 - to:
     - ipBlock:
         cidr: 0.0.0.0/0
-        {{- if or (not $netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.cloudMetadataServer) }}
         except:
-          {{- if not $netpol.egressAllowRules.privateIPs }}
-          # Don't allow outbound connections to private IP ranges
+          # As part of this rule, don't:
+          # - allow outbound connections to private IP
           - 10.0.0.0/8
           - 172.16.0.0/12
           - 192.168.0.0/16
-          {{- end }}
-
-          {{- if not $netpol.egressAllowRules.cloudMetadataServer }}
-          # Don't allow access to the cloud metadata server
+          # - allow outbound connections to the cloud metadata server
           - {{ $root.Values.singleuser.cloudMetadata.ip }}/32
-          {{- end }}
-        {{- end }}
 {{- end }}
 
-{{- if and ($netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.nonPrivateIPs) }}
-# Allow outbound connections to private IP ranges (and their DNS ports)
+{{- if $netpol.egressAllowRules.privateIPs }}
+# Allow outbound connections to private IP ranges
 - to:
     - ipBlock:
         cidr: 10.0.0.0/8
@@ -82,8 +69,8 @@
         cidr: 192.168.0.0/16
 {{- end }}
 
-{{- if and ($netpol.egressAllowRules.cloudMetadataServer) (not $netpol.egressAllowRules.nonPrivateIPs) }}
-# Allow outbound connections to cloud metadata server
+{{- if $netpol.egressAllowRules.cloudMetadataServer }}
+# Allow outbound connections to the cloud metadata server
 - to:
     - ipBlock:
         cidr: {{ $root.Values.singleuser.cloudMetadata.ip }}/32
--- a/jupyterhub/templates/_helpers-netpol.tpl
+++ b/jupyterhub/templates/_helpers-netpol.tpl
@@ -29,7 +29,6 @@
 {{- $root := index . 0 }}
 {{- $netpol := index . 1 }}
 {{- if and ($netpol.egressAllowRules.dnsPortsPrivateIPs) (not $netpol.egressAllowRules.privateIPs) }}
-
 # Allow outbound connections to the DNS port in the private IP ranges
 - ports:
     - protocol: UDP
@@ -46,7 +45,6 @@
 {{- end }}
 
 {{- if $netpol.egressAllowRules.nonPrivateIPs }}
-
 # Allow outbound connections to non-private IP ranges
 {{- if $netpol.egressAllowRules.privateIPs }}
 # Allow outbound connections to private IP ranges (and their DNS ports)
@@ -74,7 +72,6 @@
 {{- end }}
 
 {{- if and ($netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.nonPrivateIPs) }}
-
 # Allow outbound connections to private IP ranges (and their DNS ports)
 - to:
     - ipBlock:
@@ -86,7 +83,6 @@
 {{- end }}
 
 {{- if and ($netpol.egressAllowRules.cloudMetadataServer) (not $netpol.egressAllowRules.nonPrivateIPs) }}
-
 # Allow outbound connections to cloud metadata server
 - to:
     - ipBlock:
@@ -94,8 +90,7 @@
 {{- end }}
 
 {{- with $netpol.egress }}
-
 # Allow outbound connections based on user specified rules
-{{- . | toYaml }}
+{{ . | toYaml }}
 {{- end }}
 {{- end }}
--- a/jupyterhub/templates/_helpers-netpol.tpl
+++ b/jupyterhub/templates/_helpers-netpol.tpl
@@ -1,11 +1,35 @@
-{{- define "jupyterhub.networkPolicy.standardEgressConfiguration" -}}
-{{- include "jupyterhub.networkPolicy.standardEgressConfiguration.withLeadingNewLine" . | trimPrefix "\n" }}
-{{- end }}
+{{- /*
+  This named template renders egress rules for NetworkPolicy resources based on
+  common configuration.
 
-{{- define "jupyterhub.networkPolicy.standardEgressConfiguration.withLeadingNewLine" -}}
+  It is rendering based on the `egressAllowRules` and `egress` keys of the
+  passed networkPolicy config object. In practice each flag under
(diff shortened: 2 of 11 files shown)


yuvipanda:
Thanks a lot to @consideRatio for bringing this up and engaging with this intently in https://github.com/2i2c-org/infrastructure/issues/829, and to @sgibson91 for actually coming up with the network policies.

We've been running this at 2i2c and Berkeley for a while. Inspired by the REAL BAD azure security incident https://www.wiz.io/blog/chaosdb-explained-azures-cosmos-db-vulnerability-walkthrough. https://github.com/jupyterhub/kubespawner/pull/545 is the other more secure default that came out of that.

yuvipanda:
Discuss adding another rule that removes any network restriction to
anything else in the current namespace. For example, even now, you might
have to add an extra rule allowing network egress from the user pod to the
traefik component of the dask-gateway chart for dask-gateway to work (need
to verify that) - just going through the hub service is not enough. Instead of having
to figure that out on a service by service basis, we can simply enable outgoing network
access to everything else in the same namespace. This removes protection for services
without incoming networkpolicy restrictions or other means of authz in the same namespace.
A trade-off to be made between admin simplicify (NetworkPolicys are hard to debug!) vs
secure defaults. @consideRatio strongly prefers opening up this hole, I'm mildly opposed to it but
even with hole this would be a massive improvement over our current status quo. So I'm happy to
ship with that hole open by default and allow users to close it. Would love to hear your thoughts, @minrk @manics (and others?)

consideRatio:
Thanks @yuvipanda for bringing this onwards!

### Agreement clarification

Thanks to practical experience reported by @yuvipanda and the 2i2c.org tech team of using tighter egress rules, combined with some practical ideas on how to implement it - I'm now also in favor of dropping permissions to a bare minimum.

>For your information, the option I deliberated on in the past were these.
>
>1. All egress is allowed by default
>2. k8s namespace local and non-local IP egress is allowed by default
>3. Only non-local IP egress is allowed by default

### Change requests

- [ ] __Language about restrict/permit__
  As soon as there is an egress or ingress NetworkPolicy targetting a pod, its allowed egress or ingress communication drops to zero - this is the only restrictive part of a NetworkPolicy. Then, the NetworkPolicies simply add more and more permissions with various rules. I'd like our language to reflect that we grant permissions for egress or ingress specifically, rather than say that we restrict it. The act of having a network policy at all is the restriction, and then it is all about adding permissions.
- [ ] __Configuration naming and planning__
  - [ ] I suggest we create a section like `egressAllowRules` that contain for example `nonLocalIPs` and `dnsPorts` and such that toggles the predefined rules we have in addition to the custom egress rules the user can specify.
  - [ ] Decide if we should allow further configuration of these predefined allow rules, or if we go with a set of boolean on/off switches only. I suggest we go for simple boolean rules that can't be configured further.
- [ ] __Technical details__
  - [ ] We should match on the labels `component`, `app`, and `release`. Currently release is missing. We should also adopt the pattern to not hardcode these labels as done in other locations in the netpol file.
  ```yaml
  # Example on how to render the component, app, and release labels
  # for the autohttps pod.
  {{- $_ := merge (dict "componentLabel" "autohttps") . }}
  {{- include "jupyterhub.matchLabels" $_ | nindent 14 }}
  ```
  - [ ] We should ensure that the allow rule for non-local IPs won't go ahead and allow access to the metadata server.
- [ ] __Schema updates__
  - [ ] The configuration schema should reflect the new configuration options.
  - [ ] We should have a tools/templates/lint-and-validate-config.yaml entry for all added config
  - [ ] We should have a jupyterhub/values.yaml entry for all added config

consideRatio:
### Configuration proposal 1

```diff
 singleuser:
   networkPolicy:
     enabled: true
     ingress: []
+    egressAllowRules:
+      nonLocalIPs: true
+    egress: []
-    egress:
-      - to:
-          - ipBlock:
-              cidr: 0.0.0.0/0
-              except:
-                - 169.254.169.254/32
     interNamespaceAccessLabels: ignore
     allowedIngressPorts: []
```

I also suggest we make the dnsPorts configuration configurable alongside the `nonLocalIPs`, mainly because it improves comprehensibility of the setup rather than it seems to be important for end users to configure.

```diff
 singleuser:
   networkPolicy:
     enabled: true
     ingress: []
     egressAllowRules:
+      dnsPorts: true
       nonLocalIPs: true
     egress: []
     interNamespaceAccessLabels: ignore
     allowedIngressPorts: []
```

minrk:
👍 to the change. The only thing I think we need to highlight in docs is that this feature must be _disabled_ in order to block egress more strictly, because adding a more restrictive egress policy with this in place will have no effect, since this already grants access to the outside world.

We might even want to fail if networkPolicy.egress is defined while this is true, since I can't think of a situation where that would behave as intended.

consideRatio:
>  The only thing I think we need to highlight in docs is that this feature must be disabled in order to block egress more strictly, because adding a more restrictive egress policy with this in place will have no effect, since this already grants access to the outside world.

@minrk this is why I suggest a change of language where we speak of allowing instead of blocking or restricting, and a concrete suggestion on how we name our config options in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2508#issuecomment-989651310

> We might even want to fail if networkPolicy.egress is defined while this is true, since I can't think of a situation where that would behave as intended.

I think this is a problematic as it makes some special configuration transform the behavior of other configuration, and that is a pattern I've seen us revert many times as complexity increase. I suggest we stick to anything that is easy to describe and avoid introducing configuration that influences other configuration behavior. I think https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2508#issuecomment-989651310 accomplishes that as well.

minrk:
@consideRatio I think that proposal's great! Then it's a sort of 'alias' for assembling one egress rule,r athe rthan two that won't make sense together. That seems like a good way to think of it.

If we do ever have mutually exclusive config options, I do think it's reasonable to try to fail informatively rather than have undefined behavior, as long as the failure is manageable. But of course, it's ideal if we can achieve the result _without_ creating mutually exclusive config.

manics:
> Most JupyterHub users talk to the internet, but by default I think nobody really talks to the internal network.

It's probably worth clarifying what you mean by "internal"- do you mean internal to the K8s cluster, or internal to an organisation's network?

Blocking all non-public traffic makes sense for public JupyterHubs, but I'd be surprised if it was the case for internal hubs inside an organisation's network- one of the reasons for running an internal hub is to safely give access to other (non-k8s) networked resources in an organisation.

I think defaulting to blocking non-public traffic from singleuser pods is fine as it improves security by default, though I _think_ we've discussed this before so might be worth going back through past discussions to check if we've missed any points.

Ideally there'd be an option to block within-cluster traffic only but I don't think that's possible- instead we could add an example config showing how to correctly block internal k8s traffic whilst still allowing traffic to private IPs outside the cluster CIDR.

I prefer `*PrivateIP*` over `*LocalIP*`. The latter implies it refers to the Kubernetes cluster whereas it's everything on a non-public IP which could be your entire organisation!

consideRatio:
@manics I just want to warn about using or even thinking about "blocking". We may at most "not allow" with a network policy.

Regarding local etc, my definition is what wiki call private ipv4 addresses: https://en.m.wikipedia.org/wiki/Private_network

The rule we consider, in my mind, is to _allow_ all non-provate ipv4 adresses except the cloud metadata server ip.

yuvipanda:
Thank you for the extremely detailed feedback, @consideRatio! I really appreciate that. I'm very much in favor of the example you had in Configuration proposal 1, and would like to get that implemented.

yuvipanda:
Nomenclature-wise, I also prefer PrivateIPs, as the IPs we are blocking are from the RFC that defines private IP space, and calls them private IPs: https://datatracker.ietf.org/doc/html/rfc1918. Things you consider 'local' can also have *public IPs* - for example, most kubernetes masters are accessed via public IPs!

consideRatio:
@yuvipanda wiee okay! Btw 100% on replacing references to "local IPs" with references to "private IPs" as that is what seems to be their common name.

Looking at this now, I wonder if we may want to include something like `egressExtraAllowRules` rather than just `egressAllowRules` for example?

manics:
I think it'd be good to get `cloudMetadataServer` in as well, the current configuration for cloud metadata seems to confuse people.

For docs does it sounds reasonable to have 4 examples based on the use-cases discussed in this issue?
1. Allow everything except private IPs
2. Allow everything except internal Kubernetes IPs (requires a placeholder since the IP range depends on the cluster)
3. Allow everything except private IPs, but allow some internal services in the same or different namespaces
4. Block everything except for an allowlist of IPs/ports

yuvipanda:
Is there anyone else who can also help push this along? I don't want to be a blocker...

consideRatio:
I can take this on in due time @yuvipanda, if anyone wants to go for it, please do! I'll write something here if i start doing actual work later.

consideRatio:
I can take this on in due time @yuvipanda, if anyone wants to go for it, please do! I'll write something here if i start doing actual work later.

consideRatio:
@yuvipanda I'm picking up on this now!

jupyterhub/values.yaml
```diff
@@ -331,18 +331,11 @@ singleuser:
...
-      # someone with malicious intent.
-      - to:
-          - ipBlock:
-              cidr: 0.0.0.0/0
-              except:
-                - 169.254.169.254/32
+    egressAllowRules:
+      cloudMetadataServer: false
+      dnsPorts: true
+      publicIPs: true
```
yuvipanda, line 337: Maybe `onlyPublicIPs` or similar, to indicate that we are in fact blocking private IPs here?

consideRatio: Oh nooo, I really really dislike the idea of communicating that we are blocking private IPs, we are not blocking them, we are allowing public IPs!

It is such a common confusion with network policies, when enabled to target a pod with regards to egress, they are isolated and nothing is allowed. That is the only operation that restricts the pods networking, after that, it is all about allowing. This allows publicIPs, it doesn't say "only" public IPs, it just allows them. If we have another rule like "dnsPorts" we allow communication to dnsPorts as well, even they are not public IPs for example.

Do you see my point? I'm very stubborn about avoiding the use of "block", "disallow", "only", "restrict" etc, because we are only declaring what we allow here, not what we disallow.

yuvipanda: Yeah, communicating around network policies is tricky and I appreciate the thought you are putting into naming things! 

The breaking change in this PR - and hence z2jh 2.x vs 1.x - is that private IPs will not be accessible anymore. Public IP behavior has not changed, while private IP behavior changes. I want the naming to communicate this - hence my suggestion of `onlyPublicIP`.

The change in this PR has two parts:

1. Allow access to 0.0.0.0/0 (which is basically *everything*)
2. But restrict access to three private IP ranges

So from a deployer perspective, the change from before is that now we have restricted access to three private IP ranges, while public IP behavior has not changed.

I'd love to hear a bit more about this statement:

>  I'm very stubborn about avoiding the use of "block", "disallow", "only", "restrict" etc, because in this specific configuration, we are only declaring what we allow here, not what we disallow.

Particularly in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2508/files#diff-70983994b57e310d348c747400dd1ae5ea9f3e4efe63da0155789c3a6bd2a411R132, we do explicitly disallow private IPs. We even use the word 'block' to communicate that in the comment, because that's what it is doing. So understanding a bit more about your mental model there will help me.

Regardless, this is a complex to communicate change because network policies can be a bit complex sometimes, so I appreciate the thoughtfullness here.

consideRatio: Thanks for iterating with me @yuvipanda, in the typical very thoughtful and kind way!

Note that the PRs current state does not reflect the communication I intend it to include, so there is a lot of room for communication to be added in the changelog and in general about how network policy configuration can be understood and made.

I'm thinking a communication about this should exist in multiple places:
- changelog
- config reference:
  - networkpolicy level
    - egressAllowRules level
      - specific egressAllowRules level
- documentation: about network config

> The breaking change in this PR - and hence z2jh 2.x vs 1.x - is that private IPs will not be accessible anymore. Public IP behavior has not changed, while private IP behavior changes. I want the naming to communicate this - hence my suggestion of onlyPublicIP.

We are in full agreement about communicating this properly, the question is how. NetworkPolicies are like we agree very complicated to comprehend and work with, and due to that, I care about this detail a lot as well. I want all rules under the egressAllowRules to clearly convey its about ADDING permissions by steering very clear from anything indicating its about restricting permissions - we are in practice just adding the permissions of the public ip range, even though it is done technically by saying "all ips: except the private ips". To help mitigate your concern, i want to add documentation and clarifications about things as I already intended in various places, but I don't want to risk miscommunicating that this may block private IPs when it is in practice just only allowing the public IPs.

If you had a rule that said `onlyPublicIP` it indicates to me that we _disallow_ non-public IPs, while in practice, we are simply just not explicitly allowing them - there is no disallow operation in a NetworkPolicy and this is what I'm super afraid of miscommunicating about. Example to make a point related to this:

```yaml
    # Rule 1
    - to:
        - ipBlock:
            cidr: 10.0.0.0/8
            # The "except" clause within this rule, is just to
            # restrict the range of what we are allowing in
            # this specific rule, out of many rules that add
            # more and more permissions. No rule is
            # removing or restricting permissions!
            except:
              - 10.128.0.0/10
    # Rule 2: equivalent to rule 1, both allow the ranges
    # 10.0.0.0-10.127.255.255 and 10.192.0.0-10.255.255.255
    - to:
        - ipBlock:
            cidr: 10.0.0.0/10
        - ipBlock:
            cidr: 10.64.0.0/10
        - ipBlock:
            cidr: 10.192.0.0/10
```

I also end up thinking that if we name it `onlyPublicIPs`, we could just as well also name `dnsPorts` as `onlyDnsPorts` as well, because i see no technical difference between the situations. Both _add_ permissions about dnsPorts and publicIPs specifically, only about that, but none are disallowing or blocking anything.

yuvipanda: I think my confusion is around this step:

> I don't want to risk miscommunicating that this may block private IPs when it is in practice just only allowing the public IPs.

But we are explicitly disallowing private IPs. NetworkPolicies *do* have the ability to *disallow* access via the `except` block, which is what we are using. 

Perhaps the problem is that we are trying to name *one* config item to really mean two different things - allowing all IPs (the 0.0.0.0/0) and blocking all private IPs.  Perhaps we can split that into two configurations? Although I'm not sure what an allow rule for `0.0.0.0/0` *without* the `except` block will approach.

> I also end up thinking that if we name it onlyPublicIPs, we could just as well also name dnsPorts as onlyDnsPorts as well, because i see no technical difference between the situations. Both add permissions about dnsPorts and publicIPs specifically, only about that, but none are disallowing or blocking anything.

Ah, now I understand your concern with the addition of the word `only`. I am confused about the publicIP rule not blocking anything though, as we have an `except` clause there.

yuvipanda: From a deployer perspective, maybe looking through a couple config options might help.

For example, let's look at this config:

```yaml
egressAllowRules:
   publicIPs: false
```

With this config, as an admin, I expect the user pods to just not be able to communicate to any public IPs. However, in practice, I think this will just allow all outgoing traffic, to private and public IPs.

```yaml
egressAllowRules:
   publicIPs: true
```

With this, I expect the change to be that public IPs are accessible. While that is true, private IPs are now blocked - and that isn't apparent from the config.

Perhaps an option is:

```yaml
egressAllowRules:
   publicIPs: true
   privateIPs: false
```

We turn this into two config options, rather than one. We default to `publicIPs: true` and `privateIPs: false`, so same effect as now. Here is a matrix of how that might work

|  | privateIPs: true | privateIPs: false |
| - | - | - |
| publicIPs: true | No network policy setup |   Network Policy that allows *only* public IP ranges (defined as 0.0.0.0/0 - private ranges) (present behavior, default) |
| publicIPs: false | Network Policy that allows *only* private IP ranges | NetworkPolicy that prevents *all* outbound network access (how to do this?) |

Might be a bit more work, but clearer?

consideRatio: > But we are explicitly disallowing private IPs. NetworkPolicies do have the ability to disallow access via the except block, which is what we are using.

No i disagree, they can't disallow, they can just make a rule that excludes allowing something while allowing a larger set. That rule is a pure allow-rule.

> Although I'm not sure what an allow rule for 0.0.0.0/0 without the except block will approach.

Yepp, this is nonsensical. This is because it is not a disallow rule, it is a "but as part of this allow rule, don't also allow this section...".

> Ah, now I understand your concern with the addition of the word only. I am confused about the publicIP rule not blocking anything though, as we have an except clause there.

YES! This has contributed a lot to my confusion in the past, and I'm very concerned about adding to this confusion. 

> Network policies do not conflict; they are additive. If any policy or policies apply to a given pod for a given direction, the connections allowed in that direction from that pod is the union of what the applicable policies allow.

This quote was from https://kubernetes.io/docs/concepts/services-networking/network-policies/. It's all about adding permissions that add up in a "union" like operation and no subtraction operation is involved. The `except` clause is just to restrict the specific rule because its very convenient to allow a large set of IP ranges except a small subset rather than building up the range specification piece by piece. For example, if you were to specify all IPs except a single one, you need a lot of IP ranges specified - but if you were instead to say "0.0.0.0/0 except 123.123.123.123/32" it is far easier.

```yaml
    # Rule 1, a rule that allows all but 123.123.123.123/32
    # but it doesn't disallow123.123.123.123/32
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
            except:
              - 123.123.123.123/32
    # Rule 2: equivalent to rule 1, both allow the same ranges
    - to:
        - ipblock:
            cidr: 0.0.0.0/2
        - ipblock:
            cidr: 64.0.0.0/3
        - ipblock:
            cidr: 96.0.0.0/4
        - ipblock:
            cidr: 112.0.0.0/5
        - ipblock:
            cidr: 120.0.0.0/7
        - ipblock:
            cidr: 122.0.0.0/8
        - ipblock:
            cidr: 123.0.0.0/10
        - ipblock:
            cidr: 123.64.0.0/11
        - ipblock:
            cidr: 123.96.0.0/12
        - ipblock:
            cidr: 123.112.0.0/13
        - ipblock:
            cidr: 123.120.0.0/15
        - ipblock:
            cidr: 123.122.0.0/16
        - ipblock:
            cidr: 123.123.0.0/18
        - ipblock:
            cidr: 123.123.64.0/19
        - ipblock:
            cidr: 123.123.96.0/20
        - ipblock:
            cidr: 123.123.112.0/21
        - ipblock:
            cidr: 123.123.120.0/23
        - ipblock:
            cidr: 123.123.122.0/24
        - ipblock:
            cidr: 123.123.123.0/26
        - ipblock:
            cidr: 123.123.123.64/27
        - ipblock:
            cidr: 123.123.123.96/28
        - ipblock:
            cidr: 123.123.123.112/29
        - ipblock:
            cidr: 123.123.123.120/31
        - ipblock:
            cidr: 123.123.123.122/32
        - ipblock:
            cidr: 123.123.123.124/30
        - ipblock:
            cidr: 123.123.123.128/25
        - ipblock:
            cidr: 123.123.124.0/22
        - ipblock:
            cidr: 123.123.128.0/17
        - ipblock:
            cidr: 123.124.0.0/14
        - ipblock:
            cidr: 123.128.0.0/9
        - ipblock:
            cidr: 124.0.0.0/6
        - ipblock:
            cidr: 128.0.0.0/1
```

consideRatio: > For example, let's look at this config:
> 
> ```yaml
> egressAllowRules:
>    publicIPs: false
> ```
> 
> With this config, as an admin, I expect the user pods to just not be able to communicate to any public IPs. However, in practice, I think this will just allow all outgoing traffic, to private and public IPs.

Hmmm... no. We set `egress: []` as a default in this PR, which stops having a default rule that no longer adds the allow rule "0.0.0.0/0 except the cloud metadata server ip". In other words, with that set to false, there is no rule left adding permissions to public or private IPs, but there is a rule adding permissions to dns ports and important pods like the hub.

consideRatio:
Me and @yuvipanda spoke about this work and concluded a lot of concrete action points to take that are found acceptable to us.

>Implementing the following API for all network policy resources, where the
default values are a bit different between singleuser netpol and the other
three netpol resources.

```yaml
  networkPolicy:
    egressAllowRules:
      cloudMetadataServer: false
      dnsPortsPrivateIPs: true
      nonPrivateIPs: true
      privateIPs: false
```

We discussed the use of rules where the `except` clause would or wouldn't influence other rules. We concluded that for performance reasons and due to uncertainty, we should render our helm templates in a way to avoid such conflicts.

consideRatio:
@yuvipanda I've now made a second pass on this configuration changes according to what we discussed earlier!

jupyterhub/templates/_helpers-netpol.tpl
```diff
@@ -0,0 +1,73 @@
...
+    - ipBlock:
+        cidr: 172.16.0.0/12
+    - ipBlock:
+        cidr: 192.168.0.0/16
+{{- end }}
+
+{{- if $netpol.egressAllowRules.nonPrivateIPs }}
+# Allow outbound connections to non-private IP ranges
+{{- if $netpol.egressAllowRules.privateIPs }}
+# Allow outbound connections to private IP ranges (and their DNS ports)
```
yuvipanda, line 27: Are these two intentionally empty conditional blocks?

jupyterhub/templates/_helpers-netpol.tpl
```diff
@@ -0,0 +1,73 @@
...
+{{- if $netpol.egressAllowRules.cloudMetadataServer }}
+# Allow outbound connections to cloud metadata server
+{{- end }}
+- to:
+    - ipBlock:
+        cidr: 0.0.0.0/0
+        {{- if or (not $netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.cloudMetadataServer) }}
+        except:
+          {{- if not $netpol.egressAllowRules.privateIPs }}
+          # Don't allow outbound connections to private IP ranges
```
yuvipanda, line 38: Link to the RFC that lists these as the private ranges? With bare CIDRs like this I find it very helpful to figure out where they came from.

jupyterhub/templates/_helpers-netpol.tpl
```diff
@@ -0,0 +1,73 @@
...
+
+          {{- if not $netpol.egressAllowRules.cloudMetadataServer }}
+          # Don't allow access to the cloud metadata server
+          - {{ $root.Values.singleuser.cloudMetadata.ip }}/32
+          {{- end }}
+        {{- end }}
+{{- end }}
+
+{{- if and ($netpol.egressAllowRules.privateIPs) (not $netpol.egressAllowRules.nonPrivateIPs) }}
+# Allow outbound connections to private IP ranges (and their DNS ports)
```
yuvipanda, line 52: Mismatched comment?

yuvipanda (commented):
Still WIP I think, but I love this! Completely approve. Comments need some updating, but otherwise ✔️

jupyterhub/templates/_helpers-netpol.tpl
consideRatio: I made them intentionally to render the comment. I felt that without them, the performance optimization to make the rules combine, made it a bit too hard to verify the logic from the rendered result based on the configuration choices.

I'm still not feeling 100% about the performance optimization of combining rules due to the complexity to read the netpol: both as a helm template and as a rendered k8s resource manifest. I'd much prefer if it was optimized effectively by a network policy controller before implemented into rules. I have asked a question related to this in a k8s sigs slack group dedicated to network policies.

I could be convinced to remove the helm template logic to render these comments, but doing that is also a nudge towards reducing the complexity fully by having clearly separate rules. Somewhat tricky judgement calls, but this can be changed without breaking anything later at least.

yuvipanda: Ah I don't mind having them there! Maybe add a helm comment stating why they are there?

More so than performance, I think having two different sets of rules apply to the same IP range is very confusing to reason about. You need to understand enough about netpol and iptables to *know* that the 'to' block will override the 'except' block. So the complexity gets transferred from the helm templating logic to the logic of reading and understanding what the output will do. So that is the tradeoff to be made - more conditionals in the helm logic, or conflicting rule definitions in the network policy?

jupyterhub/templates/_helpers-netpol.tpl
consideRatio: I have the link in the schema.yaml / config reference. If i mention the definition of what a private IP is wherever it is mentioned it would be mentioned in too many places and I would see it as cluttering the other logic, so I think mentioning it here isn't justified.

In practice, I've opted to have each rule emit ~one line of comments about it being active (or not active in this case), and referred to the schema.yaml / config reference for a centralized and more in depth explanation.

jupyterhub/templates/_helpers-netpol.tpl
consideRatio: The "(and their DNS ports)" was a hint about `dnsPortsPrivateIPs` becomes irrelevant, as it automatically accepts communication to the DNS ports by allowing connections to these IPs independent of ports.

Do you have a change suggestion? (preferably sticking to a single line)

jupyterhub/templates/_helpers-netpol.tpl
consideRatio: > You need to understand enough about netpol and iptables to know that the 'to' block will override the 'except' block.

Hmm..... I'd say your knowledge about the practical implementation that has been used to enforce network policies makes it more complicated. For me mainly focused on the k8s understanding, it is easier.

You have a rule saying `to`, and you can choose a target by labels or an ipBlock, the ipBlock could be a cidr range, and you are also allowed to say that it should be a cidr range with some exceptions. The destination rule is still a ipBlock which is either a full cidr range or parts of a cidr range.

I'm feeling 90% confident that a correctly implemented network policy controller/enforcer will just addatively add more persmissions based on the ipBlocks.

I had helm template comments as well, but it felt like I cluttered the logic by adding them so I deleted then :p Perhaps I can add some notes at the top instead about the implementation decisions made overall?

consideRatio: > So that is the tradeoff to be made - more conditionals in the helm logic, or conflicting rule definitions in the network policy?

I'm feeling very confident that the k8s documentation is correct - network policies can't conflict - they just incrementally add permissions.

@yuvipanda you have doubt about this right? Being iptables savvy, can you verify that there can be conflicts - as I think it can't be?

consideRatio: I'll try verify this now myself practically, not by inspecting the ip tables, but instead simply trying to restrict communication between pods.

consideRatio:
@yuvipanda and I have discussed how network policies work and we had a disagreement how we understood them to work. Due to that, the following test was developed. Here are the reproduction steps and a gif demonstrating the result.

```yaml
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
            except:
              - 10.72.2.240/32
    # Will decommenting this rule make the netpol allow the pod to establish
    # a connection to 10.72.2.240? The answer was yes! NetworkPolicy
    # resources rules are additive like this!
    # - to:
    #     - ipBlock:
    #         cidr: 10.72.0.0/16
```

netpol-rule-behavior-verification

```shell
# Creation of a server exposing port 1080 to try connect against from the busybox
helm upgrade --install mockserver http://www.mock-server.com/mockserver-5.11.1.tgz

# Get the mockserver Pod IP:
kubectl describe pod -l app=mockserver | grep IP
IP=10.72.2.240

# Creation of a busybox pod (separate terminal)
kubectl run -it --rm busybox --image=busybox --restart=Never -- sh

# Run the following in the busybox pod to test the ability
# to create a connection to the webserver repeatedly
watch -tn1 "timeout 1 nc -zv 10.72.2.240 1080"

# Label the busybox pod
kubectl label pod busybox enforce-test-netpol=true

# Restrict the busybox pod to communication to anything
# besides the mockserver's ip
cat <<EOF | kubectl apply -f -
kind: NetworkPolicy
metadata:
  name: test-netpol
spec:
  podSelector:
    matchLabels:
      enforce-test-netpol: "true"
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
            except:
              - 10.72.2.240/32
    # - to:
    #     - ipBlock:
    #         cidr: 10.72.0.0/16
EOF


# Add another rule that allows communication
# to the mockservers ip
cat <<EOF | kubectl apply -f -
kind: NetworkPolicy
metadata:
  name: test-netpol
spec:
  podSelector:
    matchLabels:
      enforce-test-netpol: "true"
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
            except:
              - 10.72.2.240/32
    - to:
        - ipBlock:
            cidr: 10.72.0.0/16
EOF
```

jupyterhub/templates/_helpers-netpol.tpl
consideRatio: See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2508#issuecomment-998753082 for a followup.

jupyterhub/values.yaml
consideRatio: I think the resolution of this discussion is https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2508#issuecomment-998753082.

jupyterhub/schema.yaml
```diff
@@ -583,11 +583,86 @@ properties:
...
+                type: boolean
+                description: |
+                  Defaults to `false` for singleuser servers, but to `true` for
+                  all other network policies.
+
+                  Private IPs refer to the IP ranges `10.0.0.0/8`,
+                  `172.16.0.0/12`, `192.168.0.0/16`.
+
+                  When enabled this rule allows the respective pod(s) to
+                  establish outbound connections to the internal k8s cluster.
```
manics, line 657:
```suggestion
                  establish outbound connections to the internal k8s cluster,
                  and to other private IP ranges outside the cluster (for example
                  if your internal network uses a private IP range).
```

jupyterhub/schema.yaml
```diff
@@ -583,11 +583,86 @@ properties:
...
+                  This means users can access the internet but not (say) an
+                  unsecured prometheus server running in the same cluster.
+
+                  Since not all workloads in the k8s cluster may have
+                  NetworkPolicies setup to restrict their incoming connections,
+                  having this set to false can be a good defence against
+                  malicious intent from someone in control of software in these
+                  pods.
```
manics, line 665:
```suggestion
                  pods.
                  
                  To improve security we recommend only allowing the private IP CIDRs required
                  <TODO: insert example>
                  or namespaces
                  <TODO: insert example>
                  instead of enabling this.
```
Alternatively the example could go in to the docs https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#kubernetes-network-policies instead of embedding it here?

manics (commented):
A couple of docs suggestions, but feel free to put them somewhere else!

jupyterhub/templates/_helpers-netpol.tpl
yuvipanda: I generally agree, but I only want it exactly where the IPs themselves are listed - otherwise they look like [magic numbers](https://refactoring.guru/replace-magic-number-with-symbolic-constant). So this isn't a definition of what private IPs are, but a reference to exactly where these hard coded CIDR values come from.

yuvipanda:
Wow, that is amazing work @consideRatio! That definitely did not match what my understanding of how network policies worked, and TIL. Thank you for putting in the effort to make that happen. I'll de-confuse myself over time :) Given this, I think it's alright to have them be mentioned multiple times - I'll admit it's confusing, but that is just how network policies seem to be.

jupyterhub/templates/_helpers-netpol.tpl
yuvipanda: Just remove the comment about 'and their DNS ports'?

consideRatio:
@yuvipanda great! I'm happy to stop coalescing the rules for chart maintainability and readability, but i have no ideas about the performance impact - how concerned are you about that?

Ah you wrote on slack:

> I think we should definitely remove our notes about performance - based on the results of this, I expect there is no perf difference at all

Then I'll restructure this back to be standalone rules again!

consideRatio:
@yuvipanda @manics - i'm opinionated towards having  easy logic to maintain: separated rules

Currently we have logic that coalesce rules, which leads to easier parsing of the final result.

What are you opinionated towards doing? Refactor to have separate rules or not?

I refactored it to coalesced rules motivated by potential performance, but yuvi doesnt think there id a notable difference, so then im opinionated towards separate rules for charr maintainability and comprehensibility. The final rendered ruleset will be more complicated as a netpol ruleset, but also easier to map to the chart condig that generated it: like 4/2*3/6 is 1, and 1 is easier to understand in one way, but the expression is easier to map to the configuration that led to it being 1.

yuvipanda:
I think there should not be any noticeable performance difference, so I think having two rulesets that are easy to reason about in the helm template logic is worth the added complexity in the rendered rules.

manics:
Separate rules makes sense. I favour clarity and maintainability unless there's evidence of performance degradation. My naive(!) assumption is that firewall rules can be optimised by the K8s network controller or the Linux kernel.

jupyterhub/schema.yaml
consideRatio: This is the description of a boolean value, toggling the inclusion of a egress rule alongside other rules. I don't think an example on how to configure something else belongs here if this would be for example disabled.

I'll see if I can push on and update the docs to include details about this new configuration in general and some ideas on recommendations.

consideRatio:
@yuvipanda @manics @choldgraf this is now ready for review again. I think I've addressed all review comments, my own included, at this point.

yuvipanda:
This looks awesome to me, @consideRatio! I don't wanna merge this as I opened the PR, but I think this is good to go!

consideRatio:
Thank you @yuvipanda!! Wieee this PR merged is a big win! :)

choldgraf:
Amazing - thanks for the persistence on this 🙂

PR #2517, opened 2021-12-15
docs: document `singleuser.someConfig` by linking to `KubeSpawner.some_config` docs

manics:
When responding to https://discourse.jupyter.org/t/sql-operationalerror-with-jh-default-config/12207/3 I realised the Z2JH docs on setting the storage group ID are a bit unclear.

- [`singleuser.fsGid`](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#singleuser-fsgid) links to https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html
- https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html doesn't contain `fsGid` so searching for it doesn't work, the property is called `fs_gid`.
- `fs_gid` maps to the K8S property `fsGroup` which is relevant when investigation cloud storage volume permissions.

It'd be nice if we could link directly to `fs_gid` in https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html but it doesn't have any in-page anchors.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1788,10 +1788,9 @@ properties:
       networkPolicy: *networkPolicy-spec
       podNameTemplate:
         type: [string, "null"]
-        description: &kubespawner-native-config-description |
-          KubeSpawner native configuration, see the [KubeSpawner
-          documentation](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html)
-          for more information.
+        description: |
+          Passthrough configuration for
+          [KubeSpawner.pod_name_template](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.pod_name_template).
       cpu:
         type: object
         additionalProperties: false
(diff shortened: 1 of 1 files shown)


consideRatio:
@manics you have identified a general issue, that the configuration in kubespawner has been formatted with snake_case instead of camelCase. Due to that, a lot of properties all referencing the `kubespawner-native-config-description` YAML anchor has the same issue.

I have a somewhat bad feeling of duplicating this documentation in z2jh and kubespawner as its a maintenance practice I've seen have caused problems in the past. What do you think about generally updating this text instead?

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/5af0902212f23f227e287b24b4b76c452246c072/jupyterhub/schema.yaml#L1791-L1794

I've opened https://github.com/jupyterhub/kubespawner/issues/555 to reflect the need to support HTML `#anchors` for the configuration reference in kubespawner.

manics:
This PR was motivated by the lack of a link between `fsgid` and `fsgroup`, which I think is important to understand when configuring some dynamic storage providers. I've seen this come up multiple times on discourse, and AFAIK this isn't documented in Z2JH. The point about snakecase vs camelCase is a contributory issue, but I still think we should document this in Z2JH since it's related to integrating JupyterHub with a cloud provider rather than something specific to Jupyter.

I'm happy to move/add some docs in a different location if you can suggest somewhere.

consideRatio:
> It'd be nice if we could link directly to `fs_gid` in https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html but it doesn't have any in-page anchors.

It seems that if you hover to the right of the heading, it pops up an invisible anchor, see https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.fs_gid

> I'm happy to move/add some docs in a different location if you can suggest somewhere.

I do think this belongs to KubeSpawner, but it is indeed a bit tricky to understand how things map since this helm chart has its own names, kubespawner its own, and the pod specification its own. Anyhow, I suggest that this chart doesn't add documentation even though it would be helpful for end users, but that instead we try to improve the coupling to KubeSpawner's documentation.

I've added a commit reverting yours, and added a commit where all KubeSpawner native configuration entries are instead linking out directly to `KubeSpawner.<associated_config>` entries. What do you think about this @manics?

consideRatio:
I'm just about to add `singleuser.allowPrivilegeEscalation` to #2663 where I'd like to rely on this PRs way of doing things. I'll go for a self-merge to retain momentum. @manics feel free to ask me to adjust anything retrospectively!

PR #2518, opened 2021-12-16
DOC: add extra AWS ssh info

raybellwaves:
I learned how to do this recently and wanted to upstream it here.

### DIFF
--- a/doc/source/kubernetes/amazon/step-zero-aws.md
+++ b/doc/source/kubernetes/amazon/step-zero-aws.md
@@ -38,7 +38,7 @@ template you will use to setup and shape your cluster.
    ```
 
 3. SSH to your CI host. Instructions on how to do this are given
-   [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html)
+   [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html).
 
 4. Install kops and kubectl on your CI host
 
--- a/doc/source/kubernetes/amazon/step-zero-aws.md
+++ b/doc/source/kubernetes/amazon/step-zero-aws.md
@@ -31,9 +31,18 @@ template you will use to setup and shape your cluster.
 
    When creating it, assign the IAM role created in step 1.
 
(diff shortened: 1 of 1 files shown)


doc/source/kubernetes/amazon/step-zero-aws.md
```diff
@@ -31,9 +31,18 @@ template you will use to setup and shape your cluster.
...
+   Once created, download ssh keys (.pem file). Ensure permissions are restrictive on the file:
+
+   ```
+   chmod 400 name.pem
+   ```
+
+3. SSH to your CI host as:
+
+   ```
+   ssh -i /path/name.pem ec2-user@my-instance-Private-IPv4-address
```
raybellwaves, line 43: This may be specific to my required way of ssh’ing in

manics: It must be! A private IP shouldn't be accessible from outside AWS.

consideRatio: Is there an request for change here @manics?

If ec2-user isn't a hardcoded username, perhaps you can make it included `my-` as well to clarify its really not something predefined by AWS that others can re-use.

manics: `my-instance-Private-IPv4-address` won't work, you shouldn't be able to SSH into your private IP from outside! Unless you've got a VPN or something more advanced.

`ec2-user` is fine if you're using the Amazon Linux AMI since that's the default username.... So this depends on how much detail you want to go into. If you assume people know how to use AWS EC2 it's enough to just say bring up a VM and SSH in, if you don't then you have to go into details about the choice of AMI, security groups to allow SSH, etc.

raybellwaves: Thanks for the comments. I'll scale my code back a bit and point to documentation. It saves on being too specific

consideRatio (approved):
Thanks @raybellwaves!!!

@manics LGTY?

manics (approved):
I think linking to the AWS instructions is a good idea, since it'll always be up to date.

PR #2531, opened 2021-12-30
fix ingressClassName for kubernetes version is >= 1.18

chenshap:
Signed-off-by: chenshapira <cs030391@gmail.com>


consideRatio:
If you try render the templates, you will find that if and with functions in the same way here. The with block will not render if the value is falsy.

PR #2532, opened 2022-01-01
 Fix Ingress API check

emrahkaya:
If the full API resource is given, Helm on ArgoCD produces wrong answer for the ".Capabilities.APIVersions.Has" template.
Checking for "networking.k8s.io/v1" is enough, and works well.


consideRatio:
Is that because helm use a too outdated version, older than 3.5? I've seen these things in the past.

consideRatio:
Closed because I assume Argo has been using a too outdated version of `helm`.

PR #2544, opened 2022-01-15
Use chart logo from this repo

manics:
https://jupyter.org/assets/hublogo.svg has moved to https://jupyter.org/assets/homepage/hublogo.svg

I've copied the logo into the RTD site instead to keep everything in this repo.


consideRatio (approved):

consideRatio:
Hmmm, it seems that our built docs in RTD don't expose the never referenced hublogo.svg file - I assume that is because when built from source, it wasn't referenced.

consideRatio:
Opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2599

PR #2546, opened 2022-01-18
Enable parent chart's (binderhub etc) to use imagePullSecrets helper

consideRatio:
I saw https://github.com/jupyterhub/binderhub/pull/1444 and figured reusing the helper template in this helm chart was the way to do it, because then binderhub could make use of the functionality defined in this Helm chart to both create and reference a k8s Secret with image registry credentials for the various pods.

I think https://github.com/jupyterhub/binderhub/pull/1444 can be merged as it is, but think of this as the long term solution rather than a short term fix. BinderHub would need to make a version bump of the jupyterhub helm chart to include a release with this PR merged.

### Review notes

Note that most lines changed are indentation changes, focus your attention on 98df189c5c46e6f99b4ef10aed3c59d67b547f38 instead of the other commit.

### Related

- About using this Helm chart to create a k8s Secret with image pulling credentials: https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#imagepullsecret
- About using this Helm chart's named template ("jupyterhub.imagePullSecrets") to render a combination of what is specified under `<something>.image.pullSecrets` and what is declared in this Helm chart's `.Values.imagePullSecret` and `.Values.imagePullSecrets` fields: https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#imagepullsecrets

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -180,9 +180,30 @@ component: {{ include "jupyterhub.componentLabel" . }}
     Augments passed .pullSecrets with $.Values.imagePullSecrets
 */}}
 {{- define "jupyterhub.imagePullSecrets" -}}
+    {{- /*
+        We have implemented a trick to allow a parent chart depending on this
+        chart to call this named templates.
+
+        Caveats and notes:
+
+            1. While parent charts can reference these, grandparent charts can't.
+            2. Parent charts must not use an alias for this chart.
+            3. There is no failsafe workaround to above due to
+                https://github.com/helm/helm/issues/9214.
+            4. .Chart is of its own type (*chart.Metadata) and needs to be casted
+                using "toYaml | fromYaml" in order to be able to use normal helm
+                template functions on it.
+    */}}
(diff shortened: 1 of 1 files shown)


manics (approved):
I haven't tested this but it looks OK!

If it's ready to go why don't we bump the BinderHub dependency to a dev Z2JH version instead of merging https://github.com/jupyterhub/binderhub/pull/1444 ?

consideRatio:
@manics I've now manually tested this to not cause any change of behaviour in z2jh, and to function to be used from binderhub! I'll go for a merge!

While binderhub could update to this helm chart, there are breaking changes not documented. I wish that binderhub don't bump before there is a proper changelog to reference with regards to what needs to be considered bumping to this dev version that currently is using JupyterHub 2 etc.

Practically:
- I suggest binderhub waits bumping z2jh until a beta release is out at least
- I think it's fine to either merge the PR in question or wait for this functionality to be used
- When z2jh is bumped in binderhub, this helper to set the imagePullSecrets of all the binderhub's pod should be implemented, as well as the imagePullPolicy for all pods while we are at it.

Issue #2265, opened 2021-06-21
A better story for Authenticators not in the Hub image


minrk:
### Proposed change

I think we should have a better story for users who want to use an Authenticator not shipped with the Hub image (or any other pip-installed packages).

I don't know if this should be a *technical* solution or a documentation one. My initial inclination is that documentation *should* be enough, but we should test it out and make sure it works well and is sustainable across an updating hub.

### Alternative options

Don't!  Right now, folks can use very lightweight custom images, like:

```dockerfile
FROM jupyterhub/k8s-hub:1.0.0
RUN python3 -m pip install my-authenticator-package
```

and register in config:

```yaml
hub:
  image: my-hub-image:my-hub-tag
```

This works pretty well the first time you do it, and isn't *so* bad for folks sticking to tagged releases. But it's a pretty big pain to manage for anyone tracking chartpress-published dev versions, because the image versions aren't easily discoverable, leading to issues like #1015. And folks have to update the base tag and rebuild the image every time they bump the chart version (this is true no matter what version is installed, but at least it's easy to know when using releases).

### Who would use this feature?

- Anyone who wants to use a package (or version) not in the default Hub image
- Maintainers reviewing PRs adding little-used dependencies. If external dependencies have a good solution, it will be easier to decline additions to the default image and point to this custom-package installation than accept every PR adding a package to the default image, which *implies* support even if it's not tested, etc.

### (Optional): Suggest a solution

Possible ideas:

- support a pip-install step that works at runtime. This might be doable via documenting existing lifecycle hooks/volumes to allow a pip install to run at image start. Most of these are going to be super lightweight pip installs, so shouldn't be too bad. Support for a hash-including requirements.txt would be good, though.
- work out/document a workflow that makes it easier to keep a custom image extending the default up-to-date across chart upgrades

For instance, to get the current image, this oneliner works:

```bash
helm show values jupyterhub/jupyterhub --version 1.0.0-n015.h95486ae6 | python3 -c 'import sys, json, yaml; json.dump(yaml.safe_load(sys.stdin), sys.stdout)' | jq -r '.hub.image | [.name, .tag] | join(":")'
```

(why am I using Python to turn yaml into json and then jq to print values instead of just Python to print values? 🤷 )

Or to assume the repo and just get the tag with shyaml:

```bash
helm show values jupyterhub/jupyterhub --version 1.0.0-n015.h95486ae6 | shyaml get-value hub.image.tag
```

[...]

PR #2577, opened 2022-02-11
custos-oauthenticator-requirement

isururanawaka:
Custos is a security middleware for science gateways. See https://airavata.apache.org/custos/ for additional information.

This PR contains the dependency for the implementation of Custos Authenticator for Jupyterhub as an OIDC plugin.

### DIFF
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -107,6 +107,7 @@ oauthenticator==14.2.0
     # via
     #   -r requirements.in
     #   jupyterhub-ltiauthenticator
+custos-jupyterhub-authenticator==1.0.0
 oauthlib==3.1.1
     # via
     #   jupyterhub
--- a/images/hub/requirements.txt
+++ b/images/hub/requirements.txt
@@ -107,7 +107,7 @@ oauthenticator==14.2.0
     # via
     #   -r requirements.in
     #   jupyterhub-ltiauthenticator
-custos-jupyterhub-authenticator==1.0.0
+custos-jupyterhub-authenticator==1.0.3
 oauthlib==3.1.1
     # via
     #   jupyterhub


consideRatio:
I want to raise this for discussion with the @jupyterhub/jupyterhubteam. What policy should we have for adding an authenticator to the jupyterhub image used in this Helm chart?

Currently, we have the following:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/2d03a133eb2253e25891f3bbd59f9124cfdbdde6/images/hub/requirements.in#L10-L18

I'm thinking we should stick to those maintained by the JupyterHub team, and ask that other third party authenticators are installed by deriving a custom image or similar.

manics:
By including a package in Z2JH we're saying it works, it's safe, and it's secure. I think sticking with just JupyterHub authenticators is the most practical solution. 

_If_ we wanted to relax it slightly we could add authenticators where there's strong demand, and evidence of long term support for the package.

I think adding more docs on extending the hub image including a basic example would be nice to have though, and provides a page we can point people to if this comes up again.

willingc:
I believe that we should only include JupyterHub packages in the chart. For cases like this particular one (thanks @isururanawaka for the suggestion), I would prefer creating a page in the docs or wiki for external packages. IIRC we had a section in the JupyterHub docs for Community Provided Packages.

smarru:
@consideRatio @manics @willingc thanks if it helps, we are committed to supporting this authenticator. Custos is part of Apache Airavata, a top-level Apache Software Foundation project - https://airavata.apache.org/custos/ and is supported by NSF - https://www.nsf.gov/awardsearch/showAward?AWD_ID=1840003 our goal for this PR is for any users of the Jetstream cloud to checkout the code and have native support of federated authentication using CILogon for US research community - https://jetstream-cloud.org/

isururanawaka:
@manics Thanks for the suggestions. We will update the source and README file.

consideRatio:
Thanks for the discussion everyone, and thank you for your work to provide an authenticator for the community to use @isururanawaka!

I'll go ahead and close this PR as there seem to be agreement that we would like to stick to only having JupyterHub maintained authenticator packages in this image.

We currently have no great option on how to install an additional package on startup by adding some configuration or similar, but there is an open issue of relevance about it to keep an eye on: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2265.

PR #2578, opened 2022-02-11
Support idle culler --cull-admin-users

manics:
See https://discourse.jupyter.org/t/jupyterhub-removes-admin-when-cull-is-used/12976

Note I haven't tested this.


consideRatio:
Thanks @manics!

PR #2590, opened 2022-02-22
Update user-scheduler's kube-scheduler binary and config when in k8s clusters versioned >=1.21

consideRatio:
## Description of breaking changes

If you have configured `scheduling.userScheduler.plugins`, this is a breaking change.

- When k8s <=1.20 is used, you can no longer influence the configuration via `scheduling.userScheduler.plugins`. The previous default configuration is now hardcoded for this scenario.
- When k8s >=1.21 is used, `scheduling.userScheduler.plugins` configuration should be updated to be compatible with a KubeSchedulerConfiguration of kind `v1beta3`.

## Original PR description

As new k8s versions arrive, we should try ensure that our user-scheduler that schedules some of our pods, that relies on a k8s official kube-scheduler binary - is updated as well.

This PR bumps the kube-scheduler binary version we rely on in k8s clusters versioned 1.21 or higher to the kube-scheduler binary v1.23.4, and the associated configuration to be of a modern kind (v1beta3) incompatible with the previous kind (v1beta1). Since I learned that kube-scheduler verionsed >=1.21 also couldn't be used in k8s clusters verionsed <=1.20, I had to introduce some complexity.

I made the configuration this Helm chart provides under `scheduling.userScheduler.plugins` be updated and respected in k8s k8s clusters >=1.21, but adopted a hardcoded user-scheduler configuration in k8s clusters <=1.20 that relied on the latest known function kube-scheduler binary (1.20.15).

### DIFF
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -46,7 +46,7 @@ spec:
           {{- else }}
           # WARNING: The tag of this image is hardcoded, and the
           #          "scheduling.userScheduler.image.tag" configuration of the
-          #          Helm chart that generated this resource manifest wasn't
+          #          Helm chart that generated this resource manifest isn't
           #          respected. If you install the Helm chart in a k8s cluster
           #          versioned 1.21 or higher, your configuration will be
           #          respected.
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -47,13 +47,6 @@ spec:
           {{- end }}
           command:
             - /usr/local/bin/kube-scheduler
-            # NOTE: --leader-elect-... (new) and --lock-object-... (deprecated)
-            #       flags are silently ignored in favor of whats defined in the
-            #       passed KubeSchedulerConfiguration whenever --config is
-            #       passed.
-            #
-            # ref: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/
-            #
             # NOTE: --authentication-skip-lookup=true is used to avoid a
             #       seemingly harmless error, if we need to not skip
             #       "authentication lookup" in the future, see the linked issue.
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -61,12 +61,14 @@ spec:
           livenessProbe:
             httpGet:
               path: /healthz
-              port: 10251
+              scheme: HTTPS
+              port: 10259
             initialDelaySeconds: 15
           readinessProbe:
             httpGet:
               path: /healthz
-              port: 10251
+              scheme: HTTPS
+              port: 10259
           {{- with .Values.scheduling.userScheduler.resources }}
           resources:
             {{- . | toYaml | nindent 12 }}
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -41,7 +41,17 @@ spec:
       {{- end }}
       containers:
         - name: kube-scheduler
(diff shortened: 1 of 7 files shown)


jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -68,12 +68,12 @@ spec:
           livenessProbe:
             httpGet:
               path: /healthz
-              port: 10251
+              port: 10259
```
consideRatio, line 71:
```suggestion
              scheme: HTTPS
              port: 10259
```

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -68,12 +68,12 @@ spec:
...
             initialDelaySeconds: 15
           readinessProbe:
             httpGet:
               path: /healthz
-              port: 10251
+              port: 10259
```
consideRatio, line 76:
```suggestion
              scheme: HTTPS
              port: 10259
```

manics:
Are the failures due to old k8s versions?

I've started https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2591 anyway

consideRatio:
Rebased on top of #2600 to not have a failing unrelated test.

jupyterhub/values.yaml
```diff
@@ -398,42 +398,98 @@ scheduling:
...
-      #            that we have forked.
+      #            that we have forked in
+      #            templates/scheduling/user-scheduler/rbac.yaml.
+      #
+      #            Debugging advice:
+      #
+      #            - Is configuration of kube-scheduler broken in
+      #              templates/scheduling/user-scheduler/configmap.yaml?
+      #
+      #            - Is the kube-scheduler binary's compatebility to work
```
manics, line 473:
```suggestion
      #            - Is the kube-scheduler binary's compatibility to work
```

jupyterhub/values.yaml
```diff
@@ -398,42 +398,98 @@ scheduling:
...
+      #              against a k8s api-server that is too new or too old?
+      #
+      #            - You can update the GitHub workflow that runs tests to
+      #              include "deploy/user-scheduler" in the k8s namespace report
+      #              and reduce the user-scheduler deployments replicas to 1 in
+      #              dev-config.yaml to get relevant logs from the user-scheduler
+      #              pods. Inspect the "Kubernetes namespace report" action!
+      #
+      #            - Typical failures are that kube-scheduler fail to search for
```
manics, line 482:
```suggestion
      #            - Typical failures are that kube-scheduler fails to search for
```

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -41,19 +41,22 @@ spec:
       {{- end }}
       containers:
         - name: kube-scheduler
+          {{- if ge (atoi .Capabilities.KubeVersion.Minor) 21 }}
           image: {{ .Values.scheduling.userScheduler.image.name }}:{{ .Values.scheduling.userScheduler.image.tag }}
+          {{- else }}
+          # WARNING: The tag of this image is hardcoded, and the
+          #          "scheduling.userScheduler.image.tag" configuration of the
+          #          Helm chart that generated this resource manifest wasn't
```
manics, line 49:
```suggestion
          #          Helm chart that generated this resource manifest isn't
```

manics (commented):
Makes sense to me!

Is there anything that needs to be [documented](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/optimization.html#efficient-cluster-autoscaling), or is it fine as it is?

consideRatio:
> Is there anything that needs to be [documented](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/optimization.html#efficient-cluster-autoscaling), or is it fine as it is?

The `plugins` configuration is in my mind an advanced escape hatch. I've added some documentation in this PR description of what I think could make sense to write as part of the changelog as it is a breaking change for someone who has used this advanced configuration - which I suspect may be 0 people.

manics (approved):

PR #2594, opened 2022-02-24
Changes for asyncio kubespawner

athornton:
These are helpful for https://github.com/jupyterhub/kubespawner/pull/563

Neither commit is strictly necessary, but if you don't use the first patch, you still need the synchronous k8s client, even though it is now not a KubeSpawner dependency.

The second one is quite optional.  I'm using it because one of my goals is to reduce log noise to the point where my JupyterHub logs tell me about real problems.  I know that I am not properly closing connections, and the reasons are due to the shared reflector and the shared k8s clients in KubeSpawner, both of which bar using them as self-closing context managers or closing them when one Spawner is done with them.


athornton:
(not going to pass tests until KubeSpawner 3.x is adopted)

consideRatio:
@athornton can you document in this PRs original post a copy-pasted example of an error message supressed by this, and describe that this PR patches a library to supress those messages stemming from the upcoming switch to kubespawner using a different kubernetes api-server client libary that relies on async functions.

I'm thinking this kind of patch is the last resort solution. I want to dig into the details and come to the conclusion it really seem to be the last resort we have before approving a workaround like this in a downstream project.

Practically, I'm thinking of this as a kubespawner issue to be resolved before wanting to consider a z2jh workaround.

consideRatio:
I opened https://github.com/jupyterhub/kubespawner/issues/578, let's switch to discuss things there!

athornton:
https://github.com/jupyterhub/kubespawner/pull/579 is a much better way to do this, so I'm withdrawing the monkeypatch.

The remaining change should be uncontroversial.

consideRatio:
@athornton I'd like PR titles and descriptions to be clear and easy to understand for a) people looking into active PRs in general to contribute with review effort, b) anyone reading a changelog.

We can merge this change along with a kubespawner version bump at a later time (I don't want to put our main branch in a broken state), but please update your PR title/description!

PR #2602, opened 2022-02-24
docs: add examples about creating regional GKE clusters with high availability

Economax:
Added specifications for high availability JupyterHub with examples that are tested and confirmed to work with JupyterHub Helm version 1.1.1. as of 2022-02-24

### DIFF
--- a/docs/source/kubernetes/google/step-zero-gcp.md
+++ b/docs/source/kubernetes/google/step-zero-gcp.md
@@ -80,8 +80,7 @@ your google cloud account.
      - They also increase control plane uptime to 99.95%.
      - To avoid tripling the number of nodes while still having HA kubernetes, the `--node-locations` flag can be used to specify a single zone to use.
 
-
-   A High Availability (HA) cluster can be specified as such: 
+   A High Availability (HA) cluster can be specified as such:
 
    ```
    gcloud container clusters create \
@@ -92,9 +91,9 @@ your google cloud account.
      --cluster-version latest \
      <CLUSTERNAME>
    ```
-   
+
    For example, specifying one for scandinavia:
-   
+
    ```
    gcloud container clusters create \
      --machine-type n1-standard-2 \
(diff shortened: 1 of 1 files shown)


consideRatio (commented):
Thank you for helping maintaining this documentation!

I'm afraid of the complexity to make sure these examples are kept up to date given the level of detail they include.

A note about HA GKE clusters is that you need to pay significant amounts of money to have them now, while in the past, they were free. I think instead of promoting them further, I suggest we actually adjust the docs to help users default to something cheap.

A HA k8s api-server isn't so important unless you have a massive installation with thousands of users where new users arrive every other second and you want to ensure no downtime during that time.

I worry a bit more about having examples that become outdated, or overwhelming the reader, than leaving out more advanced details. How do you think about that?

Looking into the situation, it seems that GCP provides free credits for having one zonal cluster per account, but if a regional cluster is used - no such credits is given and it will cost ~72 USD a month for the cluster - apart from the nodes.

Economax:
Thanks for your response @consideRatio! 

You make very good points. I think you're completely right that the default settings should be cheap and simple!

My personal experience has been that more documentation is always better than less, even if out of date, because documentation is evidence/information that can lead to a solution for problems for which there currently exists no solution. 
The recent issue I commented on regarding autohttps is a prime example of that, in my view!
What do you think about that?

Regardless, I think you're right that the documentation needs to be structured well so as to neither overwhelm nor introduce greater documentation burden.

Do you see any way in which we could potentially satisfy all concerns? Maybe change the documentation structure a bit or something?

We're running this on GCP with the free trial using HA right now and we have currently not incurred any extra billing. We are still on free credits. 

Could that be due to having specified node locations? According to the documentation that avoids duplication, at the very least...

manics:
One option is to post on the Jupyter community forum
https://discourse.jupyter.org/
It's not just for discussions, there's a lot of helpful technical advice there too, and your instructions could eventually become a living document, with others adding to it.

consideRatio:
Hey @Economax, sorry for the slow followup!

> My personal experience has been that more documentation is always better than less, even if out of date, because documentation is evidence/information that can lead to a solution for problems for which there currently exists no solution.

I agree with this point!

What I struggle with is mainly the maintenance impact rather than the user impact. More things in the documentations main text makes it a bit harder to overview and update. I consider a person following the guide, but then since it now has a "red thread" that branches apart in different ways, only some parts may have been verified tested by the person. If anything is to be changed, should that person try learn about the other paths as well?

I think a compromise should involve separating the text so that a single page doesn't involve also variations about setting up a regional cluster side by side with the other parts. Practically, I think I would favor to link out to a dedicated post on discourse.jupyter.org for this.

I've tried out something similar in https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449, a post that is linked from https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-environment.html#user-dependent-profile-options.

@Economax would you like to update this PR to link out to a discourse.jupyter.org post about this?

consideRatio:
I'll close this for now, thank you for working to improve this ecosystem of software @Economax!!! :heart: :tada:

PR #2628, opened 2022-03-18
Fix failure to respect proxy.secretSync.resources configuration

jhowton-restor3d:
The `resources` property was not being set in the secret-sync container configuration. This is a configurable property that is described in the zero-to-jupyterhub configuration docs.


consideRatio:
Thanks @jhowton-restor3d this looks perfect!

PR #2635, opened 2022-03-23
Require k8s 1.20+ and small cleanups based on assuming it

### LINKED ISSUE #2591: Review supported versions of K8S for 2.0.0 release
### Proposed change
Currently we test and support K8s versions 1.17 to 1.23:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6f4609e2f73258c7e0146f8e849d31aa7fe7f09f/.github/workflows/test-chart.yaml#L127-L154

Several of these older versions have reached end-of-life according to https://endoflife.date/kubernetes
Can we drop some old versions? If we do we should ensure all the other test parameters are still tested

### Alternative options
Do nothing

Developers benefit from not needing to ensure compatibility with older K8s versions

- Drop versions that have reached EOL?
- Drop versions that reached EOL more than X years ago?

consideRatio:
This PR closes #2591 by requiring k8s 1.20+ and dropping support for k8s 1.17-1.19 that are no longer supported by the major cloud providers.

As part of this, I also do some cleanup to not test against old versions, remove misc conditionals to support old versions, and updates misc documentation references found to old versions.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -124,32 +124,22 @@ jobs:
         # k3s-version: https://github.com/rancher/k3s/tags
         # k3s-channel: https://update.k3s.io/v1-release/channels
         include:
-          - k3s-channel: v1.23
+          - k3s-channel: latest
             test: install
-          - k3s-channel: v1.22
+          - k3s-channel: stable
             test: install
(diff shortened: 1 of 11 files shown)


yuvipanda:
Thanks, @consideRatio

PR #2640, opened 2022-03-25
docs: Remove a false promise regarding CHOWN_HOME

dmigo:
The comment implies that the problem was fixed a few versions ago, but unfortunately it wasn't.


yuvipanda:
Thanks a lot, @dmigo!

PR #2574, opened 2022-02-10
Implement continuous image puller priority for better pod scheduling

a3626a:
### Changes
- Introduce `scheduling.podPriority.podPriority.continuousImagePullerPriority` in `values.yaml`
- Modify `schema.yaml` accordingly.
- Define `jupyterhub.continuous-image-puller-priority.fullname`
- Let continuous image puller demonset create pods with "continuous image puller priority"

### Purpose
- Sometimes, placeholder pods block continuous image puller pods to spawn.
image
image

- This situation can occur after increasing the number of user placeholder pods a lot(in my case, it is 100).

### Alternatives
- Set "continuous imager puller priority" as `(defaultPriority+userPlaceholderPriority)/2` automatically.

### Related Issues
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1760

PR #2681, opened 2022-05-05
Enable image-puller pods to evict user-placeholder pods

consideRatio:
I wanted to push a commit to #2574 but wasn't allowed to, due to that I decided to open a new PR built on top of #2574. @a3626a is this okay with you?

I've rebased @a3626a's commit from #2574 and added another commit with changes I wanted to see.

## Summary

This PR adds another PriorityClass resource that [hook|continuous]-image-puller pods can reference in order to acquire a priority making them considered more important than user placeholder pods but still not more important than normal user pods. In practice, this resolves a bug where we could have too many user placeholder pods to fit a image-puller pod and end up with a node that didn't get their image pre-pulled for users.

I labelled it as a bugfix, its kind of a new feature as well - intended to fix a bug in behavior.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -2259,6 +2259,11 @@ properties:
           4. Normal user pods have a higher priority than the user-placeholder
              pods
 
+          Additionally, continuous image puller pods should get a priority
+          between normal user pods and placeholder pods. Otherwise, continuous
+          image puller pods may block normal user pods to spawn, and placeholder
+          pods may block continuous image puller pods.
+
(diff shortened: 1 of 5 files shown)


a3626a:
> I wanted to push a commit to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2574 but wasn't allowed to, due to that I decided to open a new PR built on top of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2574. @a3626a is this okay with you?

Of course. Thank you

yuvipanda:
Thanks a lot, @consideRatio and @a3626a!!!

PR #2694, opened 2022-05-16
ci: add automation to update chp and traefik images

### LINKED ISSUE #2419: ci: automated updates of traefik and chp images
We sometimes update CHP and traefik as external images, and this can be automated!

The strategy for automating these can be like this.

1. Identify a latest version somehow
   This strategy needs to be reliable. The latest git tag in a repo part of the default branch may work, but it can absolutely fail. Git tags can be made on different branches missing commits, or backported tags can be newer in date on separate branches.
2. Verify if the docker image is available
   Sometimes a docker image is slow to publish after the git tag has been created. This is especially true for traefik, so before creating a PR, we should check that the image is actually available still.

consideRatio:
Closes #2419 and tested to work in a fork already.

I've added a note about the need to close/reopen PRs created like this until we have setup a personal access token to a bot account, as that is required to trigger tests automatically.

### DIFF
--- a/.github/workflows/watch-image-dependencies.yaml
+++ b/.github/workflows/watch-image-dependencies.yaml
@@ -1,0 +1,89 @@
+# This is a GitHub workflow defining a set of jobs with a set of steps.
+# ref: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
+#
+# - Watch the proxy.chp.image.tag field of values.yaml to match the latest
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
LGTM. Something that may interest you, I am currently working on generalising this action so we can bump repo2docker in 2i2c BinderHub charts :) https://github.com/sgibson91/bump-jhub-image-action

consideRatio:
Ah thanks linking out and working on these automizations @sgibson91, and thanks for the review!

PR #2713, opened 2022-05-18
Add singleuser.allowPrivilegeEscalation for KubeSpawner 2+

consideRatio:
I could not push to #2663 so I'm opening this PR to complement it with required parts. Closes #2663, ping @snickell!

---

I'm a bit unhappy about adding another dedicated passthrough config to the helm chart, but I'm also afraid of the work to champion the use of hub.config.KubeSpawner at this point. I'm afraid of the effort to clarify what can and can't be set there without trouble. Some configuration has been used both by helm chart templates, and by the jupyterhub_config.py file, and that config must still be read from the dedicated config values etc.


yuvipanda:
Thanks, @consideRatio!

Issue #2651, opened 2022-04-01
Not able to unset fsGid & containerSecurityContext values to let the (k8s) platform configure this automatically.


BlueCog:
### Proposed change
Because my team is deploying zero-to-jupyterhub-k8s to a OpenShift platform i'd would like to be able to unset **fsGid** and **containerSecurityContext** values for all deployments/containers. So that OpenShift can define them automatically in the range that is configured at the platform level. 

You would think that setting a `null` value would unset the value. But this is not possible wilt Helm v3 at the moment (when using nesterd charts). Background: https://github.com/helm/helm/issues/9136. A proposed solution would be to se an _if statement_ in the templating. 

Examples in code:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/hub/deployment.yaml#L82
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/hub/deployment.yaml#L176
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/proxy/deployment.yaml#L170

### Alternative options
- Downloading the zero-to-jupyterhub Helm template and editing it myself: I want to avoid this
- Defining values in the allowed OpenShift range: i'd to like to let OpenShift decide this (what if the range changes?)

### Who would use this feature?
At least every OpenShift user or any other platform what needs to set these ID's automatically

### (Optional): Suggest a solution

Adding some templating to be able to enable or disable fsGid & containerSecurityContext values. 

Proposed Helm solution:

Templating (hub as example):
```
{{- if .Values.hub.fsGid.enabled }}
      securityContext:
        fsGroup: {{ .Values.hub.fsGid.id }}
{{- end }}  
 
{{- if .Values.hub.containerSecurityContext.enabled }}
{{- with .Values.hub.containerSecurityContext }}
       securityContext:
        {{- . | toYaml | nindent 12 }}
{{- end }}
{{- end }}
 
```
Override:
```
hub:
  fsGid: 
     id: 
     enabled: false
  containerSecurityContext:
    runAsUser: 
    runAsGroup: 
    allowPrivilegeEscalation: 
    enabled: false
```

consideRatio:
```yaml
# this configuration will do the trick i think
hub:
  podSecurityContext:
  containerSecurityContext:
```

consideRatio:
[...]

PR #2716, opened 2022-05-20
Add the securityContext.enabled value. Will be useful for openshift deployment

nicolasiltis:
Trying to fix issue  : Background: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2651
## Proposed change
Because my team is deploying zero-to-jupyterhub-k8s to a OpenShift platform i'd would like to be able to unset fsGid and containerSecurityContext values for all deployments/containers. So that OpenShift can define them automatically in the range that is configured at the platform level.

You would think that setting a null value would unset the value. But this is not possible wilt Helm v3 at the moment (when using nesterd charts). Background: https://github.com/helm/helm/issues/9136. A proposed solution would be to se an if statement in the templating.

Examples in code:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/hub/deployment.yaml#L82
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/hub/deployment.yaml#L176
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/proxy/deployment.yaml#L170

## Alternative options
Downloading the zero-to-jupyterhub Helm template and editing it myself: I want to avoid this
Defining values in the allowed OpenShift range: i'd to like to let OpenShift decide this (what if the range changes?)
Who would use this feature?
At least every OpenShift user or any other platform what needs to set these ID's automatically

## My implementation

Adding some templating to be able to enable or disable containerSecurityContext values.

Proposed Helm solution:
```
{{- if .Values.securityContext.enabled }}
      securityContext:
        fsGroup: {{ .Values.hub.fsGid.id }}
{{- end }}  
 
{{- if .Values.securityContext.enabled }}
{{- with .Values.hub.containerSecurityContext }}
       securityContext:
        {{- . | toYaml | nindent 12 }}
{{- end }}
{{- end }}
 ```

With default value to true to avoid regression:
```
# securityContext will allow you to turn off all the securityContext (useful for the deployment on openshift)
securityContext:
  enabled: true
```

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -78,6 +78,17 @@ properties:
     description: |
       See the documentation under [`fullnameOverride`](schema_fullnameOverride).
 
+  securityContext:
+    type: object
+    required: [enabled]
+    description: |
+      This is the configuration to enable the securityContext for all the workloads. Useful for openshift.
+    properties:
+        enabled:
+          type: boolean
+          description: |
+            This is the configuration to enable the securityContext for all the workloads. Useful for openshift.
+
   imagePullSecret:
     type: object
     required: [create]
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -84,10 +84,10 @@ properties:
     description: |
       This is the configuration to enable the securityContext for all the workloads. Useful for openshift.
     properties:
-        enabled:
-          type: boolean
-          description: |
-            This is the configuration to enable the securityContext for all the workloads. Useful for openshift.
+      enabled:
+        type: boolean
+        description: |
+          This is the configuration to enable the securityContext for all the workloads. Useful for openshift.
 
   imagePullSecret:
     type: object
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -78,8 +78,10 @@ spec:
       {{- if .Values.rbac.enabled }}
       serviceAccountName: {{ include "jupyterhub.hub.fullname" . }}
       {{- end }}
+      {{- if .Values.securityContext.enabled }}
       securityContext:
         fsGroup: {{ .Values.hub.fsGid }}
+      {{- end }}
       {{- with include "jupyterhub.imagePullSecrets" (dict "root" . "image" .Values.hub.image) }}
       imagePullSecrets: {{ . }}
       {{- end }}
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by consideRatio** [comment]
@nicolasiltis I'd like to better understand in detail the requirement you have to deploy this on openshift. Is it that any pod and container `securityContext` must be undeclared, or is it that certain keys within it must be undeclared?

Would this cause trouble?

```yaml
- containers:
  - securityContext: {}
```

Would this cause trouble?

```yaml
- containers:
  - securityContext:
```

Would the default setup of openshift have troubles like this, or is it a specific setup of openshift that does things like this?

delamart:
Hi @consideRatio OpenShift sets securityContext values by default and they are restrictive by default. Therefore you usually dont' set any values in an OpenShift helm chart at all as they will be set automatically. If you do set something it's because you need it (special fs permissions, host network etc...) but then you also need admin privileges in the cluster to allow these overrides. This is why it's easiest to leave it empty if the pod/application supports it.

delamart:
@consideRatio and to answer your original question :) it looks like setting  
```
- containers:
  - securityContext: {}
 ``` 
works in openshift (same as not setting securityContext at all)

consideRatio:
Thanks for providing some details @delamart, I'd want very crisp clarity about what this workaround is about if its to be approved. I think overall, applying workarounds of various kinds quickly becomes unsustainable with regards to long term maintenance.

What `works in openshift` is vague to me overall. I'd like to have references to what openshift does and why this workaround is relevant in quite clear detail. If openshift has documentation about modifying, overwriting, etc, that is relevant to link.

manics:
@consideRatio There have been quite a few posts on Discourse about running Z2JH with openshift.

Would it be helpful to put a pause on workarounds, and instead focus on getting Z2JH on openshift tested in a GitHub workflow?
A bit of googling came up with:
- https://github.com/minishift/minishift
- https://github.com/openshift/microshift
- https://github.com/code-ready/crc

Perhaps this is something the community could help with?

consideRatio:
## Action point towards resolving this PR

- [ ] Please provide more background to motivate and document this change request. Practically by making it very clear what this workaround is meant to handle in detail within openshift, without assuming knowledge about openshift. It will include summarizing how openshift manages both pod securityContexts and container securityContexts, also linking out to openshift documentation about it.
- [ ] Validate that the consequences of no longer setting `fsGroup` in the `hub` pod in openshift, when using local storage for the database (default behavior), still will work and not lead to file access permissions errors for JupyterHub.

consideRatio (changes requested):
## Action points towards resolving this PR

- [ ] Please provide more background to motivate and document this change request. Practically by making it very clear what this workaround is meant to handle in detail within openshift, without assuming knowledge about openshift. It will include summarizing how openshift manages both pod securityContexts and container securityContexts, also linking out to openshift documentation about it.
- [ ] Validate that the consequences of no longer setting `fsGroup` in the `hub` pod in openshift, when using local storage for the database (default behavior), still will work and not lead to file access permissions errors for JupyterHub.

consideRatio:
With #2718, I'll go for a close on this PR and start working on resolving #2718.

PR #2717, opened 2022-05-22
Add opengraph tags

manics:
Jupyter readthedocs sites are often referenced on the [Jupyter Discourse](https://discourse.jupyter.org/). The lack of Open Graph tags means link previews aren't useful, e.g. in the compose mode
image

This uses [sphinxext-opengraph](https://github.com/wpilibsuite/sphinxext-opengraph) as suggested by @choldgraf in https://github.com/pydata/pydata-sphinx-theme/pull/664

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -13,4 +13,5 @@ pydata-sphinx-theme
 pyyaml
 sphinx-autobuild
 sphinx-copybutton
+sphinxext-opengraph
 sphinxext-rediraffe
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -102,6 +102,7 @@ extensions = [
     "sphinx_copybutton",
     "myst_parser",
     "sphinxext.rediraffe",
+    "sphinxext.opengraph",
 ]
 
 # List of patterns, relative to source directory, that match files and
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -173,6 +173,10 @@ rediraffe_redirects = {
     "advanced": "administrator/advanced",
 }
 
+# opengraph configuration
+# ogp_site_url/prefix is set automatically by RTD
+ogp_image = "_static/logo.png"
+ogp_use_first_image = True
 
 # -- Generate the Helm chart configuration reference from a schema file ------
 


consideRatio:
@manics wieee nice!! I observe it to not yet function though based on the PR preview. I did this on discourse.jupyter.org:

image

choldgraf:
agree w/ @consideRatio - we need to add a bit of configuration to get the base url right etc, and then the link metadata / logo should show up properly

manics:
I've added some tags for `og:image`. I didn't change the site URL because according to the docs [RTD will set it automatically](https://github.com/wpilibsuite/sphinxext-opengraph#options), and it may need to include the site prefix `en/latest/` which may cause problems with multiple versions of the docs.

https://cards-dev.twitter.com/validator refuses to load the rendered RTD pages because bots are blocked by `robots.txt`, maybe Discourse does the same?

manics:
The other option mentioned in https://github.com/pydata/pydata-sphinx-theme/pull/664#issuecomment-1125962191
>  define a jupyterhub-sphinx-theme that pre-configures some things for all of our repositories. For example, we could automatically add a link to our Discourse forum, Jupyter home page, etc (and each repository could then additionally configure it on top of that).

If we can include
```py
ogp_image = "_static/logo.png"
ogp_use_first_image = True
```
in `jupyterhub-sphinx-theme` and all JupyterHub RTD sites have `_static/logo.png` this would provide a drop-in zero config theme. We could also pin the version of `pydata-sphinx-theme` exactly but leave `jupyterhub-sphinx-theme` unpinned in all repos, so we have full central control over the base theme if there are `pydata-sphinx-theme` bugs.

consideRatio:
@manics I think you are right!

See the difference between the preview robots.txt https://zero-to-jupyterhub.org.readthedocs.build/robots.txt vs the production robots.txt https://z2jh.jupyter.org/robots.txt. robots.txt is blocked in the preview setting, but not for RTD production pages not meant for preview.

It makes sense, they should stop scrapers etc from finding preview pages. I think things may work if we merge this :)

choldgraf:
@manics if we define a jupyterhub theme we could also consider defining some extensions that are automatically installed in all of our repositories (e.g. `sphinx-copybutton`) or some theme styles to standardize (e.g. the primary color could be jupyter orange). Another option would be to use a single-page theme like the [Sphinx Book Theme](sphinx-book-theme.readthedocs.io) and then define a **topbar** in the jupyterhub theme that is shared between all projects. This is what [the Dask project does](https://docs.dask.org) as well as [the 2i2c docs](https://docs.2i2c.org).

consideRatio:
image

Things seem to work overall, there is an issue with a reference to favicon.ico but other than that, things work!

PR #2720, opened 2022-05-23
breaking: add hub.podSecurityContext, remove hub.fsGid

### LINKED ISSUE #2718: Deprecate `hub.fsGid` in favor of `hub.podSecurityContext`
This helm chart provides various pods, which can have a `securityContext` (PodSecurityContext), along with containers, which can have `securityContext` (ContainerSecurityContext).

Typically, we only set these dictionaries directly to values passed, but not for the hub pod's podsecuritycontext.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c98ba583be42636868edfbcb7ca7c7eca9a590b0/jupyterhub/templates/hub/deployment.yaml#L78-L85

By deprecating this setting of fsGroups, in favor of declaring it via a hub.podSecurityContext instead, we can avoid such exception. With such consistency, I think there is a better solution than accepting #2716 available for us that is easier to communicate and understand.

For #2716 to help not render securityContext's, one could simply instead create a values file like this to disable them for the hub pod for example.

```yaml
hub:
  podSecurityContext: null
  containerSecurityContext: null
```

## Related

- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2716

consideRatio:
This PR is motivated in #2718.

Closes #2718.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -873,14 +873,15 @@ properties:
       fsGid:
         type: [integer, "null"]
         minimum: 0
-        description:
-          The gid the hub process should be using when touching any volumes mounted.
-
-          Use this only if you are building your own image & know that a group with this gid
-          exists inside the hub container! Advanced feature, handle with care!
-
-          Defaults to 1000, which is the gid of the `jovyan` user that is present in the
-          default hub image.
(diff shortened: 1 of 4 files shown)


yuvipanda:
Thanks @consideRatio!

PR #2727, opened 2022-05-28
Fix for PDBs in k8s 1.20

### LINKED ISSUE #2649: PodDisruptionBudget not using the correct API Version in Kubernetes v1.20
### Bug description
When attempting to deploy a helm chart to the cluster, both of the default PDB resources cannot deploy giving the error:

`server could not find the requested resource`

Having a look at the helm chart itself:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/5c0d0c5c74bc97bf403844ced6b8e79cc089498d/jupyterhub/templates/scheduling/user-scheduler/pdb.yaml#L2 

It looks like this is checking if only `policy/v1` exists on the cluster - which I believe it does - however PDB v1 does not. Should this not be `policy/v1/PodDisruptionBudget`? As this does not exist on K8s version 1.20 and therefore would default to `policy/v1beta1`.

I have tried deploying both versions manually and only `policy/v1beta1` works

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  labels:
    app: jupyterhub
```

### Your personal set up
 - Kubernetes Version:
v1.20
 - Version(s):
Jupyterhub Helm Chart: 1.2.0

consideRatio:
Closes #2649 as reported by @geoffo-dev, where policy/v1 seems to be available in the k8s cluster event though policy/v1 PodDisruptionBudgets only come around in k8s 1.21.

I was on track to get this done across other repos, and went along and did it also for this repo. The actual work was not making the PR, it was to realize there is a bug like this. Thank you @geoffo-dev!

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,4 +1,4 @@
-# dependabot.yml reference: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates
+# dependabot.yml reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 #
 # Notes:
 # - Status and logs from dependabot are provided at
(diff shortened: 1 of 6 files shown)


geoffo-dev:
@consideRatio - apologies I never got around to creating a PR for this! Flagged the email to do it and seen that you already had!

Thanks! Much appreciated 😄

PR #2730, opened 2022-05-29
maint: add pre-commit isort hook, and let pyupgrade assume py38+ in hub container

consideRatio:
While this suddenly assumes the hub pod has py38+ and it could be seen as a breaking change, I don't think we should write about this in the changelog, as that version is considered internal.

What is a breaking change worth mentioning though is a transition from Python 3.8 to Python 3.9 within the hub image, see #2733.


yuvipanda:
Thanks @consideRatio!

PR #1185, opened 2019-03-11
Switch to openssl backend for pycurl

joshbode:
Builds `pycurl` as a requirement with `openssl` backend

Resolves #1102

PR #2733, opened 2022-05-29
breaking: hub image ubuntu->debian, `build-essential` removed, `--build-arg PIP_OVERRIDES=...` removed, images/hub/dependencies removed

### LINKED ISSUE #2731: hub.image currenlty based on ubuntu:20.04 - switch to python:3.9-slim-bullseye (Debian 11)?
`binderhub` use a `python:3.9-slim-bullseye` image, and I've come to find a few benefits of that making me think that perhaps we should make a switch here as well for our hub image.

1. We would get Python pre-installed in our base image (ubuntu:20.04 / ubuntu:22.04 images doesn't have it pre-installed), and choose a Python version like `3.9` instead of Py38 or Py310 that are available via apt in ubuntu:20.04 and ubuntu:22.04 respectively.
2. I think we can get rid of the hub/image/dependencies script as done in binderhub https://github.com/jupyterhub/binderhub/pull/1433 easier because the base image has python pre-installed, as we can then just start a base-image container and install `pip install pip-tools && pip-compile ...`.
3. ubuntu and debian are similar, so we hopefully won't break things for end users.

Overall though, I lack experience of Debian, and have only used Ubuntu on my local computer. Maybe @minrk @yuvipanda has some linux experience of relevance?

consideRatio:
I've trialed a switch from `ubuntu:20.04` to `python:3.9-slim-bullseye` and think this is a breaking change we should do. Closes #2731. Some of the changes in this PR are changes that could be done with `ubuntu:20.04` as well.

### Image size

Using [`dive`](https://github.com/wagoodman/dive#readme):

- Old image: 623 MB
- New image: 437 MB (70% of old image)

### About removed apt packages

- `libpq-dev` doesn't have to be installed any more as there is a pip wheel for aarch64 now making it not needed.
- `build-essential` isn't needed, but removing it makes people who install packages later that require `gcc` etc need to install it.
- `python3*` is removed as the base image provides python for us.

### About pycurl apt package dependencies

It seems that we can get openssl as a backend for `pycurl` (introduced in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1185) by referencing `libcurl4` without needing to install `libssl-dev` and `libcurl4-openssl-dev`.

If I test to run `python -c "import pycurl; print(pycurl.version_info())"` in the 1.2.0 image and compare with the output from this new Dockerfile we see this.

```shell
# pycurl.version_info() from this Dockerfile
(7, '7.74.0', 477696, 'x86_64-pc-linux-gnu', 29345693, 'OpenSSL/1.1.1n', 0, '1.2.11', ('dict', 'file', 'ftp', 'ftps', 'gopher', 'http', 'https', 'imap', 'imaps', 'ldap', 'ldaps', 'mqtt', 'pop3', 'pop3s', 'rtmp', 'rtsp', 'scp', 'sftp', 'smb', 'smbs', 'smtp', 'smtps', 'telnet', 'tftp'), None, 0, '2.3.0')
# pycurl.version_info() from the hub image version 1.2.0
(5, '7.68.0', 476160, 'x86_64-pc-linux-gnu', 12568477, 'OpenSSL/1.1.1f', 0, '1.2.11', ('dict', 'file', 'ftp', 'ftps', 'gopher', 'http', 'https', 'imap', 'imaps', 'ldap', 'ldaps', 'pop3', 'pop3s', 'rtmp', 'rtsp', 'scp', 'sftp', 'smb', 'smbs', 'smtp', 'smtps', 'telnet', 'tftp'), None, 0, '2.2.0')
```

### About removed --build-arg `PIP_OVERRIDES`

This was introduced by me in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1948. I think it should be removed now. We have requirements.txt to allow anyone pin whatever version if the repo is cloned the repo, and at worst someone would have to do a `sed` replace operation to the in the requirements.txt file or append a line if they have automation to make their own build of the hub image.

### Above removed `images/hub/dependencies` script

This script was removed for binderhub in https://github.com/jupyterhub/binderhub/pull/1433, but not for z2jh because of how we passed `--build-arg` with `PIP_OVERRIDES` something we stopped with.

The script is no longer needed, but instead we can simply do this and we are done. This script works both locally and in a GitHub CI environment as well, which makes us able to use automation to update this.

```shell
cd images/hub
docker run --rm \
    --env=CUSTOM_COMPILE_COMMAND="see README.md" \
    --volume=$PWD:/io \
    --workdir=/io \
    --user=root \
    python:3.9-bullseye \
    sh -c 'pip install pip-tools==6.* && pip-compile --upgrade'
```


minrk (approved):
Very nice!

yuvipanda:
yay! I also appreciate the addition of tini.

PR #2734, opened 2022-05-30
feat: add KubeSpawner configuration singleuser.k8sApiRequestTimeout

desaintmartin:
On some clusters, keeping the default of 3 seconds makes it really hard to start a notebook. Allows to customize it while keeping the default of 3s.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -387,6 +387,7 @@ singleuser:
     pullPolicy:
     pullSecrets: []
   startTimeout: 300
+  k8sApiRequestTimeout: 3
   cpu:
     limit:
     guarantee:
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -137,6 +137,7 @@ set_config_if_not_none(
 for trait, cfg_key in (
     ("pod_name_template", None),
     ("start_timeout", None),
+    ("k8s_api_request_timeout", None),
     ("image_pull_policy", "image.pullPolicy"),
     # ('image_pull_secrets', 'image.pullSecrets'), # Managed manually below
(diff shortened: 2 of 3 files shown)


jupyterhub/values.yaml
```diff
@@ -387,6 +387,7 @@ singleuser:
     pullPolicy:
     pullSecrets: []
   startTimeout: 300
+  k8sApiRequestTimeout: 3
```
consideRatio, line 390:
```suggestion
  k8sApiRequestTimeout:
```

consideRatio: I want kubespawners default value to be respected as long as we don't see a reason for this Helm chart to override it. I think that improves maintainability.

consideRatio:
I've hestitated a bit regarding this, because to duplicate everything kubespawner provides is a somewhat unsustainable pattern. One can use `hub.config.KubeSpawner.k8s_api_request_timeout` as well currently.

Do you think we should add this? I think long term it should be deprecated in favor of `hub.config.KubeSpawner` but its been a bit of work and complexity to do that where it is possible.

desaintmartin:
I completely agree with you: I am all in favor of delegating this to KubeSpawner except maybe the most needed options, indeed. Cons: loosing validation, Pros: much more maintainability. Let's close this, then. Thanks for your time reviewing this!

consideRatio:
> Edit: let me find the time in the next few days to improve the schema / values, it was not obvious at first read that it worked like this.

That would be great! Note that `hub.config` is a feature not well documented or advertised so far. I've been a bit worried about how it can't easily replace all configuration, for example `hub.baseUrl` as is described more about [in a note](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-config) etc, but I'd love to see more adoption of it.

Issue #2691, opened 2022-05-10
Allow provisioning service account and not rbac resources


dingobar:
### Proposed change

This somewhat relates to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2680.

It would be nice to split the creation of `Role` and `RoleBinding` out from `ServiceAccount`. In our cluster, we provision some default roles for each namespace and don't allow users to create their own roles. The reason for this is to keep complexity somewhat in check and try to standardize a bit across all namespaces.

The problem as I see it: Currently, the service account is always named `hub` afaict, and if `rbac.enabled=false` no account is created at all. This is blocking me from using the chart.

I think it is intuitive that the setting `rbac.enable` would only have an effect on the actual rbac API (currently `rbac.authorization.k8s.io/v1`), ie. just `Role` and `RoleBinding` in this case. We could introduce `hub.serviceAccount.create` separate from `rbac.enable`. This pattern is the one used in [apache/airflow](https://github.com/apache/airflow/blob/main/chart/templates/rbac/pod-launcher-role.yaml), and I find it quite intuitive. It would be a breaking change, though. We could expand this pattern to all service accounts, all defaulting to `true`.

Alternatively, it could be possible to let `hub.serviceAccount.create` overrule `rbac.enabled` so that if the former is set and true the `ServiceAccount` resource is created no matter what the value of `rbac.enabled` is. The default could be `null` or `""`  so that if `hub.serviceAccount.create` is not set, the old behavior with `rbac.enabled` determining the creation of `ServiceAccount` resources is used. I don't like this approach so much because it adds complexity on behalf of backwards compatibility which is probably not something we want to do too much of. Most users will pin the chart version too, if running production workloads.

Alternative three is to allow setting a service account name, something like `hub.serviceAccount.name`, that is heeded no matter the value of `rbac.enabled`, allowing a 3rd-party provisioned `ServiceAccount` with an arbitrary name.

~~In any case, currently setting `rbac.enabled=false` only supports a very specific pattern where a service account named `hub` must be manually created on the side.~~

What do you think?

**I don't mind taking a stab at implementing the a change here!**

### Alternative options
I want to avoid having to fork this repository to use the chart. I want to avoid having to manually deploy jupyterhub without helm.

### Who would use this feature?
Users of kubernetes clusters where the creation of `rbac` resources is limited but `ServiceAccount` resources are user-managed.

consideRatio:
[...]

PR #2736, opened 2022-06-02
feat: Implement `serviceAccount.create`

dingobar:
This feature introduces a new `.serviceAccount.create` for all
`ServiceAccount` resources, which defaults to `true`. This is a breaking
change for all users with `rbac.enabled` set to `false`, as these users will get some ServiceAccount resources created where they previously ran as the default user.

For context, please see the discussion at https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2691 .

The use-case for this feature is to allow users who use `rbac.enabled=false` to provision `serviceAccounts` without rbac resources like `Role` and `RoleBinding`. It also limits the `rbac.enabled` to only refer to resources that relate to the actual rbac API in kubernetes.

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1259,6 +1259,10 @@ properties:
           Configuration for a k8s ServiceAccount dedicated for use by the
           specific pod which this configuration is nested under.
         properties:
+          create:
+            type: boolean
+            description: |
+              Whether or not to create the `ServiceAccount` resource
           annotations:
             type: object
             additionalProperties: false
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1263,6 +1263,14 @@ properties:
             type: boolean
             description: |
               Whether or not to create the `ServiceAccount` resource
+          name:
+            type: ["string", "null"]
+            description: |
+              This configuration serves multiple purposes:
+              - It will be the `serviceAccountName` referenced by related Pods.
+              - If `create` is set, the created ServiceAccount resource will be named like this.
+              - If [`rbac.enabled`](schema_rbac.enabled) is set, the associated (Cluster)RoleBindings will bind to this name.
+              If not explicitly provided, a default name will be used.
           annotations:
             type: object
             additionalProperties: false
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -78,10 +78,10 @@
 
 {{- /* hub-serviceaccount ServiceAccount */}}
 {{- define "jupyterhub.hub-serviceaccount.fullname" -}}
-    {{- if .Values.hub.serviceAccount.name }}
-        {{- .Values.hub.serviceAccount.name }}
-    {{- else if .Values.hub.serviceAccount.create }}
-        {{- include "jupyterhub.hub.fullname" . }}
+    {{- if .Values.hub.serviceAccount.create }}
+        {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" . ) }}
+    {{- else }}
+        {{- .Values.hub.serviceAccount.name | default "default"}}
     {{- end }}
 {{- end }}
 
@@ -144,10 +144,10 @@
 
 {{- /* autohttps-serviceaccount ServiceAccount */}}
 {{- define "jupyterhub.autohttps-serviceaccount.fullname" -}}
-    {{- if .Values.proxy.traefik.serviceAccount.name }}
-        {{- .Values.proxy.traefik.serviceAccount.name }}
-    {{- else if .Values.proxy.traefik.serviceAccount.create }}
-        {{- include "jupyterhub.autohttps.fullname" . }}
+    {{- if .Values.proxy.traefik.serviceAccount.create }}
+        {{- .Values.proxy.traefik.serviceAccount.name | default (include "jupyterhub.autohttps.fullname" . ) }}
+    {{- else }}
+        {{- .Values.proxy.traefik.serviceAccount.name | default "default"}}
     {{- end }}
 {{- end }}
 
@@ -158,10 +158,10 @@
 
 {{- /* user-scheduler-serviceaccount ServiceAccount */}}
 {{- define "jupyterhub.user-scheduler-serviceaccount.fullname" -}}
-    {{- if .Values.scheduling.userScheduler.serviceAccount.name }}
-        {{- .Values.scheduling.userScheduler.serviceAccount.name }}
-    {{- else if .Values.scheduling.userScheduler.serviceAccount.create }}
-        {{- include "jupyterhub.user-scheduler-deploy.fullname" . }}
+    {{- if .Values.scheduling.userScheduler.serviceAccount.create }}
+        {{- .Values.scheduling.userScheduler.serviceAccount.name | default (include "jupyterhub.user-scheduler-deploy.fullname" . ) }}
+    {{- else }}
+        {{- .Values.scheduling.userScheduler.serviceAccount.name | default "default"}}
     {{- end }}
 {{- end }}
 
@@ -182,10 +182,10 @@
 
 {{- /* image-awaiter-serviceaccount ServiceAccount */}}
 {{- define "jupyterhub.hook-image-awaiter-serviceaccount.fullname" -}}
-    {{- if .Values.prePuller.hook.serviceAccount.name }}
-        {{- .Values.prePuller.hook.serviceAccount.name }}
-    {{- else if .Values.prePuller.hook.serviceAccount.create }}
-        {{- include "jupyterhub.hook-image-awaiter.fullname" . }}
+    {{- if .Values.prePuller.hook.serviceAccount.create }}
+        {{- .Values.prePuller.hook.serviceAccount.name | default (include "jupyterhub.hook-image-awaiter.fullname" . ) }}
+    {{- else }}
+        {{- .Values.prePuller.hook.serviceAccount.name | default "default"}}
     {{- end }}
 {{- end }}
 
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -76,6 +76,11 @@
     {{- include "jupyterhub.fullname.dash" . }}hub
 {{- end }}
 
+{{- /* hub-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hub-serviceaccount.fullname" -}}
+    {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" .) }}
+{{- end }}
+
 {{- /* hub-existing-secret Secret */}}
 {{- define "jupyterhub.hub-existing-secret.fullname" -}}
     {{- /* A hack to avoid issues from invoking this from a parent Helm chart. */}}
@@ -133,9 +138,19 @@
     {{- include "jupyterhub.fullname.dash" . }}autohttps
 {{- end }}
 
+{{- /* autohttps-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.autohttps-serviceaccount.fullname" -}}
+    {{- .Values.proxy.traefik.serviceAccount.name | default (include "jupyterhub.autohttps.fullname" .) }}
+{{- end }}
+
 {{- /* user-scheduler Deployment */}}
 {{- define "jupyterhub.user-scheduler-deploy.fullname" -}}
     {{- include "jupyterhub.fullname.dash" . }}user-scheduler
+{{- end }}
+
+{{- /* user-scheduler-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.user-scheduler-serviceaccount.fullname" -}}
+    {{- .Values.scheduling.userScheduler.serviceAccount.name | default (include "jupyterhub.user-scheduler-deploy.fullname" .) }}
 {{- end }}
 
 {{- /* user-scheduler leader election lock resource */}}
@@ -151,6 +166,11 @@
 {{- /* image-awaiter Job */}}
 {{- define "jupyterhub.hook-image-awaiter.fullname" -}}
     {{- include "jupyterhub.fullname.dash" . }}hook-image-awaiter
+{{- end }}
+
+{{- /* image-awaiter-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hook-image-awaiter-serviceaccount.fullname" -}}
+    {{- .Values.prePuller.hook.serviceAccount.name | default (include "jupyterhub.hook-image-awaiter.fullname" .) }}
 {{- end }}
 
 {{- /* hook-image-puller DaemonSet */}}
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -276,6 +276,7 @@
 fullname: {{ include "jupyterhub.fullname" . | quote }}
 fullname-dash: {{ include "jupyterhub.fullname.dash" . | quote }}
 hub: {{ include "jupyterhub.hub.fullname" . | quote }}
+hub-serviceaccount: {{ include "jupyterhub.hub-serviceaccount.fullname" . | quote }}
 hub-existing-secret: {{ include "jupyterhub.hub-existing-secret.fullname" . | quote }}
 hub-existing-secret-or-default: {{ include "jupyterhub.hub-existing-secret-or-default.fullname" . | quote }}
 hub-pvc: {{ include "jupyterhub.hub-pvc.fullname" . | quote }}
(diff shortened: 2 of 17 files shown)


dingobar:
So seems the CI pipeline adds some commits, but I think a squash merge of this PR should fix that. Please let me know if you'd prefer I rebase+force push the branch with the auto-fixes.

consideRatio:
Wiee hey @dingobar, thanks for working on this topic!

You can pull and manually fixup/squash the formatting commits if you want, but you can also leave them in. I use pre-commit locally to ensure my commits are formatted locally overall, for notes on how to do that see the notes in the .pre-commit-config.yaml file.

I've added a lengthier note about the kind of changes I think makes sense to bundle in a single PR in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2691#issuecomment-1144625118. What do you think about expanding this PR a bit to include what's described there?

jupyterhub/templates/proxy/autohttps/rbac.yaml
```diff
@@ -1,6 +1,17 @@
 {{- $HTTPS := (and .Values.proxy.https.hosts .Values.proxy.https.enabled) }}
 {{- $autoHTTPS := (and $HTTPS (eq .Values.proxy.https.type "letsencrypt")) }}
```
consideRatio, line 2:
```suggestion
{{- $HTTPS := (and .Values.proxy.https.hosts .Values.proxy.https.enabled) -}}
{{- $autoHTTPS := (and $HTTPS (eq .Values.proxy.https.type "letsencrypt")) -}}
```

jupyterhub/templates/proxy/autohttps/rbac.yaml
```diff
@@ -1,6 +1,17 @@
...
-{{- if (and $autoHTTPS .Values.rbac.enabled) -}}
+{{- if $autoHTTPS -}}
+{{- if .Values.proxy.traefik.serviceAccount.create }}
```
consideRatio, line 4: The `-` to the right and left of the `{{` and ` }}` are related to "whitespace chomping".

A rule of thumb to avoid rendering surplus new lines etc we've tried to follow is:
1. Always whitespace chomp left
2. Never whitespace chomp right, _except for_ when it's before any previous content has been rendered in the template, then always whitespace chomp right as well.

For this line, we should also whitespace chomp right to follow that rule. It is not going to make things break etc, but it makes the output contain less surplus new lines when rendering it.

jupyterhub/schema.yaml
```diff
@@ -1259,6 +1259,14 @@ properties:
           Configuration for a k8s ServiceAccount dedicated for use by the
           specific pod which this configuration is nested under.
         properties:
+          create:
+            type: boolean
```
desaintmartin, line 1263:
```suggestion
            type: boolean
            additionalProperties: false
```

desaintmartin (commented):
We collided on this one, I've just made https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2737 that has just been closed, here are my comments from the work I've done.

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -75,8 +75,8 @@ spec:
           persistentVolumeClaim:
             claimName: {{ include "jupyterhub.hub-pvc.fullname" . }}
         {{- end }}
-      {{- if .Values.rbac.enabled }}
-      serviceAccountName: {{ include "jupyterhub.hub.fullname" . }}
+      {{- if .Values.hub.serviceAccount.create }}
```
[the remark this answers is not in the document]
desaintmartin: @consideRatio I would thus amend your Change Request 1 to unconditionnaly set the serviceAccountName. If `rbac.enable: false` and `serviceAccount.create: false` and `serviceAccount.name:` then it will use `default` anyway, which is how it seems to be done in helm boilerplate.

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -76,6 +76,11 @@
     {{- include "jupyterhub.fullname.dash" . }}hub
 {{- end }}
 
+{{- /* hub-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hub-serviceaccount.fullname" -}}
+    {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" .) }}
```
desaintmartin, line 81:
```suggestion
{{- if .Values.hub.serviceAccount.create -}}
    {{ default (include "jupyterhub.hub.fullname" .) .Values.hub.serviceAccount.name }}
{{- else -}}
    {{ default "default" .Values.hub.serviceAccount.name }}
{{- end -}}
```

As it is done in helm boilerplate?

jupyterhub/schema.yaml
```diff
@@ -1259,6 +1259,14 @@ properties:
...
+            description: |
+              Whether or not to create the `ServiceAccount` resource
+          name:
+            type: ["string", "null"]
+            description: |
+              Name of the `ServiceAccount` resource, if created (if `null`, a default name is used)
```
consideRatio, line 1269:
```suggestion
              This configuration serves multiple purposes:
              - It will be the `serviceAccountName` referenced by related Pods.
              - If `create` is set, the created ServiceAccount resource will be named like this.
              - If [`rbac.enabled`](schema_rbac.enabled) is set, the associated (Cluster)RoleBindings will bind to this name.
              If not explicitly provided, a default name will be used.
```

jupyterhub/schema.yaml
dingobar: I didn't add that because it's not there for any other booleans.

jupyterhub/schema.yaml
```diff
@@ -1259,6 +1259,14 @@ properties:
           Configuration for a k8s ServiceAccount dedicated for use by the
           specific pod which this configuration is nested under.
         properties:
+          create:
+            type: boolean
+            description: |
+              Whether or not to create the `ServiceAccount` resource
+          name:
+            type: ["string", "null"]
```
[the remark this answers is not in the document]
dingobar: I didn't add that because it's not there for any other nullable strings.

jupyterhub/templates/hub/deployment.yaml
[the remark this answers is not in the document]
dingobar: I believe this case is already covered by CR1 though? If no name is given, and create=false, it won't reference any serviceAccount, thus just default to whatever the cluster default is, right? If the user wants to run as `default` even if the cluster default is something else, they can specify that in the name. Or am I missing something? There is some nested logic going on here and I'm happy it's no deeper than in is to be honest :sweat_smile:

jupyterhub/schema.yaml
dingobar: I've rebased this into the first commit where I changed the schema

dingobar:
> Thanks for documenting your work so clearly @dingobar!!
> ## Various config combinations seems ok
> 
> I'm considering various combinations of these configurations. I was a bit surprised about the idea of creating a SA with a specified name, but yeah - that is probably reasonable. Those that wants to specify an existing SA, which is something I think people have wanted historically, can would then set `create=false` and `name=something`.
> 
>     1. `rbac.enabled=true`
>        We create the Role and RoleBinding, so we need to be able to reference a SA, otherwise we have a guaranteed failure.
>        
>        
>        
>        create
>        name
>        note
>        
>        
>        
>        
>        `false`
>        `something`
>        User creates SA with specified name
>        
>        
>        `false`
>        `null`
>        Broken configuration - RoleBindings have no SA to point to
>        
>        
>        `true`
>        `something`
>        Chart creates SA with specified name
>        
>        
>        `true`
>        `null`
>        Chart creates SA with default name
> 
>     2. `rbac.enabled=false`
>        
>        
>        
>        create
>        name
>        note
>        
>        
>        
>        
>        `false`
>        `something`
>        User creates SA with specified name
>        
>        
>        `false`
>        `null`
>        Possibly not broken as this is what you would have pre-rbac permissions were around, but it is probably not a configuration people have these days.
>        
>        
>        `true`
>        `something`
>        Chart creates SA with specified name
>        
>        
>        `true`
>        `null`
>        Chart creates SA with default name
> 
> 
> These options seems fine to me. I considered if we should add logic to error loudly if this is configured to the broken configuration state, but I think such validation logic in the schema.yaml file, or having a conditional `fail` command in NOTES.txt would be a bit messy to do for all service accounts.
> ## Change request 1
> 
> Users will need to be allowed to specify a SA without the helm chart creating one. Due to that, I suggest that the following changes are made.
> 
> ```diff
> # from pod specifications where we specify serviceAccountName
> -      {{- if .Values.hub.serviceAccount.create }}
> -      serviceAccountName: {{ include "jupyterhub.hub-serviceaccount.fullname" . }}
> +      {{- with include "jupyterhub.hub-serviceaccount.fullname" . }}
> +      serviceAccountName: {{ . }}
>        {{- end }}
> ```
> 
> ```diff
> # from _helpers-names.tpl
>  {{- define "jupyterhub.hub-serviceaccount.fullname" -}}
> -    {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" .) }}
> +   {{- if .Values.hub.serviceAccount.name }}
> +   {{- .Values.hub.serviceAccount.name }}
> +   {{- else if .Values.hub.serviceAccount.create }}
> +   {{- include "jupyterhub.hub.fullname" . }}
> +   {{- end }}
>  {{- end }}
> ```
> 
> ## Change request 2
> 
> There is a section you can find by searching for `{{- define "jupyterhub.name-templates" -}}` that should have entries for all resources. Update that list with the new `.fullname` named templates you have created.
> ## Change request 3
> 
> All resources in this helm chart skip the leading `---` in the file besides a few of these you now add. Remove those that doesn't act as separators between resources.

I believe these changes are implemented now. @desaintmartin I'm not able to re-request a review from you but feel free.

jupyterhub/templates/hub/deployment.yaml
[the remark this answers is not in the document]
desaintmartin: Honestly I would just do the same way as the whole community is doing instead of having specific logic for this Chart only !

jupyterhub/templates/hub/deployment.yaml
```diff
@@ -75,8 +75,8 @@ spec:
           persistentVolumeClaim:
             claimName: {{ include "jupyterhub.hub-pvc.fullname" . }}
         {{- end }}
-      {{- if .Values.rbac.enabled }}
-      serviceAccountName: {{ include "jupyterhub.hub.fullname" . }}
+      {{- with include "jupyterhub.hub-serviceaccount.fullname" . }}
+      serviceAccountName: {{ . }}
       {{- end }}
```
consideRatio, line 80: Right now with this logic, we will provide a serviceAccountName if `.name` is specified or `.create` is set, but not if `.name` isn't set and `.create` isn't set.

We are all in agreement that we want to be allowed to set serviceAccountName if `.name` is specified, no matter if `.create` is set which we now do. Is https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2736#discussion_r888137126 also a request for something additional besides this?

jupyterhub/templates/hub/deployment.yaml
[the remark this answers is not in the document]
consideRatio: Closing this as resolved, but asked a followup question about this in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2736/files#r888795016 where we can look at the latest code.

jupyterhub/templates/_helpers-names.tpl
consideRatio: This change includes two parts of relevance:
1. serviceAccountName is always set, even when `.create` is false and `.name` is unspecified.
2. Style change, writing `default <default value> <primary value>` instead of `<primary value> | default <default value>`.
3. Style change, whitespace chomping practices

I don't want to accept the style changes as they are consistent across the Helm chart in a manner, but I think its acceptable to agree on the change 1 as it reduces the complexity of our template logic, making it easier to read.

@dingobar I'll let you decide about 1 as you also would be the person to do the work and I find both options fine.

consideRatio: Hmmm, I'm inclined to agree, favoring implementing part 1 here.

jupyterhub/templates/image-puller/serviceaccount.yaml
```diff
@@ -0,0 +1,21 @@
+{{- /*
+ServiceAccount for the pre-puller hook
```
consideRatio, line 2:
```suggestion
ServiceAccount for the pre-puller hook's image-awaiter-job
```

dingobar: Rebased it into a previous commit (the one where `serviceaccount.yaml` was created)

jupyterhub/templates/_helpers-names.tpl
dingobar: While I see the original functionality good and easy to understand, I buy the argument that since the default helm-chart base also sets "default" as the serviceAccount in the pods we should also do it.

dingobar:
I've included the last CRs now. Hopefully we can merge this now.

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -133,11 +142,29 @@
     {{- include "jupyterhub.fullname.dash" . }}autohttps
 {{- end }}
 
+{{- /* autohttps-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.autohttps-serviceaccount.fullname" -}}
+    {{- if .Values.proxy.traefik.serviceAccount.create }}
+        {{- .Values.proxy.traefik.serviceAccount.name | default (include "jupyterhub.autohttps.fullname" . ) }}
+    {{- else }}
+        {{- .Values.proxy.traefik.serviceAccount.name | default "default"}}
```
consideRatio, line 150:
```suggestion
        {{- .Values.proxy.traefik.serviceAccount.name | default "default" }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -133,11 +142,29 @@
...
 {{- define "jupyterhub.user-scheduler-deploy.fullname" -}}
     {{- include "jupyterhub.fullname.dash" . }}user-scheduler
 {{- end }}
 
+{{- /* user-scheduler-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.user-scheduler-serviceaccount.fullname" -}}
+    {{- if .Values.scheduling.userScheduler.serviceAccount.create }}
+        {{- .Values.scheduling.userScheduler.serviceAccount.name | default (include "jupyterhub.user-scheduler-deploy.fullname" . ) }}
+    {{- else }}
+        {{- .Values.scheduling.userScheduler.serviceAccount.name | default "default"}}
```
consideRatio, line 164:
```suggestion
        {{- .Values.scheduling.userScheduler.serviceAccount.name | default "default" }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -153,6 +180,15 @@
     {{- include "jupyterhub.fullname.dash" . }}hook-image-awaiter
 {{- end }}
 
+{{- /* image-awaiter-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hook-image-awaiter-serviceaccount.fullname" -}}
+    {{- if .Values.prePuller.hook.serviceAccount.create }}
+        {{- .Values.prePuller.hook.serviceAccount.name | default (include "jupyterhub.hook-image-awaiter.fullname" . ) }}
+    {{- else }}
+        {{- .Values.prePuller.hook.serviceAccount.name | default "default"}}
```
consideRatio, line 188:
```suggestion
        {{- .Values.prePuller.hook.serviceAccount.name | default "default" }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -76,6 +76,15 @@
     {{- include "jupyterhub.fullname.dash" . }}hub
 {{- end }}
 
+{{- /* hub-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hub-serviceaccount.fullname" -}}
+    {{- if .Values.hub.serviceAccount.create }}
+        {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" . ) }}
+    {{- else }}
+        {{- .Values.hub.serviceAccount.name | default "default"}}
```
consideRatio, line 84:
```suggestion
        {{- .Values.hub.serviceAccount.name | default "default" }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -76,6 +76,15 @@
     {{- include "jupyterhub.fullname.dash" . }}hub
 {{- end }}
 
+{{- /* hub-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hub-serviceaccount.fullname" -}}
+    {{- if .Values.hub.serviceAccount.create }}
+        {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" . ) }}
```
consideRatio, line 82:
```suggestion
        {{- .Values.hub.serviceAccount.name | default (include "jupyterhub.hub.fullname" .) }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -133,11 +142,29 @@
     {{- include "jupyterhub.fullname.dash" . }}autohttps
 {{- end }}
 
+{{- /* autohttps-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.autohttps-serviceaccount.fullname" -}}
+    {{- if .Values.proxy.traefik.serviceAccount.create }}
+        {{- .Values.proxy.traefik.serviceAccount.name | default (include "jupyterhub.autohttps.fullname" . ) }}
```
consideRatio, line 148:
```suggestion
        {{- .Values.proxy.traefik.serviceAccount.name | default (include "jupyterhub.autohttps.fullname" .) }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -133,11 +142,29 @@
...
+
 {{- /* user-scheduler Deployment */}}
 {{- define "jupyterhub.user-scheduler-deploy.fullname" -}}
     {{- include "jupyterhub.fullname.dash" . }}user-scheduler
 {{- end }}
 
+{{- /* user-scheduler-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.user-scheduler-serviceaccount.fullname" -}}
+    {{- if .Values.scheduling.userScheduler.serviceAccount.create }}
+        {{- .Values.scheduling.userScheduler.serviceAccount.name | default (include "jupyterhub.user-scheduler-deploy.fullname" . ) }}
```
consideRatio, line 162:
```suggestion
        {{- .Values.scheduling.userScheduler.serviceAccount.name | default (include "jupyterhub.user-scheduler-deploy.fullname" .) }}
```

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -153,6 +180,15 @@
     {{- include "jupyterhub.fullname.dash" . }}hook-image-awaiter
 {{- end }}
 
+{{- /* image-awaiter-serviceaccount ServiceAccount */}}
+{{- define "jupyterhub.hook-image-awaiter-serviceaccount.fullname" -}}
+    {{- if .Values.prePuller.hook.serviceAccount.create }}
+        {{- .Values.prePuller.hook.serviceAccount.name | default (include "jupyterhub.hook-image-awaiter.fullname" . ) }}
```
consideRatio, line 186:
```suggestion
        {{- .Values.prePuller.hook.serviceAccount.name | default (include "jupyterhub.hook-image-awaiter.fullname" .) }}
```

consideRatio (approved):

consideRatio:
Thank you @dingobar and @desaintmartin for your work on this!!!!!! :heart: :tada: :sunflower:

Issue #2680, opened 2022-05-05
Handling of ServiceAccount resources


consideRatio:
## Overview: the chart's ServiceAccount resources

These ServiceAccount resources are currently defined and created by the Helm chart. They are granted permissions by (Cluster)Role resources coupled to the ServiceAccounts via (Cluster)RoleBinding resources. They are declared to be used by the associated pods

ServiceAccount | Role type | Permissions
-|-|-
hub | Role | KubeSpawner runs in the hub pod and require permissions to work with pods etc.
image-puller | Role | The hook-image-awaiter pod requires `get` permissions of DaemonSet resources to conclude if the hook-image-puller pod's are ready.
autohttps | Role | Traefik running in this pod doing TLS termination requires certificates etc. We save them in a Secret resource, so this service account is used by an initContainer and additional container to initially load a certificate from a k8s Secret resource and sync it if updated over time.
user-scheduler | ClusterRole | `kube-scheduler` is configured for us to pack user pods on nodes tightly, but to do that it requires significant permissions related to the scheduling of pods on nodes. These permissions are ClusterRole based.

### Rename `rbac.enabled` to `rbac.create`

_Feature status: not considered worth doing as its a breaking change or introduces complexity._

> According to [Helm docs](https://docs.helm.sh/chart_best_practices/#rbac-resources-should-be-created-by-default), the best practice is to have `rbac.create` for the setting to create RBAC resources.

This would be a breaking change or add the complexity of supporting both. So far, we have opted to not make this change.

- #1797
- #921

### Let Role/RoleBinding's reference another ServiceAccount

_Feature status: plausible to implement, the first step is to open an issue outlining the suggested configuration proposal. A key piece to clarify is what to do with the ServiceAccount's associated Role/RoleBinding resources as well, should they still be created etc?_

Especially ClusterRole and ClusterRoleBinding can be troublesome for someone to acquire by their k8s admin if they have permissions in a specific namespace. When they do acquire these permissions, it could be with a custom name etc. Allowing the users to opt out creating service accounts and manually specifying them for various pods could be relevant. If we do this, let's do it for all the service account for consistency though.

- #616
- #670
- #2691

### Create ServiceAccount for _all pods_

_Feature status: not to be implemented unless motivated clearly in a dedicated issue to still be relevant given recent deprecation._

[...]

PR #2739, opened 2022-06-03
breaking, maint: replace rbac.enabled with rbac.create

consideRatio:
To use `rbac.create` instead of `rbac.enabled` has been a common suggestion over time as that is how most other Helm charts have named this configuration, so like this we will better align with the common practice.

This is a followup PR to #2736 that changed the behavior of `rbac.enabled` in a way that motivated some breaking change documentation in the changelog. As part of that, I figured it could be motivated to also transition `rbac.enabled` to `rbac.create` and force users having declared `rbac.enabled` to adjust to the change.

Previously, `rbac.enabled` was handling both the creation of Role/RoleBinding and ServiceAccount resources, while after #2736 it is only managing the Role/RoleBinding resources. With this PR, users that has configured `rbac.enabled` explicitly, either true or false, will see the following error on template rendering.

```
helm template jupyterhub jupyterhub --set rbac.enabled=false
Error: execution error at (jupyterhub/templates/NOTES.txt:152:4): 

#################################################################################
######   BREAKING: The config values passed contained no longer accepted    #####
######             options. See the messages below for more details.        #####
######                                                                      #####
######             To verify your updated config is accepted, you can use   #####
######             the `helm template` command.                             #####
#################################################################################

CHANGED: rbac.enabled must as of version 2.0.0 be configured via rbac.create and <hub|proxy.traefik|scheduling.userScheduler|prePuller.hook>.serviceAccount.create.
```

## Related
- #2736
- #2680

### DIFF
--- a/docs/source/administrator/security.md
+++ b/docs/source/administrator/security.md
@@ -205,7 +205,7 @@ If you want to disable the RBAC rules, for whatever reason, you can do so with t
 
 ```yaml
 rbac:
-  enabled: false
+  create: false
 ```
 
 We strongly **discourage disabling** the RBAC rules and remind you that this
--- a/docs/source/administrator/security.md
+++ b/docs/source/administrator/security.md
@@ -197,7 +197,7 @@ kubectl --namespace=kube-system delete rc kubernetes-dashboard
 Kubernetes supports, and often requires, using [Role Based Access Control (RBAC)](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)
 to secure which pods / users can perform what kinds of actions on the cluster. RBAC rules can be set to provide users with minimal necessary access based on their administrative needs.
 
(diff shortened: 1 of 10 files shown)


consideRatio:
@dingobar and @desaintmartin who have helped out with the recent related changes, do you think this is a reasonable change to do, and does the change technically seem sound?

yuvipanda (approved):
This only affects people who explciitly want to handle their own RBAC setup, right? No changes for folks just using the defaults?

consideRatio:
> No changes for folks just using the defaults?

Yepp, I think this only affect people that has explicitly set a value either true or false. Well I figure it could also affect the daskhub helm chart that does a seldom used [`import-values` thingy](https://github.com/dask/helm-chart/blob/bb5b91f1354e075f72936d978219215490518f8d/daskhub/Chart.yaml#L7-L13), maybe.

yuvipanda:
Thanks a lot @consideRatio

dingobar:
Seems I missed this, but LGTM! :)

PR #2740, opened 2022-06-04
Fix link to authentication guide

jdmcbr:
The current link results in a 404. It looks like is just missing the `.md` extension.

### DIFF
--- a/docs/source/jupyterhub/customizing/user-management.md
+++ b/docs/source/jupyterhub/customizing/user-management.md
@@ -114,4 +114,4 @@ hub:
 ## Authenticating Users
 
 For information on authenticating users in JupyterHub, see
-[the Authentication guide](../../administrator/authentication).
+[the Authentication guide](../../administrator/authentication.md).


consideRatio:
.md isnt needed with sphinx i think. So, maybe the link is broken for another reason?

You can test in the readthedocs preview of the website if your patch works, you find it via the tests run for this pr.

jdmcbr:
Ah, I see. The sphinx version does indeed work. I came across the github hosted markdown when looking for tips on configuring authorization (I'm trying to get a j2zh working locally, and running into auth issues): https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/docs/source/jupyterhub/customizing/user-management.md

When clicking the link to the more detailed guide on authenticating users from the version above, the lack of the `.md` extension produces a 404 error. It doesn't look like the addition of the `.md` causes any issue in the sphinx output, and fixes the navigation of documentation within github.

consideRatio:
@jdmcbr let's go for a merge then :) There will be quite a few broken links in a non-sphinx-html based version that can't be fixed in the same way though, so we can't probably optimize this much further.

It would be good if we could have a way to help users find their way to z2jh.jupyter.org if they end up reading these docs though, hmmm...

jdmcbr:
@consideRatio Thanks, it honestly should have occurred to me to try the rendered docs anyway, I've spent quite a bit of time going through those recently. I'd assumed that the markdown ought to mostly work on github too though, good to know that isn't generally the case. I don't have any great ideas on directing people to navigate through z2jh.jupyter.org myself, but I'll keep that in mind as I continue going through the docs. 

Also, thanks for all your work on this project! :pray:

PR #2762, opened 2022-06-17
Add scheduling.userPlaceholder.annotations

joncotton:
This is a generic pattern to allow user annotations on the resulting pods.

The specific use case I have is adding `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"`. A snippet from the [cluster-autoscaler docs](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node):

>    What types of pods can prevent CA from removing a node?
>    - Pods with local storage. *
>    *Unless the pod has the following annotation (supported in CA 1.0.3 or later): "cluster-autoscaler.kubernetes.io/safe-to-evict": "true"

and what I see in the autoscaler logs:
```
Fast evaluation: node ip-10-10-10-10.us-west-2.compute.internal cannot be removed: pod with local storage present: user-placeholder-0
```

### Your personal set up
- EKS 1.22
- cluster-autoscaler 1.22.2
- Zero-to-JupyterHub Chart 1.2.0

My running environment is a separate node pool for the notebook pods. I have a node running with only the `user-placeholder` Pod (and DaemonSet pods, but they don’t count. they evict just fine.) _and_ there’s room on another notebook node for the placeholder. I want the placeholder pod to be rescheduled onto the in-use node and then the autoscaler can tear down the empty node.

### DIFF
--- a/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
+++ b/jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
@@ -23,6 +23,10 @@ spec:
   serviceName: {{ include "jupyterhub.user-placeholder.fullname" . }}
   template:
     metadata:
+      {{- if .Values.scheduling.userPlaceholder.annotations }}
+      annotations:
+        {{- toYaml .Values.scheduling.userPlaceholder.annotations | nindent 8 }}
+      {{- end }}
       labels:
         {{- /* Changes here will cause the Deployment to restart the pods. */}}
         {{- include "jupyterhub.matchLabels" . | nindent 8 }}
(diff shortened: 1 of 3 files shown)


jupyterhub/templates/scheduling/user-placeholder/statefulset.yaml
```diff
@@ -23,6 +23,10 @@ spec:
   serviceName: {{ include "jupyterhub.user-placeholder.fullname" . }}
   template:
     metadata:
+      {{- if .Values.scheduling.userPlaceholder.annotations }}
+      annotations:
+        {{- toYaml .Values.scheduling.userPlaceholder.annotations | nindent 8 }}
+      {{- end }}
```
consideRatio, line 29:
```suggestion
      {{- with .Values.scheduling.userPlaceholder.annotations }}
      annotations:
        {{- . | toYaml | nindent 8 }}
      {{- end }}
```

consideRatio:
Thank you @joncotton!

joncotton:
Yay! Thank you!

PR #2770, opened 2022-06-26
docs: transition glossyar to myst

consideRatio:


consideRatio:
I'll go for a self-merge on this as it is mostly a technical change only related to documentation and I think the changelog for 2.0.0 can be finalized now.

PR #2528, opened 2021-12-27
docs: add changelog for 2.0.0 (beta.1)

consideRatio:
Updated and ready for review. We will use JupyterHub 3.0.0b1 for z2jh 2.0.0-beta.1

PR #2779, opened 2022-07-03
2.0.0 upgrade guide

manics:
The Z2JH changelog is already very long, and the upgrade details in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2528 make it even longer.

I think moving the details to a separate page and keeping the changelog concise is more useful- people who want technical details can go through the changelog, administrators can go straight to the upgrade page without wading through a large changelog.

### DIFF
--- a/docs/source/administrator/upgrade-1-to-2.md
+++ b/docs/source/administrator/upgrade-1-to-2.md
@@ -1,0 +1,142 @@
+# Major upgrade: 1.\* to 2.\*
+
+Z2JH 2 contains several breaking changes, including some that affect the security of your deployment.
+This guide will help you upgrade from 1.\* to 2.\*.
+
+## Security: breaking change to `*.networkPolicy.egress`
+
+NetworkPolicy egress rules have been extended with a new property.
+If you have configured any of:
+
+- `hub.networkPolicy.egress`
+- `proxy.chp.networkPolicy.egress`
+- `proxy.traefik.networkPolicy.egress`
+- `singleuser.networkPolicy.egress`
+
+you must review your configuration as additional default egress routes have been added.
+Previously `...networkPolicy.egress` controlled all egress but a new properties `...networkPolicy.egressAllowRules` add additional egress rules by default.
+
+If you have configured `...networkPolicy.egress` for `hub`, `proxy.chp`,
+`proxy.traefik` or `singleuser` to restrict the permissions to establish
+outbound network connections, then this upgrade is likely to _escalate those
+permissions unless you revise your configuration_. The new configuration
+`...networkPolicy.egressAllowRules` are by default granting most of the egress
+permissions previously granted by default via the `...networkPolicy.egress`
+configuration, and `...networkPolicy.egress` are now by default not providing
+any permissions.
+
+If you for example had overridden the previously very permissive default value
+of `singleuser.networkPolicy.egress` to be less permissive, you should consider
+disabling all `singleuser.networkPolicy.egressAllowRules` like this
+to not risk escalating the permissions.
+
+```yaml
+singleuser:
+  networkPolicy:
+    egressAllowRules:
+      cloudMetadataServer: false
+      dnsPortsPrivateIPs: false
+      nonPrivateIPs: false
+      privateIPs: false
+```
+
+For more details, see the documentation on [Kubernetes Network Policies](netpol)
+and the configuration reference entries under
+[`...networkPolicy.egress`](schema_hub.networkPolicy.egress) and
+[`...networkPolicy.egressAllowRules`](schema_hub.networkPolicy.egressAllowRules).
+
+## JupyterHub 2 and related hub components
+
+Z2JH 2.0.0 upgrades JupyterHub to the 2.\* series, and also upgrades all hub components.
+If you are using any custom JupyterHub services, addons, API integrations, or extra configuration, you should review the breaking changes in the
+[JupyterHub 2.0.0 changelog](https://github.com/jupyterhub/jupyterhub/blob/2.3.1/docs/source/changelog.md#200).
+
+JupyterHub 2 uses an updated database schema.
+Z2JH 2.0.0 automatically handles the upgrade, but it will not be possible to downgrade to older releases after this.
+
+JupyterHub 2 adds RBAC for managing permissions in JupyterHub.
+The old permissions model of admin/non-admin still works but you should use
+[RBAC to assign the required privileges to users or services in future](https://jupyterhub.readthedocs.io/en/stable/rbac/index.html)
+
+See
+`TODO: link to Notable dependencies updated`
+in the changelog for more information on other upgraded hub components.
+
+## JupyterLab and Jupyter Server
+
+The default singleuser server is [JupyterLab](https://jupyterlab.readthedocs.io/), running on [Jupyter server](https://jupyter-server.readthedocs.io/en/latest/).
+To switch back to Jupyter Notebook either configure/rebuild your singleuser image to default to notebook, or see [the documentation on user interfaces](user-interfaces)
+
+## Default to using the container image's command instead of `jupyterhub-singleuser` [#2449](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2449)
+
+Z2JH now launches the container's default command (equivalent to setting `CMD` in a `Dockerfile`) instead of overriding it.
+This ensures that containers that use a custom start command to configure their environment, such as some
+[Jupyter Docker Stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/)
+images, will work without any changes.
+To restore the old behaviour set:
+
+```yaml
+singleuser:
+  cmd: jupyterhub-singleuser
+```
+
+## Configuration in `jupyterhub_config.d` has a higher priority than `hub.config` [#2457](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2457)
+
+Previously if `hub.config` was used to configure some JupyterHub traitlets it would override any custom configuration files mounted into `jupyterhub_config.d` in the hub container.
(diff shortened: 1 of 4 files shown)


docs/source/administrator/upgrading.md
```diff
@@ -1,5 +1,3 @@
-(upgrading)=
```
manics, line 1: This label has the same name as the page. When I tried to use it I got this warning:
```
WARNING: more than one target found for 'myst' cross-reference upgrading: could be :std:ref:`upgrade guide` or :std:doc:`upgrade guide`
```

docs/source/administrator/upgrading.md
```diff
@@ -100,19 +76,27 @@ will be performed automatically when you do a `helm upgrade`.
 4. Do a [`helm upgrade`](helm-upgrade-command). This should perform the database upgrade needed.
```
manics, line 76: TODO: test the above MySQL/PostgreSQL still works

docs/source/administrator/upgrade-1-to-2.md
```diff
@@ -0,0 +1,141 @@
...
+    fsGroup: GROUP-ID
+```
+
+## Hub image is based on Debian instead of Ubuntu [#2733](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2733)
+
+The hub container base image has switched from `ubuntu:20.04` to `python:3.9-slim-bullseye` which is based on `debian:bullseye-slim`.
+If you have extended the Z2JH hub image please review the [hub Dockerfile](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/2.0.0/images/hub/Dockerfile).
+Note the singleuser image is not affected.
+
+## Disabling RBAC requires setting multiple properties,`rbac.enable` is removed [#2736](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2736) [#2739](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2739)
```
consideRatio, line 117:
```suggestion
## Disabling RBAC requires setting multiple properties, `rbac.enable` is removed [#2736](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2736) [#2739](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2739)
```

docs/source/administrator/upgrading.md
consideRatio: I don't know, but I wonder how many uses we have of the link? The labels can be renamed etc freely without risk of breaking something I think.

docs/source/administrator/upgrading.md
consideRatio: @yuvipanda do you happen to know if this has worked for you when updating to a jupyterhub 2 versioned z2jh with a postgresql server?

I can test the upgrade part pre-release, for now, assume it works!

docs/source/administrator/upgrading.md
```diff
@@ -100,19 +76,27 @@ will be performed automatically when you do a `helm upgrade`.
...
 
 ```Dockerfile
-RUN pip install --no-cache-dir jupyterhub==0.8.1
+RUN pip install --no-cache-dir jupyterhub==2.3.1
 ```
 
+If you are using conda or mamba:
+
+```Dockerfile
+RUN conda install -y jupyterhub-base=2.3.1
```
consideRatio, line 94: Let's point to conda-forge, I'm not sure if we are available quickly in the default channel.

```suggestion
RUN conda install --channel=conda-forge -y jupyterhub-base=2.3.1
```

consideRatio (commented):
THANK YOU THIS IS GREAT!!!

It think that if you locate the upgrading-1-to-2.md file next to other files in the source/ folder, they will all be top level structured. Those top level entries will be very visible from menu items in the top menu - and having two associated with upgrading seems too much, maybe we can have a `upgrade` or `upgrading` folder, with an `index.md` representing general instructions etc, and linking out to specific files for specific version upgrades?

It would be good to know if you agree if that makes sense to you separate from if you are willing to do it, we can get this PR merged as it is and I can get it done after if so for example if you feel that it becomes scope creep for this PR etc.

Btw, in case you don't know, `make devenv` is quite nice to have when trialing these things - you will get rebuild-on-save and a browser that reloads-on-new-build.

docs/source/administrator/upgrading/index.md
```diff
@@ -1,38 +1,40 @@
...
-```{toctree}
-:maxdepth: 1
-:caption: Upgrade Guide
-
-upgrade-1-to-2
-```
+Z2JH follows [semantic versioning](https://semver.org/), with each version taking the form `MAJOR.MINOR.PATCH`.
+Minor and patch releases should be backwards compatible, and shouldn't require changes to your deployment.
+Review the [CHANGELOG](changelog) to find out about new features or bug fixes that affect your deployment,
+then follow [](helm-upgrade-command).
```
consideRatio, line 9: I'm not sure how the `[](helm-upgrade-command)` part will render, but it seems like a forgotten title for the link - or is sphinx doing some magic here?

manics: It's magic, it uses the subheading for that label.

manics:
I'm not completely happy with the upgrade docs.... though it's a start.

docs/source/administrator/upgrading/index.md
consideRatio: Ah nice!

consideRatio:
> I'm not completely happy with the upgrade docs.... though it's a start.

I'm super happy with this incremental improvenent!!!

Is this OK to merge at this point on your end? If so, I'll go for a readthrough again tomorrow morning and go for merge.

docs/source/administrator/upgrading.md
yuvipanda: I used SQLite everywhere, no postgres. I told believe it should just work.

docs/source/administrator/upgrading/upgrade-1-to-2.md
```diff
@@ -0,0 +1,141 @@
...
+[`*.networkPolicy.egressAllowRules`](schema_hub.networkPolicy.egressAllowRules).
+
+## JupyterHub 2 and related hub components
+
+Z2JH 2.0.0 upgrades JupyterHub to the 2.\* series, and also upgrades all hub components.
+If you are using any custom JupyterHub services, addons, API integrations, or extra configuration, you should review the breaking changes in the
+[JupyterHub 2.0.0 changelog](https://github.com/jupyterhub/jupyterhub/blob/2.3.1/docs/source/changelog.md#200).
+
+JupyterHub 2 uses an updated database schema.
+Z2JH 2.0.0 automatically handles the upgrade, but it will not be possible to downgrade to older releases after this.
```
consideRatio, line 55: This isn't true generally, hub.db.upgrade=true needs to be set for external databases like MySQL and PostgreSQL I think.

manics: Updated! Feel free to merge if you're happy, and we can refine in follow-up PRs

manics:
BTW, do you know _why_ we don't auto-upgrade other DB types?

docs/source/administrator/upgrading/upgrade-1-to-2.md
```diff
@@ -0,0 +1,141 @@
...
+NetworkPolicy egress rules have been extended with a new property.
+If you have configured any of:
+
+- `hub.networkPolicy.egress`
+- `proxy.chp.networkPolicy.egress`
+- `proxy.traefik.networkPolicy.egress`
+- `singleuser.networkPolicy.egress`
+
+you must review your configuration as additional default egress routes have been added.
+Previously `*.networkPolicy.egress` controlled all egress but a new properties `*.networkPolicy.egressAllowRules` add additional egress rules by default.
```
consideRatio, line 17:
```suggestion
Previously `*.networkPolicy.egress` controlled all egress but a new property `*.networkPolicy.egressAllowRules` add additional egress rules by default.
```

docs/source/administrator/upgrading/upgrade-1-to-2.md
```diff
@@ -0,0 +1,141 @@
...
+[`*.networkPolicy.egressAllowRules`](schema_hub.networkPolicy.egressAllowRules).
+
+## JupyterHub 2 and related hub components
+
+Z2JH 2.0.0 upgrades JupyterHub to the 2.\* series, and also upgrades all hub components.
+If you are using any custom JupyterHub services, addons, API integrations, or extra configuration, you should review the breaking changes in the
+[JupyterHub 2.0.0 changelog](https://github.com/jupyterhub/jupyterhub/blob/2.3.1/docs/source/changelog.md#200).
+
+JupyterHub 2 uses an updated database schema.
+Z2JH 2.0.0 automatically handles the upgrade for SQLite databases (the default), but it will not be possible to downgrade to older releases after this.
```
consideRatio, line 55:
```suggestion
Z2JH 2.0.0 automatically handles the upgrade for SQLite databases (the default), but if you use an external database you need to configure [`hub.db.upgrade`](schema_hub.db.upgrade) to true when upgrading. It will not be possible to downgrade to older releases after this without also using a backup or resetting the database.
```

consideRatio:
> BTW, do you know why we don't auto-upgrade other DB types?

I'm not sure, I recall that Yuvi wanted the default behavior to be simple, and the "production grade" setup with external database to be a more hands on aware decision based.

Actually, I'd argue that it makes sense for the default value to be `true`, and then let those wishing to opt out have it set to `false` by default if they wish to be very well aware of when and if this is to be done.

consideRatio (approved):
Thank you so much for helping drive this onwards Simon!!

I added [b747f55](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2779/commits/b747f55ab1adb1b8b6826f23e075aad93d78060d), but also want to get this merged to unblock possible other work referencing this etc from changelog so I'll go for a merge to unblock any further work.

I deeply appreciate your work on this, I find it super complicated to communicate this effectively and that a lot of this chore refactoring was needed that you now have addressed!

Issue #2783, opened 2022-07-04
Let `hub.db.upgrade` be True by default and reduce complexity?


consideRatio:
In https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2779#issuecomment-1173835935 I guessed in response to @manics on why we have a quite advanced logic associated with `hub.db.upgrade` configuration, where the unset value means it is true for `sqlite` databases (default), but false for external databases.

I want to consider if we should make the system easier to communicate about etc, and simply let `hub.db.upgrade` be true by default, and remove the in-between behavior state where unset meant you would have upgrades for sqlite but not for other dbs?

The key value as I see it, is that the behavior is simpler. With simpler behavior, we can communicate easier about this, people understand it better, and that can be a win-win for users and maintainers.

@manics @yuvipanda @minrk what do you think about this?

manics:
I think it makes sense. Sqlite is as much "production" as an external database. If we're not completely confident that the external DB schema upgrade works then we can't release Z2JH 2.0.0 regardless of this decision.

manics:
I'm trying to add a DB upgrade test for PostgreSQL https://github.com/manics/zero-to-jupyterhub-k8s/tree/test-postgres

minrk:
It doesn't have to do with our confidence about the schemas. All are pretty thoroughly tested in JupyterHub itself. I think it's true by default for sqlite because our own sqlite upgrade makes a backup as part of the process, so recovery should always be possible. We can't backup external databases as part of upgrade (users have to do this before upgrading jupyterhub), so a database upgrade has a better chance of being destructive _and unrecoverable_ with postgres, etc.. Because "backup, then upgrade" (what everyone should always do) necessarily requires out-of-band user actions, the potentially destructive action on requires 'confirmation' of a sort (what little we have available in helm).

That's what I recall about the reasoning behind it, anyway. I'm open to changing the default, since I suspect it is a pain and most folks probably set `hub.db.upgrade` to True as soon as they encounter this once, and then leave it for the next upgrade. I would _much_ prefer a regular "confirm you're ready to upgrade the database" prompt, if we could have one. As it is, I guess docs are all we have, but that's little help for folks who need to recover their database after an upgrade they want to roll back but can't.

PR #2785, opened 2022-07-04
Test postgres schema upgrade in CI

manics:
Adds a test of `hub.db.upgrade` for PostgreSQL. This requires pinning an old chart that requires a schema upgrade, in general stable will be too close to dev for there to be an upgrade.

Adds docs on why the upgrade isn't automatic for other DB provides.
Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2783

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -146,16 +146,18 @@ jobs:
               --set hub.config.CryptKeeper.keys[0]=cccc3333
             create-k8s-test-resources: true
 
-          # We run two upgrade tests where we first install an already released
-          # Helm chart version and then upgrades to the version we are now
-          # testing. We test upgrading from the latest stable version (like
-          # 1.2.3), and one from the latest dev version (like
-          # 1.2.3-n012.h1234abc).
+          # We run three upgrade tests where we first install an already released
+          # Helm chart version and then upgrade to the version we are now
+          # testing:
+          # - latest stable version (like 1.2.3)
+          # - latest dev version (like 1.2.3-n012.h1234abc),
+          # - and an old version that requires a JupyterHub DB upgrade
           #
           # It can be very useful to see the "Helm diff" step's output from the
           # latest dev version.
           #
-          # The upgrade-from input should match the version information from
+          # The upgrade-from input should be a chart version, or match the version
+          # information from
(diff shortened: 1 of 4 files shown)


.github/workflows/test-chart.yaml
```diff
@@ -237,11 +267,23 @@ jobs:
           timeout: 150
           max-restarts: 1
 
+      - name: "(Upgrade) Install PostgreSQL server chart"
+        if: matrix.setup-postgresql-args
+        run: |
+          . ./ci/common
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm install postgresql bitnami/postgresql ${{ matrix.setup-postgresql-args }}
```
consideRatio, line 275: I think `helm repo add` would also require `helm repo update`, which updates all declared helm repos. Better to do just `--repo=...` in the upgrade command to avoid a step like this I think.

```suggestion
          helm install postgresql postgresql --repo=https://charts.bitnami.com/bitnami ${{ matrix.setup-postgresql-args }}
```

consideRatio (commented):
Thank you for your thorough work @manics!

Currently, there is once again some issues with intermittency in our tests as caused by modern versions of calico and/or k3s - unclear. With one more test running, we will experience more intermittent failures.

Key question: do you think adding this makes sense? It is tested in jupyterhub/jupyterhub already, and all we do is provide a flag to jupyterhub during upgrades. Are we going to catch something new by doing this upgrade that isn't catched by jupyterhub/jupyterhub's test suite?

If you think this is the right thing, lets do it!

manics:
I think it's worth adding because the DB upgrade is a critical step (all other steps can be downgraded), and Z2JH is a production distribution so it's worth having the additional test. It's also good to have a test of an external DB instead of juts sqlite.

If three upgrade tests is too many we could remove either the `stable` or `dev` upgrade test?

consideRatio:
Let's have then all and re-evaluate as time goes if needed!

I think maybe removing a normal test makes most sense if we are to slim these down - the different k8s versions can be covered as long as we have 5 or so tests, no matter what is done within them, if they are running an upgrade first or not etc.

consideRatio:
Thank you @manics!!!!!

PR #2791, opened 2022-07-08
Add `labels` config for `scheduling.userScheduler`, `scheduling.userPlaceholder`, and `prePuller`

ruben-rodriguez:
- Enterprise run environments often enforce labeling policies that doesn't allow to run workloads with certain mandatory labels.
- This PR adds the ability to configure labels in the chart values to customize scheduler & puller components workloads in a similar fashion to the Hub pod.


consideRatio:
We support adding annotations for these, and should absolutely support labels alongside that. Thank you @ruben-rodriguez!!

PR #2800, opened 2022-07-12
ci: workaround intermittent test failures pending upstream fix in k3s 

### LINKED ISSUE #2799: Intermittent test failures
These are tracked in https://github.com/jupyterhub/action-k3s-helm/issues/59 and https://github.com/k3s-io/k3s/issues/5633#issuecomment-1181424511.

My conclusion is that I hope k3s finds and patches something, otherwise the action-k3s-helm repo should probably take action. I don't expect these kinds of failures to be indicative of any issues in the Helm chart or similar though, this is just an issue for users of k3s as a k8s cluster provider I think - and people using `kubectl exec` with those clusters as we do in our test suite.

consideRatio:
Closes #2799 via a workaround patch. https://github.com/jupyterhub/action-k3s-helm/issues/59 is opened still to track possible further actions, but I'd like to avoid adding a workaround in that repo unless its clear this isn't going to resolve in k3s or some further time passes without resolution.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -190,7 +190,7 @@ jobs:
             local-chart-extra-args: >-
               --set hub.db.type=sqlite-pvc
               --set singleuser.storage.type=dynamic
-          - k3s-channel: v1.24
+          - k3s-channel: v1.22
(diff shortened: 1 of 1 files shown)


consideRatio:
There was one test failure, but intermittently failing for another reason related to traefik and cert acquisition.

PR #2802, opened 2022-07-14
Add revisionHistoryLimit to all deployments

mgobec:
## Description

Users need to be able to control how many old ReplicaSets are retained. Some users prefer to keep minimum number or even zero to clean up clutter (ArgoCD for example) or if they use `helm rollback` instead of kubectl rollout undo to roll back.

## Changes

Added revision history limit property to all deployments and added default of 10 to values file.

### DIFF
--- a/jupyterhub/templates/proxy/autohttps/deployment.yaml
+++ b/jupyterhub/templates/proxy/autohttps/deployment.yaml
@@ -8,7 +8,7 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  {{- if not (eq nil .Values.proxy.traefik.revisionHistoryLimit) }}
+  {{- if typeIs "int" .Values.proxy.traefik.revisionHistoryLimit }}
   revisionHistoryLimit: {{ .Values.proxy.traefik.revisionHistoryLimit }}
   {{- end }}
   replicas: 1
--- a/jupyterhub/templates/proxy/autohttps/deployment.yaml
+++ b/jupyterhub/templates/proxy/autohttps/deployment.yaml
@@ -8,8 +8,8 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  {{- if not (eq nil .Values.proxy.revisionHistoryLimit) }}
-  revisionHistoryLimit: {{ .Values.proxy.revisionHistoryLimit }}
+  {{- if not (eq nil .Values.proxy.traefik.revisionHistoryLimit) }}
+  revisionHistoryLimit: {{ .Values.proxy.traefik.revisionHistoryLimit }}
   {{- end }}
   replicas: 1
   selector:
--- a/jupyterhub/templates/proxy/autohttps/deployment.yaml
+++ b/jupyterhub/templates/proxy/autohttps/deployment.yaml
@@ -8,6 +8,9 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
+  {{- if not (eq nil .Values.proxy.revisionHistoryLimit) }}
+  revisionHistoryLimit: {{ .Values.proxy.revisionHistoryLimit }}
+  {{- end }}
   replicas: 1
   selector:
     matchLabels:
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -7,6 +7,9 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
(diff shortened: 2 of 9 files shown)


consideRatio (changes requested):
If you need this, its reasonable to support in my mind! Thanks for contributing to the jupyterhub ecosystem!

Changes requested:
- Dont let templates render this unless explicitly set by user, allow a null and integer value in schema, default to null in values. Since 0 is a falsy value, use of "with" wont work in this case but checking for the value not being nil or similar would be required.
- in the schema file, use yaml anchors like for other entries in the schema to avoid repeating the same content.
- the schema description should be limking out to k8s docs rather than describing it, see how other schema entries have linked out to k8s docs

mgobec:
@consideRatio Thanks for the code comments. I updated the PR.

consideRatio (changes requested):
Thanks for iterating on this!

> Dont let templates render this unless explicitly set by user, allow a null and integer value in schema, default to null in values. Since 0 is a falsy value, use of "with" wont work in this case but checking for the value not being nil or similar would be required.

This is probably still not fixed, you can check if you got it fixed by using `helm template` to render the outputs and see if you render a passed `0` configuration correctly (`--set hub.revisionHistoryLimit=0`. (When doing this, you can use `--show-only` flag to avoid printing the entire helm chart and only focus on a deployment's template that you trial on.)

Please also review your PR from small formatting typos.

mgobec:
@consideRatio You are right! I completely forgot that zero isn't handled like you would normally expect. I was previously testing with values other than zero. Thank you for reviewing

jupyterhub/templates/proxy/autohttps/deployment.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
+  {{- if not (eq nil .Values.proxy.revisionHistoryLimit) }}
+  revisionHistoryLimit: {{ .Values.proxy.revisionHistoryLimit }}
+  {{- end}}
```
consideRatio, line 13:
```suggestion
  {{- end }}
```

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
+  {{- if not (eq nil .Values.proxy.revisionHistoryLimit) }}
+  revisionHistoryLimit: {{ .Values.proxy.revisionHistoryLimit }}
+  {{- end}}
```
consideRatio, line 12:
```suggestion
  {{- end }}
```

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
+  {{- if not (eq nil .Values.scheduling.userScheduler.revisionHistoryLimit) }}
+  revisionHistoryLimit: {{ .Values.scheduling.userScheduler.revisionHistoryLimit }}
+  {{- end}}
```
consideRatio, line 11:
```suggestion
  {{- end }}
```

consideRatio (commented):
This configuration is available for StatefulSet resources and DaemonSet resources that the helm chart provides as well. I presume you have a need for this as well then, right?

mgobec:
I'm not sure statefulset and deamonset support revisionhistorylimit. I have to check if the spec is the same as deployment and I'll add those.

consideRatio:
> I'm not sure statefulset and deamonset support revisionhistorylimit. I have to check if the spec is the same as deployment and I'll add those.

100%!

I checked here and found it to be the same and available for those as well, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/

consideRatio:
@mgobec i wanted to get this PR resolved so i pushed two commits that changed the following:

1. The configuration now applies to statefulset and daemonset resources this helm chart provides as well
2. Test values are passed to verify the use of this configuration leads to valid k8s resources when rendering the helm templates
via `helm template --validate`
3. Configuration `proxy.revisionHistoryLimit` split into `proxy.chp.revisionHistoryLimit` and `proxy.traefik.revisionHistoryLimit` as those are two separate deployment resources.
4. Use of the comparison operator `eq` with `nil` and `5` led to errors using `helm` versions 3.5-3.7 while it worked well in 3.8+. Since this helm chart works well with 3.5.0 I started using the `kindIs "int"` instead which works also with helm 3.5.0+.

If this LGTY, let's go for a merge!

consideRatio (approved):

mgobec:
LGTM

PR #2803, opened 2022-07-14
Port old fixes from #2528 to new upgrade guide

manics:
Copies the fixes to the breaking changes/upgrade guide from @minrk in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2528/ to the new upgrade guide


manics:
On second thoughts I'll merge `main` into https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2528 and update the upgrade doc there instead to avoid broken cross-links.

PR #2805, opened 2022-07-19
Set singleuser.args to jupyterhub-singleuser

yuvipanda:
After writing
https://github.com/pangeo-data/pangeo-docker-images/pull/355/files#diff-a77643b43a7be453fa8556937bf32b27907e152a10d4c693f3e7670c66a44378,
I would like to ask that we reconsider *requiring* entrypoint be set to
a specific thing (`jupyterhub-singleuser`) in the images users are
required to produce. This makes it quite difficult to produce images
that *work across multiple tools* without having to configure each of
those tools. In this case, we want to use the *exact same image* in both
binder and apache beam. Apache beam *requires* an entrypoint be
specified, while at least as of now binder does not - so I wrote a
script that uses heuristics to determine if the caller is apache beam or
something else, and change behavior accordingly. But if I want this
image to also work with JupyterHub in the future, I'll have to modify my
script to detect if JupyterHub is calling it, and exec
`jupyterhub-singleuser`. Now if dask also wants to use this image, I'd
need to have a code path for that as well. I think this strong coupling
causes a lot of unnecessary friction here.

Another point of confusion here is that `cmd` in helm / k8s is actually
*ENTRYPOINT* in docker, while `args` in helm / k8s is *CMD* in docker.
Earlier, we were setting `cmd` to `jupyterhub-singleuser`, which
overrode any entrypoint the docker builder had set up - causing the
issues we were trying to fix. However, now that KubeSpawner doesn't
actually use `args` (k8s / helm) in any meaningful fashion, can we set
the default `args` to `jupyterhub-singleuser`? This has a few
advantages:

1. *ENTRYPOINT* set in Dockerfile is *not* overriden by what we set in
   helm, which is the current problem. This goes away
2. All existing docker images continue to work without admins having to
   set extra config or the image having to be rebuilt.
3. The same image can be used in multiple places without having to be
   aware of who is calling it.

I think this is an approach to solving the original problem that is both
*easier to document*, *easier for our users (existing and new)* and
makes life a lot less difficult for people building images.

Re-implementation of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2449

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -395,6 +395,8 @@ singleuser:
     limits: {}
     guarantees: {}
   cmd:
+  args:
+    - jupyterhub-singleuser
   defaultUrl:
   extraPodConfig: {}
   profileList: []


yuvipanda:
I can investigate the test failures and try getting this to work if y'all think the general approach is workable, @manics @minrk @consideRatio

yuvipanda:
Here's the somewhat tortured test I had to write to validate my 'multi dispatch entrypoint' works ok https://github.com/pangeo-data/pangeo-docker-images/pull/355/files#diff-6027b5fe95c80f9fe0bcead4ae1288105dfee85005ad12d2aa07d58c0d2fa8d2R35

manics:
The aim of the original PR was to ensure images that run scripts with custom setup will work without changes. For example, the docker-stacks images run some hooks to customise the environment before running jupyterhub-singleuser:
https://discourse.jupyter.org/t/pyspark-library-is-missing-from-jupyter-pyspark-notebook-when-running-with-jupyterhub-zero-to-jupyterhub-k8s/8450/2

If we override the entrypoint or command this will break the docker-stacks images again:
https://github.com/jupyter/docker-stacks/blob/57ebe55c07d06df70f6fa1166bc5d8ce22b6da34/base-notebook/Dockerfile#L144

consideRatio:
### About docker-stacks images

Docker stacks relies on the startup scripts to do misc logic, but will call `jupyterhub-singleuser` if a `JUPYTERHUB_API_TOKEN` env var is found. This is done in three steps: [step 1 - start-notebook.sh](https://github.com/jupyter/docker-stacks/blob/57ebe55c07d06df70f6fa1166bc5d8ce22b6da34/base-notebook/start-notebook.sh#L11-L14), [step 2 - start-singleuser.sh](https://github.com/jupyter/docker-stacks/blob/57ebe55c07d06df70f6fa1166bc5d8ce22b6da34/base-notebook/start-singleuser.sh#L12-L13), [step 3 - start.sh](https://github.com/jupyter/docker-stacks/blob/57ebe55c07d06df70f6fa1166bc5d8ce22b6da34/base-notebook/start.sh#L260-L261).

From a docker-stacks perspective, such images should be started with the start-notebook.sh script if users also want to rely on various environment variable influencing those scripts. They can start like that either by default via their Dockerfile's declared ENTRYPOINT+CMD, or via z2jh's configuration of kubespawner, which in turn configures the k8s `args` (but not the k8s `command` aka. Dockerfile's ENTRYPOINT).

minrk:
I've lost track of where it was discussed recently, but as long as set-to-null in a user chart to opt-in to the image's custom command is easy and doesn't require replicating the command itself (the original motivation of #2138, I think), I'm happy. As @consideRatio mentioned, changing None to be the default in KubeSpawner seeems to have made this possible, so with this change (after fixing `singleuser.args` to `singleuser.cmd`), we have:

- default behavior: always works, matches other JupyterHub contexts
- custom image CMD (i.e. docker stacks) is easy and clear to opt-in to: `singleuser.cmd = null` (solves #2138 without breaking other images or changing default behavior)

This seems the best balance to me.

manics:
OK, let's revert the breaking change then. Can you update https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-environment.html#choose-and-use-an-existing-docker-image to include `singleuser.cmd: null` in the config?

PR #2807, opened 2022-07-22
Fix user-scheduler backward compatibility for AWS EKS

a3626a:
# Introduction

user-scheduler support both kubernetes before 1.20 and after 1.21. It is done by helm conditionals `if ge (atoi .Capabilities.KubeVersion.Minor) 21`

# Problem

`.Capabilities.KubeVersion.Minor` is not a simple number in AWS EKS environment. It is something like `22+`. So backward compatibility fallback is applied to every kubernetes clusters even if its version is greater than 1.20.

# Solution

I suggest simple solution just to replace every `+` signs into empty string `''`. Helm `replace` function can do this.

# Alternative

[`semverCompare`](https://helm.sh/docs/chart_template_guide/function_list/#semvercompare) can achieve better result, but I am worried about its unexpected behaviors. It changes things a lot.

# Appendix
Refer this image for various Helm expression results.
image

### DIFF
--- a/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
@@ -28,7 +28,7 @@ data:
        and configuration (v1beta1) of kube-scheduler.
   */}}
   config.yaml: |
-    {{- if ge (atoi .Capabilities.KubeVersion.Minor) 21 }}
+    {{- if ge (atoi ( .Capabilities.KubeVersion.Minor | replace "+" "" )) 21 }}
     apiVersion: kubescheduler.config.k8s.io/v1beta3
     kind: KubeSchedulerConfiguration
     leaderElection:
--- a/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
@@ -28,7 +28,14 @@ data:
        and configuration (v1beta1) of kube-scheduler.
   */}}
   config.yaml: |
-    {{- if ge (atoi ( .Capabilities.KubeVersion.Minor | replace "+" "" )) 21 }}
+    {{- /*
+      FIXME: We have added a workaround for EKS where
+            .Capabilities.KubeVersion.Minor can return a
+            string like "22+" instead of just "22".
+
+            See https://github.com/aws/eks-distro/issues/1128.
+    */}}
+    {{- if ge (atoi (.Capabilities.KubeVersion.Minor | replace "+" "")) 21 }}
     apiVersion: kubescheduler.config.k8s.io/v1beta3
     kind: KubeSchedulerConfiguration
     leaderElection:
--- a/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
+++ b/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
@@ -47,7 +47,7 @@ spec:
       {{- end }}
       containers:
         - name: kube-scheduler
-          {{- if ge (atoi .Capabilities.KubeVersion.Minor) 21 }}
+          {{- if ge (atoi ( .Capabilities.KubeVersion.Minor | replace "+" "" )) 21 }}
           image: {{ .Values.scheduling.userScheduler.image.name }}:{{ .Values.scheduling.userScheduler.image.tag }}
           {{- else }}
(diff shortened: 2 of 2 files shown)


consideRatio:
Thanks for the investigation and report!

I'd like that something like this is followed up properly in the root cause, and then if conclude it won't be enough, we consider a workaround.

In my mind, the issue must be in `helm` and/or AWS EKS that makes `helm` provide `.Capabilities.KubeVersion.Minor` be something that isn't a number.

I could reproduce this against a EKS cluster i manage, and I found...

```
.Capabilities.KubeVersion: v1.22.9-eks-a64ea69
.Capabilities.KubeVersion.Major: 1
.Capabilities.KubeVersion.Minor: 22+
```

Reading https://github.com/helm/helm/issues/4929#issuecomment-738826749 that indicates that `helm` just reports what is reported by EKS, I try `kubectl version` and finds...

```
Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.9-eks-a64ea69", GitCommit:"540410f9a2e24b7a2a870ebfacb3212744b5f878", GitTreeState:"clean", BuildDate:"2022-05-12T19:15:31Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

In other words, `helm` has just propegated the nonsense version response from `EKS` which is something they reasonably won't fix.

I check in GKE and find this response:

```
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.12-gke.1700", GitCommit:"6c11aec6ce32cf0d66a2631eed2eb49dd65c89f8", GitTreeState:"clean", BuildDate:"2022-05-13T09:31:25Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
```

I've now reported this upstream in https://github.com/aws/eks-distro/issues/1128.

---

I note also that the GitVersion field is `v1.22.9-eks-a64ea69`, and that is not a semver version. So, if we would try to parse that as a semver version, we would fail as well.

I think it is reasonable for us to keep using `.Capabilities.KubeVersion.Minor` and try to cleanup the junk manually then because I don't expect a quick enough response from EKS. I think though we should clarify that this is a workaround and link to the issue I opened. I want to avoid workarounds like this, but if they are added, that we have a clear path to removing them again when something upstream etc is resolved.

The kind of comment to be added should be a helm template comment.

jupyterhub/templates/scheduling/user-scheduler/configmap.yaml
```diff
@@ -28,7 +28,7 @@ data:
        and configuration (v1beta1) of kube-scheduler.
   */}}
   config.yaml: |
-    {{- if ge (atoi .Capabilities.KubeVersion.Minor) 21 }}
+    {{- if ge (atoi ( .Capabilities.KubeVersion.Minor | replace "+" "" )) 21 }}
```
consideRatio, line 31:
```suggestion
    {{- /*
      FIXME: We have added a workaround for EKS where
            .Capabilities.KubeVersion.Minor can return a
            string like "22+" instead of just "22".

            See https://github.com/aws/eks-distro/issues/1128.
    */}}
    {{- if ge (atoi (.Capabilities.KubeVersion.Minor | replace "+" "")) 21 }}
```

jupyterhub/templates/scheduling/user-scheduler/deployment.yaml
```diff
@@ -47,7 +47,7 @@ spec:
       {{- end }}
       containers:
         - name: kube-scheduler
-          {{- if ge (atoi .Capabilities.KubeVersion.Minor) 21 }}
+          {{- if ge (atoi ( .Capabilities.KubeVersion.Minor | replace "+" "" )) 21 }}
```
consideRatio, line 50:
```suggestion
          {{- /*
            FIXME: We have added a workaround for EKS where
                  .Capabilities.KubeVersion.Minor can return a
                  string like "22+" instead of just "22".

                  See https://github.com/aws/eks-distro/issues/1128.
          */}}
          {{- if ge (atoi (.Capabilities.KubeVersion.Minor | replace "+" "")) 21 }}
```

consideRatio (approved):

consideRatio:
With the suggested changes accepted, I'll go for a merge of this!

a3626a:
> With the suggested changes accepted, I'll go for a merge of this!

Accepted!

consideRatio:
Thank you @a3626a!!

PR #2810, opened 2022-07-23
Add breaking KubeSpawner changes to upgrade-1-to-2

manics:
Mainly https://github.com/jupyterhub/kubespawner/pull/545

### DIFF
--- a/docs/source/administrator/upgrading/upgrade-1-to-2.md
+++ b/docs/source/administrator/upgrading/upgrade-1-to-2.md
@@ -70,9 +70,10 @@ in the changelog for more information on other upgraded hub components.
 The default singleuser server is [JupyterLab](https://jupyterlab.readthedocs.io/), running on [Jupyter server](https://jupyter-server.readthedocs.io/en/latest/).
 To switch back to Jupyter Notebook either configure/rebuild your singleuser image to default to notebook, or see [the documentation on user interfaces](user-interfaces)
 
-## KubeSpawner disallows root users by default
+## KubeSpawner prevents privilege escalation such as sudo by default
 
-KubeSpawner will prevent processes executing as root in the singleuser container by default.
+By default processes cannot escalate their privileges.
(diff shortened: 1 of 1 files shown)


docs/source/administrator/upgrading/upgrade-1-to-2.md
```diff
@@ -67,6 +70,20 @@ in the changelog for more information on other upgraded hub components.
...
+
+KubeSpawner will prevent processes executing as root in the singleuser container by default.
+You must set `singleuser.allowPrivilegeEscalation: true` to enable root users.
+For example, if you are using a [docker-stacks](https://jupyter-docker-stacks.readthedocs.io/) image and want to enable sudo:
+
+```yaml
+singleuser:
+  allowPrivilegeEscalation: true
+  uid: 0
+  extraEnv:
```
consideRatio, line 83: This is the default CMD for a docker-stacks Dockerfile, and without it, `GRANT_SUDO` doesn't work. Since this will work no matter if our default `cmd` is unset or explicit, I suggest we be explicit about this. Whoever use the docker-stacks will then better grasp that GRANT_SUDO is an env related to their script as well, making people debug possible issues better overall I think.

```suggestion
  cmd: start-notebook.sh
  extraEnv:
```

consideRatio (approved):
Thanks @manics !!! I've made a suggestion but it looks good to me besides that

docs/source/administrator/upgrading/upgrade-1-to-2.md
manics: Good point! I've made it generic instead of being specific to docker-stacks. However it might be probably worth adding the `GRANT_SUDO` as a full docker-stacks example somewhere in the main docs?

consideRatio: Yepp, very reasonable call as well!

PR #2820, opened 2022-08-01
Restore jupyterhub-singleuser as the default command

minrk:
update docs to explain explicit `cmd: null` use case, which works now (unlike when #2449 came in, I think)

reverts #2449
closes #2805

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -408,8 +408,13 @@ for key, role in get_config("hub.loadRoles", {}).items():
 
     c.JupyterHub.load_roles.append(role)
 
+# respect explicit null command (distinct from unspecified)
+# this avoids relying on KubeSpawner.cmd's default being None
+_unspecified = object()
+specified_cmd = get_config("singleuser.cmd", _unspecified)
+if specified_cmd is not _unspecified:
+    c.Spawner.cmd = specified_cmd
 
-set_config_if_not_none(c.Spawner, "cmd", "singleuser.cmd")
 set_config_if_not_none(c.Spawner, "default_url", "singleuser.defaultUrl")
 
 cloud_metadata = get_config("singleuser.cloudMetadata", {})
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -2126,7 +2126,7 @@ properties:
           [KubeSpawner.cmd](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.cmd).
           The default is "jupyterhub-singleuser".
           Use `cmd: null` to launch a custom CMD from the image,
-          which must launch jupyterhub-singleuser eventually.
+          which must launch jupyterhub-singleuser or an equivalent process eventually.
           For example: Jupyter's docker-stacks images.
       defaultUrl:
         type: [string, "null"]
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1019,7 +1019,7 @@ properties:
           extraPorts:
             type: array
             description: |
-              Extra ports to add to the Hub Service object besides `hub` / `8081`.  
+              Extra ports to add to the Hub Service object besides `hub` / `8081`.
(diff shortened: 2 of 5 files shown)


jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -408,8 +408,13 @@ def camelCaseify(s):
 
     c.JupyterHub.load_roles.append(role)
 
+# respect explicit null command (distinct from unspecified)
+# this avoids relying on KubeSpawner.cmd's default being None
+_unspecified = object()
+specified_cmd = get_config("singleuser.cmd")
+if specified_cmd is not _unspecified:
+    c.Spawner.cmd = specified_cmd
```
consideRatio, line 416: I don't understand this very well and think there may be a broken assumption here.

With `singleuser.cmd` set at all, to `null` or something else, it will be recognized. In `helm` the values will be merged, but without setting `singleuser: null` or similar can you get rid of `singleuser.cmd` being declared at all.

I think the change of relevance would be to always set config of c.Spawner.cmd, independent of if we are passed a YAML null (Python None) value or not. This means that we will never regonize the kubespawner default value of `cmd`.

```yaml
# both a and b have the `null` value when YAML is parsed below.
a:
b: null
```

minrk: > I think the change of relevance would be to always set config of c.Spawner.cmd, independent of if we are passed a YAML null (Python None) value or not. This means that we will never regonize the kubespawner default value of cmd.

That's right - it ensures that setting `singleuser.cmd: null` _explicitly_ sets the value to None, rather than _implicitly_ setting it to whatever KubeSpawner has as default (currently: None).

There are 3 ways we can do this:

1. `set_config_if_not_none` (before this PR)
2. `c.Spawner.cmd = get_config("singleuser.cmd")`
3. this PR

These will all result in the same behavior with the _current_ state of everything, specifically that KubeSpawner.cmd's default is None.

Number 1. caused problems because setting `None` is _not_ telling KubeSpawner to use None, it's telling KubeSpawner to use its default, which is _currently_ None but doesn't have to be. By avoiding this, we keep full control of the value in z2jh, which seems important when z2jh's default is not the same as KubeSpawner's default.
This change also means that KubeSpawner's default need not have been changed to allow using the underlying image.

Number 2. and 3. will generally behave the same, because:

1. singeluser.cmd will always be set, unless we change the default values.yaml to omit it for some reason
2. leaving the value unset still results in None, but via KubeSpawner's default instead

If _either_ `values.yaml` changes to leave `singleuser.cmd` actually unspecified, _or_ KubeSpawner changes its default to not be None, then 2. starts to misbehave, whereas 3. is always correct - reflecting explicit values in config.

Note: we have to choose between `cmd: null` meaning 'explicitly choose the unconfigured default' vs 'explicitly choose None', because we can't do both unless we define a special value e.g. `":default:"` to distinguish "don't set anything" from "set it to None". This issue comes up for every config value where None is a valid value and may not be the default (it's _almost_ always the default if it's legal).

jupyterhub/schema.yaml
```diff
@@ -2124,6 +2124,10 @@ properties:
         description: |
           Passthrough configuration for
           [KubeSpawner.cmd](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.cmd).
+          The default is "jupyterhub-singleuser".
+          Use `cmd: null` to launch a custom CMD from the image,
+          which must launch jupyterhub-singleuser eventually.
```
manics, line 2129:
```suggestion
          which must launch jupyterhub-singleuser or an equivalent process eventually.
```

manics (commented):
Can you also remove
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/9653da69f8e2be963402b36ab73c7126e8500fce/docs/source/administrator/upgrading/upgrade-1-to-2.md?plain=1#L84-L96

consideRatio (approved):

jupyterhub/files/hub/jupyterhub_config.py
minrk: Yup, that was a typo where it just wouldn't do what I meant it to :)

manics (approved):

PR #2833, opened 2022-08-08
ci: fix jupyterhub version bumping automation

consideRatio:
This PR makes us bump to pre-releases of jupyterhub automatically, and to as part of that PR also make sure we refreeze the dependencies in requiements.txt if jupyterhub was bumped. The refreezing part makes us duplicate code a bit, but I think it should be fine. I added a note about it.

Fixes #2828 and addresses https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2827#issuecomment-1203718287.

---

While this version bumping automation is a bit complicated, I like that it makes us codify knowledge on chore maintenance tasks so anyone can do it quickly.

### DIFF
--- a/.github/workflows/watch-dependencies.yaml
+++ b/.github/workflows/watch-dependencies.yaml
@@ -141,6 +141,9 @@ jobs:
         with:
           python-version: "3.9"
 
+      - name: Install Python dependencies
+        run: pip install requests
+
       - name: Get images/hub/requirements.in pinned version of jupyterhub
         id: local
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/watch-dependencies.yaml
```diff
@@ -149,12 +152,16 @@ jobs:
...
-              curl -s https://pypi.org/pypi/jupyterhub/json \
-            | jq -r .info.version
-          )
-          echo "::set-output name=version::$latest_version"
+          import requests
+
+          request = requests.get("https://pypi.org/pypi/jupyterhub/json")
+          data = request.json()
+          releases = data["releases"]
+          latest_version = list(releases.keys())[-1]
```
minrk, line 162: I'm not sure the order of this dict is guaranteed, or what its order is, if it is stable (e.g. it would be wrong when we do a backport release if it's by time). I think you need to sort by parsed version: `sorted(releases.keys(), key=packaging.version.Version)[-1]`.

minrk:
Great! Just ran the job and it didn't open a new PR.

PR #2845, opened 2022-08-17
Make the singleuser-sample image use python:3.9-slim-bullseye as a base image to retain arm64 support

### LINKED ISSUE #2844: jupyter/docker-stacks no longer provide multiplatform images - that influences our singleuser-sample image on arm
We have relied on jupyter/docker-stacks to provide amd64/arm64 multiplatform images, but they are no longer around but they are meant to be around. We can follow https://github.com/jupyter/docker-stacks/issues/1737 for progress and help out there.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6fc3d41fc8032b04f2ad1fd8da7ed15827cc3812/images/singleuser-sample/Dockerfile#L1

minrk:
### Update by Erik

Closes #2844 by switching to another base image for singleuser-sample that supports both amd64 and arm64.

### Original PR description

Using pre-chartpress `sed`, since chartpress doesn't have a way to influence the base image at invocation time. It's unclear to me what our level of support for running on ARM is meant to be.

Real deployments won't use the sample image anyway (closes #2844). We are publishing a `linux/arm64` tag via buildx, but it's actually an amd64 image if you try to use it.

Alternative: switch to building from a lower base image (e.g. `python:3.10`). Advantage: simpler, under our full control.


minrk:
Got farther, but actual spawn fails. Tests don't show any logging or debugging info, though.

.circleci/config.yml
```diff
@@ -50,6 +50,14 @@ jobs:
...
           name: Install dependencies
 
+      - run:
+          name: fix singleuser base image for aarch64
+          # Jupyter docker-stacks don't have multi-arch images
+          # see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2844
+          command: |
+            sed -i -E s@base-notebook:latest@base-notebook:aarch64-latest@ images/singleuser-sample/Dockerfile
+            git diff
+
```
consideRatio, line 60:
```suggestion
```

With the new base image, we can be multi-platform and this is no longer relevant.

images/singleuser-sample/Dockerfile
```diff
@@ -1,26 +1,61 @@
...
 
-# VULN_SCAN_TIME=2022-08-08_07:54:05
+# VULN_SCAN_TIME=
+
+WORKDIR /build-stage
+
+# Build wheels for packages that require gcc or other build dependencies and
+# lack wheels either for amd64 or aarch64.
+COPY requirements.txt requirements.txt
+RUN pip wheel -r requirements.txt
```
consideRatio, line 12: It is better to avoid building wheels here from a final size perspective, because when you copy and then install in another stage, the copy + install layers will be separate.

I think we won't need a build stage for this image actually.

images/singleuser-sample/Dockerfile
```diff
@@ -1,26 +1,61 @@
...
  && apt-get install -y --no-install-recommends \
+        ca-certificates \
         dnsutils \
         git \
+        locales \
         iputils-ping \
- && rm -rf /var/lib/apt/lists/*
-USER $NB_USER
+        tini \
+ && apt-get clean \
```
consideRatio, line 38:
```suggestion
```

`apt-get clean` is redundant in this situation I've learned from reading Dockerfile best-practices etc. I want to remove it to avoid introducing a hint that it is needed in other Dockerfiles etc even though its probably also harmless to have.

images/singleuser-sample/Dockerfile
```diff
@@ -1,26 +1,61 @@
...
 # conda/pip/apt install additional packages here, if desired.
+
+RUN adduser --disabled-password \
+        --gecos "Default user" \
+        --uid ${NB_UID} \
+        --home ${HOME} \
+        --force-badname \
+        ${NB_USER}
+
+USER $NB_USER
```
consideRatio, line 59: We can switch to the notebook user earlier, before running pip install - lets do it.

.circleci/config.yml
consideRatio, line 60:
```suggestion
```

consideRatio:
Sorry for the trouble @minrk - I did a `git push` on the `arm-tag` branch by mistake and pushed to your fork and not my fork in order to update https://github.com/minrk/zero-to-jupyterhub-k8s/pull/1.

Feel free to revise as you please, I have my suggested changed available locally as well.

minrk:
No worries! Feel free to iterate in this PR directly.

consideRatio:
@minrk I consider this is good to go as it is now, what do you think?

minrk:
Thanks @consideRatio!

PR #2853, opened 2022-08-26
Corrected indentation in the config.yml example

praveen-ag:
At section `Modify your config.yaml file to specify the image`. This corrects following error when running helm upgrade using incorrect config:

```Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- singleuser.image: Additional property cmd is not allowed


consideRatio:
Thank you!!!

PR #2856, opened 2022-09-01
Mount wheels from build stage instead of copying

minrk:
- avoids copying wheels, eliminating the need to track which need compilers on arm and select those for the build stage. The build stage can now be much simpler and build all wheels
- mount pip cache as a cache dir, so cache can be used without polluting the image

Requires buildkit for `--mount`

results: smaller images (when compared with the simpler `COPY --from` for all wheels, negligible difference with the more complex, optimized build stage), simpler Dockerfiles (no need to pick 'needs compile on arm' images for the build stage), faster rebuilds (cache is actually used, which saves a lot of time)

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -7,8 +7,8 @@ FROM python:3.9-bullseye as build-stage
 
 WORKDIR /build-stage
 
-# set env for pip cache,
-# but use ARG to avoid persisting in image
+# set pip's cache directory using this environment variable, and use
+# ARG instead of ENV to ensure its only set when the image is built
 ARG PIP_CACHE_DIR=/tmp/pip-cache
 
 # Build wheels
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -1,3 +1,4 @@
+# syntax = docker/dockerfile:1.3
 # The build stage
 # ---------------
 FROM python:3.9-bullseye as build-stage
@@ -6,21 +7,16 @@ FROM python:3.9-bullseye as build-stage
 
 WORKDIR /build-stage
 
-# Build wheels for packages that require gcc or other build dependencies and
-# lack wheels either for amd64 or aarch64.
-#
-#   - https://pypi.org/project/pycurl/#files, no wheels available as of 7.45.1.
-#     See https://github.com/pycurl/pycurl/issues/738.
-#
-# If you find a dependency here no longer listed in requirements.txt, remove it
-# from here as well. The more wheels we build here and copy and then install,
-# where the copy and install are separate steps, the larger the final image
-# becomes.
-#
+# set env for pip cache,
+# but use ARG to avoid persisting in image
+ARG PIP_CACHE_DIR=/tmp/pip-cache
+
+# Build wheels
+# These are mounted into the final image for installation
 COPY requirements.txt requirements.txt
-RUN pip install build \
- && pip wheel \
-        $(cat requirements.txt | grep "pycurl==")
+RUN --mount=type=cache,target=${PIP_CACHE_DIR} \
+    pip install build \
+ && pip wheel -r requirements.txt
 
 
 # The final stage
@@ -30,6 +26,7 @@ FROM python:3.9-slim-bullseye
 ARG NB_USER=jovyan
 ARG NB_UID=1000
 ARG HOME=/home/jovyan
+
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN adduser --disabled-password \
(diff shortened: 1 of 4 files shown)


images/hub/Dockerfile
```diff
@@ -56,10 +53,16 @@ RUN apt-get update && \
         tini \
  && rm -rf /var/lib/apt/lists/*
 
-COPY --from=build-stage /build-stage/*.whl /tmp/pre-built-wheels/
+# set env for pip cache,
+# but use ARG to avoid persisting in image
+ARG PIP_CACHE_DIR=/tmp/pip-cache
```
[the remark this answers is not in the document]
consideRatio: Ah I verified that to be so, nice! I didn't know this. I also wasn't sure that PIP_CACHE_DIR was respected by PIP itself or something we gave meaning to. This variation of the comment could have perhaps helped med understand this a bit quicker.

```suggestion
# set pip's cache directory with environment variable and use
# ARG instead of ENV to ensure its only set when the image is built
ARG PIP_CACHE_DIR=/tmp/pip-cache
```

images/singleuser-sample/Dockerfile
```diff
@@ -6,21 +7,15 @@ FROM python:3.9-bullseye as build-stage
...
-#   - https://pypi.org/project/psutil/#files, no wheels available as of 5.9.1
-#     for aarch64. See https://github.com/giampaolo/psutil/pull/2070.
-#
-# If you find a dependency here no longer listed in requirements.txt, remove it
-# from here as well. The more wheels we build here and copy and then install,
-# where the copy and install are separate steps, the larger the final image
-# becomes.
-#
+# set env for pip cache,
+# but use ARG to avoid persisting in image
```
consideRatio, line 11:
```suggestion
# set pip's cache directory using this environment variable, and use
# ARG instead of ENV to ensure its only set when the image is built
```

minrk: Spot on!

Probably doesn't matter and we could use ENV. I don't know what the best practice is for environment variables that you want to use in a Dockerfile, but don't necessarily want to have in the final image.

images/singleuser-sample/Dockerfile
```diff
@@ -53,10 +48,16 @@ RUN apt-get update \
...
+ARG PIP_CACHE_DIR=/tmp/pip-cache
+
+# install wheels built in the build-stage
 COPY requirements.txt /tmp/requirements.txt
-RUN pip install --no-cache-dir \
-        /tmp/pre-built-wheels/*.whl \
+RUN --mount=type=cache,target=${PIP_CACHE_DIR} \
+    --mount=type=cache,from=build-stage,source=/build-stage,target=/tmp/wheels \
+    pip install \
+        --find-links /tmp/wheels/ \
```
consideRatio, line 60: This variation helped my understanding a bit, not used `--find-links` before and wasn't sure for example that it accepted an argument or not.

```suggestion
    pip install --find-links=/tmp/wheels/ \
```

images/singleuser-sample/Dockerfile
```diff
@@ -53,10 +48,16 @@ RUN apt-get update \
         git \
  && rm -rf /var/lib/apt/lists/*
 
-COPY --from=build-stage /build-stage/*.whl /tmp/pre-built-wheels/
+# set env for pip cache,
+# but use ARG to avoid persisting in image
```
consideRatio, line 52:
```suggestion
# set pip's cache directory using this environment variable, and use
# ARG instead of ENV to ensure its only set when the image is built
```

images/hub/Dockerfile
```diff
@@ -6,21 +7,16 @@ FROM python:3.9-bullseye as build-stage
...
-#   - https://pypi.org/project/pycurl/#files, no wheels available as of 7.45.1.
-#     See https://github.com/pycurl/pycurl/issues/738.
-#
-# If you find a dependency here no longer listed in requirements.txt, remove it
-# from here as well. The more wheels we build here and copy and then install,
-# where the copy and install are separate steps, the larger the final image
-# becomes.
-#
+# set env for pip cache,
+# but use ARG to avoid persisting in image
```
consideRatio, line 11:
```suggestion
# set pip's cache directory using this environment variable, and use
# ARG instead of ENV to ensure its only set when the image is built
```

consideRatio (approved):
I missed two comment updates, four was needed.

This looks great to me though, and resolves something that has bugged me for quite a while! NICE!!!!! :D


Ping @mathbunnyru, this is a relevant strategy to be aware about in Dockerfile's!

PR #2860, opened 2022-09-07
Update chartpress to 2.1+ and adopt tbump in the release process

minrk:
- require chartpress 2 in dev-requirements (previous unbounded pinning meant we were using it already as soon as it shipped)
- set baseVersion in chartpress.yaml
- note baseVersion bump in release process

One bit of the new chartpress 2.0 process that may be useful, but I'm not sure where to write it down: since we publish versions continuously, we should bump the baseVersion e.g. 2.1.0-0.dev -> 3.0.0-0.dev in the first 'breaking' PR. It's not the _most_ important thing to get exactly right, but it will be easy to overlook. Arguably,w e could do the same for 2.0.1 -> 2.1.0 fr the first new thing. Alternately, we could always start with '3.0.0-0.dev' to indicate 'might be breaking' so we don't have to pay attention so much in mid-development.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -94,7 +94,7 @@ Also the images we build are based on some image specified in the `FROM` stateme
   - [ ] Set the next prerelease version (don't create a tag).
 
     ```bash
-    tbump --no-tag x.y+1.z-0.dev
+    tbump --no-tag x.y.z+1-0.dev
     ```
 
   - [ ] Create a GitHub release.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -65,8 +65,7 @@ Also the images we build are based on some image specified in the `FROM` stateme
     ```bash
     git checkout main
     git reset --hard <upstream>/main
-    git tag -a x.y.z-beta.1 -m x.y.z-beta.1 <commit on main>
-    git push --follow-tags <upstream> main
+    tbump x.y.z-beta.1
     ```
 
 - Announce the x.y.z-beta.1 release
@@ -89,11 +88,14 @@ Also the images we build are based on some image specified in the `FROM` stateme
     ```bash
     git checkout main
     git reset --hard <upstream>/main
-    git tag -a x.y.z -m x.y.z HEAD
-    git push --follow-tags <upstream> main
+    tbump x.y.z
     ```
 
-  - [ ] Update baseVersion in chartpress to the next release (e.g. `2.1.0-0.dev` after 2.0.0)
+  - [ ] Set the next prerelease version (don't create a tag).
+
+    ```bash
+    tbump --no-tag x.y+1.z-0.dev
+    ```
 
   - [ ] Create a GitHub release.
         Visit the [release page](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/releases) and create a new release referencing the recent tag. Add a brief text like the one below.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -93,6 +93,8 @@ Also the images we build are based on some image specified in the `FROM` stateme
     git push --follow-tags <upstream> main
     ```
 
+  - [ ] Update baseVersion in chartpress to the next release (e.g. `2.1.0-0.dev` after 2.0.0)
+
   - [ ] Create a GitHub release.
         Visit the [release page](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/releases) and create a new release referencing the recent tag. Add a brief text like the one below.
 
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -13,6 +13,7 @@ charts:
     # Dev: imagePrefix can be useful to override if you want to trial something
     #      locally developed in a remote k8s cluster.
     imagePrefix: jupyterhub/k8s-
+    baseVersion: "2.0.0-0.dev"
     repo:
       git: jupyterhub/helm-chart
       published: https://jupyterhub.github.io/helm-chart
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -13,7 +13,7 @@ charts:
     # Dev: imagePrefix can be useful to override if you want to trial something
     #      locally developed in a remote k8s cluster.
     imagePrefix: jupyterhub/k8s-
-    baseVersion: "2.0.0-0.dev"
+    baseVersion: "2.0.0-beta.1"
     repo:
       git: jupyterhub/helm-chart
       published: https://jupyterhub.github.io/helm-chart
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@
 #
 # ref: https://github.com/jupyterhub/chartpress
 #
-chartpress==2.*
+chartpress>=2.1,<3
 
 # pytest run tests that require requests and pyyaml
 pytest>=3.7.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@
 #
 # ref: https://github.com/jupyterhub/chartpress
 #
-chartpress>=1.1.0
+chartpress==2.*
 
 # pytest run tests that require requests and pyyaml
 pytest>=3.7.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -17,3 +17,6 @@ click
 
 # used to formatting and linting
 pre-commit>=2.0.0
+
+# used for tagging releases
+tbump
--- a/tbump.toml
+++ b/tbump.toml
@@ -1,4 +1,9 @@
-# tbump config sets baseVersion in chartpress.yaml
+# tbump is a tool to update version fields that we use
+# to update baseVersion in chartpress.yaml as documented
+# in RELEASE.md
+#
+# Config reference: https://github.com/your-tools/tbump#readme
+#
 [version]
 current = "2.0.0-beta.1"
 
--- a/tbump.toml
+++ b/tbump.toml
@@ -1,0 +1,35 @@
+# tbump config sets baseVersion in chartpress.yaml
+[version]
+current = "2.0.0-0.dev"
+
+# match our prerelease prefixes
+# -alpha.1
+# -beta.2
+# -0.dev
+
+regex = '''
+  (?P<major>\d+)
+  \.
+  (?P<minor>\d+)
+  \.
+  (?P<patch>\d+)
+  (\-
+    (?P<prelease>
+      (
+       (alpha|beta|rc)\.\d+|
+       0\.dev
+      )
+    )
+  )?
+  '''
+
+[git]
+message_template = "Bump to {new_version}"
+tag_template = "{new_version}"
+
+# For each file to patch, add a [[file]] config
+# section containing the path of the file, relative to the
+# tbump.toml location.
+[[file]]
+src = "chartpress.yaml"
+search = 'baseVersion: "{current_version}"'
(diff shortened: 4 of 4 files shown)


RELEASE.md
```diff
@@ -93,6 +93,8 @@ Also the images we build are based on some image specified in the `FROM` stateme
     git push --follow-tags <upstream> main
     ```
 
+  - [ ] Update baseVersion in chartpress to the next release (e.g. `2.1.0-0.dev` after 2.0.0)
+
   - [ ] Create a GitHub release.
```
minrk, line 98: @consideRatio do you know why we have 'make a GitHub Release' in our release process? To me, those are for projects that produce artifacts to attach, and/or use GitHub Releases for release notes, and we do neither of these. I'm not sure what's beneficial in what we do beyond what's already in the autogenerated releases for tags.

consideRatio: Yeah I've added it because when projects once has made them (they are for z2jh), they show up in the web UI. So, if someone makes 2.0.0 and not make a github release, it will show up as 1.2.0 there as the latest version.

So, it is a precaution. I'd wish we could make github stop presenting that at all for us really though.

chartpress.yaml
```diff
@@ -13,6 +13,7 @@ charts:
     # Dev: imagePrefix can be useful to override if you want to trial something
     #      locally developed in a remote k8s cluster.
     imagePrefix: jupyterhub/k8s-
+    baseVersion: "2.0.0-0.dev"
```
consideRatio, line 16: It currently works as it should, because we have 2.0.0-beta.1 out and then the behavior without having a baseVersion set is matching what is correct.

```suggestion
    baseVersion: "2.0.0-beta.1"
```

consideRatio: image

RELEASE.md
consideRatio: Should we remove them all? I'll open an issue about it.

chartpress.yaml
minrk: Hm. I was specifically trying to avoid needing to update baseVersion for every prerelease (i.e. only bump it on final release). The result would be that untagged dev releases would keep '0.dev' in between tagged betas. WDYT?

RELEASE.md
minrk: Yeah, I really wish it were easier (or possible) to hide unused features on GitHub. Should we delete all the releases, then? Or just edit the latest release description to point to where actual releases are published?

chartpress.yaml
consideRatio: So dev releases would always show up as `2.0.0-0.dev.git.5783.h70afc746` no matter if they are before or after a alpha/beta tag?

Not so happy about that, then we have version sorting issues once again, but between dev releases and pre-releases. I'll want to think further and read more in the PR you just opened in chartpress that I yet haven't looked at.

RELEASE.md
consideRatio: Opened #2861 about it.

chartpress.yaml
minrk: > then we have version sorting issues once again

I think that depends. Should untagged builds be preferred to the latest prerelease tag? I think it could be a _good_ thing that the latest tag is preferred to the latest untagged commit.

One nice thing about the absolute commit count is you can see clearly the build order, regardless of the prerelease tag (humans, not machines).

I'm trying to find the right balance of maintenance tasks vs clear communication, and minimize the impact of this workflow change to solve the relatively minor issue of confusing dev tags mid-development. Bumping baseVersion during prereleases adds a manual, not-automated step to making prereleases (as opposed to just stable releases). Is the semantic ordering of `beta.1[.git.5700]` vs `dev.0.git.5701` important? If so, which is the best direction of preference? I'd argue that the ordering between the affected versions is _not_ important, and the best direction is ambiguous. This issue also only comes up _during the prerelease process_, not during normal development, where I think the distinction is of relatively little consequence, and I think reducing maintainer tasks has a larger benefit than perfect semantic ordering of a small number of necessarily similar unreleased commits.

But it is just one task per prerelease tag, so if you think perfect ordering is worth the extra step (and preferring dev to tagged beta is desirable), we can add bumping baseVersion to the prerelease steps as well.

consideRatio: I'm having so many hats on now:
- Being a z2jh user and communicating to other z2jh users as a chart maintainer: I greatly appreciate if the intuition of version sorting is correctly aligning with how new something is. It can be relevant for example when viewing releases at artifacthub.io. People will get this by default I think, so anything besides this is a drawback I think.
  image
- Being chartpress user and a z2jh maintainer: I appreciate if its a) easy to communicate what to do in RELEASE.md, and b) if what to do is little work. Optimizing for a) is far more important than b) in my mind - to run longer but in a straight line is to be preferred over running a shorter distance but needing to read a map first.
- Being a chartpress developer: I appreciate simple logic that is easy to document and maintain.

As a z2jh maintainer, I'd very much like to do the work to ensure dev releases are sorting correctly also in between pre-releases and not compromise on that - even if it means to update baseVersion often. As a chartpress maintainer I'd very much like to stick to relatively simple logic that is robust and easy to understand.

---

> I think reducing maintainer tasks has a larger benefit than perfect semantic ordering of a small number of necessarily similar unreleased commits.

I see this as a trade off between more work that is easier and than less work that is more complicated. I have too many hats on at the same time to think very clearly about things, but for z2jh specifically I'm opinionated towards not doing anything with baseVersion or doing it systematically so that our releases are always version sorting newer if they are newer.

minrk: > I greatly appreciate if the intuition of version sorting is correctly aligning with how new something is. It can be relevant for example when viewing releases at artifacthub.io. People will get this by default I think, so anything besides this is a drawback I think.

I want to make sure we're clear that we don't have this now, which was the motivation for introducing the baseVersion option. Right now, _all_ dev versions are sorted incorrectly behind the last release, and (pre-2.0) inconsistently relative to each other. Only between a prerelease and a final release (where we are now) would ordering be mostly correct (but not entirely if more than 9 commits occurred), the rest of the time it would be wrong. As an example, in your screenshot that `1.1.3-n804` (and hundreds of other tags) should be between 1.2.0 and 2.0.0-beta.1, not below 1.1.3.

We've lived with "dev versions are not sortable except by humans looking at the `.n123`" field (or date) thus far. Every option with baseVersion is an improvement over the status quo in this regard. Without baseVersion, 2.0 is an improvement in that dev versions are _internally_ sortable, but still incorrect relative to stable releases.

The question right now is: _during_ the prerelease process, how should _tagged_ prereleases compare with _untagged_ dev versions. I genuinely think both answers are valid! Preferring a tagged prerelease to an untagged commit doesn't seem wrong to me. At the same time, during the beta process all commits probably are straightforward improvements over what came before.

It's a bit frustrating that the _only_ time behavior is correct without baseVersion (when the latest tag is a prerelease) is the time it's most tedious for baseVersion to behave correctly.


>  Optimizing for a) is far more important than b) in my mind - to run longer but in a straight line is to be preferred over running a shorter distance but needing to read a map first.

I definitely agree here. This is part of my main reason for omitting the baseVersion bump, but your thinking may be different! I thought it would be easier to make prereleases and not make mistakes if the only repo activity were 'push a tag'. But maybe it's more of a burden for stable releases to have _different_ steps from prereleases. WDYT?

WDYT about adopting tbump? I think any time making a release involves modifying file(s) and making a tag, having a tool do both atomically is valuable. That would mean baseVersion always matches the tag, and is updated just once.

It would _require_ https://github.com/jupyterhub/chartpress/pull/184 because without that, setting `baseVersion: 2.0.0` on the release tag would cause chartpress to fail.

minrk: Actually, I was wrong about that last statement - we already accept `baseVersion: 2.0.0` _on the tagged commit itself_, but not after. So I think that's all okay already.

minrk: Nope, sorry. I had https://github.com/jupyterhub/chartpress/pull/184 checked out when I tested locally. We do indeed need that patch in order to use tbump, which I think should probably be recommended in combination with baseVersion. It simplifies everything, _reduces_ the complexity of making a release relative to what we already have, _and_ ensures version ordering is always correct, even between prereleases.

minrk:
I've updated this PR to reflect using tbump, which I think is probably what we should recommend to anyone using chartpress with baseVersion config. The last commit was created using tbump, to get the files in sync with the latest tag.

I'm happy to revert, but at this point I don't think we should recommend using baseVersion _without_ tbump (or similar).

RELEASE.md
```diff
@@ -89,8 +88,13 @@ Also the images we build are based on some image specified in the `FROM` stateme
...
     git reset --hard <upstream>/main
-    git tag -a x.y.z -m x.y.z HEAD
-    git push --follow-tags <upstream> main
+    tbump x.y.z
+    ```
+
+  - [ ] Set the next prerelease version (don't create a tag).
+
+    ```bash
+    tbump --no-tag x.y+1.z-0.dev
```
consideRatio, line 97: I'd like to stay with `x.y.z+1-0.dev` to avoid publishing early dev releases that sort later than upcoming patch releases.

If we end up doing a backport release with a patch release, because we have also added things we don't want to release yet with a minor/major bump, we could at that time also increment the minor or major version in the baseVersion field.

What do you think?

```suggestion
    tbump --no-tag x.y.z+1-0.dev
```

dev-requirements.txt
```diff
@@ -5,7 +5,7 @@
 #
 # ref: https://github.com/jupyterhub/chartpress
 #
-chartpress>=1.1.0
+chartpress==2.*
```
consideRatio, line 8:
```suggestion
chartpress>=2.1,<3
```

Since we want to use tbump

tbump.toml
```diff
@@ -0,0 +1,35 @@
+# tbump config sets baseVersion in chartpress.yaml
```
consideRatio, line 1:
```suggestion
# tbump is a tool to update version fields that we use it
# to update baseVersion in chartpress.yaml as documented
# in RELEASE.md
#
# Config reference: https://github.com/your-tools/tbump#readme
#
```

RELEASE.md
minrk: I think that's fine. The trade off is we need to remember to bump baseVersion when we introduce the first enhancement / new feature (rather than just 'breaking' for the major version). Tracking baseVersion means every time we review a PR, we need to ask the questions:

1. have we already bumped the major baseVersion? (If yes, stop here)
2. Is this a breaking change? (If yes, bump major baseVersion)
3. have we already bumped minor baseVersion? (If yes, stop here)
4. Is this a new feature or enhancement? (If yes, bump minor baseVersion)

Let's try it and find out how we feel!

consideRatio: Yepp I agree, that would be the procedure. When we merge the first thing into the default branch that motivates something beyond a patch release, we increment the baseVersion to represent that unless its already done.

If we would fail to remember bumping the baseVersion when we merge a PR that motivates a minor or major version bump, it wouldn't be a big deal in my mind. Those dev-releases following that would not correctly reflect the situation, but they would correctly order themselves in the semver2 version sorting.

chartpress.yaml
consideRatio: :heart: :tada: soo much thorough effort into this! Thank you @minrk I'm super happy about a robust tbump resolution, now use of baseVersion makes me confident. Its fine that the initial setup is somewhat tricky, especially now that chartpress links to an example setup.

consideRatio (approved):

chartpress.yaml
minrk: Thanks for all your review! I'm much happier with the combination with tbump as well.

PR #2863, opened 2022-09-07
Fix docs for github org membership based auth

jakobkolb:
To read members of an organization, the scope has to be read:org, not read:user. Debugging this cost me half a day and I hope fixing this can save others the time.


consideRatio:
Yikes! Thank you soo much for upstreaming this @jakobkolb!!!! :heart: :tada:

PR #2870, opened 2022-09-13
fix: add string type to cpu limit and guarantee in helm schema

wilmardo:
Found this while using the chart, CPU limits and requests can be a string as well:

```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: frontend
spec:
  containers:
  - name: app
    image: images.my-company.example/app:v4
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
  - name: log-aggregator
    image: images.my-company.example/log-aggregator:v6
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
```
Source: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#example-1

In the current schema `100Mi` is an invalid value since the type is `string` instead of `number`

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -1955,9 +1955,9 @@ properties:
           for more info.
         properties:
           limit:
-            type: [number, "null"]
+            type: [number, string, "null"]
           guarantee:
-            type: [number, "null"]
+            type: [number, string, "null"]
       memory:
         type: object
         additionalProperties: false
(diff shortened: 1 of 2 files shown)


wilmardo:
Nevermind, it seems that these values are used by KubeSpawner which uses their own validation and values:

memory is `100M` vs `100Mi`:
https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.mem_limit

cpu must be a float:
https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.cpu_guarantee

PR #2893, opened 2022-09-26
docs: Remove unreleased reverted change from 2.0.0 release changelog

Uular:
PR #2449 was reverted by #2820 before release, and should not be mentioned in the changelog as a breaking change. 

The reversion should be kept in the changelog, as it brought new functionality in addition to reverting the unreleased changes. Added a note referencing the reverted change for posterity. Not sure if the note is in a good spot, feel free to suggest a better place for it.


yuvipanda:
Thank you very much, @Uular!

PR #2917, opened 2022-10-25
Replace vim with nano in hub image to reduce known vulnerabilities

### LINKED ISSUE #2916: Replace `vim` with `nano` in the hub image
Our `hub` image installs the apt package `vim` to provide one text editor, but by doing so have included a lot of unfixable known vulnerabilities that has been around for a very long time now without being fixed. The purpose of having an terminal based text editor in the hub image is just to make our life as admins easier when debugging things, and for this purpose `nano` also works.

Due to this, I'm proposing that we replace `vim` with `nano` which is as I understand the default editor for Debian Linux that we use. Due to this, and being lightweight, I expect it to result in less unpatched known vulnerabilities.

This investigationg was prompted by @fifofonix mentioning that many vulnerabilities stems from `vim` in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2586#issuecomment-1290389415.

```
# python:3.9-slim-bullseye after apt-get upgrade
Total: 84 (UNKNOWN: 0, LOW: 12, MEDIUM: 31, HIGH: 37, CRITICAL: 4)

# python:3.9-slim-bullseye after apt-get upgrade and nano installed (no change)
Total: 84 (UNKNOWN: 0, LOW: 12, MEDIUM: 31, HIGH: 37, CRITICAL: 4)

# python:3.9-slim-bullseye after apt-get upgrade and install vim (16 more critical, 300+ more high)
Total: 576 (UNKNOWN: 0, LOW: 16, MEDIUM: 111, HIGH: 429, CRITICAL: 20)
```

---

Note that artifacthub.io provides automated scanning of our Helm chart's images: https://artifacthub.io/packages/helm/jupyterhub/jupyterhub?modal=security-report

consideRatio:
I hope we don't need to discuss what editors to use in general, but agree that `nano` can do the trick for the occational debugging needed in the `hub` pod.

- Closes #2916

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -45,7 +45,7 @@ RUN apt-get update && \
         git \
         # misc other utilities
         less \
-        vim \
+        nano \
         # requirement for pycurl
         libcurl4 \
         # requirement for using a local sqlite database


yuvipanda:
I would really really really really really like to not remove vim *and* also avoid a conversation about editors :)

consideRatio:
Installing `neovim` resulted in this report summary by `trivy` (from python:3.9-slim-bullseye)

```
# python:3.9-slim-bullseye after apt-get upgrade
Total: 84 (UNKNOWN: 0, LOW: 12, MEDIUM: 31, HIGH: 37, CRITICAL: 4)

# python:3.9-slim-bullseye after apt-get upgrade and nano installed (no change)
Total: 84 (UNKNOWN: 0, LOW: 12, MEDIUM: 31, HIGH: 37, CRITICAL: 4)

# python:3.9-slim-bullseye after apt-get upgrade and install vim (16 more critical, 300+ more high)
Total: 576 (UNKNOWN: 0, LOW: 16, MEDIUM: 111, HIGH: 429, CRITICAL: 20)

# python:3.9-slim-bullseye after apt-get upgrade and install neovim (18 more critical, 100+ more high)
Total: 258 (UNKNOWN: 0, LOW: 13, MEDIUM: 63, HIGH: 160, CRITICAL: 22)
```

yuvipanda:
I do a lot of debugging work in the hub pod and removing vim completely in there would have a serious effect on that. We also don't really have conda in there so I can't really just install it either. I also personally believe these CVEs won't really apply to us in this context (which is one of the reasons I hate simple CVE scans). But I understand they cause noise which might drown out useful signals. I'm not sure what is the right thing to do :(

consideRatio:
> But I understand they cause noise which might drown out useful signals.

That you voiced this part makes me feel heard, this is the key reason for me to look into this atm!

yuvipanda:
Thank you so much @consideRatio for working on this!

minrk:
Question for @manics idea: with our chartpress setup, is it feasible to _inherit_ from one image to another? I think it might not be easy because the tag of one image isn't known or accessible in the build of another.

I was thinking the simplest version of this would be to have the bigger image FROM the smaller one, with a couple extra install lines, but I don't know how I'd do that with chartpress. Lockfiles and such aren't kind to layering either, so maybe the two having the same Dockerfile only differing by apt.txt/requirements.txt would be the easiest in practice.

consideRatio:
I'm quite strongly opinionated currently that its not worth the complexity of having two different kinds of images, and would argue that whatever isn't a good compromise in what we provide by default, users would have to adjust to.

I see the complexity related to multiple images is adds significantly to the complexity of maintaining source code, CI system, tooling, and documentation related to it.

---

I'd like to postpone opining on a choice or action point until we have a better understanding about the security reports for the `ubuntu` and `debian` images. If the `ubuntu` image is far better patched, then I'd prefer using it over using the `debian` image.

- Use `python` based on `debian` and vim from apt (current)
- Use `python` based on `debian` and nano from apt
- Use `python` based on `debian` and install `mamba` that users can use to install a text editor from conda-forge
- Use `ubuntu` and vim from apt
- Use `ubuntu` and nano from apt
- Use `ubuntu` based on `debian` and install `mamba` that users can use to install a text editor from conda-forge

Overall, I want low amounts of unpatched CVEs by default in all images we maintain for many reasons, where the key action point as I see it from us as maintainers is to make our fresh releases not show up with F rating on artifacthub.io:
  image

Currently, using `ubuntu:22.04` vs `debian:bullseye`

```
# ubuntu:22.04 after apt-get upgrade
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)

# ubuntu:22.04 after apt-get upgrade and less, nano, vim, neovim installed
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)

# debian:bullseye after apt-get upgrade and less, nano, vim, neovim installed
Total: 627 (UNKNOWN: 0, LOW: 16, MEDIUM: 123, HIGH: 454, CRITICAL: 34)
```

The `vim` apt package in ubuntu [`2:8.2.3995-1ubuntu2.1`, is patched Sept 15](https://launchpad.net/ubuntu/+source/vim/2:8.2.3995-1ubuntu2.1), and in debian [`2:8.2.2434-3+deb11u1`, is patched Dec 18 2021](https://launchpad.net/debian/+source/vim/2:8.2.2434-3+deb11u1).

## My current position

- We should go with `ubuntu` as a base image
- As part of switching to `ubuntu`, we embrace the complexity of installing Python ourselves using [micromamba](https://mamba.readthedocs.io/en/latest/installation.html#micromamba) that is very helpful in doing that from an environment without Python.
- We stick with `vim` as an apt package after using `ubuntu` again

manics:
It's not supported in Chartpress at the moment, but I think it's a reasonable addition if there's demand for a more secure minimal image.

The approach I'd take is to have a single multi-stage Dockerfile, and [tag an intermediate stage](https://stackoverflow.com/questions/58971382/how-to-give-name-or-tag-for-intermediate-image/65203301#65203301)

This ensures the multiple images are always in sync and rebuilt/tagged together, since they originate from the same Dockerfile.

minrk:
> The approach I'd take is to have a single multi-stage Dockerfile, and [tag an intermediate stage](https://stackoverflow.com/questions/58971382/how-to-give-name-or-tag-for-intermediate-image/65203301#65203301)

That sounds cool! I've not done that before.

I think it's reasonable to ask what's the role of these debugging tools in the default image:

1. who uses them? (speaks to reasonable expectations)
2. when do folks decide to use them? (i.e. does switching to a non-default image defeat the purpose because critical context is lost when the container is restarted)
3. what permissions are needed to be installed, if they aren't in the image? (i.e. can they be installed in a running container)

For instance, when I'm `kubectl exec` debugging, I regularly do some pip-installing, and that works fine because I don't need root, so there's no real benefit I see to those tools being in the image if they are pip-installable. But I _do_ need root for `apt-get install`, and, unlike `docker exec`, `kubectl exec` doesn't allow specifying users (I'm not sure how robust tools like [kubectl-execuser](https://github.com/drodbar/kubectl-execuser) are).

If we lost these tools in the default image, what would be involved for a deployment that wanted an additional apt package added to the default image? I guess this is the general 'custom hub image' case. Maybe the value of having them in the default image is because it's too much of a pain to maintain an extended image, and that's the problem we should work on?

For the more short-term choice, I think switching to ubuntu base makes perfect sense, based on @consideRatio's findings. I like the simpler, more straightforward installation of Python from apt instead of conda for simple environments like ours, but that doesn't allow easy selection of the Python version itself, so if we want to get e.g. Python 3.11 in a relatively short time frame, micromamba makes sense. We can do micromamba for just Python itself, and keep dependabot-compatible pip-compile tooling, or we could switch to [conda-lock](https://github.com/conda-incubator/conda-lock), which doesn't need to run in a container, and wouldn't have issues with e.g. pycurl compilation dependencies, because there would never be a compile step.

consideRatio:
I'm closing this as...

1. We can provide a `jupyterhub/k8s-hub-slim` version alongside a `jupyterhub/k8s-hub` image with ease.
   #2920
2. We can avoid all known vulnerabilities by switching to `ubuntu:22.04` that seem to patch things far faster.
   #2921

PR #2943, opened 2022-11-14
ci: fix deprecation of set-output in github workflows

consideRatio:
- Addresses https://github.com/jupyterhub/team-compass/issues/579 for this repo.

### DIFF
--- a/.github/workflows/watch-dependencies.yaml
+++ b/.github/workflows/watch-dependencies.yaml
@@ -162,7 +162,7 @@ jobs:
           latest_version = sorted(releases.keys(), key=packaging.version.Version)[-1]
 
           with open(os.environ["GITHUB_OUTPUT"], "a") as f:
-              f.write(f"version={latest_version}")
+              f.write(f"version={latest_version}\n")
 
       - name: Update pinned version of jupyterhub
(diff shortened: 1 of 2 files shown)


.github/workflows/watch-dependencies.yaml
```diff
@@ -160,7 +161,8 @@ jobs:
           releases = data["releases"]
           latest_version = sorted(releases.keys(), key=packaging.version.Version)[-1]
 
-          print(f"::set-output name=version::{latest_version}")
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
+              f.write(f"version={latest_version}")
```
manics, line 165:
```suggestion
              f.write(f"version={latest_version}\n")
```
I don't think `f.write` includes a newline

consideRatio: Hmmm hmmm, but we are in append mode (`a`), so I would presume it leads with a new line. Testing a bit.

consideRatio: I was considering if `echo hi >> myfile.txt` would cause a new line to be appended or prepended etc, but appending is the way to go I've confirmed now rather than thinking about prepending.

manics (approved):

PR #2956, opened 2022-11-24
docs: AWS master node size needs to be larger than micro

### LINKED ISSUE #2955: Setup on AWS: Validation error cluster not yet healthy
I was following the documentation [here](https://z2jh.jupyter.org/en/stable/kubernetes/amazon/step-zero-aws.html) to setup k8s on AWS and on step 12, at `kops validate cluster`, I was getting `Validation Failed
Error: Validation failed: cluster not yet healthy`

The debug message in `kops validate cluster` was:
```
2INSTANCE GROUPS
NAME			ROLE	MACHINETYPE	MIN	MAX	SUBNETS
master-us-east-2a	Master	t2.micro	1	1	us-east-2a
nodes-us-east-2a	Node	t2.medium	1	1	us-east-2a

NODE STATUS
NAME	ROLE	READY

VALIDATION ERRORS
KIND	NAME							MESSAGE
Machine	i-0887c7c6af5d1c854					machine "i-0887c7c6af5d1c854" has not yet joined cluster
Machine	i-0ee3b9baa855718e7					machine "i-0ee3b9baa855718e7" has not yet joined cluster
Pod	kube-system/coredns-77bcddd996-dpnct			system-cluster-critical pod "coredns-77bcddd996-dpnct" is pending
Pod	kube-system/coredns-autoscaler-975545559-xbdz4		system-cluster-critical pod "coredns-autoscaler-975545559-xbdz4" is pending
Pod	kube-system/ebs-csi-controller-7598d855cf-zdvbh		system-cluster-critical pod "ebs-csi-controller-7598d855cf-zdvbh" is not ready (csi-attacher,csi-provisioner)
Pod	kube-system/kops-controller-gs4gz			system-cluster-critical pod "kops-controller-gs4gz" is not ready (kops-controller)
Pod	kube-system/kube-controller-manager-i-0887c7c6af5d1c854	system-cluster-critical pod "kube-controller-manager-i-0887c7c6af5d1c854" is not ready (kube-controller-manager)

Validation Failed
Error: Validation failed: cluster not yet healthy
```

The issue might be that since master is `t2.micro`, it has low resources and not getting started up. After seeing the comment in [this](https://github.com/kubernetes/kops/issues/7438#issuecomment-535880472) issue, I fixed it by updating master to `t2.medium`.

arunppsg:
Closes #2955

### DIFF
--- a/docs/source/kubernetes/amazon/step-zero-aws.md
+++ b/docs/source/kubernetes/amazon/step-zero-aws.md
@@ -100,9 +100,9 @@ template you will use to setup and shape your cluster.
     kops create cluster $NAME \
     --zones "$ZONES" \
     --authorization RBAC \
-    --master-size t2.medium \
+    --master-size t3a.small \
     --master-volume-size 10 \
-    --node-size t2.medium \
+    --node-size t3.medium \
     --node-volume-size 10 \
     --yes
     ```
--- a/docs/source/kubernetes/amazon/step-zero-aws.md
+++ b/docs/source/kubernetes/amazon/step-zero-aws.md
@@ -100,7 +100,7 @@ template you will use to setup and shape your cluster.
     kops create cluster $NAME \
     --zones "$ZONES" \
     --authorization RBAC \
-    --master-size t2.micro \
+    --master-size t2.medium \
     --master-volume-size 10 \
     --node-size t2.medium \
     --node-volume-size 10 \


consideRatio:
It is good if the recommendation is as cheap as possible. Is micro too small - does it never come online? Or, is the micro node size just slow to start, and that is unexpected?

If this is increased from micro to medium, isn't there a "small" or similar that works? I figure one wants to have something as cheap as possible that still works well as the recommended setup.

manics:
_If_ we're moving away from the free instance then `t3` instances should work out cheaper than the `t2` equivalents, and `t3a` is even cheaper. Main caveat is t2 and t3 use burstable CPU, but if you hit the burst limit t2 throttles your CPU usage whilst t3 defaults to charging extra (but can be changed to throttle). In practice since most compute happens at startup I think this would be fine for the master.

arunppsg:
> It is good if the recommendation is as cheap as possible. Is micro too small - does it never come online? Or, is the micro node size just slow to start, and that is unexpected?

I waited for about 30-40 mins and micro did not come online for me.

> If this is increased from micro to medium, isn't there a "small" or similar that works? I figure one wants to have something as cheap as possible that still works well as the recommended setup.

I will test out a couple of options - small, t3 and update here on which instance has lowest cost @manics.

arunppsg:
@consideRatio  @manics `small` instances are able to get online and they cost less than medium instance. I updated t2.medium to use t3a.small.

consideRatio:
I'm not confident on the change to t3 for the other servers, but could make sense. I'm unsure, @manics, what do you think?

manics:
We're straying into optimising costs, which depends on usage and quickly gets complicated....

If I was spinning up a demo cluster and wanted to get costs down as much as possible I'd probably go for `t3a.small` for the nodes.

## The longer answer

There's loads of docs on burstable instances
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances.html

TLDR: For typical interactive/intermittent workloads burstable works because on average CPU usage is pretty low. With `t3*` if you exceed the instance's CPU "quota" for too long (e.g. 100% CPU for prolonged periods) there's a surcharge, though you can turn this off:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-unlimited-mode-concepts.html#unlimited-mode-enabling
If you do so then `t3*` should always be cheaper (since CPU is throttled if you hit the quota) but still more performant than  `t2`.

Off topic: Skimming through https://z2jh.jupyter.org/en/stable/kubernetes/amazon/step-zero-aws.html there's a few other options for reducing costs further, such as replacing `gp2` storage with `gp3`. And the first line "AWS does not have native support for Kubernetes, however there are many organizations that have put together their own solutions and guides for setting up Kubernetes on AWS." is incorrect 😄

consideRatio:
Thanks @manics, I'll go for a merge here, thinking its an incremental improvement for now!

PR #3002, opened 2023-01-26
Documentation fix for running k8s-singleuser-sample locally

pnasrat:
Local docker run needs to bind externally for expose to work.

Remove outdated instructions running on old/new interface.

Reported via https://discourse.jupyter.org/t/how-to-test-k8s-singleuser-sample-alone-and-locally-without-deploying-to-eks/17694/1


consideRatio (approved):
Thank you @pnasrat !!

PR #3003, opened 2023-01-26
docs: Update custom image docs to reflect root requirement

pnasrat:


docs/source/administrator/services.md
```diff
@@ -18,11 +18,15 @@ In the following snippet, I'm using a custom image that copies over the applicat
 
 ```Dockerfile
 # Dockerfile
-# 0.11.1 is latest stable release at the time of this writing
-FROM jupyterhub/k8s-hub:0.11.1
+# 2.0.0 is latest stable release at the time of this writing
+FROM jupyterhub/k8s-hub:2.0.0
 
+# The k8s-hub uses a multi-stage build to modify packages your build steps may need to run as root
```
consideRatio, line 24:
```suggestion
# Depending on version, the k8s-hub image may have installed
# pip packages as root, forcing you to install as root as well
```

consideRatio (commented):
The example update LGTM given that we have what I believe would be a bug in 2.0.0 - forcing use of root to install a pip package.

I opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3039 about this as well.

---

What do you think about my code suggestion updating the comment you have provided?

docs/source/administrator/services.md
```diff
@@ -18,11 +18,15 @@ In the following snippet, I'm using a custom image that copies over the applicat
 
 ```Dockerfile
 # Dockerfile
-# 0.11.1 is latest stable release at the time of this writing
-FROM jupyterhub/k8s-hub:0.11.1
+# 2.0.0 is latest stable release at the time of this writing
+FROM jupyterhub/k8s-hub:2.0.0
```
manics, line 22: We've got `chart_version` as a variable
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/29b041ced53678357d062c188f52e8983bfb1f03/docs/source/conf.py#L71-L72

@consideRatio Do you know of a simple way we could get `last_released_chart_version` to avoid having to keep this up to date?

consideRatio: Hmmmm, I recall there is a complexity around having a substitution inside a triple-backtick example. So, there may be two aspects to consider here.

But getting the latest released chart version is something I typically like getting done via `helm` and `yq`, which isn't available everywhere without trouble.

```
helm show chart --repo=https://hub.jupyter.org/helm-chart jupyterhub | yq eval ".version" -
```

For our charts specifically, we can use the `info.json` file we host alongside the Helm required `index.yaml` that summarizes the latest stable version of the charts: https://hub.jupyter.org/helm-chart/info.json, that way all we would need is a web request.

consideRatio:
Thank you @pnasrat for spotting this issue!!

Issue #3010, opened 2023-02-02
User pod launching timeout with large single user image when cluster scaling up


xcompass:
### Bug description

During the cluster scaling up, once the node is provisioned and is ready in k8s, the user pods may get scheduled to the newly provisioned node. However, if the continuous image puller is still pulling the single user image because the image is large (a few GB) or slow network, the user will get timeout error message from UI if the image pulling is not completed by the set timeout. Increasing the timeout might be a quick fix but it is not a good user experience as user may have to wait long time to get access (for us, it may take 10mins).

#### Expected behaviour

The user pods should not be scheduled to the new node before image pulling is done.

#### Actual behaviour

See Bug description

### How to reproduce

1. Create a large single user image or use a slower registry/network
2. Trigger a scaling up event
3. Spawn a few user pods, some of them will be scheduled to the new node when the node is ready
4. The user will see timeout from UI

### Our current workaround

Adding a taint before starting image puller or when provisioning the node to prevent the user pod being schedule to the node. Then remove taint after image puller is done. 

We have to code ready to send a PR. But want to see if there is a better way to solve it. Happy to send the PR anytime.

consideRatio:
KubeSpawner makes a pod be registered with the k8s api-server, at that point in time, the jupyterhub's server startup timeout starts. You can configure this timeout, and typically I increase it to 20 minutes when large images are involved.

https://z2jh.jupyter.org/en/stable/resources/reference.html#singleuser-starttimeout

xcompass:
Thanks for the reference. We tried to increase the timeout, but our user has to wait a long time (in our case around 10mins) to be able to get to the UI after logging in. However, we still have resources in the existing node to assign user. This happens when user placeholder is enabled with node density scheduling or default scheduling without placeholder enabled. As new node has no pods running, the user pod will be scheduled there as soon as it is ready (before single user image is pulled). 

This is kind like an edge case as it is not being triggered very often. It has to satisfy the following conditions:
1. A new node being provisioned (scaling up)
2. User placeholders are being used (if using node density scheduling)
3. Large single user image is being used and pre-pulled by continuous image puller
4. A user is trying to login after the new node is ready but before single user image is pulled.

consideRatio:
This is a tricky topic, I've considered in the past and I recall I couldn't conclude a strategy to resolve it that was simple and robust enough to think it was worth working towards.

- If the strategy used to solve this isn't very robust, it may cause other problems for users.
[...]

PR #3011, opened 2023-02-03
Adding taint during image pre-pulling to prevent user pod being scheduled

xcompass:
This is to solve the issue in #3010.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -609,6 +609,17 @@ prePuller:
       annotations: {}
   continuous:
     enabled: true
+  taintmanager:
+    enabled: true
+    image:
+      name: lthub/taintmanager
+      tag: master
+    # taint/tolerations used for preventing user pods being scheduled while user image is being pulled
+    # only first element in the list is being used
+    tolerations:
+      - key: hub.jupyter.org/imagepulling
+        operator: Exists
+        effect: NoExecute
   pullProfileListImages: true
   extraImages: {}
   pause:
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -559,6 +559,9 @@ scheduling:
         operator: Equal
         value: core
         effect: NoSchedule
+      - key: hub.jupyter.org/imagepulling
+        operator: Exists
+        effect: NoExecute
     nodeAffinity:
       matchNodePurpose: prefer
   userPods:
--- a/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
+++ b/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
@@ -51,6 +51,8 @@ spec:
         per node limit all k8s clusters have and have a higher priority
         than user-placeholder pods that could block an entire node.
       */}}
+      serviceAccount: taintmanager
+      serviceAccountName: taintmanager
       {{- if .Values.scheduling.podPriority.enabled }}
       priorityClassName: {{ include "jupyterhub.image-puller-priority.fullname" . }}
       {{- end }}
@@ -58,7 +60,7 @@ spec:
       nodeSelector:
         {{- . | toYaml | nindent 8 }}
       {{- end }}
-      {{- with concat .Values.scheduling.userPods.tolerations .Values.singleuser.extraTolerations .Values.prePuller.extraTolerations }}
+      {{- with concat .Values.scheduling.userPods.tolerations .Values.singleuser.extraTolerations .Values.prePuller.extraTolerations .Values.prePuller.taintmanager.tolerations }}
       tolerations:
         {{- . | toYaml | nindent 8 }}
       {{- end }}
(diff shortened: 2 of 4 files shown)


consideRatio:
Can you what you wish to accomplish, and how how it is accomplished in the PR description?

Also, the docker image - where is the Dockerfile and its bundled source code defined?

This will be a quite advanced PR to review, so we will need all the help we can get to consider this.

xcompass:
@consideRatio Please checkout the PR description. The docker image is linked in the description as well. Let me know if there is anything else you would like me to add.

xcompass:
Tagging @yuvipanda @donotpush @Lastcysa for visibility

xcompass:
I'll work on failed tests next a few days.

yuvipanda:
Ok, so I absolutely love the idea! I think if we can implement this cleanly, it's a major win to cold-start problems that exist. I too can't seem to find the source for the images, etc though - they should probably live in this repo as well, and be built by chartpress.

The two things to test are:

1. How does the *autoscaler* react when the taint is present? Does it go on an infinite loop of adding new nodes because the user nodes can't be scheduled on the tainted nodes? I think figuring this out is going to be important.
2. What is the failure mode? What happens when the code crashes, and leaves the taint in after the pull too? If the situation resolves itself soon (probably by autoscaler claiming the failed node), I think this is fine.
3. Is there a latency penalty where the user pod is scheduled onto the node before the tainting can happen?

Anyway, thank you so much for working on this!

consideRatio:
Thank you @xcompass, now I understand the strategy and think your PR description helps others understand the strategy.

I'd like review effort to remain at a very high level still. I'll switch back to #3010 to help narrow down exactly what you see as the problem that you want to address with this. I don't think the issue description captures that at the moment.

yuvipanda:
I would say that importing https://github.com/ubc/taintmanager into https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/main/images, and shaping it to be *as similar* to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/main/images/image-awaiter as possible would also help a lot with getting this reviewed. We have limited go expertise in the jupyterhub community, and making this new go thing be *as close* to the existing go thing we have will help a lot!

yuvipanda:
The trick that I'm happy to see here and I hadn't considered before is that because initContainers execute *sequentially*, we can use it to:

1. Taint nodes,
2. Pull user image
3. untaint node

Which massively simplifies the process vs a full controller as I was thinking of before.

yuvipanda:
sorry, I didn't see your comments @consideRatio! I'll let @xcompass provide more info now, as I'd assume they have a more up-to-date understanding of the problem right now than I do :)

images/taint-manager/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# syntax=docker/dockerfile:1
```
yuvipanda, line 1: Atleast to start with and to keep the amount of required go knowledge to a minimum, what do you think about doing this build the same way as https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/images/image-awaiter/Dockerfile? I think primarily that means using scratch as the final target rather than distroless.

images/taint-manager/taintmanager.go
```diff
@@ -0,0 +1,132 @@
...
+
+    if errAdd != nil && errRemove != nil {
+        log.Println("Please specify at least one option -add or -remove")
+        log.Println("Usage: taintmanager [-add TAINT] [-remove TAINT]")
+        flag.PrintDefaults()
+        os.Exit(1)
+    }
+
+    // creates the in-cluster config
+    config, err := rest.InClusterConfig()
```
yuvipanda, line 33: Can you fallback to kubeconfig here if inclusterconfig is not available? That would allow us to test this locally *much* more easily, and get rid of all the extra YAML bits and tiltfile required to simply test it.

yuvipanda (changes requested):
Thanks for adding those here! I am going to be off for a few days, but left a couple preliminary comments to simplify this further, keeping in mind our ability to maintain this long term with limited go expertise.

I think by falling back to kubeconfig auth if we aren't in cluster, we can get rid of the tiltfile as well as all the extra YAML bits needed for testing and do quick tests locally. We will need to add some tests that exercise a z2jh config with this tainting though.

Excited to see progress on this!

images/taint-manager/Dockerfile
xcompass: the reason I'm using distroless is to run taintmanager as nonroot user to be more secure. If it is not a concern, I can change it to scratch.

images/taint-manager/taintmanager.go
xcompass: Good idea. out-cluster kubeconfig and node name parameter are added.

images/taint-manager/test/clusterrole.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
yuvipanda, line 1: Given we can test this by running the taintmanager locally, are these test files still needed?

jupyterhub/values.yaml
```diff
@@ -559,6 +559,9 @@ scheduling:
         operator: Equal
         value: core
         effect: NoSchedule
+      - key: hub.jupyter.org/imagepulling
```
yuvipanda, line 562: image-pull-in-progress maybe?

yuvipanda (changes requested):
Thank you! My next question is how do you think we can write a unit / integration test for this here?

jupyterhub/schema.yaml
```diff
@@ -2754,6 +2754,20 @@ properties:
         properties:
           enabled:
             type: boolean
+      taintmanager:
+        type: object
+        additionalProperties: false
+        required: [enabled]
+        description: |
+          Manage the taints during the single user image prepulling. It will add
+          sepcified taint before image pulling and remove it after it's done.
```
tfmark, line 2763: "sepcified" -> "specified"

xcompass:
I'll bring it up-to-date people are still interested in it.

yuvipanda:
Yeah i still think this would be pretty useful if we can get it to work correctly!

PR #3051, opened 2023-03-05
Summarise linkcheck CI output

manics:
It takes too long to read through the entire output to find out what needs fixing


consideRatio (approved):
Great idea and 👍👍👍👍👍 for including a docstring like comment in the top of the script file!

I also appreciate how the solution was made a standalone step in the job

PR #3066, opened 2023-03-22
compile psycopg2 in hub image

minrk:
psycopg2-binary is [not recommended for production](https://www.psycopg.org/docs/install.html#change-in-binary-packages-between-psycopg-2-7-and-2-8), so compile and link to the host libpq.

I expected to need `libpq-dev` in build, too, but apparently it's already in the non-slim base image for some reason.

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -55,6 +55,8 @@ RUN apt-get update \
  && apt-get install -y --no-install-recommends \
         # requirement for pycurl
         libcurl4 \
+        # requirement for using postgres database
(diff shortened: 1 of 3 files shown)


consideRatio (approved):

consideRatio:
@minrk libpq-dev comes from `python:3.11-bullseye`'s [base image `buildpack-deps`, as defined here](https://github.com/docker-library/buildpack-deps/blob/598c0e2a0ff9ea3a4b8cb30e95007157d542da74/Dockerfile.template#L32).

PR #3068, opened 2023-03-24
ci: fix deprecation of set-output and use ubuntu 22.04 and py311

consideRatio:
- A deprecated `set-output` call was around apparently (https://github.com/jupyterhub/team-compass/issues/579)
- We used ubuntu-20.04, ubuntu-latest, and ubuntu-22.04, now ubuntu-22.04 systematically
- An small opionated refactoring of a script snippet

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,7 +71,8 @@ jobs:
           ):
               publishing = "true"
               print("Publishing chart")
-          print(f"::set-output name=publishing::{publishing}")
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
+              f.write(f"publishing={publishing}\n")
 
       - name: Set up QEMU (for docker buildx)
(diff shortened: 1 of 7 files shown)


.github/workflows/publish.yml
```diff
@@ -44,34 +44,29 @@ jobs:
...
-          EVENT: ${{ github.event_name }}
-          REF: ${{ github.event.ref }}
         run: |
           import os
-          repo = os.environ["REPO"]
-          event = os.environ["EVENT"]
-          ref = os.environ["REF"]
+          repo = "${{ github.repository }}"
+          event = "${{ github.event_name }}"
+          ref = "${{ github.event.ref }}"
```
consideRatio, line 56: Provides a bit more readability when inspecting this, and now that we write to a file based on `GITHUB_OUTPUT`, this is no longer safe to run locally anyhow when it could be nice to replace env vars etc.

PR #3078, opened 2023-03-31
Refactor of image-awaiter's dockerfile

alekseyolg:
small revision of the dockerfile - removed the extra layer


consideRatio (approved):
The changes are in the build stage of this Dockerfile and shouldn't infleunce the final stage that starts from `scratch`, but its a simplification so lets go for it!

PR #3079, opened 2023-03-31
doc: singleuser.uid default is always 1000

manics:
The `singleuser.uid` documentation refers to https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.uid which says
> If set to None, the user specified with the USER directive in the container metadata is used.

In Z2JH the UID always defaults to 1000.

Alternatively, could we unset
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/26fb33da64b4914a192e37e03156eb0f9e72d863/jupyterhub/values.yaml#L367
so it uses the image's default?

### DIFF
--- a/jupyterhub/schema.yaml
+++ b/jupyterhub/schema.yaml
@@ -2343,6 +2343,8 @@ properties:
           root, enabling it from the container in a startup script, and then
           transitioning to the normal user.
 
+          Default is 1000, set to null to use the container's default.
+
   scheduling:
     type: object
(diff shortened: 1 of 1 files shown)


consideRatio:
> Alternatively, could we unset

I think us sticking with 1000 as a default is a good convention for z2jh at the moment. To use the image's default would open up a box of complexity with little benefit I figure.

PR #3084, opened 2023-04-04
Bump to 3.0.0dev

yuvipanda:
We're already on JupyterHub 4.x, so this 
makes sense from a semver perspective. Useful
in generating better versions in https://hub.jupyter.org/helm-chart/.

Credit to @pnasrat for pointing this out.

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -13,7 +13,7 @@ charts:
     # Dev: imagePrefix can be useful to override if you want to trial something
     #      locally developed in a remote k8s cluster.
     imagePrefix: jupyterhub/k8s-
-    baseVersion: "2.0.1-0.dev"
+    baseVersion: "3.0.0-0.dev"
     repo:
       git: jupyterhub/helm-chart
       published: https://jupyterhub.github.io/helm-chart


yuvipanda:
Even if we are not on JupyterHub 4.x, bumping to next version.dev after each release makes sense, and we do that for our other projects.

consideRatio:
I think the test failures are intermittent, I'll re-run the tests later!

consideRatio:
I figure the pattern should be to bump baseVersion to next minor with `-0.dev` suffix if we have merged a new feature, or next major with `-0.dev` suffix if we have merged a breaking change.

yuvipanda:
@consideRatio yep makes sense re: pattern. Should we document that someplace?

chartpress.yaml
```diff
@@ -13,7 +13,7 @@ charts:
     # Dev: imagePrefix can be useful to override if you want to trial something
     #      locally developed in a remote k8s cluster.
     imagePrefix: jupyterhub/k8s-
-    baseVersion: "2.0.1-0.dev"
+    baseVersion: "3.0.0-0.dev"
```
consideRatio, line 16:
```suggestion
    # baseVersion should be a -0.dev suffixed version, where the version should
    # be the next major, minor, or patch version depending on what we have
    # merged so far into the main branch. If for example we have merged a
    # breaking change it should be the next major version, like 3.0.0-0.dev.
    baseVersion: "3.0.0-0.dev"
```

consideRatio:
Thank you @pnasrat and @yuvipanda!!

PR #3134, opened 2023-06-06
Adding support for extraManifests for the chart

JunaidChaudry:
Adding an additional feature to support `extraManifests`, which allows end-users to deploy any additional k8s resources that they may want to manage along with the helm chart deployment. 

Some of the really common use-cases include creating a secret (or secretProviderClass) which the users might be passing on to the hub, or a configmap creation which might then be used with `extraVolume`, or anything else for that matter.

### DIFF
--- a/jupyterhub/templates/extraManifests.yaml
+++ b/jupyterhub/templates/extraManifests.yaml
@@ -1,0 +1,4 @@
+{{ range .Values.extraObjects }}
+---
+{{ tpl (toYaml .) $ }}
+{{ end }}
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -2926,7 +2926,7 @@ properties:
           deprecation message that contain censored content that you wish to
           reveal.
 
-  extraObjects: 
+  extraObjects:
     type: array
     description: |
       Any additional K8s manifests to deploy with the chart
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -2925,3 +2925,8 @@ properties:
           A flag that should only be set to true temporarily when experiencing a
           deprecation message that contain censored content that you wish to
           reveal.
+
+  extraObjects: 
+    type: array
+    description: |
+      Any additional K8s manifests to deploy with the chart
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -662,23 +662,23 @@ global:
 
 # -- Array of extra K8s manifests to deploy
 extraObjects: []
-  # example: 
-    # - apiVersion: secrets-store.csi.x-k8s.io/v1
-    #   kind: SecretProviderClass
-    #   metadata:
-    #     name: singleuser-hub-secrets-store
-    #   spec:
-    #     provider: aws
-    #     parameters:
-    #       objects: |
-    #         - objectName: "singleuser-secret"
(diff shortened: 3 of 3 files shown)


JunaidChaudry:
@consideRatio  Can you please take a look at this feature request as well?

consideRatio:
I'd say this is out of scope for the helm chart to suppor this feature. If you want to provide extra manifests etc, you may end up also wanting to provide entire helm charts as well.

The pattern I'd recommend to handle this situation is to create a new chart and add extra things to it, then let it depend on the jupyterhub chart, and install that. This is for example [what jupyterhub/mybinder.org-deploy does](https://github.com/jupyterhub/mybinder.org-deploy/tree/main/mybinder), its a repo with the deployment configuration for https://mybinder.org. It defines a helm chart that depends on the binderhub helm chart and other helm charts.

Another more lightweight pattern would be to simply do `kubectl apply -f some-k8s-resource-manifest.yaml` once.

ofirshtrull:
I also think this is relevant @consideRatio 
in today's world everything should get fully Gitops I manage x cluster prod and dev, to advise that running a single `kubectl apply -f` is bad practice in today's world 

and to have an umbrella chart only is a big overhead instead of just adding a manifest or two to the main one

please lets add this support like all the major 3rd party charts it will be very helpfull

JunaidChaudry:
Any updates here @consideRatio

aBandicoot:
this is a very important feature, and would help us a lot with managing everything (like secrets) in one place with our jupyterhub deployment values

frittentheke:
Introducing a values variable that could contain any type of other resource is honestly messy antipattern in the world of Helm charts and templating. If people want to add additional manifests to this (or a) chart, the Helm way would be to create a parent chart and then add this chart as subchart. This then allows to use a single values file (https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), proper schemas and validation for all of the resources.

ofirshtrull:
> Introducing a values variable that could contain any type of other resource is honestly messy antipattern in the world of Helm charts and templating. If people want to add additional manifests to this (or a) chart, the Helm way would be to create a parent chart and then add this chart as subchart. This then allows to use a single values file (https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), proper schemas and validation for all of the resources.

I disagree, this option makes users maintain another chart just to add a secret 
many mainstream charts are starting to use this cool trick to help such users 
i.e a few of them 
- https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml
- https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/extra-manifests.yaml
- https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/extra-objects.yaml

this helps keep using the main chart and just add dependent yamls without the mess of wrapper charts and dependencies

frittentheke:
Let's say there is a middle ground, as in adding something basic like a single secret or some Prometheus rule, where "your" approach is more lightweight yes. Maybe calling it an anti-pattern was a little harsh.

I'd still argue that more extensive additions, should rather be their own chart (a parent).

manics:
Sorry for leaving this for so long. Clearly there's good arguments for having and not having it. I'll close this for now.

We can revisit this in future, though now that [kustomize](https://github.com/kubernetes-sigs/kustomize) is getting more popular that might be a better way to add customisations to this Helm chart.

PR #3137, opened 2023-06-12
Update changelog for 3.0.0-beta.1

consideRatio:
We made an alpha.1 release, and I was thinking it should be alpha.1 because we were to also add oauthenticator 16 to that later - another breaking change.

I figure now, in order to unblock this release and let's not force a timeline on oauthenticator 16 and instead do another major version release once oauthenticator 16 lands.

I'd like to see us release 3.0.0 or so this friday or early next week.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -158,7 +158,7 @@ jobs:
           #
           # The upgrade-from input should be a chart version, or match the version
           # information from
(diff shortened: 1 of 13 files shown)


minrk (approved):

yuvipanda (approved):

yuvipanda:
Generally +1 on not blocking this on oauthenticator 16

PR #3142, opened 2023-06-28
Enhance keycloak configuration example

LucasVanHaaren:
Hello everybody,

I just started using JupyterHub and I had hard times understanding how to configure keycloak autorisation in order to use keycloak's client roles as JupyterHub internal's groups.
What I wanted to achieve is to have only keycloak users which have `user` role assigned allowed to connect and use Jupyter and also defining Jupyter's administrator through keycloak via an `admin` role.

I have to sneak into a tons of threads and post and no one describing how to do this. 
I finally had to go through the OAuthenticator documentation to find an example of code which achieve this (see [the docs](https://github.com/jupyterhub/oauthenticator/tree/main/examples/generic#example-configuration-with-authorization-enabled)).

I think anybody who wants to deploy it in production alongside keycloak auth wants to have this type of authorization (maybe I'm wrong). So I just added the mandatory lines in the example in docs : 

```yaml
hub:
  config:
    GenericOAuthenticator:
      [ ... ]
      claim_groups_key: roles
      allowed_groups:
        - user
      admin_groups:
        - admin
```

Maybe in addition we can link this example in the zero-to-jupyterhub-k8s docs ? What do you think about this ?

Thanks in advance

### DIFF
--- a/docs/source/administrator/authentication.md
+++ b/docs/source/administrator/authentication.md
@@ -397,6 +397,12 @@ hub:
       username_key: preferred_username
       userdata_params:
         state: state
+      # In order to use keycloak client's roles as autorisation layer
+      claim_groups_key: roles
+      allowed_groups:
+        - user
+      admin_groups:
+        - admin
     JupyterHub:
       authenticator_class: generic-oauth
 ```
--- a/docs/source/administrator/authentication.md
+++ b/docs/source/administrator/authentication.md
@@ -397,7 +397,7 @@ hub:
       username_key: preferred_username
       userdata_params:
         state: state
-      # In order to use keycloak client's roles as autorisation layer
+      # In order to use keycloak client's roles as authorization layer
       claim_groups_key: roles
       allowed_groups:
         - user


docs/source/administrator/authentication.md
```diff
@@ -397,6 +397,12 @@ hub:
       username_key: preferred_username
       userdata_params:
         state: state
+      # In order to use keycloak client's roles as autorisation layer
```
consideRatio, line 400:
```suggestion
      # In order to use keycloak client's roles as authorization layer
```

consideRatio:
Thank you @LucasVanHaaren for upstreaming your insights!

I think this overall should live in jupyterhub/oauthenticator, and this project as a distribution of jupyterhub - like for example jupyterhub/the-littlest-jupyterhub also is, overall instead should reference. That way we can document this for:
- this distribution
- the-littlest-jupyterhub distribution
- anyone setting up jupyterhub directly

But overall, I figure we should just accept this PR as an incremental improvement for now, and in the future we may be able to relocate this docs to jupyterhub/oauthenticator.

consideRatio (approved):
Thank you @LucasVanHaaren!!! :heart: :tada: :sunflower:

LucasVanHaaren:
> I think this overall should live in jupyterhub/oauthenticator, and this project as a distribution of jupyterhub - like for example jupyterhub/the-littlest-jupyterhub also is, overall instead should reference. [ ... ]

You're absolutely right, this information should be in the OAuthenticator repo too!

Thanks for your prompt reply, I'll take a look at the OAuthenticator doc in the next few days, to see if I can improve this point.

consideRatio:
> Thanks for your prompt reply, I'll take a look at the OAuthenticator doc in the next few days, to see if I can improve this point.

I actually have a massive documentation PR I'm about to open updating a lot of documentation there, to avoid conflicts, please wait a bit :)

PR #3149, opened 2023-07-05
Add changelog for 3.0.0-beta.2

consideRatio:
I updated the chart to include oauthenticator 16, and that merits another beta release I think because its another breaking change.


minrk (approved):
👍 for beta 2

PR #3162, opened 2023-07-19
Allow `enabled` config, for use by charts depending on this chart conditionally

### LINKED ISSUE #3160: Update values schema to allow key `enabled`
### Proposed change

Update values schema to allow key `enabled` so the z2jh helm chart can be included as a dependency in other helm charts and be enabled using the preffered syntax

Usually, when including a chart as a helm dependency, the chart can be turned on and off by setting the charts `enabled` key to `true` or `false`.

This is the preffered method according to the [helm docs](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags)

For example, if jupyterhub is a helm dependency, then it would look like this in the parent chart's `values.yaml` file:
```yaml
jupyterhub:
    enabled: true
    hub:
        ...
```

As of now, if you try to include the `enabled` key in a values.yaml, the jupyterhub values scheme will throw a validation error:
```
failed to upgrade release '<name>' from chart '<chart>': values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- (root): Additional property enabled is not allowed
```

The values schema should be updated to allow for the enabled key so z2jh can be used as a dependency

### Alternative options

Unsure

Users who want to incorporate jupyterhub as a dependency in their own helm charts

Solution in PR #3162

monoakg:
Linked issue #3160 - Update values schema to allow key `enabled`

Closes #3160

### DIFF
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -28,7 +28,7 @@ required:
   - global
 properties:
   enabled:
-    type: boolean
+    type: [boolean, "null"]
     description: |
       enabled allows you to enable or disable jupyterhub using the preffered method
       if it is included as a dependency in a parent helm chart
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -27,6 +27,11 @@ required:
   - rbac
   - global
 properties:
+  enabled:
+    type: boolean
+    description: |
+      enabled allows you to enable or disable jupyterhub using the preffered method
+      if it is included as a dependency in a parent helm chart
   fullnameOverride:
     type: [string, "null"]
     description: |
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -30,8 +30,9 @@ properties:
   enabled:
     type: [boolean, "null"]
     description: |
-      enabled allows you to enable or disable jupyterhub using the preferred method
-      if it is included as a dependency in a parent helm chart
+      `enabled` is ignored by the jupyterhub chart itself, but a chart depending
+      on the jupyterhub chart conditionally can make use this config option as
+      the condition.
   fullnameOverride:
     type: [string, "null"]
     description: |
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -30,7 +30,7 @@ properties:
   enabled:
     type: [boolean, "null"]
     description: |
-      enabled allows you to enable or disable jupyterhub using the preffered method
+      enabled allows you to enable or disable jupyterhub using the preferred method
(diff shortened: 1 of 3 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -27,6 +27,11 @@ required:
   - rbac
   - global
 properties:
+  enabled:
+    type: [boolean, "null"]
+    description: |
+      enabled allows you to enable or disable jupyterhub using the preffered method
```
manics, line 33:
```suggestion
      enabled allows you to enable or disable jupyterhub using the preferred method
```

jupyterhub/values.yaml
```diff
@@ -3,6 +3,9 @@
 fullnameOverride: ""
 nameOverride:
 
+# enable or disable jupyterhub if it is included as a dependency in a parent helm chart
+enabled:
```
manics, line 7: What do other charts typically do? Do they leave this null, or default to `true`?

consideRatio: I suggest its made null as default here, true/false has no meaning by this chart anyhow so making it null seems less confusing, and both false/true and can be defaults for a chart depending on this - such chart should define enabled to false/true explicitly.

consideRatio: Most charts dont need an entry for this, but we have a strict schema.

jupyterhub/values.schema.yaml
```diff
@@ -27,6 +27,11 @@ required:
   - rbac
   - global
 properties:
+  enabled:
+    type: [boolean, "null"]
+    description: |
+      enabled allows you to enable or disable jupyterhub using the preferred method
+      if it is included as a dependency in a parent helm chart
```
consideRatio, line 34:
```suggestion
      `enabled` is ignored by the jupyterhub chart itself, but a chart depending
      on the jupyterhub chart conditionally can make use this config option as
      the condition.
```

jupyterhub/values.yaml
```diff
@@ -3,6 +3,9 @@
 fullnameOverride: ""
 nameOverride:
 
+# enable or disable jupyterhub if it is included as a dependency in a parent helm chart
```
consideRatio, line 6:
```suggestion
# enabled is ignored by the jupyterhub chart itself, but a chart depending on
# the jupyterhub chart conditionally can make use this config option as the
# condition.
```

consideRatio (approved):
Thanks @monoakg!!

From experience and examples [from helm's documentation](https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies), its clear that `enabled` indeed is the common naming choice for this.

I've updated the language to make it really clear that this is a feature implemented in a chart conditionally depending on this chart

Issue #3161, opened 2023-07-19
resourceLock = endpoints, endpointsleases or leases ?


yilas:
### Bug description

That issue is close to the issue https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3128.

The value of `resourceLock` used is not correct. Indeed, I have that value in the ConfigMap `user-scheduler` : 

```yaml
apiVersion: v1
data:
  config.yaml: |
    apiVersion: kubescheduler.config.k8s.io/v1beta3
    kind: KubeSchedulerConfiguration
    leaderElection:
      resourceLock: endpointsleases
      resourceName: user-scheduler-lock
      resourceNamespace: "jupyterhub"
(...)
```

but the log from the pod `user-scheduler` returns : 

```bash
(...)
I0719 09:20:29.519934       1 serving.go:348] Generated self-signed cert in-memory
I0719 09:20:29.521209       1 configfile.go:59] "KubeSchedulerConfiguration v1beta3 is deprecated in v1.26, will be removed in v1.29"
E0719 09:20:29.521499       1 run.go:74] "command failed" err="leaderElection.resourceLock: Invalid value: \"endpointsleases\": resourceLock value must be \"leases\""
```

#### Expected behaviour

That value shoud be `leases` as returned by the log.

The value is [hardcoded](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blame/main/jupyterhub/templates/scheduling/user-scheduler/configmap.yaml#L32). Maybe a comparison with the k8s version should be done like this ⬇️  but I do not know from which version of k8s the value `leases` should be used.

```
    {{- if semverCompare ">=1.24.0-0" .Capabilities.KubeVersion.Version }}
    apiVersion: kubescheduler.config.k8s.io/v1
    {{- else }}
```

#### Actual behaviour

Change the value of `resourceLock` to `leases`.

### Your personal set up

Here is the version of k8s (eks) I'm using : 

```bash
clientVersion:
  buildDate: "2023-06-14T09:53:42Z"
  compiler: gc
  gitCommit: 25b4e43193bcda6c7328a6d147b1fb73a33f1598
  gitTreeState: clean
  gitVersion: v1.27.3
  goVersion: go1.20.5
  major: "1"
  minor: "27"
  platform: linux/amd64
kustomizeVersion: v5.0.1
serverVersion:
  buildDate: "2023-06-16T17:32:40Z"
  compiler: gc
[...]

PR #3165, opened 2023-07-20
Adapt resourceLock according to k8s version

yilas:
Issue from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3161


consideRatio (changes requested):
I'm not ready to merge this because I suspect this may not be the right fix. Discussion needs to be resolved in #3161 first.

consideRatio:
Preliminary closing this, but we can re-open easily!

Issue #3167, opened 2023-07-21
Installation on GKE (standard & autopilot) fails with Dataplane V2 & kube-dns(Cilium)


vizeit:
### Bug description
This is the continuation of the ticket [3163](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3163). It was closed with the comment to search on https://discourse.jupyter.org for similar issues. This issue reported here did not match with any existing discussions on the discourse for error 599.
The issue is that when GKE cluster, either standard or Autopilot, is created with Dataplane V2, the hub fails to start with the following error,

```
[W 2023-07-21 17:35:47.524 JupyterHub proxy:899] api_request to the proxy failed with status code 599, retrying...
[E 2023-07-21 17:35:47.525 JupyterHub app:3382]
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/app.py", line 3380, in launch_instance_async
        await self.start()
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/app.py", line 3146, in start
        await self.proxy.get_all_routes()
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/proxy.py", line 946, in get_all_routes
        resp = await self.api_request('', client=client)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/proxy.py", line 910, in api_request
        result = await exponential_backoff(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/utils.py", line 237, in exponential_backoff
        raise asyncio.TimeoutError(fail_message)
    TimeoutError: Repeated api_request to proxy path "" failed.
```

#### Expected behaviour

The installation should complete successfully

#### Actual behaviour

The installation fails with hub pod in crashbackoff state. The logs show the following error message,
api_request to the proxy failed with status code 599, retrying...

### How to reproduce

GKE Autopilot:
1. Create GKE autopilot cluster (Dateplane V2 & kube-dns is enabled by default) 
2. Install Jupyterhub
```
helm upgrade --cleanup-on-fail   --install testjupyterhub jupyterhub/jupyterhub   --namespace testjupyterhubdev   --create-namespace   --version=3.0.0-beta.3   --values config.yaml
```
config.yaml content:

[...]

PR #3185, opened 2023-08-02
Let singleuser.cloudMetadata.blockWithIpTables only block TCP port 80

### LINKED ISSUE #3180: Let `singleuser.cloudMetadata.blockWithIptables` block HTTP ports only
It seems that the cloud metadata server can also be an IP for DNS queries (GKE clusters with Cloud DNS configured), and I don't think we should block these with `singleuser.cloudMetadata.blockWithIptables` by default.

I suggest we just block the HTTP ports, 80 and possibly also 443. In practice it seems only port 80 is used but I'm not 100% sure. I've searched the internet and found no examples of some other port than port 80 being exposed to serve the metadata API.

If we do this, what we ship with by default in 3.0.0 would work for GKE clusters with Cloud DNS, GKE clusters with kube-dns and nodelocal cache, with GKE/EKS/AKS clusters with pods in k8s clusters, with Cilium (GKE dataplane v2) as a network policy enforcer - and no additional config exceptions would be required.

Currently without this, users of GKE's Cloud DNS would need to disable all blocking of the cloud metadata server, including the metadata API, just to still be able to access the DNS on the same IP.

### LINKED ISSUE #3184: Addressing review feedback about docs for `singleuser.cloudMetadata.blockWithIpTables`
I merged #3179 without addressing a comment by @yuvipanda (below) since resolving #3180 would influence that specific documentation anyhow. So, this issue represents the idea to address Yuvi's comment, but after acting on #3180.

> One thing I'd like added is a reference to `dnsPortsCloudMetadataServer` in the documentation for `blockWithIptables`, saying that even when it is true, DNS access is allowed and controlled by `dnsPortsCloudMetadataServer`. With that, this looks good to me.

The gist I'm putting it off to document this quickly in #3179 is because I'm not confident on the general truth of the because: I don't understand iptables that well, and partially because the network policy enforcement can be done in different ways and I'm not sure about the consequences of that. For example Calico use iptables (apparently prepended rules rather than appended rules by default), and Cilium use eBPF to manage networking permissions etc. How does all of this couple?

consideRatio:
- Closes #3180
  By blocking only TCP Port 80 on the cloud metadata server where the sensitive metadata API is exposed, as compared to all traffic which could include relevant DNS traffic for example when a GKE cluster is configured to use Cloud DNS instead of kube-dns (default).
- Closes #3184
  By adding a versionchanged note, and mentioning TCP Port 80 specifically in docs, and by failing if a config is ambiguous by both blocking via iptables and allowing access via NetworkPolicy.
  
### Notes thinking about this

- I've learned we can do `--destination-ports`, in plural instead of singular, allowing us to specify 80 and 443 (EDIT: didn't get this to work)
- Is the metadata API by GKE/EKS/AKS exposed only on TCP port 80? I'm not 100%, but it seems so.
  - EDIT: GKE, EKS, AKS doesn't have anything on port 443 in clusters I inspected
- Should we block UDP port 80 as well since there are things like QUIC protocol to send HTTP traffic also etc, or is it safe to just assume there is no other ports open etc?
- I've learned we can use an exclamation mark before a flag to say the opposite, like `! --destination-port 53` to block traffic except to port 53.
  - EDIT: But this makes us need to specify a --protocol, either tcp or udp, one at at the time, not `all` or similar.

I currently think that blocking only traffic to 169.254.169.254 TCP port 80 this is a clean solution thats easy to document, but I'd like quite good faith in that. Going to check this against GKE EKS AKS

### DIFF
--- a/docs/source/administrator/security.md
+++ b/docs/source/administrator/security.md
@@ -279,23 +279,26 @@ only need one.
 
 (block-metadata-netpol)=
 
-### Block metadata with a NetworkPolicy enforced by a NetworkPolicy controller
+### Block cloud metadata API with a NetworkPolicy enforced by a NetworkPolicy controller
 
-If you have _NetworkPolicy controller_ such as Calico in the Kubernetes cluster,
-it will enforce the NetworkPolicy resource created by this chart
-(`singleuser.networkPolicy.*`) that blocks user access to the metadata server.
+If you have _NetworkPolicy controller_ such as Calico or Cilium in the
+Kubernetes cluster, it will enforce the NetworkPolicy resource created by this
+chart (`singleuser.networkPolicy.*`) that blocks (by never allowing) user access
+to the cloud metadata API exposed on TCP port 80 on the IP `169.254.169.254`.
+
 We recommend relying on this approach if you you had a NetworkPolicy controller,
 and then you can disable the other option.
 
 (block-metadata-iptables)=
 
-### Block metadata with a privileged initContainer running `iptables`
+### Block cloud metadata API with a privileged initContainer running `iptables`
 
-If you can't rely on the NetworkPolicy approach to block access to the metadata
-server, we suggest relying on this option. When
+If you can't rely on the NetworkPolicy approach to block access to the cloud
+metadata API on TCP port 80, we suggest relying on this option instead. When
 `singleuser.cloudMetadata.blockWithIptables` is true as it is by default, an
 `initContainer` is added to the user pods. It will run with elevated privileges
-and use the `iptables` command line tool to block access to the metadata server.
+and use the `iptables` command line tool to block access to the cloud metadata
+API on port 80.
 
 ```yaml
 # default configuration
@@ -303,6 +306,13 @@ singleuser:
   cloudMetadata:
     blockWithIptables: true
     ip: 169.254.169.254
+```
+
+```{versionchanged} 3.0.0
+Since 3.0.0 this rule only blocks TCP port 80, and this configuration is no longer allowed
+to be configured true at the same time as
+[`singleuser.networkPolicy.egressAllowRules.cloudMetadataServer`](schema_singleuser.networkPolicy.egressAllowRules.cloudMetadataServer)
+is configured true to avoid ambiguous configurations.
 ```
 
 (netpol)=
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -22,6 +22,10 @@ This is a beta release for testing before the 3.0.0 release.
 Since 3.0.0-beta.1 release 2023-06-12, another breaking change was made by
 upgrading OAuthenticator from version 15.1.0 to 16.0.0. Please read to the
 [OAuthenticator changelog]'s breaking changes before upgrading from 3.0.0-beta.1.
+
+Since 3.0.0-beta.3 release 2023-07-06, default networking rules related to the
+cloud metadata API and DNS ports has changed slightly as documented in the
+breaking changes below.
 ```
 
 #### Breaking changes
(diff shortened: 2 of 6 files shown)


manics:
> I've learned we can use an exclamation mark before a flag to say the opposite, like ! --destination-port 53 to block traffic except to port 53

This sounds safer and less likely to have unexpected consequences, but I'm happy to go with the consensus.

consideRatio:
Its safer, but its harder to get done. To use `! --destination-port 53`, I have to specify the protocol, and I can't use `all` because then the flag to specify destination-port isn't available. So, what I'll end up specifying is `tcp` or `udp` for which the flag is available.

I've created three rules below inside a dummy container started like `docker run -it --rm --privileged alpine:3`. Maybe there are some variations to consider here, I'm not confident enough about `iptables`.

```
# what we currently do
iptables --append OUTPUT --jump DROP --destination 169.254.169.254

# options to what we currently do
# option 1 - drop just TCP port 80, this is currently in PR #3184
iptables --append OUTPUT --protocol tcp --jump DROP --destination 169.254.169.254 --destination-port 80

# option 2 - drop either only TCP or TCP and UDP for all ports except 53
iptables --append OUTPUT --protocol tcp --jump DROP --destination 169.254.169.254 ! --destination-port 53
iptables --append OUTPUT --protocol udp --jump DROP --destination 169.254.169.254 ! --destination-port 53
```

```
/ # iptables --list
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
DROP       tcp  --  anywhere             169.254.169.254      tcp dpt:http
DROP       all  --  anywhere             169.254.169.254     
DROP       tcp  --  anywhere             169.254.169.254      tcp dpt:!domain
DROP       udp  --  anywhere             169.254.169.254      udp dpt:!domain
```

consideRatio:
Looking through notes in https://book.hacktricks.xyz/pentesting-web/ssrf-server-side-request-forgery/cloud-ssrf, and testing that only port 80 seems to be relevant among 80 and 443 in GKE/EKS/AKS by practical tests, I'm starting to feel confident on just blocking TCP 80.

I'll ask Yuvi later today about this.

consideRatio:
I think that strategy is quite clean, so i appreciate it more than others in a way. But, would doing this possibly make port 53 access to that ip allowed when it perhaps isnt to be 
allowed though, and is that a problem?

consideRatio:
## Options overview

0. Let `iptables` remain having a "drop all" rule (current behavior)
1. Let `iptables` add two accept rules for TCP/UDP port 53 before the the "drop all" rule (see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3185#issuecomment-1662383005)
2. Let `iptables` not add a "drop all" rule, but instead a drop TCP port 80 rule (currently in PR 3185)
3. Let `iptables` add a "drop all" rule excluding port 53 for TCP or for TCP/UDP (see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3185#issuecomment-1662346757)

consideRatio:
I chatted with Yuvi who were inclined towards the safest approach, and a varying behavior based on configuration. I was agreeing when we chatted yesterday. Now I'm having doubts again thinking about the growing complexity of things taking this path. The growing complexity stems from use of `iptables`, and it makes me re-think committing further to use of `iptables`.

By working with `iptables` in the first place, we can't predict or sustainably research and document the interactions with different NetworkPolicy enforcements. Due to that, I think we shouldn't couple further to it, but instead just declare `singleuser.cloudMetadata.blockWithIpTables` as a crude failsafe and highlight NetworkPolicy enforcement as the way to go. That `blockWithIpTables` should be disabled in favor of having a NetworkPolicy controller, and rely on the charts NetworkPolicy rules that doesn't allow user servers to communicate with the cloud metadata server besides using port 53 by default.

What's clear to me is that I'm not comfortable with locking us into further maintenance effort involving `iptables` low level details and interactions with NetworkPolicy rules and its varying enforcement implementations.

manics:
I'm happy with deprecating (though not removing) iptables, and saying that
- everyone _should_ use network policies
- anyone with non-default requirements _must_ use network policies

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3167 is related to Dataplane V2. According to
https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#overview
> Network policy enforcement is built into GKE Dataplane V2. You do not need to enable network policy enforcement in clusters that use GKE Dataplane V2.

So we know that issue can be safely resolved without iptables.

manics:
A belated thought regarding `dnsPortsCloudMetadataServer`- how likely is it that other ports will be required in the future? Is it worth making this not specific to DNS, and taking a list of ports, so that we don't need to add another config option in future?

consideRatio:
> A belated thought regarding `dnsPortsCloudMetadataServer`- how likely is it that other ports will be required in the future? Is it worth making this not specific to DNS, and taking a list of ports, so that we don't need to add another config option in future?

I'm happy with the choice to provide a pre-defined rule specifically for 169.254.169.254:53 and not going more general in this case, thinking the combination of simple pre-defined boolean togglable rules + full freedom `egress` config does the trick quite well.

PR #3208, opened 2023-08-22
Add 2i2c as an institutional supporter

yuvipanda:
I think there are now multiple people contributing significantly to z2jh from 2i2c, and I'd love for this callout to be here!

I'd also love for other contributors to add their own institutions here :)


yuvipanda:
I would definitely like someone *not* working at 2i2c to merge this, if deemed mergeworthy!

consideRatio (approved):

PR #3217, opened 2023-09-11
Pull images from `singleuser.profileList` found in `profile_options.choices`

manfuin:
Currently prePuller, when it comes to obtaining list of images from `singleuser.profileList`, does not support the [profile_options](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list) way of redefining the images. 

This PR aims to add such a support. 

Current mitigation without such a support, is to define images second time in `prePuller.extraImages`, but this is not really convenient and overcomplicates maintenance of the actual profile list.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -175,6 +175,7 @@ with open("resources/reference.md", "w") as f:
 intersphinx_mapping = {
     "jupyterhub": ("https://jupyterhub.readthedocs.io/en/stable/", None),
     "oauthenticator": ("https://oauthenticator.readthedocs.io/en/stable/", None),
+    "kubespawner": ("https://jupyterhub-kubespawner.readthedocs.io/en/stable/", None),
 }
 
 # intersphinx_disabled_reftypes set based on recommendation in
--- a/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
+++ b/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
@@ -134,6 +134,7 @@ spec:
         {{- /* --- Conditionally pull profileList images --- */}}
         {{- if .Values.prePuller.pullProfileListImages }}
         {{- range $k, $container := .Values.singleuser.profileList }}
+        {{- /* profile's kubespawner_override */}}
         {{- if $container.kubespawner_override }}
         {{- if $container.kubespawner_override.image }}
         - name: image-pull-singleuser-profilelist-{{ $k }}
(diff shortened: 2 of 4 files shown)


yuvipanda:
ooo, I actually really love this! It's very clever and solves a problem pretty well :)

How do you think we can test this, @manfuin?

consideRatio:
### About testing

> Correct me if I am wrong, but I see no mentions of either `prePuller` or `profile_options` in the existing tests, so I don't think it is easy to add this specific feature testing without developing a common tests base for prePuller.

There are the lint-and-validate-values.yaml file, it is used here in the tests:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/84ebc254a8854ade3c1ed2f6f72243ba5be59329/.github/workflows/test-chart.yaml#L55-L59

The lint-and-validate.py script's docstring:

> Lints and validates the chart's template files and their rendered output without
> any cluster interaction. For this script to function, you must install yamllint.

This doesn't ensure that the rendered resources are valid k8s resources though, but the test below doest against a k8s api-server using `helm template`.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/84ebc254a8854ade3c1ed2f6f72243ba5be59329/.github/workflows/test-chart.yaml#L252-L260

To make use of these, just populate `lint-and-validate-values.yaml` with singleuser.profileList entries that makes sense to test this PR.

You could also manually inspect the changes like this:

```
helm template jupyterhub ./jupyterhub --values tools/templates/lint-and-validate-values.yaml --show-only templates/image-puller/daemonset-continuous.yam
```

consideRatio (commented):
This makes sense, thank you @manfuin!

### Action points

- [x] Add an entry or two with profile_options in lint-and-validate-values.yaml, where at least one includes also the `unlisted_choice` functionality so we validate not try pre-pull an invalid image specification from that.
- [ ] Opinie and decide about configuration
  Should we add a configuration besides [`prePuller.pullProfileListImages`](https://z2jh.jupyter.org/en/stable/resources/reference.html#prepuller-pullprofilelistimages) focused around pulling these profile_option specified images, or let it follow from having `prePuller.pullProfileListImages` set to true?
- [ ] Update the configuration reference for [`prePuller.pullProfileListImages` found in in schema.values.yaml](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/84ebc254a8854ade3c1ed2f6f72243ba5be59329/jupyterhub/values.schema.yaml#L2850-L2861) one way or another to mention image's specified via `profile_options`.

manfuin:
Thanks for the good action points, @consideRatio!

I have added entry to `lint-and-validate-values.yaml`.

I was thinking about necessity to turn this functionality separately during template rewrite already, but can't imagine a real use-case, when one wants to pre-pull only subset of the images, Jupyterhub installation is offering to end-users. Especially based on the way (profile_options vs generic override) they defined.
So in my opinion following the [prePuller.pullProfileListImages](https://z2jh.jupyter.org/en/stable/resources/reference.html#prepuller-pullprofilelistimages) is enough.

consideRatio:
Thank you @manfuin!

I pushed a commit to update the docs as well, and with that this looks good to me to merge.

consideRatio (approved):

PR #3259, opened 2023-10-27
Move note box to before list of cloud providers.

rgaiacs:
Because we expect the readers to click in one of the cloud providers, the note box should come before the list of links.


consideRatio (approved):

consideRatio:
Thank you @rgaiacs!!! ❤️ 🎉

PR #3260, opened 2023-10-27
Document k8s cluster setup using minikube (for learning and development)

rgaiacs:
As an alternative for people that want to learn more about BinderHub.

### DIFF
--- a/docs/source/kubernetes/minikube/step-zero-minikube.md
+++ b/docs/source/kubernetes/minikube/step-zero-minikube.md
@@ -1,12 +1,14 @@
 (minikube-k8s)
 
-# Kubernetes on minikube (for trial and development only)
+# Kubernetes on minikube (for learning and development only)
 
-[minikube](minikube) can setup a Kubernetes cluster on a single computer. Using minikube can be suitable trial for new Kubernetes users and for development purposes, but its not meant to be used for production purposes.
+[minikube] can setup a Kubernetes cluster on a single computer. minikube be suitable in order to learn about Kubernetes and to develop and test changes, but its not meant to be used for production purposes.
 
 ```{important}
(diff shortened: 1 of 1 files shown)


docs/source/kubernetes/setup-kubernetes.md
```diff
@@ -11,6 +11,7 @@ Choose one option and proceed.
 ```{toctree}
 :titlesonly:
 
+step-zero-minikube
```
consideRatio, line 14:
```suggestion
minikube/step-zero-minikube
```

docs/source/kubernetes/minikube/step-zero-minikube.md
```diff
@@ -0,0 +1,46 @@
...
+
+Follow the installation steps in the [minikube's "Get Started!" page](https://minikube.sigs.k8s.io/docs/start/).
+
+## Kubernetes cluster creation
+
+From a terminal, run
+
+```bash
+minikube start \
+--kubernetes-version v1.26.1 \
```
consideRatio, line 28: Not using a pre-defined version makes this docs more long lived, we could omit it entirely and rely on the default being "stable", but its kind of nice to explicitly pass this as well so that its known that this can be configured at this point in time also.

```suggestion
--kubernetes-version stable \
```

docs/source/kubernetes/minikube/step-zero-minikube.md
```diff
@@ -0,0 +1,46 @@
+(minikube-k8s)
+
+# Kubernetes on minikube (for development only)
+
+[minikube](minikube) is a implementation of Kubernetes as a local cluster primarily target to the development of applications.
```
consideRatio, line 5: I figure "development only" excludes the excellent use of minikube for trial purposes.

```suggestion
# Kubernetes on minikube

[minikube](minikube) can setup a Kubernetes cluster on a single computer. Using minikube can be suitable trial for new Kubernetes users and for development purposes, but its not meant to be used for production purposes.
```

consideRatio (commented):
I think this makes sense to include in our docs! I appreciate that the PR currently links out to external docs about details of things instead of introducing that here - which would have added long term maintenance load.

I've added a few comments, but otherwise this is ready to be merged I think!

PR #3266, opened 2023-11-08
Add `hub.customScopes` directly mapping to `c.JupyterHub.custom_scopes`

### LINKED ISSUE #3265: Add support for `custom_scopes`
### Proposed change

Add support for `custom_scopes` as described here:
- https://jupyterhub.readthedocs.io/en/stable/rbac/scopes.html#custom-scopes

Allowing to configure those directly in the chart instead of using `extraConfig` would be less error prone. 

Users using `custom_scopes`

Ph0tonic:
This is a first draft to support `custom_scopes`, I am not sure how to test it. Any feedback and help is welcome. 

Fixes #3265

\cc @consideRatio and @minrk

### DIFF
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -1410,6 +1410,31 @@ properties:
               An alias for api_token provided for backward compatibility by
               the JupyterHub Helm chart that will be transformed to
               api_token.
+      customScopes:
+        type: object
+        additionalProperties: true
+        description: |
+          This is where you should define JupyterHub custom scopes and apply them to
+          JupyterHub users, groups, and services to grant them additional
+          permissions as defined in JupyterHub's RBAC system.
+
+          Complement this documentation with [JupyterHub's
+          documentation](https://jupyterhub.readthedocs.io/en/stable/rbac/scopes.html#custom-scopes)
+          about `custom_scopes`.
+
+          ```yaml
+          hub:
+            customScopes:
+              "custom:myservice:read":
+                description: read-only access to myservice
+
+              "custom:myservice:write":
+                description: write access to myservice 
+                subscopes: ["custom:myservice:read"]
+          ```
+
+          When configuring JupyterHub custom scopes via this Helm chart, the `name`
+          field can be omitted as it can be implied by the dictionary key.
       loadRoles:
         type: object
         additionalProperties: true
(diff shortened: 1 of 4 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -1410,6 +1410,31 @@ properties:
...
+              "custom:myservice:read":
+                description: read-only access to myservice
+
+              "custom:myservice:write":
+                description: write access to myservice 
+                subscopes: ["custom:myservice:read"]
+          ```
+
+          When configuring JupyterHub custom scopes via this Helm chart, the `name`
+          field can be omitted as it can be implied by the dictionary key.
```
consideRatio, line 1437: There doesn't seem to be a `name` field for custom_scopes, instead its the dictionary key. So while `JupyterHub.load_roles` is a list and `hub.loadRoles` is a dict translated to a list, the `JupyterHub.custom_scopes` is a dict and doesn't need a translation.

This also means that we can without issues configure this directly via `hub.config.JupyterHub.custom_scopes` as well without providing a dedicated option for this I think.

@Ph0tonic could you update this description to not mention the `name` field?

consideRatio:
@minrk do you think z2jh should duplicate this, or ask that users rely on `hub.config.JupyterHub.custom_scopes` ? I think this makes it more findable, but a bit less sustainable.

manics:
I think we should encourage use of passthrough as it reduces maintenance. We should update the docs though, there's still a warning `Currently intended only for auth config`:
https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-config

Ph0tonic:
Hello,
Thanks for your informations.
Not sure however to have understood everything. So the idea is to directly use :
```yaml
hub:
  config:
    JupyterHub:
      custom_scopes:
        ...
```

So shall I close this issue ? And update the documentation ?

consideRatio:
> So the idea is to directly use :
> 
> ```yaml
> hub:
>   config:
>     JupyterHub:
>       custom_scopes:
>         ...
> ```

Yes, but I'm not sure where to document that use of `hub.config` passthrough should be used for `JupyterHub.custom_scopes` but not for `hub.loadRoles` for example. Hmmm...

manics:
I think we can either say
- use `hub.config` where there is no explicit chart config (easiest for us)
- or list the properties than can be set with `hub.config` (more work as we need to keep it up to date)

consideRatio:
I opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3271, I think the path forward is to settle for updating docs about `hub.config`, and that there is no remaining action point here.

PR #3313, opened 2024-01-11
ci: update circleci workflow for arm64, test with latest k3s

consideRatio:


.circleci/config.yml
```diff
@@ -1,72 +1,70 @@
...
+      # The k3s setup should be kept similar to how we do it in
+      # https://github.com/jupyterhub/action-k3s-helm.
       - run:
-          command: uname -a
-          name: Check architecture
-
-      - run:
-          # NOTE: we can't use k3s 1.24 and --docker unless we also install for
-          #       example cri-dockerd as done in
-          #       https://github.com/jupyterhub/action-k3s-helm.
```
consideRatio, line 23: k3s 1.24+ bundle with cri-dockerd once again, so we can rely on that!

manics (approved):

PR #3337, opened 2024-02-03
Fix documented example for proxy.chp.extraCommandLineFlags

consideRatio:
Thank @ryanlovett for spotting this!

### DIFF
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -1559,7 +1559,7 @@ properties:
                 chp:
                   extraCommandLineFlags:
                     - "--auto-rewrite"
-                    - "--custom-header {{ .Values.myCustomStuff }}"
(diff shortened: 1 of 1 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -1559,7 +1559,7 @@ properties:
                 chp:
                   extraCommandLineFlags:
                     - "--auto-rewrite"
-                    - "--custom-header {{ .Values.myCustomStuff }}"
+                    - "--custom-header={{ .Values.custom.myStuff }}"
```
consideRatio, line 1562: Without the `=` replacing the space, this example would lead to "unrecognized flag" where it parsed the full line as the flag name.

And also the Helm chart schema doesn't allow for .Values.myCustomStuff, only for `.Values.custom.myStuff` etc - nested under `custom`.

manics (approved):

PR #3357, opened 2024-03-07
Fix previously ignored revisionHistoryLimit config

SchutteJan:
Sometimes Helm reports type float64 for integer numbers set in `values.yaml`. See this Helm issue for an example:
https://github.com/helm/helm/issues/4982

This fix replaces the type check with a simple check + integer cast.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  {{- if typeIs "int" .Values.hub.revisionHistoryLimit }}
+  {{- if not (typeIs "<nil>" .Values.hub.revisionHistoryLimit) }}
   revisionHistoryLimit: {{ .Values.hub.revisionHistoryLimit }}
   {{- end }}
   replicas: 1
--- a/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
+++ b/jupyterhub/templates/image-puller/_helpers-daemonset.tpl
@@ -34,7 +34,7 @@ spec:
     type: RollingUpdate
     rollingUpdate:
       maxUnavailable: 100%
-  {{- if typeIs "int" .Values.prePuller.revisionHistoryLimit }}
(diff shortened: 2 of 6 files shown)


jupyterhub/templates/hub/deployment.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
 spec:
-  {{- if typeIs "int" .Values.hub.revisionHistoryLimit }}
-  revisionHistoryLimit: {{ .Values.hub.revisionHistoryLimit }}
+  {{- if .Values.hub.revisionHistoryLimit }}
+  revisionHistoryLimit: {{ .Values.hub.revisionHistoryLimit | int }}
```
consideRatio, line 9: We should be [able to set this to 0 as well](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy), and that was the point of having the `typeIs` check. We could also change this check to a "not null" check since we have a schema for the helm chart enforcing that makes it either null or int.

Do you have an example reproduction of the bug you look to fix with this change?

SchutteJan: Good point, forgot that 0 was treated as a false in this case. I've changed the check to check for not-nil.

Setting the following value file will not render the revisionLimitHistory in the chart:
```yaml
hub:
  revisionHistoryLimit: 1
```

This is on `helm version`:
```
version.BuildInfo{Version:"v3.14.2", GitCommit:"c309b6f0ff63856811846ce18f3bdc93d2b4d54b", GitTreeState:"clean", GoVersion:"go1.21.7"}
```

consideRatio:
Nice catch @SchutteJan! I could reproduce the issue and concluded that this PR fixes it!

Reproduction steps I took:

```bash
tools/generate-json-schema.py
# manually update jupyterhub/values.yaml to provide various defaults for hub.revisionHistoryLimit
helm template jupyterhub --show-only templates/hub/deployment.yaml | grep revision
```

consideRatio:
I also verified this was fixing the issue in all places, and that there wasn't any other issues related to using `typeIs "int"` checks on integers.

Thank you @SchutteJan! :heart: :tada: :sunflower:

PR #3371, opened 2024-03-21
unpin pycurl

minrk:
and make sure we build it from source with `--no-binary pycurl`

Also make sure to use build-stage wheels by adding `--no-index` to pip install.

avoids installing higher-priority wheels from PyPI instead of what we built (i.e. pycurl)

`--no-index` was the part missing in #3365, because the `manylinux_2_28` wheel from PyPI was preferred at install time to the `linux_arch` wheel that we built. `--no-index` ensures no contact with PyPI at all at install time.

This is a low priority fix to unpin versions going forward, no need to trigger a new patch release.

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -21,9 +21,11 @@ FROM python:3.11-bullseye as build-stage
 COPY requirements.txt requirements.txt
 ARG PIP_CACHE_DIR=/tmp/pip-cache
 RUN --mount=type=cache,target=${PIP_CACHE_DIR} \
-    pip install build \
- && pip wheel \
+    pip wheel \
         --wheel-dir=/tmp/wheels \
+        # pycurl 7.45.3 has wheels, but they aren't portable
+        # https://github.com/pycurl/pycurl/issues/834
+        --no-binary pycurl \
         -r requirements.txt \
         # Additional wheels for default-stage. Updates below should be repeated
         # in default-stage.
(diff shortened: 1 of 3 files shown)


images/hub/Dockerfile
```diff
@@ -21,9 +21,11 @@ FROM python:3.11-bullseye as build-stage
 COPY requirements.txt requirements.txt
 ARG PIP_CACHE_DIR=/tmp/pip-cache
 RUN --mount=type=cache,target=${PIP_CACHE_DIR} \
-    pip install build \
```
consideRatio, line 24: I'm not confident why we installed `build` here, so I'm not confident on seeing it removed - can you motivate the change?

minrk: It was added but never used without comment in [#2733](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2733/files#diff-b55155cca69fb062e0d3eb1511e7754d1099bedf339294ea738cfd925a490753R14).

I believe it was added by mistake, copying what we do in binderhub, which includes building the binderhub wheel itself, which used to be done with `python3 -m build --wheel .` (not anymore). If you're not building _local_ packages from source via `python3 -m build` (and not `pip wheel`, which can also do the same thing), `build` is never used.

From what I can find, it's here because we used to [invoke it here in binderhub](https://github.com/jupyterhub/binderhub/commit/9bf1418c62c20a10d565e6e3996e7b24edb98f8a#diff-d91e821ae7349a5e3cce784da75cb2695c349a4096193500bede2da5010987e3L28), but since  https://github.com/jupyterhub/binderhub/pull/1535/files switched to invoking `pip wheel`, `build` has not been used in either binderhub or z2jh. It's a tiny unused pure-python package in the build stage, so nobody noticed the very tiny bit of wasted time.

consideRatio: Ah, thank you @minrk!

consideRatio (approved):

Issue #1867, opened 2020-10-23
Helm2 -> Helm3 labeling practices transition


consideRatio:
In the past, Helm had best practices on how to label pods, but since then Kubernetes introduced a set of best practices which now Helm recommends. These have been around for more than a year now. It would be nice to transition to use the k8s official labeling system.

The crux is that it is a breaking change to modify these labels, at least those used as "matchLabels" for Deployment resources and are part of PodDisruptionBudgets. So, when we change these, we should change them all in one go to avoid multiple sequential breaking changes.

When we do make the breaking change, I think we need to manually delete some resources or use `helm --force` which in Helm 3 is very dangerous I think. So, for this change to be made, we need:

1. The change in this Helm chart to use the new labels.
2. Upgrade instructions that works for Helm 2 and 3 if Helm 2 is still used.
3. We may also want to update the labels set by KubeSpawner at the same time.

## Transition
Old | New | Status | Description from Helm docs
-|-|-|-
app: | app.kubernetes.io/name | REC | This should be the app name, reflecting the entire app. Usually {{ template "name" . }} is used for this. This is used by many Kubernetes manifests, and is not Helm-specific.
chart: | helm.sh/chart | REC | This should be the chart name and version: {{ .Chart.Name }}-{{ .Chart.Version \| replace "+" "_" }}.
heritage: | app.kubernetes.io/managed-by | REC | This should always be set to {{ .Release.Service }}. It is for finding all things managed by Helm.
release: | app.kubernetes.io/instance | REC | This should be the {{ .Release.Name }}. It aids in differentiating between different instances of the same application.
-| app.kubernetes.io/version | OPT | The version of the app and can be set to {{ .Chart.AppVersion }}.
component: | app.kubernetes.io/component | OPT | This is a common label for marking the different roles that pieces may play in an application. For example, app.kubernetes.io/component: frontend.
-| app.kubernetes.io/part-of | OPT | When multiple charts or pieces of software are used together to make one application. For example, application software and a database to produce a website. This can be set to the top level application being supported.

## Related
- #1866
- Usage of [component:](https://github.com/helm/charts/search?q=component:), and [`app.kubernetes.io/component:`](https://github.com/helm/charts/search?q=app.kubernetes.io%2Finstance)
- [Helm3 docs on labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels)
[...]

PR #3404, opened 2024-04-21
Add recommended chart labels alongside old labels (`app.kubernetes.io/...`, `helm.sh/chart`)

consideRatio:
This enables us to make a breaking change in z2jh 5 where user servers restarted once or more during z2jh 4 doesn't have to be restarted as part of the breaking change then. If we did it right away, network policy enforcement would fail for old servers for example.

Before z2jh 4, we should also get kubespawner to specify `app.kubernetes.io/component` alongisde `component`. This is done in https://github.com/jupyterhub/kubespawner/pull/835.

Also, its nice to be able to provide time for people to adjust to the new labels with a period of overlapping labels. For example, grafana dashboards working with prometheus metrics could keep working in this period and at any time transition to refering to new labels, and then when we get z2jh 5 released such dashboards could have already adjusted.

---

- Work towards #1867

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -131,9 +131,6 @@ if chart_name and chart_version:
     common_labels["helm.sh/chart"] = common_labels["chart"] = (
         f"{chart_name}-{chart_version.replace('+', '_')}"
     )
-chart_app_version = get_config("Chart.AppVersion")
-if chart_app_version:
-    common_labels["app.kubernetes.io/version"] = chart_app_version
 common_labels["app.kubernetes.io/managed-by"] = "kubespawner"
 
 c.KubeSpawner.namespace = os.environ.get("POD_NAMESPACE", "default")
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -107,28 +107,36 @@ c.JupyterHub.hub_connect_url = (
 )
 
 # implement common labels
-# this duplicates the jupyterhub.commonLabels helper
+# This mimics the jupyterhub.commonLabels helper, but declares managed-by to
+# kubespawner instead of helm.
+#
+# The labels app and release are old labels enabled to be deleted in z2jh 5, but
+# for now retained to avoid a breaking change in z2jh 4 that would force user
+# server restarts. Restarts would be required because NetworkPolicy resources
+# must select old/new pods with labels that then needs to be seen on both
+# old/new pods, and we want these resources to keep functioning for old/new user
+# server pods during an upgrade.
+#
 common_labels = c.KubeSpawner.common_labels = {}
-common_labels["app"] = get_config(
+common_labels["app.kubernetes.io/name"] = common_labels["app"] = get_config(
     "nameOverride",
     default=get_config("Chart.Name", "jupyterhub"),
 )
-common_labels["heritage"] = "jupyterhub"
+release = get_config("Release.Name")
+if release:
+    common_labels["app.kubernetes.io/instance"] = common_labels["release"] = release
 chart_name = get_config("Chart.Name")
 chart_version = get_config("Chart.Version")
 if chart_name and chart_version:
-    common_labels["chart"] = "{}-{}".format(
-        chart_name,
-        chart_version.replace("+", "_"),
-    )
-release = get_config("Release.Name")
-if release:
-    common_labels["release"] = release
+    common_labels["helm.sh/chart"] = f"{chart_name}-{chart_version.replace('+', '_')}"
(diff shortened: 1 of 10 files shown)


manics:
I think I understand what your intentions are.

[`test_spawn_basic`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c5f66ecfb2e4f835c4cc50d5330a1c90d5bbb133/tests/test_spawn.py#L56-L72) already inspects some properties of the pod, do you think you could also verify the full set of labels are present? This will also make the impact of the future removal clearer.

consideRatio:
Done @manics - I also fixed unrelated test failures via https://github.com/jupyterhub/action-k3s-helm/pull/116,

jupyterhub/templates/_helpers.tpl
```diff
@@ -112,31 +111,65 @@
...
+
+    Provides old labels:
+      app
+      release
+      chart (omitted for matchLabels)
+      heritage (omitted for matchLabels)
+    Provides modern labels (omitted for matchLabels):
+      app.kubernetes.io/name ("app")
+      app.kubernetes.io/version
+      app.kubernetes.io/instance release ("release")
```
manics, line 123:
```suggestion
      app.kubernetes.io/instance ("release")
```
Is this a typo?

jupyterhub/files/hub/jupyterhub_config.py
```diff
@@ -107,28 +107,38 @@ def camelCaseify(s):
...
-        chart_version.replace("+", "_"),
+    common_labels["helm.sh/chart"] = common_labels["chart"] = (
+        f"{chart_name}-{chart_version.replace('+', '_')}"
     )
-release = get_config("Release.Name")
-if release:
-    common_labels["release"] = release
+chart_app_version = get_config("Chart.AppVersion")
+if chart_app_version:
+    common_labels["app.kubernetes.io/version"] = chart_app_version
```
manics, line 136: I'm not sure about this. The app version in the singleuser pods will often be different to the app version in the hub. Is it still useful to have the hub version on the singleuser pods?

manics (commented):
Thanks for the updates!

jupyterhub/files/hub/jupyterhub_config.py
consideRatio: To know a version when kubespawner created resource could be relevant, but we still could get this from helm.sh/chart that includes the chart version. For example if we document an instruction that requires restarts of user servers started with z2jh/jupyterhub/kubespawner of some version that was too old, then a z2jh/jupyterhub/kubespawner version field could help us not restart all user servers but only some.

But, app.kubernetes.io/version for a kubespawner created resource - should that be version of jupyterhub server, jupyterhub inside user server, or kubespawner? The ambiguity for this situation makes it less valuable to provide a version at all.

- I think if we have version set, it should be the jupyterhub version deployed by the helm chart in hub pod, I believe that [matches the labels intent best according to k8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) - at least if we have `app.kubernetes.io/name=jupyterhub`.
- If we'd set `app.kubernetes.io/name=kubespawner` then the version should be set to the version of kubespawner, and we should use `app.kubernetes.io/part-of=jupyterhub`

---

Messy...

I think I lean towards either retaining version as it is, or removing it atm.

I've consiered also changing the .../name label to kubespawner, but I'm hesitant towards doing it as searching for labels .../name=jupyterhub .../instance=my-helm-release can successfully identify all chart resources - directly created or indirectly via kubespawner, while also allowing for filtering to those managed by helm and kubespawner via the .../managed-by label.

@manics lets make a call - do we remove it? I think whatever we do, its not a big decision and its fine to do either.

manics: > The metadata is organized around the concept of an application. Kubernetes is not a platform as a service (PaaS) and doesn't have or enforce a formal notion of an application. Instead, applications are informal and described with metadata. The definition of what an application contains is loose.

The problem is we don't have a single "application". Intuitively I think treating `version` as the `component` version is helpful, but since that's going to be too complicated maybe we remove it for now, or just add it to the hub pod only?

consideRatio: I systematically removed it from all chart resources and kubespawner resources created in 5099dfc!

jupyterhub/templates/_helpers.tpl
manics: Latest changes look good! Just this nitpick/question from me, then this is good to merge.

consideRatio: 100% typo - thank you for looking carefully at this @manics!

consideRatio:
Thank you for reviewing @manics!!! :heart: :tada: :sunflower:

PR #3441, opened 2024-06-20
fix(daemonset): Allow the definition a ServiceAccount in DaemonSets

samyuh:
This pull request introduces the option to define a ServiceAccount in DaemonSets used in the prepuller.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3442


samyuh:
This bug is fixed by https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3441

samyuh:
Hey @manics could you take a look at this?

jupyterhub/values.yaml
```diff
@@ -618,7 +618,7 @@ prePuller:
     tolerations: []
     resources: {}
     serviceAccount:
-      create: true
+      create: false
```
manics, line 621: What's the reason for changing this to false?

samyuh: This was just to not create a service account by default, but I can change it to true again tomorrow (I'm on vacations and away from my PC until tomorrow)

samyuh: Just reverted to the old config

manics: If we're providing the ability to create a service account we should create one with the correct permissions.

samyuh: It turns out that the service account already had the correct permissions. However, due to the annotations present in that SA, it was getting deleted and throwing the error that the SA did not exist. I changed the SA and now everything works.

Feel free to make more suggestions, since my knowledge in k8s is still limited :)

samyuh: One test still failing, I will look into that

jupyterhub/templates/image-puller/serviceaccount.yaml
```diff
@@ -2,18 +2,16 @@
 ServiceAccount for the pre-puller hook's image-awaiter-job
 */}}
 {{- if .Values.prePuller.hook.serviceAccount.create -}}
-{{- if (include "jupyterhub.imagePuller.daemonset.hook.install" .) -}}
+{{- if or (include "jupyterhub.imagePuller.daemonset.hook.install" .) .Values.prePuller.continuous.enabled -}}
```
samyuh, line 5: There are two types of image-pullers (daemonset-continuous and daemonset-hook). If either of them is enabled, we need to create the service account.

jupyterhub/templates/image-puller/serviceaccount.yaml
```diff
@@ -2,18 +2,16 @@
...
 metadata:
   name: {{ include "jupyterhub.hook-image-awaiter-serviceaccount.fullname" . }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
-    hub.jupyter.org/deletable: "true"
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
-    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-weight": "-30"
```
samyuh, line 14: This is to force the creation of a service account before the prepullers start.

jupyterhub/templates/image-puller/serviceaccount.yaml
```diff
@@ -2,18 +2,16 @@
...
 {{- if .Values.prePuller.hook.serviceAccount.create -}}
-{{- if (include "jupyterhub.imagePuller.daemonset.hook.install" .) -}}
+{{- if or (include "jupyterhub.imagePuller.daemonset.hook.install" .) .Values.prePuller.continuous.enabled -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "jupyterhub.hook-image-awaiter-serviceaccount.fullname" . }}
   labels:
     {{- include "jupyterhub.labels" . | nindent 4 }}
-    hub.jupyter.org/deletable: "true"
```
samyuh, line 12: We don't want to delete the service account if the daemonsets are enabled

jupyterhub/values.yaml
samyuh: Everything working, you can take a look again. Thanks for your time

samyuh:
@manics check this again when you can, everything is working as expected

jupyterhub/templates/image-puller/serviceaccount.yaml
manics: Does this need to be conditional on whether daemonset-hook but not daemonset-continuous is enabled?

jupyterhub/templates/image-puller/serviceaccount.yaml
manics: Other objects such as RBAC roles use weight `0`. If the ServiceAccount uses weight `0` does this break?

manics (commented):
Other than a couple of queries this looks good to me, thanks for working on this.

@consideRatio would you mind reviewing this? I'm not fully confident about all the changes.

jupyterhub/templates/image-puller/serviceaccount.yaml
samyuh: I don't think so. 

We don't want to delete the service account if either daemonset-hook or daemonset-continuous is enabled.

jupyterhub/templates/image-puller/serviceaccount.yaml
samyuh: Yes. If we use weight `0` this breaks, probably because the pre-puller hook starts before the service account is created.

samyuh:
@consideRatio please take a look when you can :)

samyuh:
@manics can you push this to have the second review and be merged?

consideRatio:
I'll find time to review this this PR this week

consideRatio:
I failed to make time for this, but it remains on my todo list.

samyuh:
Thanks! Please don't forget to review this since it's open since June :)

consideRatio:
Thank you for engaging and trying to figure things out here @samyuh!!

## Background and overview

There are two "pre puller" machineries:
- one that runs before a `helm upgrade` command is run which is temporary - it starts and stops. Its referred to as the hook pre puller (it involves a helm hook to run "before upgrade")
- one that runs continuously, it never stops, so its called the contiunous pre puller.

Both involve a daemonset to ensure each node gets images pulled, but the hook pre puller needs to stop, so it includes a k8s job that starts and finishes. The k8s job pod needs k8s api permissions, because it needs to read the state of the daemonset that was created, so it can say if all its pods are ready etc. Its just waiting for that to happen and then shuts down.

Anyhow, this is why the daemonset pods doesn't have a service account, they haven't needed one. The one used for the hook pre puller's k8s job pod is specifically for that k8s job pod and comes with relevant permissions, re-using it for daemonset's pods won't be the correct call no matter what as it provides irrelevant permissions to those pods.

## Decision

I'll opt to close this pull request in its current form for now as I don't want to have the hook pre puller machinery's k8s job's service account re-used for the daemonsets' pods, however:

If there a need to configure the pre-puller pods with a k8s Service Account, and possibly also have it be created by the chart as well, it could be considered. For this, please describe what is to be accomplished by configuring the daemonset pod's serviceAccountName and optionally also if there is value in getting a blank k8s ServiceAccount created by the chart.

samyuh:
Hey @consideRatio, thanks for your detailed response. 

> If there a need to configure the pre-puller pods with a k8s Service Account, and possibly also have it be created by the chart as well, it could be considered. For this, please describe what is to be accomplished by configuring the daemonset pod's serviceAccountName and optionally also if there is value in getting a blank k8s ServiceAccount created by the chart.

We really need to have every component in our deployment with a specific service account due to security compliance. In this merge request, I used the existing pre-puller service account as I thought there would be no problem. We can just go with the option of setting  the `serviceAccountName` with a value from the chart. Are you ok with that alternative? Can you reopen this merge request if you're okay with these changes?

Thanks again :)

PR #3457, opened 2024-07-12
Bump debian distribution for images

SchutteJan:
Bumping the base-image from debian bullseye (11) to bookworm (12)


manics (approved):
Thanks!

PR #3464, opened 2024-07-19
Security context hardening

lahwaacz:
This should allow running JupyterHub in a namespace with a `restricted` security profile according to the [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) (assuming an appropriate configuration for the singleuser pods in KubeSpawner).

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -91,10 +91,14 @@ hub:
   podSecurityContext:
     runAsNonRoot: true
     fsGroup: 1000
+    seccompProfile:
+      type: "RuntimeDefault"
   containerSecurityContext:
     runAsUser: 1000
     runAsGroup: 1000
     allowPrivilegeEscalation: false
+    capabilities:
+      drop: ["ALL"]
   lifecycle: {}
   loadRoles: {}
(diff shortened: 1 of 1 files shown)


lahwaacz:
Kubernetes default for `seccompProfile` is `Unconfined` (unless the [`seccompDefault` kubelet option](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) is enabled). `RuntimeDefault` means that the default seccomp profile of the container runtime (e.g. CRI-O or containerd) is used. You can find more details here: https://kubernetes.io/docs/tutorials/security/seccomp/

The default capabilities also depend on the container runtime. For example, CRI-O has `CHOWN, DAC_OVERRIDE, FSETID, FOWNER, SETGID , SETUID, SETPCAP, NET_BIND_SERVICE, KILL` [by default](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md). Docker (and containerd) [has more](https://github.com/moby/moby/blob/master/oci/caps/defaults.go#L6-L19).

manics (approved):
Thanks!

PR #3473, opened 2024-07-31
Add changelog for 3.3.8

consideRatio:


manics:
Shouldn't this be against the 3.x branch? Then we can drop the unreleased changes section.

consideRatio:
We should update the changelog in main and for 3.x, and I mean to cherry-pick this commit into 3.x after merging to mimic what was done for other patch releases. It could be done the other way around as well though.

The unreleased section changes was unrelated to the PR, and I could extract them but I figured was OK to bundle in this.

consideRatio:
@manics thank you for the neat "summarize linkcheck issues" btw!!

image

consideRatio:
Preview:
image

manics:
Looks like we've got unreleased changes in the changelogs for other backported relaseses. Given we need to make this release quickly it's fine for now, but in future I think a changelog (which forms part of a tag) shouldn't containfuture changes.

PR #3492, opened 2024-09-04
Add `ingress.extraPaths` config

alxyok:
Introduces the ability to configure custom paths for the Ingress resource in the Helm chart, allowing users to define additional paths and corresponding backend services directly in the values.yaml file.

### DIFF
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -24,6 +24,15 @@ spec:
                 name: {{ include "jupyterhub.proxy-public.fullname" $ }}
                 port:
                   name: http
+          {{- range $path := $.Values.ingress.customPaths }}
+          - path: {{ $path.path }}
+            pathType: {{ $.Values.ingress.pathType }}
+            backend:
+              service:
+                name: {{ $path.backend.serviceName }}
+                port:
+                  number: {{ $path.backend.servicePort }}
+          {{- end }}
       {{- if $host }}
       host: {{ $host | quote }}
       {{- end }}
--- a/jupyterhub/templates/ingress.yaml
+++ b/jupyterhub/templates/ingress.yaml
@@ -26,12 +26,16 @@ spec:
                   name: http
           {{- range $path := $.Values.ingress.customPaths }}
           - path: {{ $path.path }}
-            pathType: {{ $.Values.ingress.pathType }}
+            pathType: {{ $path.pathType }}
             backend:
               service:
-                name: {{ $path.backend.serviceName }}
+                name: {{ $path.backend.service.name }}
                 port:
-                  number: {{ $path.backend.servicePort }}
+                  {{- if $path.backend.service.port.number }}
+                  number: {{ $path.backend.service.port.number }}
+                  {{- else if $path.backend.service.port.name }}
+                  name: {{ $path.backend.service.port.name }}
+                  {{- end }}
           {{- end }}
       {{- if $host }}
       host: {{ $host | quote }}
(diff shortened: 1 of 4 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -2757,6 +2757,12 @@ properties:
           See [the Kubernetes
           documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls)
           for more details about annotations.
+      customPaths:
```
manics, line 2760: We've generally named properties that extend an existing object `extra*` rather than `custom*`

jupyterhub/values.schema.yaml
```diff
@@ -2757,6 +2757,12 @@ properties:
...
+        type: array
+        description: |
+          A list of custom paths to be added to the ingress configuration. 
+          Each custom path allows traffic to be routed to a specific backend service.
+          For each path, you can specify the service name, and either a service port number or a service port name.
```
manics, line 2765: Since the admin has to essentially provide the raw K8s config for a path can you add a link to the relevant part of the K8s docs?

manics (commented):
I've left a few comments but the idea seems reasonable to me.

Can you also add a test config to
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/tools%2Ftemplates%2Flint-and-validate-values.yaml
so it's tested against the K8s API? Thanks!

jupyterhub/values.schema.yaml
alxyok: of course makes sense

jupyterhub/values.schema.yaml
alxyok: sure i'll add the URL to the kubernetes doc

jupyterhub/templates/ingress.yaml
```diff
@@ -24,6 +24,19 @@ spec:
...
+            backend:
+              service:
+                name: {{ $path.backend.service.name }}
+                port:
+                  {{- if $path.backend.service.port.number }}
+                  number: {{ $path.backend.service.port.number }}
+                  {{- else if $path.backend.service.port.name }}
+                  name: {{ $path.backend.service.port.name }}
+                  {{- end }}
+          {{- end }}
```
consideRatio, line 39: Does this work as well? It would allow us to not need to keep updated with all options we can specify, for example having `backend.resource` which is valid as well.

```suggestion
          {{- with $.Values.ingress.extraPaths }}
          {{- . | toYaml | nindent 10 }}
          {{- end }}
```

You could trial this a bit by adding an entry for `ingress.extraPaths` in the lint-and-validate-config.yaml file, which also helps us test this a bit.

jupyterhub/values.schema.yaml
```diff
@@ -2757,6 +2757,15 @@ properties:
           See [the Kubernetes
           documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls)
           for more details about annotations.
+      extraPaths:
+        type: array
+        description: |
+          A list of custom paths to be added to the ingress configuration. 
+          Each custom path allows traffic to be routed to a specific backend service.
+          For each path, you can specify the service name, and either a service port number or a service port name.
```
consideRatio, line 2765: This can probably be omitted if we render like `{{- . | toYaml | nindent 10 }}`, and then we can lean on the kubernetes docs entirely.

jupyterhub/templates/ingress.yaml
alxyok: it's even better indeed, i'll test this

consideRatio (approved):
Nice, looks great!

PR #3546, opened 2024-10-18
Ensure hub container is first by appending instead of prepending hub.extraContainers

consideRatio:
By doing this, you'd be able to do `kubectl logs deploy/hub` and get the
hub containers logs no matter what. Before, you'd get the first listed
`hub.extraContainer` instead and have to add `--container=hub` to ensure
getting the hub containers logs.

I'm maintaining a deployment where we use hub.extraContainers, and by
doing so, have been required to specify `-c hub` anytime I want logs,
because its not the first container in the pod specs list of containers.

### DIFF
--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -93,9 +93,6 @@ spec:
         {{- . | toYaml | nindent 8 }}
       {{- end }}
       containers:
-        {{- with .Values.hub.extraContainers }}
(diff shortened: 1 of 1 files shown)


consideRatio:
@yuvipanda @sgibson91 @GeorgianaElena the need to specify `-c hub` when getting hub logs for 2i2c hubs has bugged me a while, this fixes it and makes us instead need to specify `hub.extraContainers` in the hub pod by name if we want their logs.

manics (approved):

PR #3548, opened 2024-10-19
Support subdomain_host (CHP needs --host-routing)

manics:
The [JupyterHub v5 upgrade doc](https://jupyterhub.readthedocs.io/en/stable/howto/upgrading-v5.html#user-subdomains) says
> All JupyterHub deployments which care about protecting users from each other are encouraged to enable per-user domains, if possible, as this provides the best isolation between user servers.

which sounds like a strong recommendation. This is not currently supported by Z2JH as CHP needs the `--host-routing` flag. Since CHP is not managed by JupyterHub this isn't automatically configured
https://github.com/jupyterhub/jupyterhub/blob/ab43f6beb8571e2b831801089d61144b15951b85/jupyterhub/proxy.py#L738-L739

https://discourse.jupyter.org/t/user-subdomains-oauth-state-missing/29328/4

The main downsides of using `hub.config.subdomain_host` as the condition instead of a dedicated parameter such as `hub.subdomainHost` (analogous to `hub.baseUrl` qhich also requires configuring multiple chart manifests) are it's not part of the schema so won't appear in the generated chart documentation, and it also adds dependencies across manifests with `hub.config` which I don't think we have, so this increases the maintenance complexity.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -100,6 +100,9 @@ spec:
             {{- if .Values.debug.enabled }}
             - --log-level=debug
             {{- end }}
+            {{- if .Values.hub.config.subdomain_host }}
+            - --host-routing
+            {{- end }}
             {{- range .Values.proxy.chp.extraCommandLineFlags }}
             - {{ tpl . $ }}
             {{- end }}
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -100,7 +100,7 @@ spec:
             {{- if .Values.debug.enabled }}
             - --log-level=debug
             {{- end }}
-            {{- if .Values.hub.config.subdomain_host }}
+            {{- if .Values.hub.config.JupyterHub.subdomain_host }}
             - --host-routing
             {{- end }}
             {{- range .Values.proxy.chp.extraCommandLineFlags }}
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -137,13 +137,14 @@ jobs:
               --set hub.image.name=quay.io/jupyterhub/k8s-hub-slim
               --set prePuller.hook.enabled=true
               --set prePuller.hook.pullOnlyOnChanges=true
-          - k3s-channel: v1.31 # also test hub.existingSecret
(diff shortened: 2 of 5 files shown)


minrk:
> it's not part of the schema so won't appear in the generated chart documentation

Can we add a couple of fields to the schema in hub.config that have significant consequences (with `additional_properties: true`)?

It probably makes sense to have a prose docs section on enabling this and the consequences.

manics:
Docs preview https://zero-to-jupyterhub--3548.org.readthedocs.build/en/3548/resources/reference.html#hub-config-jupyterhub

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -100,6 +100,9 @@ spec:
             {{- if .Values.debug.enabled }}
             - --log-level=debug
             {{- end }}
+            {{- if .Values.hub.config.subdomain_host }}
```
consideRatio, line 103:
```suggestion
            {{- if .Values.hub.config.JupyterHub.subdomain_host }}
```

consideRatio: This can error if we dont always have JupyterHub defined as a dict etc, maybe we can assume we have? I dont remember.

manics: It's already defined 😄 
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/92ef7265bbd1c03b3fbef9be82c2f7f38518b8df/jupyterhub/values.yaml#L35-L38

minrk:
If we just want a really simple test, I think we can start a cluster with subdomains enabled and make some requests to the cluster ip setting the Host header to the expected domain, e.g. 

```python
# server_url from hub api will have the user's domain in it when subdomains are enabled
# server_url = "http://user.subdomain:port/user/name"
user_host = urlparse(server_url).netloc
user_path = urlparse(server_url).path
requests.get(f"http://{proxy_ip}:{port}{path}", headers={"Host": user_host})
```

Or if we can expose a localhost port to the service, we could use `*.localhost.jovyan.org` like we do in the jupyterhub subdomain tests.

manics:
Thanks for the suggestion- I've added a test using the `Host` header.

minrk (approved):
Nice! Looks like it works. Anything more you want to add?

manics:
No, I think this is good enough!

PR #3557, opened 2024-10-24
v4 upgrade doc, changelog for 4.0

minrk:
adds 3-to-4 upgrade doc, which summarizes the relevant changes in dependencies and upgrade strategies in the chart config. I could find no changes in the chart itself that really needed upgrade docs, does anyone disagree, are there any new features we want to highlight in the upgrade doc in addition to possible breakages?

To prepare for final release, this collapses beta changelog entries into a single entry for 4.0, since I don't think we should keep changelog entries for prereleases.

I think theres some redundancy in the "version upgrade table" between the changelog and the upgrade doc. I think some redundancy is fine, but I don't have a great guide in my head for what level of info belongs where.

### DIFF
--- a/docs/source/administrator/upgrading/upgrade-3-to-4.md
+++ b/docs/source/administrator/upgrading/upgrade-3-to-4.md
@@ -2,4 +2,84 @@
 
 # Major upgrade: 3.\* to 4.\*
 
-This is currently just a placeholder file for until we have written upgrade docs.
+zero-to-jupyterhub 4.0 is a major upgrade that may require some changes to your configuration,
+depending on what features you may use.
+This mostly comes in the form of some upgraded packages, described below.
+
+See the [changelog](changelog) for details of upgraded packages.
+
+## JupyterHub 5
+
+zero-to-jupyterhub 4.0 upgrades the JupyterHub version from 4.1.6 to 5.2.1.
+
+:::{seealso}
+For more detailed changes in JupyterHub, see JupyterHub's own documentation on upgrading to version 5:
+
+Especially if you use features like per-user subdomains or custom page templates.
+
+- [jupyterhub changelog](https://jupyterhub.readthedocs.io/en/5.2.1/reference/changelog.html)
+- [Upgrading to JupyterHub 5](https://jupyterhub.readthedocs.io/en/5.2.1/howto/upgrading-v5.html)
+
+:::
+
+### Allowing users
+
+JupyterHub 5 promotes the `allow_all` and `allow_existing_users` configuration used on OAuthenticator to all other Authenticators.
+If you have no explicitly allowed users, the default is now to allow no users (this was already the default if you were using OAuth).
+If you are using an Authenticator where all users who can successfully authenticate should have access, set:
+
+```yaml
+hub:
+  config:
+    Authenticator:
+      allow_all: true
+```
+
+to explicitly opt in to this behavior that was previously the default for some authenticators.
+
+## KubeSpawner 7
+
+zero-to-jupyterhub 4.0 upgrades the KubeSpawner version from 6 to 7.
+The main relevant change here is the "slug scheme" used to compute names of resources such as pods and PVC (user storage) is changed.
+
+The new scheme is called "safe" and is the default, whereas the old scheme is called "escape".
+If you do not have any custom templated fields, it is _unlikely_ that anything should change for you and everything should work.
+If, however, you specify custom templated fields such as volume mounts or environment variables,
+especially those that have the `{username}` or `{servername}` fields,
+those values are likely to change under the new scheme for some usernames.
+In particular, there are new fields that should make things easier:
+
+- `{user_server}` combines the username and server name, and is equivalent to `{username}{servername}` in the old escape scheme
+- `{pod_name}`, `{pvc_name}`
+
+You can opt in to the kubespawner 6 behavior with:
+
+```yaml
+hub:
+  config:
+    KubeSpawner:
+      slug_scheme: escape
+```
+
+which _should_ result in no changes for you from previous behavior.
+
+:::{seealso}
+
+- [KubeSpawner changelog](https://jupyterhub-kubespawner.readthedocs.io/en/latest/changelog.html)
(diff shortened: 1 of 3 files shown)


minrk:
also needs kubespawner 7 before final, but that's ready, I think: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3557

docs/source/administrator/upgrading/upgrade-3-to-4.md
```diff
@@ -2,4 +2,84 @@
...
+
+- [jupyterhub changelog](https://jupyterhub.readthedocs.io/en/5.2.1/reference/changelog.html)
+- [Upgrading to JupyterHub 5](https://jupyterhub.readthedocs.io/en/5.2.1/howto/upgrading-v5.html)
+
+:::
+
+### Allowing users
+
+JupyterHub 5 promotes the `allow_all` and `allow_existing_users` configuration used on OAuthenticator to all other Authenticators.
+If you have no explicitly allowed users, the default is now to allow no users (this was already the default if you were using OAuth).
```
manics, line 28:
```suggestion
If you have not explicitly allowed users, the default is to allow no users (this was already the default if you were using OAuth).
```

manics:
I think we should get this ldapauthenticator fix https://github.com/jupyterhub/ldapauthenticator/pull/289 into a patch release that we can include in Z2JH.

consideRatio:
## Thoughts about what to document where etc

> I think theres some redundancy in the "version upgrade table" between the changelog and the upgrade doc. I think some redundancy is fine, but I don't have a great guide in my head for what level of info belongs where.

I've thought about this, and have a model in my head where I think things should go:

- I generally think of reference documentation (diataxis) as the lowest level building block of documentation to be referenced by higher level building blocks.
  If something fits in reference docs or generally in a lower level building block, it should be written there and repetition should be avoided to improve maintainability and to enable high level docs to be briefer which in turn improves readability.
- For a major release, the following documentation pieces are relevant
  - Changelog (lowest level building block)
  - General upgrade docs
  - Version specific upgrade docs (highest level building block)

For distributions like TLJH and Z2JH, with default configuration of various dependencies and ways of configuring those dependencies, what should its changelog / general upgrade docs / version specific upgrade docs mention if a dependency has a major version bump?

I think _at least_:
- A link to the dependecy's changelog
- Evaluation if the dependency bump is a breaking change at all in the context of the distribution. If its not, a comment about its not breaking in the context of the distribution should be added.
- Evaluation if the dependency bump's breaking change documentation in the upstream project _needs to be augmented with distribution specific context/details_.
  As practical examples:
  - I think LDAPAuthenticator mentioning deprecation of `c.LDAPAuthenticator.use_ssl` doesn't need to be augmented in z2jh with a mention of `hub.config.LDAPAuthenticator.use_ssl`, because use of `hub.config` can be assumed or documented under "General upgrade docs"
  - I think KubeSpawner's changes with escape are relevant to mention in version specific upgrade docs, where for example `singleuser.storage.static.subPath = "{username}"` is a default value making users of `singleuser.storage.type = "static"` directly impacted in a way motivating that z2jh augments documentation in KubeSpawner.

We have had examples of documentations in z2jh / tljh that caused unsustainable maintenance, because they became outdated even though the dependency project documented things clearly. That makes more docs possibly worse than less. Due to that, my mental model is to opt to lean as heavily as possible on dependent project docs and avoiding repetition as much as possible.

docs/source/administrator/upgrading/upgrade-3-to-4.md
```diff
@@ -2,4 +2,84 @@
...
+The main changes are related to using group information from OAuth providers.
+If you used or would like to use groups for authentication,
+check out the [OAuthenticator changelog](https://oauthenticator.readthedocs.io/en/stable/reference/changelog.html)
+
+## Other package upgrades
+
+- Python is upgraded from 3.11 to 3.12 in the Hub image
+- LDAPAuthenticator is upgraded from 1 to 2 ([changelog](https://github.com/jupyterhub/ldapauthenticator/blob/2.0.0/CHANGELOG.md#200---2024-10-18))
+- FirstUseAuthenticator is upgraded from 1.0 to 1.1 ([changelog](https://github.com/jupyterhub/firstuseauthenticator/blob/1.1.0/CHANGELOG.md))
+- idle culler is upgraded from 1.3.1 to 1.4.0 ([changelog](https://github.com/jupyterhub/jupyterhub-idle-culler/blob/1.4.0/CHANGELOG.md))
```
consideRatio, line 85: This is duplicated from the changelog, and I think its better to not duplicate it here.

consideRatio (commented):
I think the following are essential pieces for this docs:

- a link to z2jh general upgrade docs
- mention z2jhs default of `singleuser.storage.static.subPath="{username}"` for users of `singleuser.storage.static.type="static"` could need to re-configured to `"{escaped_username}"`

I figure its fine to retain JupyterHub and KubeSpawner headings, but I think it can be an improvement to cut out OAuthenticator and Other package upgrades headings as it otherwise overlaps with the changelog docs without fully capturing the changelogs docs either.

minrk:
Thanks for thinking so much about this! I think I've updated with what you suggested.

manics:
There's a regression in LDAPauthenticator which should be fixed by https://github.com/jupyterhub/ldapauthenticator/pull/294
We can either wait for confirmation from the bug reporter, or merge and release ldapauthenticator now so as not to further delay the Z2JH release.

minrk:
No problem waiting a bit more. We can publish once that PR lands and the hub image gets its bump.

consideRatio:
ldapauthenticator 2.0.2 is out now, bumped in z2jh's hub image, and I updated the changelog entries in this PR to reflect some recent PRs etc.

Release date updated to 2024-11-07 -- go for release tomorrow thursday?

consideRatio (approved):

minrk:
Great, let's do it!

minrk:
I'll merge this and publish 4.0 shortly.

minrk:
tag published, drafting announcements

PR #3564, opened 2024-10-31
refactor: no pins in hub image's requirements.in

consideRatio:
I explored if we could have dependabot bump major versions if we let this include major version pins. I think that would been an improvement, as it would help us distinguish major version bumps from the typical version bumps which I think makes sense to do in isolation.

However, this isn't made easy with dependabot, so I opted to not think about it for now.
- Files must be named `*requirements.txt` (https://stackoverflow.com/a/75909900/2220152)
- To ignore files and only focus on one requirements file, they must be put in different directories (https://github.com/dependabot/dependabot-core/issues/2883)

### DIFF
--- a/images/hub/requirements.in
+++ b/images/hub/requirements.in
@@ -1,32 +1,30 @@
-# This file is the input to requirements.txt,
-# which is a frozen version of this. To update
-# requirements.txt, use the "Run workflow" button at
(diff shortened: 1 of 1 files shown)


consideRatio:
Closing in favor of #3565

PR #3565, opened 2024-10-31
Pin and automate doing isolated bumps of hub image dependencies' major versions

consideRatio:
I had to rename and isolate requirements.in to `unfrozen/requirements.txt` because it was the only way for me to ensure dependabot worked as wanted.

> * Files must be named `*requirements.txt` (https://stackoverflow.com/a/75909900/2220152)
> * To ignore files and only focus on one requirements file, they must be put in different directories ([Allow/Ignore specific manifests dependabot/dependabot-core#2883](https://github.com/dependabot/dependabot-core/issues/2883))

I also had to add some arguments in ci/refreeze specifying files.

---

Dependabot function verified in a fork, and ci/refreeze function verified manually. No remnants of `requirements.in` found when searching.

image


.github/dependabot.yaml
```diff
@@ -15,3 +15,15 @@ updates:
       interval: monthly
       time: "05:00"
       timezone: Etc/UTC
+  - package-ecosystem: pip
+    directory: /images/hub/unfrozen
+    labels: [breaking]
```
minrk, line 20: nice isolation!

images/hub/unfrozen/requirements.txt
```diff
@@ -0,0 +1,33 @@
...
+jupyterhub-ltiauthenticator==1.*
+jupyterhub-nativeauthenticator==1.*
+jupyterhub-tmpauthenticator==1.*
+oauthenticator[googlegroups,mediawiki]==17.*
+
+# JupyterHub service shutting servers after a period of inactivity
+jupyterhub-idle-culler==1.*
+
+# Other optional dependencies for additional features
+pymysql==1.*  # mysql
```
minrk, line 29: fwiw, I wouldn't consider a bump of any of these to be a breaking change, but that's not a big deal. Perhaps they shouldn't be pinned?

minrk (approved):
Nice and makes sense! I'm not sure every bump of these will actually mean a breaking change for the chart, but I think it will be good to be prompted to make a choice when they come in.

images/hub/unfrozen/requirements.txt
consideRatio: Yeah hmmm I agree, they aren't meant to be a breaking change for the user.

It could be relevant to be given a chance to read the changelog before merging though. I'm _fine with removing pins or retaining them_, but I agree they are reasonably not to be listed as breaking in the end.

I think for all major bumps here, some may not be breaking in the end, so generally a manual consideration about breaking label or not this is probably needed.

images/hub/unfrozen/requirements.txt
```diff
@@ -0,0 +1,33 @@
...
+psycopg2==2.*  # postgres
+pycurl==7.*  # internal http requests handle more load with pycurl
+sqlalchemy-cockroachdb==2.* # cocroachdb
+statsd==4.*  # statsd metrics collection (TODO: remove soon, since folks use prometheus)
```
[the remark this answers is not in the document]
manics: I think we should treat most transient dependencies that are internal to JupyterHub as an implementation detail, unless our documentation assumes they've available for extension.

Potentially we could treat `pymysql` `psycopg2` as possibly breaking since they affect the integration of Z2JH with an external database (e.g. a new release of these libraries could prevent Z2JH being upgraded if someone's running a very old DB), but given that this isn't something we've considered before, and in practice it's rare to have breaking changes in database APIs, we can make maintenance easier by ignoring it.

PR #3571, opened 2024-11-09
Update Keycloak example (currently broken)

manics:
- `claim_groups_key` is deprecated, replaced by `auth_state_groups_key`
- `manage_groups` must be used with the above, otherwise JupyterHub fails to start

### DIFF
--- a/docs/source/administrator/authentication.md
+++ b/docs/source/administrator/authentication.md
@@ -425,7 +425,7 @@ documentation in this project also.
 [KeyCloak](https://www.keycloak.org) is an open source based provider of
 identity management that you can host yourself. Below is an example on how you
 can configure the GenericOAuthenticator class to authenticate against a KeyCloak
-server (version 17 or later).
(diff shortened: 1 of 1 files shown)


consideRatio:
To fix these docs are an improvement and happy to merge it, but an alternative improvement would be to delete them from z2jh and link to docs like this in oauthenticator instead. I figure that project benefits from a well maintained keycloak example, but that its hard to maintain it in z2jh/tljh/oauthenticator

consideRatio (approved):
Whats in the PR looks fine currently, since this is docs only and what we currently have is broken, i figure momentum is more important than deep scrutiny so approving already even though its in draft

manics:
We could make this a very minimal Keycloak example, I've got a CI tested example in https://github.com/manics/zero-to-jupyterhub-k8s-examples/tree/main/keycloak

manics:
I decided to take out managed groups/roles completely, as it's not very useful unless you also configure Keycloak.

PR #3573, opened 2024-11-11
documentation update on postStart hook use

k-popov:
Added a note on postStart hook is asynchronous from ENTRYPOINT execution.


consideRatio:
Thank you @k-popov!! I'm sorry for the slow turnaround!

PR #3588, opened 2024-12-11
add option to expose the automount_service_account_token option

icehunt:
This MR exposes the kubespawner option automount_service_account_token_option in the values.yaml for singleuser pods. This is useful if you want a sidecar to have access to the service account token but don't want the actual notebook container to have access to it.


consideRatio:
Thank you for engaging with the project @icehunt!!

There are too many options to manually propegate like this, so we implemented hub.config to pass config in the end. So you can already configure this like: hub.config.KubeSpawner.any_kubespanwer_config_here

See https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-config

PR #3594, opened 2025-01-10
Add dummy service accounts to hook-image-puller and continuous-image-puller pods

samyuh:
This MR adds two separate dummy ServiceAccounts for hook-image-puller and continuous-hook-image-puller.
- The continuous image puller runs constantly to ensure images are always available.
- The hook image puller is temporary, used during tasks like upgrades, and is deleted afterward.

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3545

### DIFF
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -198,11 +198,7 @@
 
 {{- /* hook-image-puller ServiceAccount */}}
 {{- define "jupyterhub.hook-image-puller-serviceaccount.fullname" -}}
-    {{- if .Values.prePuller.hook.serviceAccountImagePuller.create }}
-        {{- .Values.prePuller.hook.serviceAccountImagePuller.name | default (include "jupyterhub.hook-image-puller.fullname" .) }}
-    {{- else }}
-        {{- .Values.prePuller.hook.serviceAccountImagePuller.name | default "default" }}
-    {{- end }}
+    {{- .Values.prePuller.hook.serviceAccountImagePuller.name | default (include "jupyterhub.hook-image-puller.fullname" .) }}
 {{- end }}
 
 {{- /* continuous-image-puller DaemonSet */}}
@@ -212,11 +208,7 @@
 
 {{- /* continuous-image-puller ServiceAccount */}}
 {{- define "jupyterhub.continuous-image-puller-serviceaccount.fullname" -}}
-    {{- if .Values.prePuller.continuous.serviceAccount.create }}
-        {{- .Values.prePuller.continuous.serviceAccount.name | default (include "jupyterhub.continuous-image-puller.fullname" .) }}
-    {{- else }}
-        {{- .Values.prePuller.continuous.serviceAccount.name | default "default" }}
-    {{- end }}
+    {{- .Values.prePuller.continuous.serviceAccount.name | default (include "jupyterhub.continuous-image-puller.fullname" .) }}
 {{- end }}
 
 {{- /* singleuser NetworkPolicy */}}
--- a/jupyterhub/templates/_helpers-names.tpl
+++ b/jupyterhub/templates/_helpers-names.tpl
@@ -308,7 +308,9 @@ image-puller-priority: {{ include "jupyterhub.image-puller-priority.fullname" . 
 hook-image-awaiter: {{ include "jupyterhub.hook-image-awaiter.fullname" . | quote }}
 hook-image-awaiter-serviceaccount: {{ include "jupyterhub.hook-image-awaiter-serviceaccount.fullname" . | quote }}
 hook-image-puller: {{ include "jupyterhub.hook-image-puller.fullname" . | quote }}
+hook-image-puller-serviceaccount: {{ include "jupyterhub.hook-image-puller-serviceaccount.fullname" . | quote }}
 continuous-image-puller: {{ include "jupyterhub.continuous-image-puller.fullname" . | quote }}
+continuous-image-puller-serviceaccount: {{ include "jupyterhub.continuous-image-puller-serviceaccount.fullname" . | quote }}
 singleuser: {{ include "jupyterhub.singleuser.fullname" . | quote }}
 image-pull-secret: {{ include "jupyterhub.image-pull-secret.fullname" . | quote }}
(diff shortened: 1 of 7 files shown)


jupyterhub/templates/image-puller/serviceaccount-continuous.yaml
```diff
@@ -0,0 +1,16 @@
...
+{{- if .Values.prePuller.continuous.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "continuous-image-puller-serviceaccount"
+  labels:
+    {{- include "jupyterhub.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "0"
```
consideRatio, line 14: The hook annotations should be removed from here - this should be a normal reaource

jupyterhub/templates/image-puller/_helpers-daemonset.tpl
```diff
@@ -70,6 +70,11 @@ spec:
               {{- include "jupyterhub.userNodeAffinityRequired" . | nindent 14 }}
       {{- end }}
       terminationGracePeriodSeconds: 0
+      {{- if .hook }}
+      serviceAccountName: "hook-image-puller-serviceaccount"
+      {{- else }}
+      serviceAccountName: "continuous-image-puller-serviceaccount"
+      {{- end }}
```
consideRatio, line 77: The introduction of hardcoded names can cause issues in some edge case installation, so we have avoided it. Can you mimic how other resources have been named?

consideRatio:
I'm super low on time, but lets try this - do your best to address the review comments, and i'll try get it through to a merge by finishing details this weekend.

Details if you want to try work those as well, from the top of my head would include:
- ensure lint and validate values can trial configure this if new config is introduced
- render and check labels etc align with how other continious/hook resources have labels
I'm not able to really think properly about this now as i'm taking care of my toddler, sneaking in a review effort quickly from my mobile

samyuh:
Thanks for the feedback, I will address everything today.

jupyterhub/templates/image-puller/_helpers-daemonset.tpl
consideRatio: The name is set the same no matter what, this is the issue, so its done differently from how other resources get their name.

The daemonset name is for comparison named in this way:

name: {{ include "jupyterhub.hook-image-puller.fullname" . }}

jupyterhub/templates/image-puller/serviceaccount-continuous.yaml
```diff
@@ -0,0 +1,13 @@
+{{- /*
+ServiceAccount for the continuous image-puller daemonset
+*/}}
+{{- if .Values.prePuller.hook.serviceAccount.create -}}
```
consideRatio, line 4: This should be entirely unrelated to hook stuff

jupyterhub/templates/image-puller/_helpers-daemonset.tpl
samyuh: Thanks. 

I already updated the code and created a new entry, `.Values.prePuller.hook.serviceAccountImagePuller`. I did this so the users can have more flexibility defining serviceAccounts. 

Another alternative is to keep just one `.Values.prePuller.hook.serviceAccount` and do something similar to what is done to the daemonset name (by appending for example -image-puller at the end of serviceAccount)

Which approach do you prefer? I can change it to the second one if you like it more :)

consideRatio: Lets keep them separate, because this new one isn't supposed to be granted any permissions, while the one for the image-awaiter needs permissions inspecting daemonsets for example.

consideRatio:
@samyuh great work! I pushed to detail commits.

@manics do you have time for a quick gut-check evaluation? This PR adds two dummy service accounts to be used by the the image-puller pods to please various security checks as discussed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3545#issuecomment-2421643185.

Do we make them enabled by default, or not?

I'm leaning towards enabled by default to reduce the hassle for anyone needing to comply with these benchmarks etc, but I'm not fully confident it has no drawbacks for users in other situations.

consideRatio (approved):

samyuh:
> @samyuh great work! I pushed two detail commits.
> 
> @manics do you have time for a quick gut-check evaluation? This PR adds two dummy service accounts to be used by the the image-puller pods to please various security checks as discussed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3545#issuecomment-2421643185.
> 
> Do we make them enabled by default, or not?
> 
> I'm leaning towards enabled by default to reduce the hassle for anyone needing to comply with these benchmarks etc, but I'm not fully confident it has no drawbacks for users in other situations.

Thanks for your effort and your review!

jupyterhub/templates/_helpers-names.tpl
```diff
@@ -196,11 +196,29 @@
...
 {{- end }}
 
+{{- /* continuous-image-puller ServiceAccount */}}
+{{- define "jupyterhub.continuous-image-puller-serviceaccount.fullname" -}}
+    {{- if .Values.prePuller.continuous.serviceAccount.create }}
+        {{- .Values.prePuller.continuous.serviceAccount.name | default (include "jupyterhub.continuous-image-puller.fullname" .) }}
+    {{- else }}
+        {{- .Values.prePuller.continuous.serviceAccount.name | default "default" }}
+    {{- end }}
+{{- end }}
```
consideRatio: This mimiced some old practice, but I figure it doesn't make sense at least for these resources, so I didn't follow the practice here, and made this into::

```
{{- /* continuous-image-puller ServiceAccount */}}
{{- define "jupyterhub.continuous-image-puller-serviceaccount.fullname" -}}
    {{- .Values.prePuller.continuous.serviceAccount.name | default (include "jupyterhub.continuous-image-puller.fullname" .) }}
{{- end }}
```

samyuh:
I rebased my commits to keep the history clean and avoid including unnecessary commits.

consideRatio:
Thank you @samyuh, lets get this merged!!! :heart: :tada:

PR #3596, opened 2025-01-10
Add modern labels to pods controlled by deployments etc.

### LINKED ISSUE #3595: Modern labels not yet introduced for the pods controlled by deployments etc
The following labels are missing from the hub pod but are present in the singleuser pod. Is this the intended behavior?

```
app.kubernetes.io/name
app.kubernetes.io/instance
app.kubernetes.io/managed-by
```

You can verify this behavior by describing the resource or checking it manually with the command: 
```
helm template chart-release . --values values.yaml
```

samyuh:
Some labels are missing in the pods, such as:
```
    app.kubernetes.io/name: "jupyterhub"
    app.kubernetes.io/instance: "release-name"
```

The issue occurred because 'jupyterhub.matchLabels' was imported instead of 'jupyterhub.labels'.

Closes #3595

### DIFF
--- a/jupyterhub/templates/_helpers.tpl
+++ b/jupyterhub/templates/_helpers.tpl
@@ -124,52 +124,113 @@
       app.kubernetes.io/managed-by ("heritage")
 */}}
 {{- define "jupyterhub.commonLabels" -}}
+{{- if .legacyLabels -}}
 app: {{ .appLabel | default (include "jupyterhub.appLabel" .) | quote }}
 release: {{ .Release.Name | quote }}
 {{- if not .matchLabels }}
 chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 heritage: {{ .Release.Service }}
+{{- end }}
+{{- end }}
+{{- if and .legacyLabels .modernLabels -}}
+{{ printf "\n" }}
+{{- end }}
+{{- if .modernLabels -}}
 app.kubernetes.io/name: {{ .appLabel | default (include "jupyterhub.appLabel" .) | quote }}
 app.kubernetes.io/instance: {{ .Release.Name | quote }}
+{{- if not .matchLabels }}
 helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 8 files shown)


consideRatio:
Oh there were plenty of resources bugged like this - oops

samyuh:
Yup, I reviewed all of them :)

consideRatio:
Screenshot_20250110-213931

Hmmm okay so this is the catch and why it wasnt done. We don't want to always restart pods just because we upgrade, and some labels added now will trigger a restart now no matter what - those with chart versions in them

samyuh:
Yeah it makes sense. I will look into a way to have this working without having chart and heritage

consideRatio:
I'll refer to legacy and modern "match labels" below, with it I mean:
- legacy match labels: `component`, `app`, `release`
- modern match labels: `app.kubernetes.io/component`, `app.kubernetes.io/name`, `app.kubernetes.io/instance`

Note that I don't think pods should have the old `heritage` label or the modern `app.kubernetes.io/managed-by` label, because they are actually managed by the Deployment etc, which in turn is managed by Helm. Currently they only have the legacy match labels not including `heritage`.

I think currently we have a need for:
- specifying legacy match labels on `spec.selector.matchLabels` as we already do, which is done by including the `jupyterhub.matchLabels` helper.
- specifying legacy and new match labels on `spec.template.metadata.labels`, where we currently just provide the legacy labels, but want to provide both legacy and modern.

I think in the future, we'll have a need for specifying just the new match labels on `spec.selector.matchLabels`.

---

With this in mind, I think practically, we should have one new helper function named for example like `jupyterhub.modernMatchLabels` which is added to be included next to `jupyterhub.matchLabels` in `spec.template.metadata.labels`, and in the future, we'll replace use of `jupyterhub.matchLabels` in `spec.selector.matchLabels` with `jupyterhub.modernMatchLabels`.

consideRatio:
@samyuh I hope I didn't cause trouble pushing a commit without a heads up - I was worried I wouldn't find time to review future work in the near future and wanted to get this resolved.

image

samyuh:
No worries! I'm a bit tied up this weekend and haven't had a chance to work on this yet. I really appreciate your help! Thanks again for taking the time to fix this!

PR #3605, opened 2025-01-16
Use GitHub hosted Ubuntu arm64 runners instead of CircleCI

manics:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

This is on top of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3604

Currently there are some bugs in the runner image:
- https://github.com/actions/partner-runner-images/issues/27
- https://github.com/actions/partner-runner-images/issues/25

This results in helm attempting to create a config file n a location it doesn't have access to. We could fix the environment variable `XDG_CONFIG_HOME`, but let's see if it's fixed upstream.

If we want to merge this we'll need to disable the CircleCI check which is now failing since I've deleted `.circleci/config.yml`


consideRatio:
Oooooh nice fix transitioning from ancient minica image -> openssl image!!

```
certificate is valid for localhost, pebblepebble.defaultpebble.default.svcpebble.default.svc.cluster.local, not pebble" 
```

Seems like `pebble` and `pebble.default` etc was merged into a single string unintentionally.

manics:
Done in https://github.com/jupyterhub/pebble-helm-chart/pull/44/commits/aa6fc2b2a1fb10315f41f7a932b65cd073946bc5

consideRatio:
GitHub pages cache seems to serve the latest stuff already, we should be good to go here without using your fork again!

```
helm show chart --repo=https://hub.jupyter.org/helm-chart/ pebble | yq e '.version' -
1.3.0
```

consideRatio (approved):
Going to bed, but this looks great assuming tests pass after going back to using the pebble 1.3.0 version with you fix!

manics:
Green!

PR #3661, opened 2025-04-30
Add support for K8S Gateway API (`httpRoute` config)

manics:
@Kyrremann I hope you don't mind, but I've rebased your PR https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3222 and updated it with v1 instead of v1beta, removed the GKE specific resource, and added some tests.

### DIFF
--- a/.github/workflows/test-chart.yaml
+++ b/.github/workflows/test-chart.yaml
@@ -132,17 +132,13 @@ jobs:
             test: install
             # Also test arm64
             os: ubuntu-24.04-arm
-          - k3s-channel: stable # also test hub-slim, prePuller.hook, gateway
+          - k3s-channel: stable # also test hub-slim, and prePuller.hook
             test: install
             local-chart-extra-args: >-
               --set hub.image.name=quay.io/jupyterhub/k8s-hub-slim
               --set prePuller.hook.enabled=true
               --set prePuller.hook.pullOnlyOnChanges=true
-              --set httpRoute.enabled=true
-              --set httpRoute.gateway.name=traefik-gateway
-              --set httpRoute.gateway.namespace=kube-system
             os: ubuntu-24.04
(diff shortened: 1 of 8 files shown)


.github/workflows/test-chart.yaml
```diff
@@ -394,3 +394,92 @@ jobs:
         if: failure() && matrix.setup-postgresql-args
         run: |
           kubectl logs statefulset/postgresql
+
+  test-gateway-ingress:
+    # The main test configures autohttps, but this conflicts with testing
+    # ingress/gateway which assume they can directly connect to proxy-public
```
manics, line 400: In theory it should be possible to have autohttps working, but that would only be used for internal connections between the traefik ingress/gateway controller and the autohttps/proxy pod which effectively means it's no use, since the certificate is designed for external connections.

The certificate for external connections has to be managed by the gateway or ingress controller, not autohttps.
https://gateway-api.sigs.k8s.io/guides/tls/

manics: Rereading https://gateway-api.sigs.k8s.io/guides/tls/#tls-configuration there's a `Passthrough` mode, but this is _Experimental_ and I don't see what benefit it offers for autohttps.

consideRatio (approved):
Thank you @Kyrremann and @manics! This looks right and sound!

I pushed two commits and think this could be ready to merge. What do you think?

I was thinking if it made sense to allow both ingress and httproute, but I think it does - right? I'm thinking there could be an ingress controller next to a gateway controller, and they both could help facilitate inbound connections to some k8s Service's endpoints.

manics:
Thanks for the additions! Both gateway and ingress can be used together, in the same way that you can have multiple ingresses for the same service. It'll be confusing for other admins so not recommended unless you really need it.

PR #3683, opened 2025-05-30
Add `proxy.service.externalIPs` config

holzman:
This adds the ability to configure an externalIP array for the proxy service.

(One particular use case: using external L4 filtering/firewalling for a shared k8s cluster so we can whitelist jupyterhub)

### DIFF
--- a/tools/templates/lint-and-validate-values.yaml
+++ b/tools/templates/lint-and-validate-values.yaml
@@ -223,6 +223,9 @@ proxy:
     nodePorts:
       http:
       https:
+    externalIPs:
+      - 123.123.123.123
+      - 123.123.123.124
   chp:
     revisionHistoryLimit: 1
     extraCommandLineFlags:


manics (approved):

jupyterhub/templates/proxy/service.yaml
```diff
@@ -81,3 +81,8 @@ spec:
     {{- . | toYaml | nindent 4 }}
   {{- end }}
   {{- end }}
+  {{- with .Values.proxy.service.externalIPs }}
+  externalIPs: {{- range . }}
```
consideRatio, line 85: This probably works fine in this case, but is a risky pattern since number looking strings becomes numbers like this, while using toYaml wouldnt.

Can you switch this to use {{ . | toYaml | nindent 4 }} instead? I'm on mobile and failed to provide a code suggestion spanning multiple lines.

consideRatio: I pushed a commit for this

Issue #2746, opened 2022-06-08
Consider autogenerating schema for hub.config passthrough


manics:
### Proposed change
Thinking aloud.... `hub.config.<class>` allows arbitrary pass through of traitlets. The downside is it bypasses the Helm chart schema validation. Is it worth trying to autogenerate `hub.config.<class>` schemas for a subset of "supported" classes?

### Alternative options
Do nothing.

### Who would use this feature?
Anyone wanting to use pass-through configuration but still keep schema validation.

### (Optional): Suggest a solution
See e.g. https://gist.github.com/manics/ce5ce8d05492ee3d1cf8217da14829c7 for a very preliminary proof-of-concept.

We could probably modify schema.yaml to contain custom metadata to indicate traitlets should be looked-up.
As part of this we could also include a deny-list of properties such as `JupyterHub.base_url`

minrk:
I see the sample script uses `getmembers`, but there's an easier way to get all configurable traits on any class:

```python
config_traits = cls.class_traits(config=True)
```

> I figure the logic would be to allow unknown classes, but to be strict about a few known classes?

Yes, I think so. Since folks can install additional packages, additional classes need to be allowed, though mistyping a class name is a likely source of typos the schema would like to catch.

I think it's also good to be loose about any traitlets we don't know how to handle. So give a type for the simple str/number/dict cases (already done by @manics), and then accept any type for any traitlet you don't recognize. I suspect the type validation is far less valuable than validating that you are using recognized keys. If the types are wrong, JupyterHub startup will fail informatively. It's the unrecognized keys where behavior is a lot harder to figure out and both more useful and easier to check via a schema.

PR #3721, opened 2025-08-06
raise on invalid keys in hub.config

minrk:
Traitlets requires that valid sub-config keys must be class names starting with uppercase. If they are lowercase, something is wrong and config will be ignored. Probably best not to do so silently, which can lead to long, frustrating debugging before realizing that config is being ignored.

e.g. if you put something like `hub.loadRoles` under `hub.config.loadRoles`, it will silently have no effect. This change makes it an informative, fatal error.

In the unlikely event that this shouldn't be fatal somewhere, we could reduce this to a warning, but I think it's safe to make it fatal.

We should probably raise (or at least warn) on this in traitlets itself, too. There's a design constraint there that a top-level Config object doesn't distinguish from sub-config (i.e. `type(c.JupyterHub) is type(c)`), so raising _on assignment_ may be hard, but that's not insurmountable since it is known _at load time_, but we know more about what `hub.config` is and this is a more likely mistake in our mixed helm/traitlets config situation than any other traitlets config application.

cc @agoose77

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -478,16 +478,33 @@ c.JupyterHub.cookie_secret = get_secret_value("hub.config.JupyterHub.cookie_secr
 c.CryptKeeper.keys = get_secret_value("hub.config.CryptKeeper.keys").split(";")
 
 # load hub.config values, except potentially seeded secrets already loaded
-for app, cfg in get_config("hub.config", {}).items():
-    if app == "JupyterHub":
+for section, cfg in get_config("hub.config", {}).items():
+    if section == "JupyterHub":
         cfg.pop("proxy_auth_token", None)
         cfg.pop("cookie_secret", None)
         cfg.pop("services", None)
-    elif app == "ConfigurableHTTPProxy":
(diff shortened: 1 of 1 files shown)


minrk:
This doesn't go nearly as far as #2746, which I think is a cool idea, but it now validates the bare minimum to be valid traitlets pass-through config:

- top-level keys start with a capital
- top-level values are objects

Produces the validation output with this config:

```yaml
hub:
  config:
    JupyterHub:
      foo: bar
    
    UnrecognizedClass:
      is: okay

    String: "string"
    
    Array:
      - list
      - of
      - items

    loadRoles:
      sec: key
```

output:

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- hub.config.Array: Invalid type. Expected: object, given: array
- hub.config.String: Invalid type. Expected: object, given: string
- hub.config: Additional property loadRoles is not allowed
```

which I think is about right, catching the three definitely-invalid cases, and leaving the maybe-a-typo case alone, since it can't be guaranteed to be wrong at validation time.

manics:
One broken linkcheck is due to https://github.com/jupyterhub/repo2docker/issues/1435 , the other is probably temporary

PR #3763, opened 2025-11-08
Set working directory for JupyterHub deployment

haobibo:
In case the image's default working dir is not set to /srv/


minrk:
closing this as awaiting response for a few months. Feel free to reopen with further discussion if/when you have time and want to pursue this further.

PR #3797, opened 2026-01-15
Test with helm version 4

yuvipanda:
Tests with https://github.com/jupyterhub/action-k3s-helm/pull/126


yuvipanda:
These issues are real I think.

.github/workflows/test-chart.yaml
```diff
@@ -433,12 +445,13 @@ jobs:
       # kubectl and helm
       #
       # ref: https://github.com/jupyterhub/action-k3s-helm/
-      - uses: jupyterhub/action-k3s-helm@v4
+      - uses: jupyterhub/action-k3s-helm@main
```
manics:
```suggestion
      - uses: jupyterhub/action-k3s-helm@v4
```

PR #3800, opened 2026-01-21
Support `httpRoute.gateway.sectionName` for Gateway listener selection

groundsada:
### Summary
Add optional support for setting `spec.parentRefs[].sectionName` on generated `HTTPRoute` resources.

### Motivation
Gateways often expose multiple listeners (e.g. `http`, `https`, or custom names).  
This allows explicitly binding an `HTTPRoute` to the intended listener.

### Changes
- Render `sectionName` in `jupyterhub/templates/httproute.yaml` when provided
- Extend schema with `httpRoute.gateway.sectionName`
- Document option in `values.yaml`

### Usage
```yaml
httpRoute:
  enabled: true
  gateway:
    name: your-gateway
    namespace: your-gateway-namespace
    sectionName: https
```

### Compatibility
Backwards compatible. Behavior is unchanged if `sectionName` is not set.

### Testing
- Verified rendered `HTTPRoute` includes `sectionName` when configured
- Verified no change in output when unset

### DIFF
--- a/jupyterhub/templates/httproute.yaml
+++ b/jupyterhub/templates/httproute.yaml
@@ -12,12 +12,8 @@ metadata:
 spec:
   parentRefs:
     - kind: Gateway
-      name: {{ .Values.httpRoute.gateway.name }}
-      {{- with .Values.httpRoute.gateway.namespace }}
-      namespace: "{{ . }}"
-      {{- end }}
-      {{- with .Values.httpRoute.gateway.sectionName }}
-      sectionName: "{{ . }}"
+      {{- with .Values.httpRoute.gateway }}
+      {{- . | toYaml | nindent 6 }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


jupyterhub/values.yaml
```diff
@@ -728,6 +728,7 @@ httpRoute:
   gateway:
     name: your-gateway
     namespace: "" # required for a gateway resource in another namespace
+    # sectionName: https # optional: specify Gateway listener section (e.g., 'http', 'https')
```
```suggestion
    # sectionName is optional: specify Gateway listener section (e.g., "http", "https")
    sectionName: ""
```

Change to let it be empty string by default instead of undefined to align with chart practices better

consideRatio (approved):
Configuration that maps directly without fuss, seems fine to add! @manics does it look right?

PR #3869, opened 2026-04-10
Features: Add per-singleuser Service/Route exposure"

Salamastik:
## Summary

Adds a new `singleuser.exposure` feature that creates a dedicated external resource per single-user server.

### What it does
- Enables per-user Service exposure for `LoadBalancer` or OpenShift `Route`
- Keeps a stable name per user server
- Supports optional SSH port mapping
- Supports templated labels, annotations, and Route hostnames
- Cleans up the created Service/Route when the server stops

### Implementation
- Added dedicated Python modules for the exposure logic
- Wired the feature into hub configuration via post-extra config loading
- Added RBAC rules for Service and Route management
- Added schema and example values for the new configuration

### Notes
- `LoadBalancer` mode works on Kubernetes clusters that support it
- `Route` mode is intended for OpenShift clusters with the Route API

### DIFF
--- a/jupyterhub/files/hub/jupyterhub_config.py
+++ b/jupyterhub/files/hub/jupyterhub_config.py
@@ -16,9 +16,13 @@ for configuration_directory in (
     os.path.dirname(os.path.realpath(__file__)),
     "/usr/local/etc/jupyterhub/config",
 ):
-    if os.path.isdir(configuration_directory) and configuration_directory not in sys.path:
+    if (
+        os.path.isdir(configuration_directory)
+        and configuration_directory not in sys.path
+    ):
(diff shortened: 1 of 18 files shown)


minrk:
Sorry we haven't communicated this clearly, yet, but we are working on updating our policy to prohibit LLM-generated changes in the JupyterHub org. Since this PR contains a lot of changes in inappropriate places, attributes the changes to a different user than the one that submitted the PR, both of which are indicative of an LLM, and would cost a lot of time to review and direct to the right repos, I think we should close this rather than provide a detailed review. If you want to propose a feature and discuss how it might be implemented, feel free to open an Issue.

Issue #3719, opened 2025-07-31
singleuser.networkPolicy.egressAllowRules[nonPrivateIPs: fals,privateIPs: false] bug


JiangDonglai98:
### How to reproduce

use this config:
singleuser:
  image:
    name: 11.50.52.80:5000/jupyter/notebook-ubuntu-py-3-11-6
    tag: "v1"
    pullPolicy: IfNotPresent
  cmd: null
  storage:
    capacity: 300Gi
    type: dynamic
    dynamic:
      storageClass: managed-nfs-storage
      pvcNameTemplate: claim-{username}{servername}
      volumeNameTemplate: volume-{username}{servername}
      storageAccessModes: [ReadWriteMany]
    extraVolumes: 
      - name: volume-common-dataset
        persistentVolumeClaim:
          claimName: claim-common-dataset
    extraVolumeMounts:
      - mountPath: /home/dataset
        name: volume-common-dataset
  networkPolicy:
    egressAllowRules:
      nonPrivateIPs: false
      privateIPs: false

hub:
  config:
    JupyterHub:
      admin_access: true   
      authenticator_class: dummy   
  
  db: 
    type: mysql
    url: mysql+pymysql://root:root-jupyterhub@11.50.52.80:3306/jupyterhub
    
  image:
    name: 11.50.52.80:5000/jupyterhub/k8s-hub
    tag: "4.2.0"
    pullPolicy: IfNotPresent
  networkPolicy:
    enabled: true
    egress:
      - to:
          - ipBlock:
              cidr: 11.50.52.80/32
        ports:
          - port: 3306
            protocol: TCP
then do: helm upgrade --cleanup-on-fail --install jd-jupyterhub /home/jupyterhub-setup/jupyterhub --namespace public-jupyter-notebooks --create-namespace --values /home/jupyterhub-setup/jupyterhub/config.yaml --debug
everything is fine and have: kubectl get pod -n public-jupyter-notebooks
public-jupyter-notebooks   continuous-image-puller-5lnqp              1/1     Running   0          16m
public-jupyter-notebooks   continuous-image-puller-km54c              1/1     Running   0          16m
public-jupyter-notebooks   hub-78dfb457b6-fjvrp                       1/1     Running   0          16m
public-jupyter-notebooks   nfs-client-provisioner-757cc64846-zddcd    1/1     Running   0          7h20m
[...]

PR #3892, opened 2026-05-08
NetworkPolicies: Treat link-local IPs as private IPs

mittler-works:
When using a cluster with [NodeLocal DNSCache](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/), the DNS Server used by pods is on a link local address (`169.254.0.0/16`), by default on `169.254.20.10/32`.

This subnet is not yet whitelisted for Port 53 UDP/TCP (required for DNS). If you have set `*.networkPolicy.egressAllowRules.nonPrivateIPs` to `true`, it doesn't stand out, because it will be allowed through the `0.0.0.0/0` rule, because the link local subnet is not in the exclusion list. (<-- maybe I should add another PR for that, too?)

Anyway, if you set `nonPrivateIPs` to `false` on a cluster that's using NodeLocal DNSCache, you cannot reach your Jupyter Notebooks anymore and your setup is broken.

This PR adds an option to whitelist link local addresses for Port 53 UDP/TCP (all of them, as it does not have to be exactly `169.254.20.10/32`). It's set to `true` by default.

Edit: there was also [an issue (#3719)](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3719) with this problem.

### DIFF
--- a/jupyterhub/templates/_helpers-netpol.tpl
+++ b/jupyterhub/templates/_helpers-netpol.tpl
@@ -30,7 +30,7 @@
 {{- define "jupyterhub.networkPolicy.renderEgressRules" -}}
 {{- $root := index . 0 }}
 {{- $netpol := index . 1 }}
-{{- if or (or $netpol.egressAllowRules.dnsPortsCloudMetadataServer $netpol.egressAllowRules.dnsPortsKubeSystemNamespace) $netpol.egressAllowRules.dnsPortsPrivateIPs }}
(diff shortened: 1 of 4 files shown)


mittler-works:
It would probably be more straight forward to simply classify linklocal IPs under "Private IPs" as well - which, fundamentally, is exactly what they are.
This also renders additional configuration options unnecessary and does not further bloat the setup.
I'll edit this PR accordingly.

Issue #2887, opened 2022-09-21
Automatic secret generation triggers constant redeploy (ArgoCD)


BlueCog:
### Scenario

Deploying [zero-to-jupyterhub-k8s](https://github.com/jupyterhub/zero-to-jupyterhub-k8s) via [ArgoCD ](https://argo-cd.readthedocs.io/en/stable/)

ArgoCD checks now and then on changes in the code base. If this code base is Helm ArgoCD will use the `helm template` command. 

Because the tokens are automatically generated. For example: [schema-proxy-secrettoken ](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#schema-proxy-secrettoken )ArgoCD constantly thinks there is a new set of code (tokens) and will try to apply the latest state what in turn will result in a new deployment. This process repeats constantly.

Now the docs note that I can a default token via override file. But we don't want that because this will mean the secret is in plain text in the code base. We'd like to apply the secret for example via the [existingsecret](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-existingsecret) method. 

### Proposed change

Be able to set ALL secrets via [existingsecret](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-existingsecret)

### Who would use this feature?
Everyone who works from a code base and don't want to commit secrets.

consideRatio:
`hub.existingSecret` is a reference to a self managed k8s Secret with configuration.

This works for most things, because in the `hub` pod we can optionally mounts the optional k8s Secret declared in `hub.existingSecret` and the Helm chart managed k8s Secret and then use Python to merge these. But, specifically with regards to the configuration `proxy.secretToken`, this is tricky. It is needed to be known by both the `hub` pod and the `proxy` pod - its the key they share between them to let the `hub` control the `proxy` via a REST API - it must be known by both.

So, we would need similar logic of merging configuration also in the proxy pod as we have in the hub pod if this is to be supported. Currently, the proxy pod reads the k8s Secret managed by the Helm chart like this - and doesn't respect `hub.existingSecret`.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/dfeb8ea4a610a5c7d91d21b6d3c2d39e20f2d290/jupyterhub/templates/proxy/deployment.yaml#L116-L124

[...]

PR #3903, opened 2026-05-28
fix: make hub.existingSecret fully replace the chart managed Secret

erwinschliske:
When `hub.existingSecret` is set, the chart still renders its own hub Secret
with randomly generated tokens. In ArgoCD/GitOps setups where `lookup()` has
no cluster access, these tokens change on every render, which rolls the hub pod
on every reconcile.

This PR makes `hub.existingSecret` actually replace the token keys:

- Omit the random token keys from the hub Secret when `existingSecret` is set (values.yaml, db.password and extraFiles are still rendered — those are deterministic and needed for volume mounts)
- Point `CONFIGPROXY_AUTH_TOKEN` in hub and proxy Deployments to the user-managed Secret
- The `checksum/secret` annotation now stabilises because the hub Secret no longer contains random values

The `hub-existing-secret-or-default` helper already existed — it just wasn't
wired up in the right places.

No change when `hub.existingSecret` is not set.

### DIFF
--- a/ci/test-hub-existing-secret.yaml
+++ b/ci/test-hub-existing-secret.yaml
@@ -7,6 +7,8 @@ metadata:
 type: Opaque
 stringData:
   hub.services.test-hub-existing-secret.apiToken: ffff9999
+  hub.services.test.apiToken: give-pytest-control
+  hub.services.test-with-scoped-access.apiToken: give-pytest-scoped-control
   hub.config.ConfigurableHTTPProxy.auth_token: ffff9999
   hub.config.JupyterHub.cookie_secret: ffff9999
   hub.config.CryptKeeper.keys: ffff9999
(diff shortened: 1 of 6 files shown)


erwinschliske:
Closes #2887

minrk:
Sorry for the delay!

I think this makes sense. It is a breaking change because the docs say it will be _merged_ and given highest priority, whereas this changes it to require being _exclusively_ used, so prior `existingSecrets` that set a subset of keys used to work and will no longer work after this.

I think that's okay, we just need to update [the schema](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/a5b4c44d4c0b3defa855f76c510887b573d64f76/jupyterhub/values.schema.yaml#L1220) so this change shows up in documentation.

mauriziomartino-px:
I have been having an issue with exactly this behaviour and would love to see this released. @erwinschliske I am happy to update the docs if you don't have time, as this is quite urgent for us.

Issue #3937, opened 2026-07-17
Allow overriding the `chp` container's `command` in the proxy Deployment (to support minimal/hardened `configurable-http-proxy` images)


Upanshu11:
## Summary
 
The proxy Deployment template hardcodes the `chp` container's `command` to invoke `configurable-http-proxy` as a bare executable name:
 
```yaml
# jupyterhub/templates/proxy/deployment.yaml
containers:
- name: chp
  image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
  command:
    - configurable-http-proxy
    - "--ip="
    - "--api-ip="
    - --api-port=8001
    - --default-target={{ ... }}
    - --error-target={{ ... }}
    - --port=8000
    ...
```
 
Because Kubernetes' `command:` field fully replaces a container's image `ENTRYPOINT` (rather than only supplying args), this hardcodes how CHP must be launched regardless of what a given `proxy.chp.image` actually expects. This works fine for the standard `jupyterhub/configurable-http-proxy` image, since it's Node/Alpine-based, has a shell and `/usr/bin/env`, and the CLI script's own shebang resolves at exec time.
 
It breaks for minimal/hardened CHP images (e.g. Docker Hardened Images) that ship without a shell. Those images intentionally define their own entrypoint as `["node", "/path/to/configurable-http-proxy"]`, specifically to avoid depending on shebang resolution — but the chart's hardcoded `command:` bypasses that entrypoint entirely and execs the bare script by name instead, so it fails with an error like:
 
```
exec /srv/configurable-http-proxy/bin/configurable-http-proxy: no such file or directory
```
 
even though the file exists — the actual cause is that the script's `#!/usr/bin/env node` shebang can't resolve without a shell/`env` present.
 
## Steps to reproduce
 
1. Set `proxy.chp.image.name`/`proxy.chp.image.tag` to a minimal/hardened CHP image that has no shell and expects to be launched via `node <script>` rather than as a standalone executable.
2. `helm upgrade` the release.
3. The `chp` pod fails to start / CrashLoopBackOffs with `exec ...: no such file or directory`.
## Expected behavior
 
The chart should provide a values.yaml field to override (or extend) the `chp` container's `command`, similar in spirit to `proxy.chp.extraCommandLineFlags` but for the base command/interpreter itself — for example:
 
```yaml
proxy:
  chp:
[...]

PR #3938, opened 2026-07-17
Add `proxy.chp.command` config

Upanshu11:
For feature #3937 - separate out `command` and `args` in proxy's deployment, so that a different `command` can be used.

### DIFF
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -75,8 +75,8 @@ spec:
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
           {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
-          command:
-            - configurable-http-proxy
+          command: {{ .Values.proxy.chp.image.command }}
+          args:
             - "--ip="
             - "--api-ip="
             - --api-port=8001
--- a/jupyterhub/templates/proxy/deployment.yaml
+++ b/jupyterhub/templates/proxy/deployment.yaml
@@ -75,7 +75,8 @@ spec:
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
           {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
-          command: {{ .Values.proxy.chp.image.command }}
+          command:
+            {{- .Values.proxy.chp.image.command | toYaml | nindent 12 }}
           args:
             - "--ip="
             - "--api-ip="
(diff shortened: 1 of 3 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -1591,6 +1591,31 @@ properties:
...
           revisionHistoryLimit: *revisionHistoryLimit
           networkPolicy: *networkPolicy-spec
+          command:
+            type: array
+            description: |
+              A list of strings to be added as the command.
+              In general, it needs to start with
+              [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy)
+              that will be expanded with Helm's template function `tpl` which
+              can render Helm template logic inside curly braces (`{{ ... }}`).
```
consideRatio: command isnt being expanded with tpl, so this needs an update one way or another. Removing comments about expanding with tpl here for example

jupyterhub/values.schema.yaml
```diff
@@ -1591,6 +1591,31 @@ properties:
...
+              ```
+
+              Note that this is added so that docker hardened image can also be used,
+              where the expectation is:
+              ```yaml
+              proxy:
+                chp:
+                  command:
+                    - "node"
+                    - "/srv/configurable-http-proxy/bin/configurable-http-proxy"
```
consideRatio: This example was hard for me to understand since its like the default behavior, perhaps better omitted

jupyterhub/values.schema.yaml
Upanshu11: removed

jupyterhub/values.schema.yaml
Upanshu11: removed

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -75,8 +75,8 @@ spec:
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
           {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
-          command:
-            - configurable-http-proxy
+          command: {{ .Values.proxy.chp.image.command }}
```
consideRatio: Not sure how this template renders without piping into | toYaml or | toJson.

Tested, this would break things.

```
helm template jupyterhub --show-only templates/proxy/deployment.yaml | grep -A10 command
          command: 
          args:
            - "--ip="
            - "--api-ip="
            - --api-port=8001
            - --default-target=http://hub:$(HUB_SERVICE_PORT)
            - --error-target=http://hub:$(HUB_SERVICE_PORT)/hub/error
            - --port=8000
```

I think this may do the trick, try yourself!

```suggestion
          command:
            {{- .Values.proxy.chp.command | toYaml | nindent 12 }}
```

Upanshu11: Makes sense

jupyterhub/templates/proxy/deployment.yaml
```diff
@@ -75,7 +75,8 @@ spec:
           image: {{ .Values.proxy.chp.image.name }}:{{ .Values.proxy.chp.image.tag }}
           {{- $hubNameAsEnv := include "jupyterhub.hub.fullname" . | upper | replace "-" "_" }}
           {{- $hubHost := printf "http://%s:$(%s_SERVICE_PORT)" (include "jupyterhub.hub.fullname" .) $hubNameAsEnv }}
-          command: {{ .Values.proxy.chp.image.command }}
+          command:
+            {{- .Values.proxy.chp.image.command | toYaml | nindent 12 }}
```
consideRatio: I made a mistake, it shouldn't be chp.image.command, but just chp.command here.

consideRatio:
It seems that what it was actually also worked, at least with helm 4 (it rendered as json) - I made a mistake while testing!

jupyterhub/templates/proxy/deployment.yaml
Upanshu11: oh yaa, already fixed it

consideRatio (approved):

consideRatio:
Thank you @Upanshu11!! :heart: :tada: :sunflower:

Upanshu11:
@consideRatio thanks for merging it :)

Issue #3769, opened 2025-11-21
Setting `singleuser.storage.extraVolumes` and `extraVolumeMounts` as arrays in subchart, fails


GeorgianaElena:
### Bug description

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3690 introduced the option to define `singleuser.storage.extraVolumes` and `extraVolumeMounts` as dicts, while also allowing it to be an array according to [the schema](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0a3b1bad94d3e0fd31c3d1e207588b7e81bda4aa/jupyterhub/values.schema.yaml#L2451-L2473).

However, in practice, when I tried to upgrade a hub from z2jh 4.0.0 to 4.3.1, without updating  `singleuser.storage.extraVolumes` to a map, I saw a bunch of helm warnings about it because of the type mismatch. 

This is because z2jh values.yaml sets the default value to be `{}`, so only dicts will be accepted, otherwise, even though the upgrade happens, the configuration just gets discarded if it's set as  an array I beleive.

```
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumeMounts is a table. Ignoring non-table value ([map[mountPath:/home/jovyan/shared name:home readOnly:true subPath:_shared] map[mountPath:/dev/shm name:dev-shm] map[mountPath:/home/rstudio name:home subPath:{escaped_username}] map[mountPath:/home/rstudio/shared name:home readOnly:true subPath:_shared]])
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumes is a table. Ignoring non-table value ([map[emptyDir:map[medium:Memory] name:dev-shm]])
coalesce.go:220: warning: cannot overwrite table with non table for basehub.jupyterhub.singleuser.storage.extraVolumeMounts (map[])
coalesce.go:220: warning: cannot overwrite table with non table for basehub.jupyterhub.singleuser.storage.extraVolumes (map[])
[...]

PR #3690, opened 2025-06-12
Support dict values for volumes, volume_mounts and other list based configs that support dict values

sunu:
KubeSpawner now supports dict values for volumes, volume_mounts etc. This allows for easy overriding when required. ref: https://github.com/jupyterhub/kubespawner/pull/843 and https://github.com/jupyterhub/kubespawner/pull/845

This PR makes sure `jupyterhub_config.py` also support defining these values as dictionaries and changes the default of these list based configurations to be dict.

PR #3943, opened 2026-07-23
Support dict values for profileList

### LINKED ISSUE #6552: Add support for dict-based profile list definitions to zero-to-jupyterhub-k8s
> [!Important]
> Follow-on from #6941 

This PR https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3690 adds support for dict-based configuration of list items in zero-to-jupyterhub-k8s. It is not exhaustive, and in particular does not add this option to profile list definitions.

In a recent conversation about how to design hubs, @yuvipanda confirmed that we want to lift as much configuration into common configuration files (per cluster) as possible. By adding support for defining profile lists as dictionaries, it will be possible to define extensible profile lists in the common configuration.

### Steps
- Modify the `profile_list` schema to accept a dictionary of dictionaries, or array of dictionaries.
- Add handling logic to `jupyterhub_config.py` (e.g. https://github.com/sunu/zero-to-jupyterhub-k8s/blob/9fca3ce184d161483fa62c87d7a5c41f6dc15a5a/jupyterhub/files/hub/jupyterhub_config.py#L343-L348)
- Add test to ensure this new config works
- Update z2jh docs to reflect this new capability

### Definition of Done
- [x] A PR has been authored that adds support for dict-based profile lists
- [x] The PR has been tested
- [x] The PR has been reviewed
- [x] The PR has docs
- [x] An issue has been created to track applying this change to our clusters https://github.com/2i2c-org/infrastructure/issues/8822

GeorgianaElena:
### Checklist

- [x] I am the author of this work

### What does this PR do?

Type of change:
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [ ] Other

This is a continuation of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3690.|
Fixes https://github.com/2i2c-org/infrastructure/issues/6552

No

A test has been added.

### What should a reviewer concentrate their feedback on?

This PR should be pretty straightforward:
- profileList now accepts a dict, a list, or null
- profileList now defaults to null to avoid https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3769
- All examples in docs have been updated to show profileList as dict.

### DIFF
--- a/jupyterhub/values.yaml
+++ b/jupyterhub/values.yaml
@@ -441,7 +441,7 @@ singleuser:
   cmd: jupyterhub-singleuser
   defaultUrl:
   extraPodConfig: {}
-  profileList: []
+  profileList:
 
 # scheduling relates to the user-scheduler pods and user-placeholder pods.
 scheduling:
--- a/dev-config-local-chart-extra-config.yaml
+++ b/dev-config-local-chart-extra-config.yaml
@@ -19,7 +19,14 @@ singleuser:
       test-volume-mount:
         name: test-volume
         mountPath: /test-volume
-
+  profileList:
+    default:
+      display_name: "Default environment"
+      description: "Here just for testing"
+      default: true
+      kubespawner_override:
+        environment:
+          AM_I_HERE: "yes"
 proxy:
   traefik:
     # This startup delay can help the k8s container network interface setup
--- a/docs/source/administrator/optimization.md
+++ b/docs/source/administrator/optimization.md
@@ -154,10 +154,12 @@ singleuser:
     name: jupyter/minimal-notebook
     tag: 2343e33dec46
   profileList:
-    - display_name: "Minimal environment"
+    minimal:
+      display_name: "Minimal environment"
       description: "To avoid too much bells and whistles: Python."
       default: true
-    - display_name: "Datascience environment"
+    datascience:
+      display_name: "Datascience environment"
       description: "If you want the additional bells and whistles: Python, R, and Julia."
       kubespawner_override:
         image: jupyter/datascience-notebook:2343e33dec46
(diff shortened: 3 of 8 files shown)


consideRatio:
Hey @GeorgianaElena!!

The prePuller pulls images ahead of time, and i recall it could also pull things from this list. Now that its an object or list or null, does it need an update as well?

GeorgianaElena:
@consideRatio, you mean this part, right?
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/54be386fae2ba55fef5a58d49bb8335a5f7b3aa5/jupyterhub/templates/image-puller/_helpers-daemonset.tpl#L154-L204

I believe the way that's written will work for both dict and null, unless I'm missing something. I did try to override the image in the test and I've seen it trying to get pre-pulled [here](https://github.com/GeorgianaElena/zero-to-jupyterhub-k8s/actions/runs/30003780720/job/89194847342).

jupyterhub/values.yaml
```diff
@@ -441,7 +441,7 @@ singleuser:
   cmd: jupyterhub-singleuser
   defaultUrl:
   extraPodConfig: {}
-  profileList: []
```
consideRatio, line 444: Not confident what makes sense here, is null more suitable than {} or []? I think it maybe is. We would want to avoid warnings related to these things showing up during template rendering for any user consistently using one type (list or dict) in passed values.

consideRatio (approved):
Looks right! Not confident on the way to avoid warnings, but i think you maybe have considered and avoided that already! 

Feel free to self merge if you think its OK to go!

jupyterhub/values.yaml
GeorgianaElena: @consideRatio, yes, I believe null is right default to use here. I've seen `{}` failing when we added support for dict `extraVolumes` and `extraVolumeMounts` in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3769

PR #3973, opened 2026-09-05
Add proxy.chp.enabled flag to allow disabling the chp proxy pod

### LINKED ISSUE #3481: Ability to disable proxy
### Proposed change
I would like the ability to disable the current proxy solution. The reason being that I implemented our own Traefik + Redis solution using the currently unreleased 2.0.0 [Traefik Proxy](https://github.com/jupyterhub/traefik-proxy) because #1162 was closed for the time being. In the clusters I manage, we're seeing the same memory leaks described in [Memory leak in proxy?](https://github.com/jupyterhub/configurable-http-proxy/issues/388) and had to switch away from CHP as a result.

I understand that disabling CHP in the current chart leads to an incomplete deployment but because #1162 isn't ready yet, it would be nice to be able to disable CHP anyway to allow custom solutions. I currently extract the files from the .tgz chart, nuke the proxy folder under templates, repackage using helm and then place random values into the PROXY_API_SERVICE_PORT and PROXY_PUBLIC_SERVICE_PORT env vars for the hub pod to disable the proxy properly.

### Alternative options

Replace CHP with Traefik would be an alternative option but I understand that this is a lot of work and not ready yet. Allowing custom proxy solutions could be a nice stopgap.

People suffering from issues with CHP that want to implement their own proxy solution.

Probably just adding some {{- if .Values.proxy.enabled }} in a few places.

jabbera:
Gates the chp Deployment, PodDisruptionBudget, and NetworkPolicy so deployments providing their own proxy can skip the built-in CHP pod.

### Checklist

- [X] I am the author of this work - Claude Sonnet used as well

### What does this PR do?
For users who want to use traefik proxy this is an easy way to disable it as opposed to a helm post render hook.   This is an advanced use case so the feature is bare minimum.

Type of change:
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [ ] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

Fixes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3481

I know that there was some thought around a different way to do this, but that was 2 years ago and this changes seems pretty safe.

No

```
helm template jupyterhub jupyterhub/ | grep -A2 "^kind: Deployment" | grep -i proxy
```

```
helm template jupyterhub jupyterhub/ \
  --set proxy.chp.enabled=false \
  --set proxy.chp.pdb.enabled=true \
  --set proxy.chp.networkPolicy.enabled=true \
  > /tmp/chp-disabled.yaml
grep "^# Source:" /tmp/chp-disabled.yaml | grep proxy
```

### DIFF
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -1594,10 +1594,7 @@ properties:
             description: |
               Default `true`.
 
-              Whether to deploy the `proxy` pod running
-              [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy),
-              along with its associated k8s resources (PodDisruptionBudget,
-              NetworkPolicy).
+              Whether to deploy configurable-http-proxy.
 
               Disable this only if you are providing your own proxy
               implementation for JupyterHub to manage instead of chp.
--- a/jupyterhub/values.schema.yaml
+++ b/jupyterhub/values.schema.yaml
@@ -1589,6 +1589,18 @@ properties:
           Configure the configurable-http-proxy (chp) pod managed by jupyterhub to route traffic
           both to itself and to user pods.
         properties:
+          enabled:
+            type: boolean
+            description: |
+              Default `true`.
+
+              Whether to deploy the `proxy` pod running
+              [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy),
+              along with its associated k8s resources (PodDisruptionBudget,
+              NetworkPolicy).
+
+              Disable this only if you are providing your own proxy
+              implementation for JupyterHub to manage instead of chp.
           revisionHistoryLimit: *revisionHistoryLimit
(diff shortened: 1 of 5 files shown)


jupyterhub/values.schema.yaml
```diff
@@ -1589,6 +1589,18 @@ properties:
...
         properties:
+          enabled:
+            type: boolean
+            description: |
+              Default `true`.
+
+              Whether to deploy the `proxy` pod running
+              [configurable-http-proxy](https://github.com/jupyterhub/configurable-http-proxy),
+              along with its associated k8s resources (PodDisruptionBudget,
+              NetworkPolicy).
```
manics:
```suggestion
              Whether to deploy configurable-http-proxy.
```
We don't need to be so verbose.

yuvipanda:
Thank you, @jabbera!

PR #3978, opened 2026-09-09
Adding UV to speed up package builds

alekseyolg:
### Checklist

- [x] I am the author of this work
The code was created using Claude, and the build was verified locally.
### What does this PR do?

Type of change:
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [x] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

Switches the wheel-install step in `images/singleuser-sample/Dockerfile`
and `images/hub/Dockerfile` from `pip install` to `uv pip install`, to
speed up image builds. Wheel *building* (`pip wheel ...` in `build-stage`)
stays on plain pip, since `uv` has no equivalent to `pip wheel
--wheel-dir`.

Not tied to an existing issue — just a build-speed improvement.

No. Only the tool used to install already-built wheels changes; the
resulting images should be unchanged.

Build both Dockerfiles locally (`docker build images/singleuser-sample`
and `docker build --target slim-stage` / `--target default-stage` for
`images/hub`) and confirm they succeed and produce working images.
Comparing build time with a cold and a warm BuildKit cache before/after
should show the speedup.

### What should a reviewer concentrate their feedback on?

- Whether pinning `uv` via a hand-written `COPY --from=ghcr.io/astral-sh/uv:X.Y.Z`
  is the preferred approach here, or if it should instead go through
  `ci/refreeze` like the Python base image version.
- Whether `--system` / `--no-index` flags on `uv pip install` are the right
  equivalents for this repo's use case.

### Other information

Used an LLM (Claude) to help draft the Dockerfile changes and this description;
reviewed and tested them myself before opening the PR.

### DIFF
--- a/images/hub/Dockerfile
+++ b/images/hub/Dockerfile
@@ -18,6 +18,12 @@ FROM python:3.14-trixie as build-stage
 # ephemeral docker cache (--mount=type=cache,target=${PIP_CACHE_DIR}). We use
 # the same technique for the directory /tmp/wheels.
 #
+# NOTE: wheel-building stays on plain pip -- uv's `pip` interface has no
+#       `wheel` subcommand, so there's nothing to speed up here with uv.
+#       The install steps below (which run far more often, on every rebuild
+#       that touches requirements.txt) are where uv pays off, so uv is
+#       installed and used only in the later stages.
+#
 COPY requirements.txt requirements.txt
 ARG PIP_CACHE_DIR=/tmp/pip-cache
 RUN --mount=type=cache,target=${PIP_CACHE_DIR} \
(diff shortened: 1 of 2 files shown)


manics:
UV is a nice package manager, but unlike pip it's managed by a single company and hasn't been around for as long, so there's a high bar for switching.

Looking at the most recent build times for PRs that required a rebuild of the image in the chartpress step:
- This PR: **1m50s** https://github.com/jupyterhub/zero-to-jupyterhub-k8s/actions/runs/34366634988/job/102516929780?pr=3978
- Older PR that bumped JupyterHub and triggered an image rebuild: **2m1s** https://github.com/jupyterhub/zero-to-jupyterhub-k8s/actions/runs/33593443781/job/100131981081

Given that the image build times are quite short already I think we need a stronger justificaition to switch.

minrk:
Yup, also +1 sticking to community tools and not OpenAI-owned uv, which doesn't save an appreciable amount of time in our case (32s vs 35s in local no-cache testing, so less than 10%).

Thanks for the PR, though!

